PR #6, opened 2026-02-08
Add Linux auto-update mechanism and update README

green2grey:
- Add linux-updater.js: self-contained update module that checks the official appcast feed, patches the Check for Updates menu, runs background checks every 15 minutes, and handles install-app-update IPC messages by running install.sh --force and relaunching
- Modify install.sh patch_asar() to inject linux-updater.js into the app.asar build and prepend require() to main.js entry point
- Modify start.sh generation to export CODEX_LINUX_INSTALLER_PATH
- Update README with fork attribution, changelog, and updater docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,30 @@
 
 Run [OpenAI Codex Desktop](https://openai.com/codex/) on Linux.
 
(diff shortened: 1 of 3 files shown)


ilysenko:
Closing for now instead of merging a large auto-update refactor into current main. The updater approach here conflicts with fixes already merged in #12 and #14; if we revisit Linux auto-update, it should come back as a fresh PR based on current main.

ilysenko:
Thank you, Babken, for putting real work into the Linux auto-update direction here. I closed this PR because current main has moved on after #12 and #14, and I did not want to merge a large refactor on top of conflicting installer behavior; if we revisit Linux auto-update, this deserves to come back as a fresh PR from current main.

PR #11, opened 2026-02-12
fix(installer): extract only .app bundle from DMG to avoid 7z symlink…

zuuxuux:
I was getting a failure on my system (ubuntu 24.04.3 LT).

This fixes it

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -88,19 +88,23 @@ get_dmg() {
 
 # ---- Extract app from DMG ----
 extract_dmg() {
-    local dmg_path="$1"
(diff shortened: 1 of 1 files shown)


ilysenko:
Closing as superseded by the DMG extraction fixes already merged in #14. If we want to revisit app-bundle-only extraction later, it should be a fresh PR on top of current main.

ilysenko:
Thank you for reporting the Ubuntu 24.04 extraction failure and for proposing a concrete fix here. I closed this PR because the extraction issue was already addressed in #14 on current main, but the report and attempted fix were definitely useful.

PR #18, opened 2026-03-27
feat: add user-local desktop integration helper bundle

moabualruz:
Title: feat: add user-local desktop integration helper bundle

Body:

## Summary

This adds an optional `contrib/user-local-install` bundle for users who want to treat `codex-desktop-linux` like a managed desktop app instead of a one-off build folder.

The bundle includes:

- `~/.local/opt/codex-desktop-linux` as the stable repo/install location
- `~/.local/bin/codex-desktop` launcher
- `codex-desktop-check-update`, `codex-desktop-update`, and `codex-desktop-version` helper commands
- a `.desktop` launcher under `~/.local/share/applications`
- icon extraction from the local `Codex.dmg`
- metadata tracking for both the wrapper repo and the upstream `Codex.dmg`
- a weekly `systemd --user` timer for unattended update checks/rebuilds

## Why

The current installer works well for initial setup, but users may reasonably want:

- a meaningful install location outside `Downloads`
- a menu entry
- a stable launcher command
- a simple way to check/update versions later
- basic automation for keeping the wrapper current

This keeps that logic isolated in `contrib/` so the main installer stays simple.

## Included files

- `contrib/user-local-install/README.md`
- `contrib/user-local-install/install-user-local.sh`
- `contrib/user-local-install/files/.local/bin/codex-desktop`
- `contrib/user-local-install/files/.local/bin/codex-desktop-check-update`
- `contrib/user-local-install/files/.local/bin/codex-desktop-update`
- `contrib/user-local-install/files/.local/bin/codex-desktop-version`
- `contrib/user-local-install/files/.local/lib/codex-desktop-linux/common.sh`
- `contrib/user-local-install/files/.local/share/applications/codex-desktop.desktop`
- `contrib/user-local-install/files/.config/systemd/user/codex-desktop-update.service`
- `contrib/user-local-install/files/.config/systemd/user/codex-desktop-update.timer`

## Notes

- This is additive only.
- The icon is generated locally from `Codex.dmg` rather than committed as a binary asset.
- The desktop entry is templated during install so it resolves to the current user's home path cleanly.

### DIFF
--- a/contrib/user-local-install/README.md
+++ b/contrib/user-local-install/README.md
@@ -5,7 +5,8 @@ This folder packages a user-local install layout for `codex-desktop-linux`.
 It adds:
 
 - a stable install root under `~/.local/opt/codex-desktop-linux`
-- launch/check/update/version commands under `~/.local/bin`
+- self-contained maintenance scripts under `~/.local/opt/codex-desktop-linux/bin`
+- thin launch/check/update/version wrappers under `~/.local/bin`
 - a desktop entry under `~/.local/share/applications`
 - an icon extracted from the local `Codex.dmg`
 - metadata tracking for the wrapper repo and cached `Codex.dmg`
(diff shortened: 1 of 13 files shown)


ilysenko:
Thanks for the contribution\! Nice work on the user-local install bundle. Could you please resolve the merge conflicts and check whether this is still relevant given the recent commits on main (especially the packaging refactor and runtime helper changes)? If it's still good to go, would be great to get it merged. Thanks\!

ilysenko (changes requested):
Thanks for the work on this! A few issues need to be addressed before merging:

1. **Breaking change**: `--class=codex-desktop` in `install.sh` mismatches `StartupWMClass=Codex` in the existing packaging desktop file — this breaks taskbar icon grouping for .deb/.rpm users. Please update `packaging/linux/codex-desktop.desktop` to match, or keep the existing class name.
2. **`patch-linux-window-ui.js`**: Searching for minified variable names (e.g. `D`) will break on any upstream update. Please make it fail gracefully (warn + skip) instead of `throw Error`.
3. **Systemd unit files**: The `.service` and `.timer` files are never copied to `~/.config/systemd/user/`, so the auto-update timer doesn't actually work.
4. **Unattended `git pull` + `./install.sh`**: The weekly timer gives the upstream repo silent code execution. Consider making this opt-in.

Suggestion: split into two PRs — (A) `install.sh` improvements + desktop file fix + graceful patching; (B) the contrib bundle with fixed systemd install and opt-in timer.

moabualruz:
Done ;)

PR #27, opened 2026-04-05
fix: auto-bootstrap 7zz when system 7-zip is too old

prideofszeged:
## Summary

- The hardcoded `7z2409` URL in `install.sh` 404s — that version no longer exists on 7-zip.org
- Instead of erroring with a stale URL, the installer now dynamically detects the latest 7zz version from the download page and installs it automatically
- Handles x86_64, aarch64, and armv7l architectures
- Tries `/usr/local/bin` first, falls back to `~/.local/bin` (no sudo required) with a PATH warning
- Updates README to note auto-bootstrap behavior and fix the example URL to current version (2600)

## Test plan

- [ ] Run `install.sh` on a system with only `p7zip-full 16.02` — should auto-download 7zz and continue
- [ ] Run on a system without sudo access — should install to `~/.local/bin` and warn about PATH
- [ ] Run on a system with modern `7zz` already installed — should skip bootstrap entirely
- [ ] Verify `bash -n install.sh` passes (syntax check)
- [ ] Test on aarch64 if available

### DIFF
--- a/scripts/install-deps.sh
+++ b/scripts/install-deps.sh
@@ -74,17 +74,6 @@ install_pacman() {
 # Pinned versions — prepend new entries as upstream releases them.
 # ---------------------------------------------------------------------------
 bootstrap_7zz() {
-    local sevenzip_arch
-    case "$ARCH" in
-        x86_64)  sevenzip_arch="x64"   ;;
-        aarch64) sevenzip_arch="arm64"  ;;
-        armv7l)  sevenzip_arch="arm"    ;;
-        *)
-            warn "Skipping 7zz bootstrap: unsupported architecture '$ARCH'"
-            return 0
-            ;;
-    esac
-
     # Already present and functional
     if command -v 7zz &>/dev/null && 7zz 2>&1 | grep -qm 1 "7-Zip"; then
         info "7zz already available ($(command -v 7zz))"
@@ -96,6 +85,17 @@ bootstrap_7zz() {
         info "System 7z is already new enough; skipping 7zz bootstrap"
         return 0
     fi
+
+    local sevenzip_arch
+    case "$ARCH" in
+        x86_64)  sevenzip_arch="x64"   ;;
+        aarch64) sevenzip_arch="arm64"  ;;
+        armv7l)  sevenzip_arch="arm"    ;;
+        *)
+            warn "Skipping 7zz bootstrap: unsupported architecture '$ARCH'"
+            return 0
+            ;;
+    esac
 
     local install_dir="$HOME/.local/bin"
     if [ "${SEVENZIP_SYSTEM_INSTALL:-0}" = "1" ]; then
--- a/scripts/install-deps.sh
+++ b/scripts/install-deps.sh
@@ -8,6 +8,7 @@ RED='\033[0;31m'
 GREEN='\033[0;32m'
 YELLOW='\033[1;33m'
 NC='\033[0m'
+ARCH="$(uname -m)"
 
 info()  { echo -e "${GREEN}[INFO]${NC}  $*"; }
 warn()  { echo -e "${YELLOW}[WARN]${NC}  $*"; }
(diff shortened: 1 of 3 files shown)


Leay15 (approved):

ilysenko (commented):
Thanks for picking this up — the dead `7z2409` URL is a real blocker for Ubuntu/Pop!_OS users, so the intent here is good. A few things I'd like to see addressed before merging:

**1. Web scraping is fragile.** `grep -oP "7z\K[0-9]+" | head -1` depends on the HTML structure of 7-zip.org/download.html. If they reorder the page or the first `7zXXXX` token in the HTML stops being the latest version, the bootstrap silently breaks. Could you instead try a small list of pinned known-good versions (e.g. `2500`, `2409`) with HEAD verification, falling back from newest to oldest? That removes the dependency on page structure entirely.

**2. Bootstrap logic belongs in `scripts/install-deps.sh`.** That script is the centralized dependency installer (apt/dnf/pacman). The 7zz bootstrap is conceptually a dependency install, not part of the DMG-extraction flow in `install.sh`. Could you move `bootstrap_7zz()` there and have `install.sh` just re-check after the user runs `install-deps.sh` (or hint at it in the error)?

**3. Don't silently write to `/usr/local/bin`.** Trying it first and falling back is surprising if the user happened to run `install.sh` with elevated privileges. Please default to `~/.local/bin` and only use `/usr/local/bin` with an explicit opt-in (env var or flag).

**4. Pinned README version will go stale.** The `7z2600` example URL has the same problem the original `7z2409` had — it becomes wrong as soon as 7-zip ships a new release. Better to say "replace with the current version from https://www.7-zip.org/download.html" without a specific number.

If you're short on time, let me know and I can take this over and finish it. Otherwise happy to re-review once these are addressed.

prideofszeged:
Hi @ilysenko — addressed all four points from your review:

1. **Web scraping removed** — replaced `grep -oP` HTML scraping with a pinned version list `(2600, 2500, 2409)` tried newest-first with `curl -fsI` HEAD verification. No HTML parsing.

2. **Moved to `scripts/install-deps.sh`** — `bootstrap_7zz()` now lives there. `install.sh` just detects the old version and points users to run `bash scripts/install-deps.sh`, including a PATH export hint for the `~/.local/bin` case.

3. **`~/.local/bin` by default** — `SEVENZIP_SYSTEM_INSTALL=1` is the explicit opt-in for `/usr/local/bin`. Updated README accordingly.

4. **README version unpinned** — replaced `7z2409`/`7z2600` example URLs with a `<VERSION>` placeholder.

Also fixed a few bugs caught during review:
- Version probe was checking line 1 (blank on modern 7z); now uses `grep -m 1 "7-Zip"` in both files
- `trap ... RETURN` doesn't fire on `errexit`; switched to `trap ... EXIT`
- `7zz` presence check now also verifies the binary is functional

@Leay15 would appreciate a re-review when you get a chance too!

scripts/install-deps.sh
```diff
@@ -68,6 +69,81 @@ install_pacman() {
...
+    local sevenzip_arch
+    case "$ARCH" in
+        x86_64)  sevenzip_arch="x64"   ;;
+        aarch64) sevenzip_arch="arm64"  ;;
+        armv7l)  sevenzip_arch="arm"    ;;
+        *)
+            warn "Skipping 7zz bootstrap: unsupported architecture '$ARCH'"
+            return 0
+            ;;
+    esac
```
Leay15: This code segment remains unutilized if Sevenzip is already present and represents a more recent iteration.

Please move it before it will use `sevenzip_arch`

Leay15 (changes requested):
@prideofszeged Done!

Just one comment, everything else looks good for me 😄

ilysenko:
Pushed a small follow-up to your branch (4735576) addressing @Leay15's comment — moved the arch detection after the early-return checks so that an unsupported architecture doesn't warn-and-skip when 7zz is already installed.

Thanks for the great work on this one — merging now.

PR #33, opened 2026-04-11
fix(updater): fix mixed-manager updater packaging

gadicc:
On Arch systems that also had rpm installed, PackageKind::detect() could fall through to the rpm path even when codex-desktop was actually installed via pacman. That made the updater rebuild .rpm artifacts, report the installed version as unknown, and fail the privileged install path on a working pacman install.

We also found that scripts/build-pacman.sh reused an existing target/release/codex-update-manager binary without checking whether the updater sources had changed, so rebuilding and reinstalling the pacman package could still ship the old updater binary.

Prefer the package manager that already owns codex-desktop, then fall back to /etc/os-release before raw command presence. Add regression tests for mixed package-manager systems, and rebuild the packaged updater binary whenever Cargo metadata or files under updater/ are newer than the existing release binary.

Co-Authored-By: Codex, gpt-5.4, xhigh.

Additional notes (not in commit message):

- Tested working on my Arch system which had this issue because rpm was installed too
- The new approach of course is of course useful for non-Arch systems too

### DIFF
--- a/scripts/lib/package-common.sh
+++ b/scripts/lib/package-common.sh
@@ -20,8 +20,29 @@ ensure_app_layout() {
     [ -x "$APP_DIR/start.sh" ] || error "Missing launcher: $APP_DIR/start.sh"
 }
 
+updater_binary_is_stale() {
+    local binary="$1"
+
+    [ -x "$binary" ] || return 0
+
+    local source
+    for source in "$REPO_DIR/Cargo.toml" "$REPO_DIR/Cargo.lock"; do
+        if [ -f "$source" ] && [ "$source" -nt "$binary" ]; then
+            return 0
+        fi
+    done
+
+    while IFS= read -r -d '' source; do
+        if [ "$source" -nt "$binary" ]; then
+            return 0
+        fi
+    done < <(find "$REPO_DIR/updater" -type f -print0 2>/dev/null)
+
+    return 1
+}
+
 ensure_updater_binary() {
-    if [ -x "$UPDATER_BINARY_SOURCE" ]; then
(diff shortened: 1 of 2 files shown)


Leay15:
@gadicc nice catch!

Could you please check if this behavior always occur on #34 ?

I found a similar behavior due to mismatching SHA signs, worked in hardening that part but I wasn't able to reproduce it anymore (but I'm not quite sure it is fixed)

gadicc:
Thanks, @Leay15, although most of the credit goes to Codex :stuck_out_tongue: 

There are two issues addressed by this PR:

* The updater could build for the wrong package manager if multiple package managers were installed.

* The updater binary itself, `codex-update-manager`, was not rebuilt even if its sources were changed, i.e. `Cargo.toml`, `Cargo.lock`, or `updater/*`.

I discovered the latter after "fixing" the first issue, but still seeing the same problem, because, it still shipped the previously built updater without rebuilding - even though its sources had changed.

So, I think the problems are tangential rather than overlapping, and both fixes are required:

* Stale upstream **DMG package** - caught by SHA hashing in your #34

* Stale **codex-update-manager** (due to newer sources/deps) - caught by this PR

There is a tiny bit of overlap though, as, I think your fixes wouldn't make it into the newly built package without the fix here, i.e., the newly built package would contain code from the new DMG, but, if the updater in the repo had been built previously, the newly built package would contain the previosly built updater without the SHA checking code.

But to be honest, I did only skim through your code and didn't actually try to test with only #34 applied.

In any event, happy to be collaborating on this fun project, and thanks for your numerous PRs and investigations until now! :pray:

Leay15:
@gadicc makes sense for me!

Let's wait for @ilysenko review and comments about 😁


It's a pleasure contributing with projects like this 😁

PR #35, opened 2026-04-14
Update nix flake's hash and fix patch-linux-window-ui.js problem

RFLXN:
## Problem

On NixOS with `nix run .`:

- the pinned `Codex.dmg` hash was outdated
- the installer failed with `Cannot find module '/tmp/.../scripts/patch-linux-
window-ui.js'`

## Cause

The DMG hash in `flake.nix` no longer matched the current upstream file.    

Also, the Nix wrapper only copied `install.sh` and `Codex.dmg` into a temporary directory:

```nix
cp ${./install.sh} "$workdir/install.sh"
cp ${codexDmg} "$workdir/Codex.dmg"
chmod +x "$workdir/install.sh"
```

But install.sh uses repo-relative files from scripts/ and assets/, so those files were missing under `nix run .`

## Changes

- updated the pinned Codex.dmg hash
- changed the Nix wrapper to copy a cleaned source tree instead of only install.sh

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -10,10 +10,19 @@
     flake-utils.lib.eachDefaultSystem (system:
       let
         pkgs = import nixpkgs { inherit system; };
+        sourceRoot = pkgs.lib.cleanSourceWith {
+          src = ./.;
+          filter = path: type:
+            pkgs.lib.cleanSourceFilter path type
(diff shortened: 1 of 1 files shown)


ilysenko (approved):
Thanks!

PR #38, opened 2026-04-20
v0.4.1

Leay15:
This pull request introduces improvements to the Linux packaging and update manager for Codex Desktop, focusing on more robust user service management and enhanced status reporting for the updater and CLI. The changes ensure that the `codex-update-manager` user service is reliably enabled and started for all users during installation or upgrade, and provide more detailed and accurate CLI status information. 

Additionally, notification logic is improved to avoid duplicate notifications, and several scripts and the updater itself are updated to support these enhancements.

**Linux packaging and service management improvements:**

* Added a new `codex-update-manager.postinst` script to ensure the user service is enabled and started for all users during DEB package installation. The build scripts and packaging logic (`build-deb.sh`, `package-common.sh`) are updated to include and install this script. 
* Refactored the RPM and Arch packaging logic (`codex-desktop.spec`, `codex-desktop.install`) to include robust logic for enabling and starting the user service for all users with a session, mirroring the logic in the new DEB postinst script. 

**Updater and CLI status enhancements:**

* Improved the updater's status reporting: the `run_status` command now refreshes CLI status, includes error messages in the output, and provides more accurate status details.
* Added a new `refresh_status` function in `codex_cli.rs` to accurately detect the CLI's presence, version, and errors, and updated the logic for determining when to check for new versions. 
* Updated how the updater handles CLI version checks and status caching, improving reliability and reducing unnecessary network requests.
**Notification and state management improvements:**

* Refactored notification logic to deduplicate "installed" notifications and ensure users are only notified once per installed version, even after restarts or recovery. Added tests for this behavior. 

**Version bump:**

* Bumped the `codex-update-manager` crate version from 0.4.0 to 0.4.1 in both `Cargo.toml` and documentation.


ilysenko:
Thanks for the patch — merges cleanly and the packaging logic mirrors the existing `prerm`/`postrm` style. Two things worth a look (both fine as follow-ups, not blockers for this PR):

1. **`run_status` is now a writer of `state.json`.** Previously `Commands::Status` was read-only; after this change every ad-hoc `codex-update-manager status` invocation calls `refresh_status` and persists, racing with the daemon on the same file. Could we confirm `PersistedState::save` writes atomically (temp file + `rename`) so a concurrent status call can't leave a truncated/partial JSON for the daemon to read back? If not, a small atomic-write helper would be a good follow-up.

2. **Triple-duplicated systemd-user enable/start block** across `codex-update-manager.postinst`, `codex-desktop.spec` (`%post`), and `codex-desktop.install` (`post_install`). Not a safety issue, but the three copies will drift. A follow-up extracting this into a single shared snippet (e.g. a shell helper sourced by all three, or generated at package-build time) would make future changes much easier.

Leay15:
Hi @ilysenko, thanks for the comments 🙂

Nice catch—here’s a quick summary:

1. You were right about the write safety issue. After the earlier change, `codex-update-manager status` was no longer read-only, and `PersistedState::save` still used `fs::write`, which could lead to truncated or partial `state.json` writes.  
   I fixed this by implementing atomic writes in `updater/src/state.rs` (temp file + `sync_all()` + `rename()`), and added a regression test to verify the behavior.

2. You were also right about the duplication. The same `systemd --user` logic was repeated across Debian, RPM, and pacman scripts.  
   I extracted it into a shared helper (`packaging/linux/codex-update-manager-user-service.sh`) and updated all package scripts to reuse it. I also replaced the temporary `ShellCheck` suppression with a proper `shellcheck source=...` directive.

PR #44, opened 2026-04-23
[codex] fix CLI preflight updates

danielwolfman:
## Summary
- update preflight when the cached latest CLI version is newer than the installed version
- retry npm install with the prefix that owns an explicit CLI path when the global npm prefix updates a different install
- add regression coverage for both failure modes

## Validation
- cargo fmt --check

### DIFF
--- a/updater/src/codex_cli.rs
+++ b/updater/src/codex_cli.rs
@@ -8,9 +8,11 @@ use anyhow::{anyhow, Context, Result};
 use chrono::{Duration, Utc};
 use std::{
     ffi::OsString,
-    fs,
+    fs::{self, OpenOptions},
     path::{Path, PathBuf},
+    process,
     process::{Command, Output},
+    time::{SystemTime, UNIX_EPOCH},
 };
 use tracing::{info, warn};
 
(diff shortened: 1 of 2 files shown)


ilysenko:
Picked this up while it's still in draft — flagging early so you have it when you're ready to move it forward. The intent is right (the cached-latest > installed case really did silently skip the upgrade), but two things on the new prefix-aware reinstall:

1. **`install_latest_cli_for_requested_path` doesn't validate the computed prefix.** `npm_prefix_for_cli_path` happily returns `/usr`, `/usr/local`, or `/opt/codex-desktop/codex-app` if the resolved `codex` binary lives there, and `npm install -g --prefix <that>` will then either need root or overwrite files owned by the distro package / our own RPM/DEB payload. Worth gating on something like "prefix is user-writable and already contains `lib/node_modules/codex-cli`" so we only retry into a prefix that genuinely owns the install.

2. **The new early-branch skips the TTL.** `cached_latest_version_matches_install` doesn't look at `cli_last_check_at`, so a stale cached `latest` (e.g. several hours old) will trigger an upgrade to that stale version instead of refreshing first. Probably rare given PR #46's launch-time `check-now`, but easy to close: either require TTL freshness here too, or fall through to `read_latest_version()` before upgrading.

Minor: the `path_env_lock` mutex around `std::env::set_var` works today, but anything else in the crate touching `PATH` would race with these tests. Subprocess isolation or `--test-threads=1` for that category would be sturdier.

Otherwise the refactor of `upgrade_cli_to_latest` reads cleanly and the test coverage for the two new paths is solid. No rush from my side — happy to re-review when you flip it out of draft.

PR #45, opened 2026-04-24
v0.4.2: Enhance Codex CLI installation and update DMG handling

Leay15:
This pull request improves the automatic installation flow for the Codex CLI, enhance user notifications, and clarify documentation. The main changes include making the CLI preflight smarter by prompting users before attempting automatic installs, improving notification handling when the CLI is missing, and updating the build process to better handle DMG downloads. The documentation and tests have also been updated to reflect these changes.

**Improvements to CLI installation and preflight:**
- The launcher now prompts the user before attempting to auto-install the Codex CLI if it is missing and the terminal is interactive, instead of always installing automatically. This avoids unexpected installs and improves user experience. 
- The CLI preflight logic now uses `install.sh` to determine when missing-CLI installation is allowed, rather than always enabling auto-installation. 

**Notification and updater enhancements:**
- The `codex-update-manager` now refreshes CLI status and shows a desktop notification if the Codex CLI is missing, with deduplication and automatic clearing of the notification marker once the CLI is installed. 
- Added tests to ensure CLI missing notifications are deduplicated and cleared after recovery.

**Build and packaging process improvements:**
- `make build-app` now defaults to letting `install.sh` reuse or download `Codex.dmg` if no override is provided, improving the experience for fresh checkouts.

**Documentation updates:**
- Updated `README.md` and `AGENTS.md` to clarify the new CLI installation flow, version bumps, and improved notification behavior. 

**Version bumps:**
- Bumped `codex-update-manager` crate version to `0.4.2` across the codebase and documentation.


ilysenko:
Net-positive PR, two notes worth flagging — neither blocking:

1. **GUI launcher has no install path for a missing CLI.** `prompt_install_missing_cli` requires a TTY (`[ -t 0 ]`), so double-click / desktop-entry launches skip it and hit the `exit 1` branch. Behavior matches the pre-PR state, but the new desktop notification from the daemon becomes the only signal to GUI users — and only if a notification daemon is reachable. A `zenity`/`kdialog` fallback for the prompt, or making the notification actionable, would close the loop. Fine as a follow-up.

2. **`refresh_status` now runs on every daemon start and every `check-now`.** It persists `state.json` two-three times and shells out to `codex --version` unconditionally (the npm-registry TTL only guards the latest-version lookup). Combined with PR #46's launch-time `check-now`, that's an extra ~200ms + several state writes per app launch. Probably fine, but worth keeping in mind if startup latency starts to matter.

Everything else looks good — the dedup + auto-clear semantics on `cli_missing` are right, the Makefile simplification is reasonable, and plumbing `--allow-install-missing` through install.sh under explicit user consent is the right call.

Heads-up: there will be a small textual conflict with #46 in `app.rs` (different functions, but adjacent context). Whoever lands second can rebase.

Leay15:
@ilysenko Thanks for the comments.

I’ll work on this, but I think it will require a new minor version. I’m considering moving the GUI behavior and changing the startup cache behavior by separating responsibilities.

I’ll push a new branch for this.

We can merge this if everything looks good to you 🙂

PR #46, opened 2026-04-25
fix(updater): Trigger safe launch-time update checks

pradigmaz:
## Summary
- start the packaged updater service without restarting it on every app launch, then trigger a non-blocking launch-time `check-now`
- serialize updater checks with a state-directory lock so daemon, periodic, and launch-triggered checks do not duplicate downloads/builds
- align RPM staging with the shared package payload so RPM installs include `.codex-linux/codex-packaged-runtime.sh` and the full update-builder bundle
- compare generated update versions after stripping RPM release suffixes, so interrupted installs can recover when the candidate is already installed

## Why
While validating the Fedora RPM install path, the updater could download and rebuild a candidate but RPM installs missed files that the packaged runtime and update-builder expected. App-launch service restarts could also interrupt an active updater build. This keeps launch checks best-effort and non-blocking while letting the updater skip duplicate live checks safely.

## Testing
- `bash -n packaging/linux/codex-packaged-runtime.sh scripts/build-rpm.sh tests/scripts_smoke.sh`
- `cargo test -p codex-update-manager` (65 passed)
- `PACKAGE_VERSION=2026.04.25.054929+prcheck APP_DIR_OVERRIDE=<path-to-extracted-codex-app> ./scripts/build-rpm.sh`
- `rpm -qlp dist/codex-desktop-2026.04.25.054929-prcheck.x86_64.rpm | rg '/opt/codex-desktop/(\.codex-linux/codex-packaged-runtime\.sh|update-builder/(packaging/linux/codex-packaged-runtime\.sh|scripts/patch-linux-window-ui\.js|scripts/lib/package-common\.sh))'`

Known unrelated checks:
- `bash tests/scripts_smoke.sh` reaches the Debian/RPM packaging and launcher marker checks, then fails in the existing `test_linux_translucent_sidebar_default_patch_smoke` fixture.
- `cargo fmt --check` fails on pre-existing formatting in `updater/src/codex_cli.rs` and `updater/src/state.rs`; this PR does not touch those files.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -407,6 +407,7 @@ dependencies = [
  "chrono",
  "clap",
  "directories",
+ "fs4",
  "futures-util",
  "notify-rust",
  "reqwest",
@@ -673,6 +674,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cb4cb245038516f5f85277875cdaa4f7d2c9a0fa0468de06ed190163b1581fcf"
 dependencies = [
  "percent-encoding",
+]
+
+[[package]]
+name = "fs4"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8640e34b88f7652208ce9e88b1a37a2ae95227d84abec377ccd3c5cfeb141ed4"
+dependencies = [
+ "rustix",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


ilysenko:
The RPM staging refactor and the version-suffix fix look right. One concern before merge:

`check.lock` can become permanently stale. The PID is written but never validated. If the holder dies non-gracefully (SIGKILL, OOM, kernel panic, power loss), `Drop` doesn't run, the file persists, and every subsequent `run_check_cycle` — including the daemon's 6-hour ticks — silently logs `skipping upstream check` and exits. Only manual deletion of `<state_dir>/check.lock` recovers. Silent forever-skip on an unattended service is a bad failure mode.

Suggest switching to `flock(2)` (e.g. `fs4` or `nix`) — kernel releases the lock on process death regardless of cause.

Other pieces look good. Happy to land once the lock swap is in, or as a tracked follow-up.

pradigmaz:
Thanks, agreed. I replaced the PID-file guard with a kernel-backed exclusive file lock via fs4. The lock file may remain on disk, but it no longer controls exclusion; stale files are overwritten only after the flock is acquired, and the kernel releases the lock if the process exits or dies non-gracefully.

I also added regression coverage for both cases:
- a stale check.lock file without a held kernel lock no longer blocks the next check
- an actively held check lock still prevents a concurrent check until the holder drops it

Verified with cargo test -p codex-update-manager.

PR #50, opened 2026-04-28
fix(updater): Use polkit for package installs

pradigmaz:
## Summary
- ship a polkit action policy for the updater install subcommands and call `pkexec` with `--disable-internal-agent`, so package installs use the desktop authentication agent instead of falling back to a textual prompt
- keep cancelled or unavailable `pkexec` authentication retryable: the updater no longer marks the candidate as permanently failed, and it suppresses repeated prompts until the app has run again or a new candidate appears
- reject non-newer RPM packages before invoking DNF/Zypper, matching the existing DEB/pacman downgrade guard behavior for generated package versions
- keep the user updater service running during RPM upgrades by only stopping/disabling it from `%preun` on package erase
- include the polkit policy in both the installed package payload and the update-builder bundle, and add the required polkit dependencies for RPM/DEB/pacman packages

## Why
The updater needs a privileged package install step, but unattended service retries should not get stuck after a cancelled authentication prompt or repeatedly launch a text-based auth flow. RPM installs also needed the same non-newer package guard that DEB and pacman already had, and RPM upgrades should not stop the user updater service between package scriptlets.

## Testing
- `rustfmt --edition 2021 --check updater/src/app.rs updater/src/install.rs`
- `cargo test -p codex-update-manager` (76 passed)
- `git diff --check`
- `bash -n scripts/lib/package-common.sh scripts/build-rpm.sh scripts/build-deb.sh scripts/build-pacman.sh`
- `python3 -m xml.etree.ElementTree packaging/linux/com.github.ilysenko.codex-desktop-linux.update.policy`
- built an RPM from the current branch and verified:
  - the policy is installed under `/usr/share/polkit-1/actions/`
  - the policy is included in the update-builder bundle
  - `/usr/bin/codex-update-manager` is included
  - `rpm -qpR` includes `polkit`
  - `%preun` only stops/disables the user service on erase
- live RPM validation on Fedora:
  - the polkit action registered through `pkaction`
  - `pkexec --disable-internal-agent codex-update-manager install-rpm --path <newer-rpm>` installed a newer test package
  - an older generated RPM version was rejected before DNF/Zypper
  - a fixed-to-fixed RPM upgrade kept `codex-update-manager.service` active

## Notes
- Full `cargo fmt --check` still reports existing formatting drift in `updater/src/codex_cli.rs` and `updater/src/state.rs`; this PR does not touch those files, so the targeted Rust format check above was used.
- DEB and pacman package builds were not run locally because the environment does not have `dpkg-deb` or `makepkg`; their shared shell staging path was syntax-checked, and the RPM build validated the shared policy staging.

### DIFF
--- a/packaging/linux/PKGBUILD.template
+++ b/packaging/linux/PKGBUILD.template
@@ -11,6 +11,7 @@ depends=(
     'npm'
     'python'
     'p7zip'
+    'polkit'
     'curl'
     'unzip'
     'gcc'
--- a/packaging/linux/codex-desktop.spec
+++ b/packaging/linux/codex-desktop.spec
@@ -5,7 +5,7 @@ Summary:        Codex Desktop for Linux
 License:        Proprietary
 ExclusiveArch:  __ARCH__
 
-Requires:       nodejs, npm, python3, p7zip, curl, unzip, gcc-c++, make
+Requires:       nodejs, npm, python3, p7zip, polkit, curl, unzip, gcc-c++, make
 Requires:       alsa-lib, at-spi2-atk, atk, glib2, gtk3, libdrm
 Requires:       nspr, nss, pango, libstdc++, libX11, libxcb
 Requires:       libXcomposite, libXdamage, libXext, libXfixes, libxkbcommon, libXrandr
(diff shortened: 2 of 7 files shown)


ilysenko:
Solid PR — polkit policy + retryable auth + RPM downgrade guard + `%preun` only-on-erase all look right, tests cover the new state-machine paths.

One pre-merge note on the DEB deps: `Depends: pkexec, polkitd` will fail to install on Ubuntu 22.04 LTS (and Mint 21.x / Pop!_OS 22.04 / Elementary based on it). On 22.04 there is no `polkitd` or standalone `pkexec` package — both binaries live in `policykit-1`. The split happened in 24.04 / Debian 12.

One-line fix in `packaging/linux/control` widens the dep to satisfy both:

```diff
-Depends: ..., pkexec, polkitd, ...
+Depends: ..., pkexec | policykit-1, polkitd | policykit-1, ...
```

Not strictly blocking — dpkg fails loudly and the affected audience is narrower than the project's typical Arch/Fedora/Nix install path — but 22.04 is supported through 2027 and there's no clean apt-side workaround once the dep mismatches, so worth handling here or as a quick follow-up.

PR #51, opened 2026-04-29
Harden updater packaging runtime

AmeerJ97:
## Summary
- prevent packaged app launches from starting a duplicate launch-time update check when the user daemon is already active
- wrap privileged package installs with a timeout while preserving upstream polkit auth-deferral behavior for permission-denied/cancelled prompts
- prune stale updater rebuild workspaces after successful builds
- normalize package payload permissions during build and post-install upgrades
- prefer executable user-local Codex CLI candidates and validate updater builder script/tool permissions

## Upstream interaction
- rebased on current `main` after the polkit install changes
- keeps `pkexec --disable-internal-agent` from upstream and applies the timeout around that command path

## Validation
- `bash -n scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh scripts/lib/package-common.sh packaging/linux/codex-packaged-runtime.sh`
- `sh -n packaging/linux/codex-update-manager.postinst packaging/linux/codex-update-manager.prerm packaging/linux/codex-update-manager.postrm packaging/linux/codex-update-manager-user-service.sh packaging/linux/codex-desktop.install`
- `cargo test -p codex-update-manager -- --test-threads=1` -> 83 passed, 0 failed
- `git diff --check`
- `git merge-tree upstream/main origin/codex/updater-package-runtime` -> clean

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -68,7 +68,7 @@ The current working flow is:
 - `~/.local/state/codex-update-manager/service.log`
   Updater service log.
 - `~/.cache/codex-update-manager/`
-  Downloaded DMGs, rebuild workspaces, staged package artifacts, and build logs.
+  Downloaded DMGs, rebuild workspaces, staged package artifacts, and build logs. Successful rebuilds prune older workspaces and keep the current workspace plus the newest previous one.
 
 Do not assume `codex-app/` is pristine. If behavior differs from `install.sh`, prefer updating `install.sh` and then regenerating the app.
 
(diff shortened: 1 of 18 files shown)


ilysenko:

The underlying issues you're catching (interrupted privileged installs hanging in `Installing`, unbounded workspace growth in `~/.cache`) are real. Two things worth landing in this PR before flipping to ready:

- The pkexec timeout (300s) is bounded but I don't see a test mocking the timeout path. A small `run_command_with_timeout` unit test (e.g. spawn `sleep 60` with a 10ms timeout, assert it kills + bubbles up the right error) would lock in the contract.
- For `prune_old_workspaces` — what happens if the current workspace is *also* the newest "previous" one (e.g. retry of the same candidate)? Worth an explicit test that we never delete the workspace we just successfully built into.

The "prefer executable user-local CLI" change is a nice fit alongside #54 — might want to coordinate landing order to avoid a textual conflict in `codex_cli.rs::resolve_cli_path`.

AmeerJ97:
Closing this for now to reduce PR noise. The updater/package runtime work is better re-cut later as a smaller, more focused follow-up after the patcher PR is reviewed.

PR #52, opened 2026-04-29
Harden Linux launcher runtime detection

AmeerJ97:
## Summary
- detect the Electron version from the upstream DMG/app metadata before downloading the Linux Electron runtime
- clean stale macOS/Windows/intermediate native-module artifacts before repacking app.asar
- prefer executable user-local Codex CLI installs when GUI launch PATH is reduced
- add `CODEX_GPU_MODE=conservative|auto|nvidia|off` launcher modes and GPU diagnostics

## Validation
- `bash -n install.sh`
- `git diff --check`
- `git merge-tree upstream/main origin/codex/launcher-gpu-cli` -> clean

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -214,6 +214,78 @@ extract_dmg() {
     echo "$app_dir"
 }
 
+# ---- Detect Electron version from DMG ----
+sanitize_electron_version() {
+    local value="$1"
+    value="${value#v}"
+    value="${value#^}"
+    value="${value#~}"
+
+    if [[ "$value" =~ ^[0-9]+(\.[0-9]+){2}([.-][0-9A-Za-z]+)*$ ]]; then
+        echo "$value"
+        return 0
+    fi
+
+    return 1
+}
+
+detect_electron_version() {
+    local app_dir="$1"
+    local plist_file="$app_dir/Contents/Frameworks/Electron Framework.framework/Versions/A/Resources/Info.plist"
+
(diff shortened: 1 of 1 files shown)


ilysenko:

The hardcoded `ELECTRON_VERSION="41.3.0"` has been creating friction and `detect_electron_version` would close that. A few things worth nailing down before ready:

- `bash -n install.sh` alone is too thin for a 188-line change of this scope. At minimum a fixture-style smoke test: feed `detect_electron_version` a synthetic upstream metadata file with a known version and assert the parse + `sanitize_electron_version` round-trip.
- Fallback path when detection fails (CDN down, malformed JSON) — does the launcher exit, or fall back to a pinned constant? A pinned fallback keeps fresh checkouts buildable when upstream metadata is unreachable.
- `CODEX_GPU_MODE=nvidia` exports `__GLX_VENDOR_LIBRARY_NAME=nvidia`. Worth a guard that `nvidia-smi` is reachable before forcing it; otherwise users without the proprietary stack on a "nvidia" pick get a worse experience than `auto`.

Sw-render diagnostics via `nvidia-smi --query-gpu` is a nice add for troubleshooting tickets.

AmeerJ97:
Closing this for now to reduce PR noise. The launcher runtime work is better re-cut later as a consolidated, cleaner follow-up after the patcher PR is reviewed.

PR #53, opened 2026-04-29
Make Linux UI patching fail-soft and testable

AmeerJ97:
## Summary
- Make Linux ASAR UI patches apply independently so missing icon assets do not block non-icon Linux fixes.
- Make the Linux file-manager patch less dependent on exact minified local variable names.
- Preserve Linux icon behavior by keeping the `ready-to-show` `setIcon()` patch independent from the BrowserWindow `icon:` option patch.
- Add Node and smoke coverage for file-manager patching, Linux menu hiding, icon/setIcon handling, tray support, single-instance behavior, idempotency, and missing-icon fail-soft behavior.

## Validation
- `node --test scripts/patch-linux-window-ui.test.js` -> 8 passed, 0 failed
- `bash tests/scripts_smoke.sh` -> All script smoke tests passed
- `git merge-tree upstream/main HEAD` -> clean


ilysenko:

Cleanest of the three drafts — the independence refactor is the right direction (we've seen needles miss on every upstream Codex update, and one missing icon shouldn't block the rest). The new `patch-linux-window-ui.test.js` with 4 tests is a great start.

Before ready, I'd extend test coverage to match the patcher's full surface so we don't regress on PR #49's tray work or PR #38's setIcon path:

- tray guard (`process.platform!==win32&&process.platform!==darwin`)
- single-instance lock (`requestSingleInstanceLock`)
- idempotency: every test should run the patcher twice and assert the second pass adds nothing (we already have one such assertion via `assert_occurrence_count` in `tests/scripts_smoke.sh`, would be nice in Node tests too)

`findIconAsset` returning `null` propagating through `applyLinuxWindowOptionsPatch` / `applyLinuxSetIconPatch` is the correct pattern. Just confirm callers downstream of `patchMainBundleSource` don't assume `iconAsset != null`.

AmeerJ97:
Thanks again. I pushed a cleaned-up update to this PR so the patcher change can stand on its own.

This update addresses the requested coverage: tray guard, single-instance lock, idempotency via Node tests, and the `iconAsset == null` path through `patchMainBundleSource`. While tightening it, I also fixed a `setIcon()` regression where the BrowserWindow `icon:` option could suppress the `ready-to-show` `setIcon()` path.

The PR body now has the updated validation: `node --test scripts/patch-linux-window-ui.test.js` is 8/8 passing, and `bash tests/scripts_smoke.sh` passes. I’ll keep the launcher/updater packaging work out of this PR and bring it back separately after this one settles.

PR #57, opened 2026-04-29
Detect Electron version from app metadata

AmeerJ97:
## Summary
- Detect the Electron runtime version from the extracted Codex.app framework metadata before rebuilding native modules and downloading the Linux Electron runtime.
- Keep the pinned `41.3.0` fallback when metadata is missing or malformed, so fresh checkouts remain buildable.
- Add smoke coverage for valid app metadata and malformed metadata fallback behavior.

## Validation
- `bash -n install.sh`
- `bash tests/scripts_smoke.sh` -> All script smoke tests passed
- `git merge-tree upstream/main HEAD` -> clean

## Notes
This is a focused re-cut of the Electron-version part of the earlier launcher draft. GPU-mode and CLI-path changes are intentionally left out for separate follow-ups.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -214,6 +214,66 @@ extract_dmg() {
     echo "$app_dir"
 }
 
+# ---- Detect Electron version from DMG ----
+sanitize_electron_version() {
+    local value="$1"
+    value="${value#v}"
+    value="${value#^}"
+    value="${value#~}"
+
+    if [[ "$value" =~ ^[0-9]+(\.[0-9]+){2}([.-][0-9A-Za-z]+)*$ ]]; then
+        echo "$value"
+        return 0
+    fi
+
+    return 1
+}
+
+detect_electron_version() {
+    local app_dir="$1"
+    local detected=""
+    local detected_version=""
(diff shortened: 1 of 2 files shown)


ilysenko:
Clean follow-up — exactly the focused re-cut I was hoping for. Smoke coverage for both the happy path and malformed-metadata fallback is solid, the `CODEX_INSTALLER_SOURCE_ONLY` source-only guard is a nice pattern for unit-style testing of bash functions, and the two-layer detection (plist → asar package.json → pinned fallback) makes the failure mode robust.

One small nit, follow-up territory at most:

- The plist branch reads `CFBundleVersion`, which on Electron frameworks is sometimes a 4-component build identifier (e.g. `41.3.0.4`) rather than the public release version. The current regex accepts that, but the resulting download URL `electron-v41.3.0.4-linux-...zip` doesn't exist on GitHub releases. `CFBundleShortVersionString` is the public marketing version and matches the GitHub release tag; preferring it (with `CFBundleVersion` as a backup) would make the plist path more reliable on real-world DMGs. The asar package.json fallback already saves us if this misses, so not blocking.

Approving and merging.

PR #59, opened 2026-04-30
Fix apt Node dependency bootstrap

tjdonley:
## Summary
- Try a compatible apt `nodejs`/`npm` candidate first, then bootstrap NodeSource Node.js 22 when apt would install an unsupported Node line.
- Add Node.js version floors to Debian, RPM, and pacman package metadata while preserving upstream package dependency additions.
- Document the apt behavior and add an Ubuntu/Debian CI matrix for the dependency bootstrap path.

## Testing
- `bash -n install.sh scripts/install-deps.sh scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh tests/scripts_smoke.sh`
- `./tests/scripts_smoke.sh`
- `podman run` regression checks on `ubuntu:24.04`, `ubuntu:22.04`, and `debian:12` confirmed Node 22.22.2 with npm/npx and `./install.sh /tmp/nope.dmg` reaching `Provided DMG not found`
- Temporary `.deb` build confirmed `nodejs (>= 20)` plus upstream polkit dependencies
- Temporary RPM build confirmed `nodejs >= 20`, `npm`, and `polkit` requirements

### DIFF
--- a/.github/workflows/install-deps.yml
+++ b/.github/workflows/install-deps.yml
@@ -1,0 +1,61 @@
+name: Install Dependencies
+
+on:
+  pull_request:
+    paths:
+      - install.sh
+      - scripts/install-deps.sh
+      - packaging/linux/control
+      - .github/workflows/install-deps.yml
+  push:
+    branches:
+      - main
+    paths:
+      - install.sh
+      - scripts/install-deps.sh
+      - packaging/linux/control
+      - .github/workflows/install-deps.yml
+
+jobs:
+  apt-node-bootstrap:
+    name: ${{ matrix.image }}
+    runs-on: ubuntu-latest
+    container:
+      image: ${{ matrix.image }}
+      options: --user root
+    strategy:
+      fail-fast: false
+      matrix:
+        image:
+          - docker.io/library/ubuntu:22.04
+          - docker.io/library/ubuntu:24.04
+          - docker.io/library/debian:12
+    env:
+      DEBIAN_FRONTEND: noninteractive
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Bootstrap sudo
+        run: |
+          apt-get update -qq >/dev/null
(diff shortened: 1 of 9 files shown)


ilysenko:
Excellent first contribution. This closes a long-standing pain point — `apt install nodejs` on stock Ubuntu 22.04/24.04 and Debian 12 silently delivers Node 18, which then surfaces as a confusing native-module / install.sh failure two steps later. The fix is exactly the right shape: try the distro candidate first, fall back to NodeSource only when needed, and gate the whole flow behind `has_compatible_nodejs` so reruns are idempotent.

A few things specifically worth calling out:

- **Modern apt practices** — keyring under `/etc/apt/keyrings/` with `signed-by=` rather than the deprecated `apt-key`. Architecture validation (`amd64|arm64|armhf`) cleanly errors on unsupported targets instead of producing a broken install.
- **Package metadata floors are the right call.** Bumping `nodejs` to `>=20` in DEB/RPM/pacman makes the runtime requirement explicit at install time rather than letting it surface during `make build-app`. Same shape as the polkit dep tightening from #50, intentionally hard-stopping vanilla 22.04 setups until the user runs `install-deps.sh`.
- **CI matrix is the first-class regression gate this project was missing.** Containerized Ubuntu 22.04 / 24.04 / Debian 12 runs with Node major + npm/npx verification will catch a future NodeSource layout change or distro repo regression instantly. Nice.
- **`NODEJS_MAJOR` override + `validate_nodejs_major` floor at 22** future-proofs the bootstrap when the next NodeSource line drops; the explicit error beats a confusing apt failure.

One small thing for follow-up territory, not blocking: `apt-cache policy` parsing assumes English output (`Candidate:`). On non-English locales the helper would fall through to NodeSource — still correct behavior, just slightly noisier than necessary. Setting `LC_ALL=C` around the call would make it locale-agnostic.

Merging — thanks for the careful work.

PR #54, opened 2026-04-29
v0.6.0: Enhance installation process

Leay15:
## Summary

This pull request introduces support for GUI-based prompts to install the Codex CLI when it is missing, improves the preflight and installation flow for both terminal and GUI launches, and enhances the overall developer and user experience with more robust CI and packaging.

Key highlights include a new GitHub Actions CI workflow, improved CLI preflight logic with GUI dialog helpers (`kdialog`/`zenity`), and updated packaging to recommend or install these helpers. Documentation and tests have been updated accordingly.

## GUI Prompt and CLI Installation Improvements

- Added a GUI-based prompt flow for installing the Codex CLI using `codex-update-manager`. It prioritizes `kdialog` on KDE/Plasma, then `zenity`, and falls back to a desktop notification if no dialog helper is available. Terminal launches continue to prompt inline.
- The CLI preflight now leverages cached local CLI state for faster startup and performs heavier checks only when necessary.

## Packaging and Dependency Changes

- Updated Debian, RPM, and pacman packaging to recommend or mark GUI dialog helpers (`zenity`, `kdialog`) as optional dependencies.
- `scripts/install-deps.sh` now detects the desktop environment and automatically installs the appropriate dialog helper.

## Continuous Integration Enhancements

- Introduced a comprehensive GitHub Actions CI workflow that runs Rust checks, script smoke tests, and validates Debian, RPM, and pacman package builds, including job summaries.

## Documentation and Versioning Updates

- Updated documentation and changelogs to reflect the new GUI prompt flow, CLI preflight improvements, and the current crate version (`0.6.0`).

## Testing

- Added and updated smoke tests to verify the presence of the new prompt logic and dialog helper references in the codebase (`tests/scripts_smoke.sh`).

## Fix

- Fixes a `better-sqlite3` regression caused by an Electron update in PR #49.

PR #60, opened 2026-04-30
Add Linux Computer Use support

avifenesh:
## Summary

- add a Linux `computer-use` bundled plugin backed by a native Rust MCP server
- stage/build the backend from `install.sh` and package source/templates for Debian/RPM/pacman/update-builder flows
- patch the Electron bundle so Computer Use is enabled on Linux and installed when missing
- add side-by-side lab app support (`make build-dev-app`) so Computer Use can be tested without touching the default app state

## Linux Computer Use backend

The new `codex-computer-use-linux` backend supports:

- `doctor` and `setup_accessibility` readiness/setup tools
- AT-SPI app listing and accessibility-tree snapshots
- screenshot capture through GNOME Shell when allowed, falling back to XDG Desktop Portal
- coordinate and element-index click/scroll plus key/text/drag input through `ydotool`

Native AT-SPI action/value mutation remains the next layer; current element-index click/scroll resolves through cached node bounds from the latest app state.

## Duplicate PR check

I checked open upstream PRs before opening this. The only open PR was #54 (`v0.5.0: Enhance installation process`), which focuses on installer UX, CLI preflight, CI, and packaging helper changes. It does not add Linux Computer Use or the bundled Computer Use backend.

## Validation

- `cargo fmt -p codex-computer-use-linux`
- `bash -n install.sh tests/scripts_smoke.sh scripts/build-deb.sh scripts/build-pacman.sh scripts/lib/package-common.sh`
- `git diff --check`

Live lab validation on Ubuntu/GNOME:

- `make build-dev-app`
- launched `./bin/codex-cua-lab` side-by-side on port 5176
- launcher log changed from `bundled_plugin_install_skipped_missing pluginName=computer-use` to `bundled_plugin_install_requested pluginName=computer-use reason=missing`
- installed cache backend `doctor` reported no readiness blockers
- installed cache backend `screenshot` captured via `xdg-desktop-portal`
- installed cache backend `state Codex` returned AT-SPI Electron frame nodes with bounds


avifenesh:
Rebased on current upstream main (97686c7) and resolved conflicts in install.sh plus scripts/patch-linux-window-ui.js. Re-ran validation after the rebase: cargo check/test for codex-computer-use-linux, node patch tests, script smoke tests, bash -n, and git diff --check. Rebuilt and relaunched the codex-cua-lab side-by-side app; Computer Use remains staged/installed, doctor reports no blockers, screenshot capture works via xdg-desktop-portal, and AT-SPI state returns Codex window nodes.

ilysenko:
@avifenesh — first off, thank you for the substantial work here. A native Rust MCP backend for Computer Use on Linux with AT-SPI tree, GNOME Shell + portal screenshot, and ydotool input synthesis is real engineering and the fail-soft `applyLinuxComputerUsePluginGatePatch` shows you read the patcher conventions. The lab/`build-dev-app` side-by-side path is a very thoughtful design choice — it means we can land an early version without risking the default install state.

That said, this PR is meaningfully bigger in *scope* than the project has accepted before, so I'd like to open a short discussion before merging rather than rubber-stamping it. Pinging everyone in `CONTRIBUTORS.md` and recent active reviewers — would appreciate your read on whether this fits the project's direction:

@Leay15 @pradigmaz @babafish12 @AmeerJ97 @tjdonley @ProfFlow @gadicc @RFLXN @prichardsondev @monkey-w1n5t0n @GuimelGMC @quantmind-br @moabualruz @fndov @khushveer007 @DhanushSantosh @LAP87 @lalalandau

The shape question: this project so far has been "make the macOS Codex DMG runnable on Linux" — patch ASAR, swap Electron, package, auto-update. This PR steps further into "implement a Linux-native version of an upstream feature that uses macOS-specific OS APIs". That's a different kind of investment, with its own ongoing maintenance surface (libatspi, ydotool, xdg-desktop-portal, GNOME Shell DBus). It's a defensible direction, but worth deciding deliberately rather than by accident.

Specific questions for @avifenesh that would help everyone evaluate:

1. **Security / consent model.** Computer Use is "the LLM clicks on the user's screen". On macOS upstream this is mediated by OS-level accessibility permissions plus an explicit per-app allowlist. How does the Linux backend decide which apps Codex can read accessibility metadata from and inject input into? Is there an allowlist persisted somewhere, an explicit consent prompt before the first action against an app, or is the gating purely "Codex Electron asks the MCP server"? A short paragraph in the PR description on the threat model would help reviewers a lot.

2. **System-dependency additions.** The runtime now needs `ydotool` (with `uinput` access via root or the `input` group), libatspi at runtime, and a working screenshot path (GNOME Shell DBus *or* an XDG Desktop Portal implementation reachable from the user session). Should `scripts/install-deps.sh` learn to install these, and what's the desired UX when one of them is missing (warn-only at install, or a doctor-style readiness check that blocks the plugin from registering)? Also worth noting which compositors / DEs you've actually validated against — the PR description mentions Ubuntu/GNOME, but KDE, sway, Hyprland, X11 fallback, etc. behave differently around portals and AT-SPI.

3. **License.** `plugins/openai-bundled/plugins/computer-use/.codex-plugin/plugin.json` declares `"license": "Proprietary"` and `"author.name": "OpenAI"`. The rest of the repo is MIT (per `LICENSE` and the PKGBUILD template). Is the proprietary marker intentional because it mirrors upstream OpenAI plugin conventions, or should the new code (the Rust crate especially) carry the project's MIT license? Worth being explicit so we don't ship a license-mixed package by accident.

Final note: this isn't an "I want to reject this" comment. The technical quality is good and Linux Computer Use is genuinely useful. I just want the merge decision to be deliberate, especially on the scope and security questions. Thanks again for the careful work — looking forward to landing this once we've talked through the above.

ilysenko:
@avifenesh — gave the PR another careful look. Genuinely useful functionality, well-built. Two questions before merge:

1. **Should this be opt-in at install time?** Computer Use injects keystrokes and reads accessibility data from any running app, with no per-app permission model on Linux. Wouldn't it be safer if users had to explicitly turn it on (e.g. `CODEX_INSTALL_COMPUTER_USE=1 ./install.sh` or `--with-computer-use`) rather than getting it by default? Putting the enable decision in the user's hands feels right for a feature with this surface area.

2. **License.** `plugin.json` declares `"license": "Proprietary"` while the rest of the repo is MIT. Intentional, or should the new crate ship under MIT?

Otherwise good to merge. Thanks for the work.

ilysenko:
Quick follow-up — checked locally why the plugin doesn't appear in the UI even with the patcher applied. The platform gate is now passing, but `e.computerUse` itself is gated by a Statsig flag (`Qf('1506311413')` in the webview bundle). That's per-account rollout, same mechanism as `gpt-5.5` and other recent OpenAI feature drops — purely cohort-based, independent of platform or subscription tier. Your account is in the cohort, mine isn't, and most users probably aren't either.

Means in practice the PR works for the small subset of accounts that received the flag and is invisible to everyone else.

Curious what you think the right move is. Two paths I see:

1. **Land as-is** — feature ships dormant for most users, lights up automatically when OpenAI rolls the flag wider. Lowest friction, but value-on-merge is small until that happens.

2. **Add a second ASAR patch** that forces `Qf('1506311413')` true locally, so Linux Computer Use becomes available to all users of this build. Mirrors how we already patch the platform gate. But it's a deliberate Statsig bypass of OpenAI's rollout, and I'm a bit nervous about the optics — could be read as a ToS-y thing and put the project in OpenAI's crosshairs.

avifenesh:
@ilysenko Give me a little time to iterate. My little concern is the PR size creep.
I can split it into different tools and merge it as a separate PR.
Regarding whether the license should be MIT or proprietary, I genuinely don't know, I will look into it. It is not a deliberate choice.
Whether it will be a problem with OpenAI is a question of where the line passes. I will be happy to hear opinions.
I will respect any decision, I did it since I wanted to have computer use, and I believe others do as well, but whether I should keep it to myself or give it back is passed to the maintainer field after I decided to open it.

avifenesh:
@ilysenko since the whole Linux scope is not OpenAI act, and since internal workers of OpenAI share on X how they are rebuilding Codex from source to have many apps running simultaneously, I believe that at least ship-as-is is safe. Whether bypassing the flag is ok is a real question.

avifenesh:
But computer use was released a long time ago and is available in both macOS and Windows. It didn't show up in my view when I built your code, it did show up only after adding the engine code.

avifenesh:
@ilysenko I actually was iterating and theres few bugs with the click coordination. Will create a follow-up soon.

PR #61, opened 2026-04-30
[codex] Fix Linux Computer Use coordinate input

avifenesh:
## Summary
- cache screenshot dimensions from `get_app_state` captures and convert screenshot pixel coordinates into ydotool absolute coordinates for click, scroll target, and drag input
- add a `--` separator for wheel scroll deltas so negative values are not parsed as options
- add regression tests for PNG dimension extraction, coordinate normalization, and negative wheel deltas

## Why
During a fresh app restart test on Linux/GNOME, the Computer Use MCP transport loaded correctly and `doctor` was green, but direct input exposed two concrete runtime issues: screenshot pixels needed conversion before ydotool absolute input, and `mousemove --wheel 0 -3` failed because the negative delta was parsed as an option.

## Validation
- `cargo fmt -p codex-computer-use-linux`
- `git diff --check`
- live app smoke: after restart, app-exposed `mcp__computer_use__` tools loaded, `doctor` reported screenshot/AT-SPI/ydotool ready, click at the visible Codex prompt typed `cu2`, and the prompt was cleared without submitting

### DIFF
--- a/computer-use-linux/src/screenshot.rs
+++ b/computer-use-linux/src/screenshot.rs
@@ -20,6 +20,8 @@ pub struct ScreenshotCapture {
     pub mime_type: String,
     pub data_url: String,
     pub source: String,
+    pub width: u32,
+    pub height: u32,
 }
 
 pub async fn capture_screenshot() -> Result<ScreenshotCapture> {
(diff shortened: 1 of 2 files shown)


avifenesh:
@ilysenko please let me know if you want anything in addition.

PR #63, opened 2026-04-30
Improve Electron mirrors and ydotool diagnostics

baktun95827:
## Summary

  - Use `artifacts.electronjs.org` as the default Electron headers endpoint for `@electron/rebuild`
  - Add `ELECTRON_MIRROR` support for downloading the Linux Electron runtime from a mirror
  - Improve Linux Computer Use diagnostics by detecting an existing `/tmp/.ydotool_socket`
  - Document Ubuntu `ydotoold` packaging behavior and Electron mirror troubleshooting

  ## Validation

  - `bash -n install.sh`
  - `git diff --check`

  ## Notes

  This fixes a rebuild failure where `@electron/rebuild` times out while trying to reach `www.electronjs.org`, and improves `ydotoold` readiness detection on Ubuntu systems where the daemon listens on `/tmp/.ydotool_socket`.


ilysenko:
Thanks for the fix — the `artifacts.electronjs.org` switch and the `/tmp/.ydotool_socket` detection both close real holes I've seen reported, so good catch. PR is conflicting on `install.sh` after recent merges (#54 reshaped that file). Could you do a rebase on current `main` when you get a chance?

Issue #74, opened 2026-05-01
2026.05.01.113446 does not launch


dcelasun:
With this version, `codex-desktop` command spawns `codex-update-manager daemon` and exits. No output.

There were skipped patches during the build, that might be related:
```
[INFO] Patching Linux window and shell behavior...
WARN: Could not find Linux settings state file marker — skipping settings persistence patch
WARN: Could not find browser annotation screenshot element highlight — skipping screenshot anchor patch
WARN: Could not find browser annotation screenshot markers — skipping screenshot marker patch
WARN: Could not find code theme bundle in /tmp/tmp.cPaf6TnX4z/app-extracted/webview/assets — skipping translucent sidebar default patch
Patched Linux window, shell, and appearance behavior: {
  target: '/tmp/tmp.cPaf6TnX4z/app-extracted/.vite/build/main-SLemWUtC.js',
  mainBundle: 'main-SLemWUtC.js',
  iconAsset: 'app-D0g8sCle.png',
  desktopName: 'codex-desktop.desktop'
}
[INFO] Repacking app.asar...
[INFO] app.asar patched
[INFO] Downloading Electron v41.2.0 for Linux...
[INFO] Using cached Electron archive: /home/dcelasun/.cache/codex-desktop/electron/electron-v41.2.0-linux-x64.zip
[INFO] Electron ready
[INFO] Webview files copied
[INFO] app.asar installed
```
I saw #68, but warning instead of throwing is a mistake IMHO. It lets a package built without knowing if it'll run at all.

Leay15:
@dcelasun could you please give us more info?

We're trying to catch this issue due to some patches that are performed by regex, that might be causing this issue but being honest I'm not quite sure

`codex-update-manager status` will show you if it is app error or cli error giving you more info about where to look

dcelasun:
`codex-update-manager status` was very useful. Turns out Codex CLI isn't included in the package and must be installed separately. Can you add `openai-codex` as a dependency on Arch? It starts correctly with that installed.

Leay15:
I don’t think making `@openai/codex` a hard package dependency is the right fix here.

At the moment the app expects the CLI to be a runtime dependency managed via `npm`, and the launcher/updater logic is built around that. If we turn it into a direct package dependency, we’d probably create more edge cases than we solve:

we’d no longer be using the same source of truth as the current CLI version checks
[...]

PR #76, opened 2026-05-01
Fix/v0.6.2

### LINKED ISSUE #67: System tray "Quit" option can leave Codex Desktop running
Selecting **Quit** from the Codex Desktop tray menu does not reliably close the app after interacting with any threads.

The tray menu item itself is wired correctly: it calls Electron `app.quit()`. The issue appears to happen later in the app quit lifecycle.

## Observed behavior

After selecting tray **Quit**:

- the UI closes
- shutdown starts
- logs show app-server transport cleanup beginning
- the Electron process can remain alive afterward

Example log shape:

```text
[AppServerConnection] Stopping app-server transport connectionId=1 transport=stdio
[AppServerConnection] app_server_connection.state_changed cause=stop_process ...
[worker-manager] Worker exited code=1
```

But ps may still show `codex-desktop-linux/codex-app/electron` and it will remain in the taskbar (or potentially running in the background but I was unable to reproduce this).

## System observed on

Operating System: Bazzite 43
KDE Plasma Version: 6.6.4
KDE Frameworks Version: 6.25.0
Qt Version: 6.10.3
Kernel Version: 6.17.7-ba29.fc43.x86_64 (64-bit)
Graphics Platform: Wayland

Leay15:
This pull request introduces several improvements and fixes to the Codex Desktop application, focusing on more robust handling of missing CLI dependencies, improved webview server management, and enhanced resilience and guard logic for Linux tray and single-instance behavior. It also updates documentation and versioning to reflect these changes.

**Dependency handling and user experience improvements:** 

* Missing Codex CLI is now explicitly reported as `cli_status: NotInstalled` in updater status and persisted state, rather than overloading the `Unknown` state. Automatic installation of a missing CLI is now strictly launcher-scoped; the daemon and `codex-update-manager status` only report and notify, but do not attempt installation themselves. First approach for #74 
* Launcher error messages now clearly distinguish between a CLI that is missing and an automatic installation that failed, providing clearer instructions for manual recovery. 

**Webview server management:**

* The launcher now detects and terminates "stale" webview server processes—those serving from deleted directories—to prevent conflicts and ensure a clean startup. New helper functions (`pid_has_webview_server_cmdline`, `pid_is_stale_webview_server`, etc.) and logic were added to manage this. 

**Linux tray and single-instance guard logic:** Closes #67 

* The Linux tray and single-instance handling patches now include explicit quit-state guards. This prevents tray menu actions and second-instance handlers from running during app shutdown, improving reliability and preventing race conditions. New patch logic and supporting helpers were added in `scripts/patch-linux-window-ui.js`. 
* Improved detection of disposable variables for patch injection, making the patching process more robust to upstream code changes.

**Documentation and versioning:**

* Updated `CHANGELOG.md`, `README.md`, and `AGENTS.md` to document the above changes and bump the updater crate version to `0.6.2`.


ilysenko:
Thanks for this — landing a real fix for #67 and #74, plus the stale-webview cleanup, all in one PR is great. CI is green and I'm happy to merge as-is, but I want to flag two latent issues that will likely surface later, so we don't get caught off guard:

**1. `codex_cli::tests::reconcile_if_present_upgrades_outdated_cli` is environment-dependent.** It mocks `npm` by writing a fixture binary to a temp dir and prepending it to `PATH`, but it doesn't `env_clear()` first. On the CI runner this is harmless because `rust-and-smoke` doesn't run `setup-node`, so there's no real `npm` in `PATH`. On any contributor's dev machine with `nvm` installed, `which npm` resolves to the real one before the fixture path, and the test fails:

```
assertion `left == right` failed
  left: "/home/user/.nvm/versions/node/v22.22.0/bin/npm"
 right: "/tmp/.tmp7coxaa/bin/npm"
```

Fix is one line — `Command::env_clear().env("PATH", &fake_bin_dir)` (or whatever the equivalent is in your test harness's `Command` builder).

**2. Four `clippy::uninlined_format_args` errors that the current CI clippy doesn't fire.** On a newer rustc (1.80+ where this lint moved from `pedantic` into `style` and became warn-by-default), `cargo clippy --workspace --all-targets -- -D warnings` errors out on:

- `updater/src/codex_cli.rs:82`, `:129`, `:193`, `:244`

The CI runner's bundled rustc is currently old enough that the lint is still allow-by-default, so this passes silently — but `ubuntu-latest`'s toolchain rolls forward periodically, and when it does this PR's commits will start failing CI on `main`. Trivial fix:

```
cargo clippy --fix --allow-dirty -p codex-update-manager
```

Either fix in this PR, or open a follow-up — both fine. Just want it on record so neither of us is surprised when CI flips red on an unrelated change later.

ilysenko:
Hey @Leay15 — pushed two commits on top of your branch covering the items I'd flagged earlier in this PR (the env-leaking tests + the `clippy::uninlined_format_args` warnings), and resolved the merge conflict against `main`. Locally `cargo test -p codex-update-manager` is 85/85 and `cargo clippy --all-targets -- -D warnings` is clean.

Take a look when you have a minute. Totally fine if you'd rather revert my commits and do this in your own style — happy to defer.

Leay15:
@ilysenko awesome, everything looks great, go ahead no problem by my side

PR #77, opened 2026-05-01
fix(computer-use): enable Linux UI gates and Wayland pointer input

avifenesh:
## Summary
- fix Linux pointer input by preferring the Wayland remote desktop portal for click, scroll, and drag
- keep raw pixel coordinates for ydotool fallback instead of remapping screenshot pixels into 0..65535 absolute space
- enable the remaining Linux Computer Use desktop/UI/install gates in the app patcher

## What changed
- add `computer-use-linux/src/remote_desktop.rs` for `org.freedesktop.portal.RemoteDesktop` + `ScreenCast` pointer sessions
- update `click`, `scroll`, and `drag` in `computer-use-linux/src/server.rs` to use the portal on Wayland and fall back to ydotool otherwise
- add a small inter-command delay for ydotool pointer sequences
- bump the bundled plugin version to `0.1.1-linux-alpha1`
- patch the Electron app bundles so Linux enables Computer Use desktop features and shows the UI/install flow

## Validation
- `cargo check --manifest-path computer-use-linux/Cargo.toml`
- `cargo test --manifest-path computer-use-linux/Cargo.toml` (`44/44`)
- `cargo build --release --manifest-path computer-use-linux/Cargo.toml`
- `node --test scripts/patch-linux-window-ui.test.js` (`31/31`)

## Live verification
Tested locally on GNOME 50.1 / Wayland:
- the previous ydotool path could jump into the GNOME hot corner / overview because the fallback was using remapped absolute coordinates
- with the portal path warm, `click`, `scroll`, and `drag` produced real mouse event bytes inside a local Ghostty terminal instead of landing in the top-left overview corner

### DIFF
--- a/computer-use-linux/src/main.rs
+++ b/computer-use-linux/src/main.rs
@@ -1,6 +1,7 @@
 mod atspi_tree;
 mod diagnostics;
 mod gnome_extension;
+mod remote_desktop;
 mod screenshot;
 mod server;
 mod terminal;
--- a/computer-use-linux/src/remote_desktop.rs
+++ b/computer-use-linux/src/remote_desktop.rs
@@ -1,0 +1,543 @@
+use crate::diagnostics::hydrate_session_bus_env;
+use anyhow::{bail, Context, Result};
+use futures_util::StreamExt;
+use std::{collections::HashMap, time::Duration};
+use zbus::{
+    proxy::SignalStream,
+    zvariant::{OwnedObjectPath, OwnedValue, Value},
+    Connection, Proxy,
+};
+
+const PORTAL_DESKTOP_SERVICE: &str = "org.freedesktop.portal.Desktop";
+const PORTAL_DESKTOP_PATH: &str = "/org/freedesktop/portal/desktop";
+const PORTAL_REMOTE_DESKTOP_INTERFACE: &str = "org.freedesktop.portal.RemoteDesktop";
+const PORTAL_SCREENCAST_INTERFACE: &str = "org.freedesktop.portal.ScreenCast";
+const PORTAL_REQUEST_INTERFACE: &str = "org.freedesktop.portal.Request";
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(120);
+
+const DEVICE_POINTER: u32 = 2;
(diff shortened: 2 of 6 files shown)


ilysenko:
Thanks for the correction (re: #78) — you're right, the per-platform vs. per-account framing in the README is wrong, and that whole caveat needs to be rewritten based on what you actually observed (Mac → Win → Linux progression on a single account, not a Statsig per-account rollout).

Re your PRs — please keep them coming, I'll merge them. Linux-control-surface patches like the UI gate work in this PR sit in exactly the same category as the existing `applyLinuxComputerUsePluginGatePatch` we already do in the main bundle (`darwin → darwin || linux`); doing the same in the webview asset is a natural extension, not a new risk surface. Same goes for whatever else you have lined up on the input/control side.

The one place I'd want us to be deliberate is **feature-flag-bypass-style patches** if any of those come up in the future — anything that forces a Statsig flag to `true` regardless of what the server returns to the client (think: forcing `computerUse: true` per-user, or unlocking a model rollout). That's a different category than "make the Linux platform path work," and shipping it on by default in our installer feels like territory where the project gets DMCA'd or otherwise nuked. For that kind of patch I'd suggest one of:

1. **Opt-in env var** (`CODEX_LINUX_FORCE_X=1`) — patch ships in code but doesn't apply by default; user has to set the flag explicitly.
2. **Standalone PR we don't merge** — leaves it as a public reference, anyone who wants it builds from your fork.
3. **Separate branch** (`experimental/feature-flags`) — documented in README as "if you want to take the risk, build from here."

But that's a bridge to cross when we get there. Everything you've described in #78 is in the safe lane — ship it.

One small heads-up on this PR: the `Rust and Smoke Tests` job is red on two `clippy::unnecessary_cast` errors in `computer-use-linux/src/server.rs:284` and `:306` (`as u32` on a value that's already `u32`). Drop those two casts and CI will go green. Once that's fixed I'll merge.

P.S. After this lands I'll fix the misleading Statsig caveat in the README based on what your patch actually unlocks.

avifenesh:
@ilysenko Gotcha.
Fixed.

PR #79, opened 2026-05-01
feat(patcher): gate the three Computer Use UI patches behind opt-in flag

ilysenko:
## Summary

The three patcher functions added in #77 — `applyLinuxComputerUseFeaturePatch`, `applyLinuxComputerUseRendererAvailabilityPatch`, `applyLinuxComputerUseInstallFlowPatch` — reach beyond pure platform-port glue. The install-flow patch in particular falls back to `navigator.userAgent.includes(\`Linux\`)` as an OR-clause against the upstream `computer_use` Statsig flag — that is closer to a feature-flag override than the long-standing `darwin → darwin || linux` plugin manifest gate, which we keep on by default.

To keep the project defensible if OpenAI ever objects, the three new patches are now only applied when the user explicitly opts in.

## Opt-in surfaces

| Path | How user opts in |
|------|------------------|
| `make build-app` (manual) | `CODEX_LINUX_ENABLE_COMPUTER_USE_UI=1 make build-app` |
| `make package` → `make install` (Debian/RPM/Pacman manual flow) | Same env var on build, packaging just wraps `codex-app/` |
| Auto-updater rebuild via `codex-update-manager` systemd user service | Persist `{"codex-linux-computer-use-ui-enabled": true}` in `~/.config/codex-desktop/settings.json` once. The systemd service does not inherit interactive shell env, so the file-based flag is the right surface for it. |

The new helper `isComputerUseUiEnabled()` returns `true` when **either** surface is set; orchestrator gating short-circuits the three patcher calls otherwise. The bundled-plugin manifest gate (`applyLinuxComputerUsePluginGatePatch`) stays unconditional — that's pure platform-port glue that has shipped on by default since the project's first release.

## What changes for users on 0.6.1

If you built `0.6.1` and Computer Use controls were showing up in the Codex Desktop UI, the next auto-updater rebuild on this version will hide them again unless you set the persisted settings.json flag. The MCP backend (plugin manifest gate) still registers; only the UI controls are gated. Quick fix:

```bash
mkdir -p ~/.config/codex-desktop
python3 -c 'import json,os,pathlib;p=pathlib.Path(os.path.expanduser(\"~/.config/codex-desktop/settings.json\"));d=json.loads(p.read_text())if p.exists()else{};d[\"codex-linux-computer-use-ui-enabled\"]=True;p.write_text(json.dumps(d,indent=2)+\"\\n\")'
```

## Tests

- 39 unit tests pass (8 new): helper across env var, settings flag, malformed JSON, missing files; orchestrator branches.
- New smoke test `test_linux_computer_use_ui_opt_in_smoke` exercises all three branches end-to-end against fake extracted ASAR fixtures (default off, env-var on, settings.json on).
- Existing per-function unit tests for the three patchers still apply unchanged — the patcher functions themselves did not move.

## Test plan

- [x] `node --test scripts/patch-linux-window-ui.test.js` (39 passed)
- [x] `bash tests/scripts_smoke.sh` (all branches including new opt-in smoke pass)
- [x] `cargo build --release -p codex-update-manager`
- [x] `cargo test -p codex-update-manager` (81 passed)
- [ ] Manual: `make build-app` without opt-in, confirm Computer Use UI hidden in Codex Desktop
- [ ] Manual: `CODEX_LINUX_ENABLE_COMPUTER_USE_UI=1 make build-app`, confirm Computer Use UI visible
- [ ] Manual: drop settings.json flag, run `make build-app` without env var, confirm UI visible (auto-updater path)

## Notes

- `codex-update-manager` bumped to `0.6.2`. README, AGENTS.md, CHANGELOG updated. The "How the platform gating works" README section now describes both opt-in surfaces explicitly.
- Pre-existing `clippy::uninlined_format_args` errors in `updater/src/codex_cli.rs` are unrelated to this PR (they trip on rustc 1.80+; CI's older rustc still passes). Tracked separately.


ilysenko:
@avifenesh — heads up, this PR walks back the default-on behaviour of your three UI gate patches from #77. MCP backend stays default-on (plugin manifest gate unchanged), but `featureGates`, renderer availability, and the install-flow `userAgent.includes("Linux")` override are now opt-in (env var or settings.json flag).

The install-flow patch sits in OR with the upstream `computer_use` Statsig flag, and I'd rather have that be an explicit user choice than a default. None of your patcher functions changed — only the orchestrator now consults `isComputerUseUiEnabled()` before invoking those three.

What do you think?

avifenesh:
> @avifenesh — heads up, this PR walks back the default-on behaviour of your three UI gate patches from #77. MCP backend stays default-on (plugin manifest gate unchanged), but `featureGates`, renderer availability, and the install-flow `userAgent.includes("Linux")` override are now opt-in (env var or settings.json flag).
> 
> The install-flow patch sits in OR with the upstream `computer_use` Statsig flag, and I'd rather have that be an explicit user choice than a default. None of your patcher functions changed — only the orchestrator now consults `isComputerUseUiEnabled()` before invoking those three.
> 
> What do you think?

Absolutely, I also feel more comfortable with it. If there's any chance that it will cause issues for users, it should be a clear decision done intentionally. It's a very useful feature that the project gives, but it's user-decision gated. It does not require great effort to get it running but does require reading the heads-up.

PR #86, opened 2026-05-02
Fix Keybinds settings patch after upstream bundle drift

baktun95827:
## Summary

- Make the Keybinds settings index patch tolerate upstream minified variable name drift
- Match the settings route/icon/nav structures instead of hard-coded `c_e` / `Xge` / `Zge` names
- Add regression coverage for the current upstream bundle shape

## Validation

### DIFF
--- a/scripts/patch-linux-window-ui.js
+++ b/scripts/patch-linux-window-ui.js
@@ -430,39 +430,34 @@ function applyLinuxKeybindOverridesRuntimePatch(currentSource) {
 function applyKeybindsSettingsIndexPatch(currentSource) {
   let patchedSource = currentSource;
 
-  if (
-    !patchedSource.includes("var c_e=") &&
-    !patchedSource.includes("Xge=") &&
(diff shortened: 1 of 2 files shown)


ilysenko:
Hey @baktun95827 — looked at the failing smoke test. The new `routePattern` matches the first `var X={"general-settings":` in the bundle, which is actually `Xge` (the icon map), not `c_e` (the route map). After the patch, `Xge` ends up with a lazy-loader where the icon should be, and `c_e` stays unchanged — so the icon patch then can't find a clean target.

One-line fix: tighten the route pattern to require `(0,X.lazy)` in the value, which is unique to the route map:

```js
const routePattern = /var ([A-Za-z_$][\w$]*)=\{"general-settings":\(0,[A-Za-z_$][\w$]*\.lazy\)/;
```

baktun95827:
Thanks, fixed. I tightened the route matcher to only target the lazy-loaded settings route map and added coverage where the icon map appears before the route map.

Validation:

PR #89, opened 2026-05-02
Fix Browser Use Node runtime selection on Linux

baktun95827:
## Summary
- validate the Browser Use / node_repl Node runtime against the observed 22.22.0+ requirement
- prefer compatible runtimes from packaged resources, the primary runtime cache, system Node, and nvm, then export both `CODEX_BROWSER_USE_NODE_PATH` and `NODE_REPL_NODE_PATH`
- stop hard-depending native packages on distro `nodejs`/`npm`; keep them as suggested/recommended runtime tooling for Browser Use and local rebuilds

## Validation
- `bash -n install.sh scripts/lib/install-helpers.sh scripts/lib/process-detection.sh scripts/lib/dmg.sh scripts/lib/native-modules.sh scripts/lib/asar-patch.sh scripts/lib/webview-install.sh scripts/lib/bundled-plugins.sh scripts/lib/package-common.sh launcher/start.sh.template scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh`
- `./scripts/build-deb.sh`
- `dpkg-deb -I dist/codex-desktop_2026.05.02.115627_amd64.deb`

## Notes
This fixes Linux desktop launches that inherit an older nvm Node path, causing Browser Use to report `browser_use_node_repl_missing` even when a compatible system or primary-runtime Node is available.


ilysenko:
Solid PR, thanks — real bug, launcher logic is clean.

Two small things before I merge:

**1. `Suggests` → `Recommends` for DEB.** In `packaging/linux/control` the new line is `Suggests: nodejs (>= 22.22.0), npm`. On Debian/Ubuntu, `Suggests` is not installed by default, while `Recommends` is — so bare-system users (no nvm, no distro Node) silently lose Browser Use after `apt install`. The RPM spec already uses `Recommends:` here, so making the DEB symmetric is the consistent move. nvm users can still opt out with `--no-install-recommends`.

**2. Make the not-found message actionable.** Right now the launcher prints:
```
Browser Use Node.js 22.22.0+ runtime not found; in-app browser automation may be unavailable.
```
Could you append a one-line install hint? Something like:
```
Install: sudo apt install nodejs npm   (or nvm: https://github.com/nvm-sh/nvm)
```
So users without Node know exactly what to do without digging through docs.

Happy to merge once those two are in.

baktun95827:
Thanks for the review. I addressed both requested changes:

- DEB now uses `Recommends: nodejs (>= 22.22.0), npm` instead of `Suggests`, matching the RPM behavior.
- The launcher warning now includes an install hint, so users without a compatible Node runtime get an actionable next step.

While doing local Browser Use validation, I also found and fixed a couple of real runtime issues in the new Linux `node_repl` path:

- The Linux package now bundles a real `resources/node_repl` ELF wrapper plus `node_repl.mjs`, instead of relying on the macOS DMG resource.
- The runtime handles local Browser Use origin approval itself for localhost/file URLs, avoiding an unsupported elicitation request that closed the MCP transport.
- Stale Browser Use native pipe sockets under `/tmp/codex-browser-use` no longer crash `node_repl`; failed socket connections now reject cleanly so backend discovery can continue.


I tested Browser Use end-to-end on my Linux machine:
  - `node_repl/js` works.
  - `setupAtlasRuntime({ backend: "iab" })` works.
  - A new in-app browser tab opened a local `127.0.0.1` smoke page.
  - DOM snapshot saw the expected title/button.
  - Screenshot capture also worked.

So Browser Use is now working locally with the packaged Linux `node_repl` runtime.

ilysenko:
Thanks for the updates. The two things I asked for are covered now: DEB uses Recommends, and the launcher warning has an install hint. Appreciate you also digging into the Browser Use runtime issues while testing locally.

Before we merge, I’d like to clarify a few parts of the broader runtime change, since this PR is now doing more than Node selection.

1. Is Browser Use on Linux expected to automate external websites, or only local/file URLs for now? I noticed the new node_repl accepts file/local origins but rejects external ones, and the smoke test asserts that https://example.com is rejected. If external website automation is expected, we should adjust that before merging.

2. Could you explain the compatibility story for the new Linux node_repl.mjs? I understand why the macOS binary from the DMG can’t be used directly, but this PR effectively adds a Linux implementation of the Browser Use runtime, so I’d like to understand which upstream node_repl behavior it is matching.

3. Could you tighten the fallback path for existing /opt/.../resources/node_repl? The new wrapper expects node_repl.mjs next to it, so if the bundled build fails and we reuse an older node_repl without copying/checking the sibling .mjs file, we may install a runtime that looks executable but fails at launch.

4. Could you also validate the auto-update path through codex-update-manager? Since this PR adds browser-use-node-repl to the update-builder bundle, I want to make sure a future updater rebuild from a new upstream DMG still produces a package with a working Linux node_repl + node_repl.mjs. This is especially important because the updater runs from a systemd --user service environment, which may not have the same PATH, Node/npm/npx, or compiler availability as an interactive shell.

Your local Browser Use validation is helpful. Could you confirm whether that was done from an installed native package launched through the desktop/GUI entry, or from the repo/dev launch path? I’m asking because this PR mainly affects launcher environment resolution and packaged resources, and GUI launches often have a different PATH/nvm/systemd environment than an interactive shell.

If it’s easy to split, I’d also prefer keeping the Node selection / package dependency fix separate from the new bundled node_repl runtime. The first part is small and clearly safe; the second is valuable, but has a bigger compatibility surface.

ilysenko:
Thanks again for digging into this. We ended up opening a smaller fallback-focused PR here: https://github.com/ilysenko/codex-desktop-linux/pull/94

It restores the old behavior by installing a known-good Linux `node_repl` from the official primary runtime when the DMG only contains macOS binaries, without introducing a custom Linux `node_repl` implementation.

Could you take a quick look and let us know if this covers the failure case you were fixing in this PR?

baktun95827:
Thanks, I took a look at #94 and I think it covers the Browser Use failure case I was hitting.

The concrete failure on my machine was that the DMG-provided `node_repl` was not a Linux ELF, so the Linux package ended up without `/opt/codex-desktop/resources/node_repl`. After that, Browser Use could not expose the `node_repl/js` tool.

#94 fixes that path by installing a verified Linux `node_repl` from the official primary runtime when the DMG resource is not usable. That is a cleaner compatibility story than the custom Linux `node_repl.mjs` implementation in my PR.

I also ran the #94 script smoke tests locally and they passed.

So I’m fine with #94 superseding the bundled/custom `node_repl` runtime part of my PR. The only remaining question is whether you still want any of the smaller Node selection / package dependency changes from #89, or whether #94 is enough for the merge path.

ilysenko:
Closing this in favor of #94 (already merged — installs the official Linux `node_repl` from the upstream `codex-primary-runtime` archive when the DMG ships only a macOS Mach-O binary) and #103 (documents the Node.js 20+ / 22.22+ requirements upfront so users don't hit the "package installs but Browser Use / auto-update silently fails" trap).

Thanks again for the original investigation and the local end-to-end Browser Use validation — that helped us understand the actual failure mode and pick the cleaner path.

Issue #75, opened 2026-05-01
codex crashing


Ziya-Wagih-husaini:
An error has occurred
Codex crashed with the following error:

  (code=null, signal=SIGSEGV).
Most recent error: None
Some things to try:

Check your config.toml for invalid settings
Try updating your IDE settings
Try downloading a different version of the extension
Click reload to restart the Codex extension, or visit our documentation for additional help.
i
I am getting this issue when trying to open the app

ilysenko:
Thanks for the report. Could you share a screenshot of the crash dialog (and any additional context window if it appears) so I can see the exact error layout? The full text from the dialog as you copied it is helpful, but a screenshot will tell me whether this is the Electron renderer crash dialog, the codex CLI crash, or something else.

Also, if you can paste the contents of these logs after the crash, that would speed up triage:

```
~/.cache/codex-desktop/launcher.log
~/.local/state/codex-update-manager/service.log
```

And which distro/version (`/etc/os-release`) plus desktop (GNOME/KDE/sway, X11 vs Wayland).

PR #90, opened 2026-05-02
Improve Electron runtime launch robustness

diegomarzaa:
## Summary
- validate the downloaded Linux Electron runtime after extraction so a wrong runtime fails during build instead of launching broken
- set chrome-sandbox to 4755 both after runtime extraction and during native package staging
- include common npm global install prefixes in launcher CLI discovery for GUI sessions

## Testing
- bash -n scripts/lib/native-modules.sh
- bash -n scripts/lib/package-common.sh
- make build-app && make package
- verified rebuilt .deb contains setuid chrome-sandbox and Electron runtime reports 41.2.0 via ELECTRON_RUN_AS_NODE

Related to #74 and #75.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -106,6 +106,8 @@ Do not assume `codex-app/` is pristine. If behavior differs from `install.sh`, p
   `7z` can return a non-zero status for the `/Applications` symlink inside the DMG. This is currently treated as a warning if a `.app` bundle was still extracted successfully.
 - Launcher and `nvm`:
   GUI launchers often do not inherit the user's shell `PATH`. The generated `start.sh` explicitly searches for `codex`, including common `nvm` locations.
+- Electron runtime validation:
+  Validate downloaded Electron with `ELECTRON_RUN_AS_NODE=1 "$INSTALL_DIR/electron" -p 'process.versions.electron || ""'`; plain `electron --version` can launch the bundled app or report the embedded Node version and is not reliable.
 - CLI preflight:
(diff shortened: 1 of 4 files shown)


ilysenko:
Thanks for this — the Electron validation in particular is the right call, and likely fixes the SIGSEGV class from #75.

Three things I'd like before merge:

**1. Reorder the new CLI paths.** Putting `~/.npm-global/bin/codex` and `~/.npm-packages/bin/codex` first in `find_codex_cli` means a stale legacy npm-prefix install will hijack a newer nvm one. Real scenario: user had codex via `npm config set prefix ~/.npm-global` long ago, then moved to nvm — old binary still sitting there will now win. Could you move both new paths **after** `/usr/bin/codex` (i.e. last in the for loop)? They become a fallback instead of a hijack.

**2. Don't swallow Electron's stderr in the validation.** Right now `2>/dev/null || true` hides the actual reason if Electron can't load (missing libs on NixOS/Alpine, glibc mismatch, etc.) and the user just sees `expected 41.2.0, got not-electron`. Could you capture stderr to a temp file and append it to the error message? Saves hours of debugging when this fires.

**3. Assert the setuid bit actually got set.** `chmod 4755 ... 2>/dev/null || true` silently no-ops on `nosuid`-mounted homes or restrictive containers, and we ship a package without the bit. A quick `[ -u "$path" ] || error "..."` after the chmod (at least in `package-common.sh`, where it matters most) would catch this.

RPM/Pacman setuid preservation we can verify in a follow-up smoke test — not a blocker for this PR.

ilysenko:
Thanks for the PR and for digging into the launch failure path.

Since this was opened, the original #74 missing-CLI scenario has been covered by #76 and #105: the launcher/updater now reports missing Codex CLI explicitly, packaged launches can recover by installing it, and native packages bundle the managed Node runtime needed for CLI install/update flows.

I’m going to close this PR for now rather than merge it as-is. The Electron runtime validation idea is still useful, but it should land as a smaller follow-up with stderr included in the validation error. The chrome-sandbox and extra CLI path changes need more care to avoid changing runtime selection in surprising ways.

Thanks again for the work here.

ilysenko:
Closing this as discussed above. Thanks again.

diegomarzaa:
Okay! No problem, not my area of development this one. It was a patch made by codex so I understand it probably wasn't the best way.

PR #91, opened 2026-05-02
[codex] fix Fedora package format detection

DhanushSantosh:
## Summary

This fixes package-format auto-detection on Fedora.

On Fedora systems that also have Debian tooling installed, the project could choose the Debian packaging path just because `dpkg` or `dpkg-deb` existed on the machine. That led `make package` to build a `.deb` instead of an `.rpm`, and `make install` to run `dpkg -i` instead of the native RPM install flow.

## Root cause

Package selection was giving too much weight to command availability and not enough weight to the actual host distribution.

That created two related problems:

- The shell-side Makefile logic could prefer DEB on Fedora if Debian tools were present.
- The Rust updater could also prefer the wrong package format if a package from another ecosystem had been installed previously.

## Fix

This change makes package detection distro-first.

- `Makefile` now checks `/etc/os-release` first and picks the native package format for the current distro.
- Fedora-family systems now prefer RPM, Debian-family systems prefer DEB, and Arch-family systems prefer pacman.
- `make install` now follows the same decision and uses the native package manager for the chosen format.
- On Fedora, that means using the RPM path with `dnf` instead of incorrectly routing through `dpkg`.
- `updater/src/install.rs` now mirrors the same distro-first logic so the updater stays aligned with the Makefile behavior.

## Why this matters

This prevents Fedora hosts from building or installing the wrong package format just because cross-distro tooling happens to be installed. It also prevents an already-installed `.deb` from teaching the updater to keep preferring Debian behavior on an RPM-based system.

## Validation

I verified the change with:

- `cargo test -p codex-update-manager detection_`
- `make -n package`
- `make -n install`

On this Fedora machine, the dry runs now resolve to the RPM build and install paths as intended.

## Notes

I did not use the full `cargo test -p codex-update-manager` suite as the primary validation signal for this PR because this environment still has unrelated CLI-path test failures around an existing `/usr/local/bin/codex`. Those failures are outside the scope of this packaging fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,12 +16,23 @@ PACMAN_GLOB := $(CURDIR)/dist/$(PACKAGE_NAME)-[0-9]*.pkg.tar.*
 .DEFAULT_GOAL := help
 
 NATIVE_PKG_FORMAT_CMD = format=""; \
+os_release_token_match() { \
+	local expected token; \
+	for token in $${ID:-} $${ID_LIKE:-}; do \
+		for expected in "$$@"; do \
+			if [ "$$token" = "$$expected" ]; then \
+				return 0; \
+			fi; \
+		done; \
+	done; \
+	return 1; \
+}; \
 if [ -r /etc/os-release ]; then . /etc/os-release; \
-	if command -v pacman >/dev/null 2>&1 && [[ " $${ID:-} $${ID_LIKE:-} " =~ (arch|archlinux|manjaro|endeavouros|artix) ]]; then \
+	if os_release_token_match arch archlinux manjaro endeavouros artix; then \
 		format="pacman"; \
-	elif command -v rpmbuild >/dev/null 2>&1 && [[ " $${ID:-} $${ID_LIKE:-} " =~ (fedora|rhel|centos|rocky|almalinux|ol|sles|suse|opensuse) ]]; then \
+	elif os_release_token_match fedora rhel centos rocky almalinux ol sles suse opensuse; then \
 		format="rpm"; \
-	elif command -v dpkg-deb >/dev/null 2>&1 && [[ " $${ID:-} $${ID_LIKE:-} " =~ (debian|ubuntu|linuxmint|pop|elementary|zorin) ]]; then \
+	elif os_release_token_match debian ubuntu linuxmint pop elementary zorin; then \
 		format="deb"; \
 	fi; \
 fi; \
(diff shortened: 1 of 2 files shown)


ilysenko:
Thanks for tackling this — distro-first detection is the right direction.

Two issues from a second review pass that I'd like fixed before merge:

**High: `make install` can still pick `deb` on Fedora without `rpmbuild`.** The os-release branches in `NATIVE_PKG_FORMAT_CMD` are gated on the *build* tool (`rpmbuild`/`dpkg-deb`/`pacman`), but the same detector is reused for `install` where only `dnf`/`zypper`/`rpm` matter. Concrete failure: on Fedora with `dnf` + a previously-built `.rpm` + `dpkg-deb` (no `rpmbuild`), the os-release section skips both RPM and DEB branches, falls through to the fallback, picks `deb` from `dpkg-deb` presence, and `make install` runs `dpkg -i`. That's the original bug still present on the install path.

Two ways to fix:
- Drop the `command -v <buildtool>` guards from the os-release section so distro identity wins on its own. `make package` can then fail clearly if the build tool is missing, while `make install` works correctly off distro identity alone.
- Or split into two detectors: one for build (needs build tool), one for install (needs an installer).

First option is simpler and covers both paths.

**Medium: bash detector matches by substring, Rust detector matches by token.** `os_release_matches` in `updater/src/install.rs` does `field.split_whitespace().any(|token| expected.contains(&token))` — proper token equality. The Makefile uses `[[ " $ID $ID_LIKE " =~ (arch|...|ol|...) ]]` — substring regex, so `ol` could match inside an unrelated ID and the two detectors can disagree on edge inputs. Could you make the bash side iterate `$ID $ID_LIKE` tokens and compare for exact match? Keeps the two implementations actually in sync, which is what the PR claims.

Once those are in I'll re-review. Also: please run the full `cargo test -p codex-update-manager` (CI will catch it but worth running locally too — the env-related failures you mentioned should be unrelated to this change).

DhanushSantosh:
@ilysenko  ill get it done in a few minutes.

DhanushSantosh:
I fixed both issues you called out.

I removed the build-tool guards from the `/etc/os-release` branch in the Makefile detector, so Fedora now stays on the RPM path for `make install` even if `rpmbuild` is missing and `dpkg-deb` is present.

I also changed the bash-side distro matching to use exact token checks instead of substring regex matching, so it now stays aligned with the Rust `os_release_matches` behavior.

I verified this with:

- `make -n package`
- `make -n install`
- a Fedora simulation with `dnf` + `dpkg-deb` present and no `rpmbuild`, which now resolves to `rpm`

The full Rust test suite still hits the same 3 existing CLI-path failures around `/usr/local/bin/codex` in this environment but that might be from my side. the packaging and detector-related tests are passing.

PR #100, opened 2026-05-03
[codex] fix Linux close-to-tray quit guard

DhanushSantosh:
## Summary

This fixes Linux close-to-tray behavior so Codex keeps running in the background after the main window is closed.

With tray/background mode enabled, closing the main window should hide the app to the system tray and leave the background process alive. Instead, some Linux builds could throw a main-process exception and fully exit:

`ReferenceError: codexLinuxIsQuitInProgress is not defined`

After this change:

- closing the main window with tray enabled keeps Codex alive in the background
- Codex remains accessible from the system tray
- explicit quit still exits the full app cleanly

## Root cause

The Linux patcher could inject tray and launch-action code that called `codexLinuxIsQuitInProgress()`, but helper injection still depended on a narrower bundle-shape insertion point.

That meant some upstream Electron bundle variants could end up with call sites that referenced `codexLinuxIsQuitInProgress()` without also containing the helper definition. When the user closed the main window, the close-to-tray path threw a `ReferenceError` in the Electron main process and the whole app exited instead of staying resident in the tray.

## Fix

This change makes the Linux quit-state patching more defensive and resilient to upstream bundle drift.

The helper injection no longer depends on one exact top-level bundle shape. It now supports:

- the current combined `require('electron')` / `node:path` / `node:fs` prelude
- the split `require(...)` variant
- the second-instance / launch-action bootstrap anchor when that is the stable insertion point

It also makes every generated quit-state-dependent path safe:

- tray close handling now uses a guarded quit check instead of a bare helper call
- tray menu refresh logic now uses the same guarded quit check
- launch-action handling now uses the same guarded quit check
- the `before-quit` path safely marks quit-in-progress only when the helper exists
- the launch-action patch reuses an existing quit helper instead of redeclaring it

## Why this matters

This restores the expected Linux tray workflow.

Without this fix, the app could advertise tray/background support but still fully exit when the user closed the window, depending on the exact upstream minified bundle shape that got patched. With this fix, the Linux build behaves the way users expect: window close hides to tray, while explicit quit actually exits.

## Validation

I validated the fix with:

- `make build-app`

I also added regression coverage for:

- a fixture where the tray close patch matches even though the original helper insertion needle is absent
- safe close-to-tray behavior when only the second-instance bootstrap anchor is available
- quit-in-progress handling for launch-action and `before-quit` paths
- smoke assertions that the generated bundle contains safe quit guards and only one quit-state helper definition

## Notes

I did not attach the local screenshot directly because the GitHub CLI flow here cannot upload a local image into the PR body, but the exact error text from that screenshot is included above for context.


ilysenko:
Thanks — we hit the same bug from a different angle in #99 (already merged), which adds a fallback helper insertion point for bundles where the original anchor doesn't match. That covers the concrete crash that was being reported.

The defensive `typeof codexLinuxIsQuitInProgress === 'function'` guards at call sites are a good idea conceptually, but the rest of this PR (helper-function refactor, third anchor variant, generator pattern) is a much larger surface area than the residual risk justifies in this hot patcher file. If you'd like to land just the call-site `typeof` guards as a small follow-up — that would be welcome on top of #99. Closing this in favor of that smaller scope?

DhanushSantosh:
ill wait for the Ci checks to be completed and ill look into the guard checks shortly.

DhanushSantosh:
@ilysenko "helper-function refactor, third anchor variant, generator pattern" Wouldnt these be helpful later on?

DhanushSantosh:
nvm ill just open up a smaller folllow up pr, just as you asked.

PR #101, opened 2026-05-03
Browser Use Node runtime selection (subset of #89)

ilysenko:
## Summary

This is the runtime-selection portion of @baktun95827's #89, cherry-picked on top of `main` so we can land it without taking on the bundled-runtime scope changes in that PR (which #94 already supersedes).

Original commits, with author preserved:

- `Fix Browser Use Node runtime selection` — adds Node 22.22.0+ version detection and a `find_compatible_node_runtime` cascade across packaged resources, the primary-runtime cache, system Node, and `nvm` installs. Exports `CODEX_BROWSER_USE_NODE_PATH` and `NODE_REPL_NODE_PATH` to the selected runtime and prepends it to `PATH`.
- `Update smoke test for Node runtime packaging` — packaging-side smoke coverage for the new selection logic.
- `Address Browser Use Node packaging feedback` — DEB now uses `Recommends: nodejs (>= 22.22.0), npm` instead of `Suggests:` (matches the RPM spec; `apt` installs Recommends by default, while nvm users can opt out with `--no-install-recommends`). Launcher's "Browser Use Node.js runtime not found" warning now appends an actionable install hint.

## What's intentionally not included

The bundled custom `node_repl` runtime parts of #89 (`browser-use-node-repl/node_repl.mjs`, `node_repl_wrapper.c`, related `bundled-plugins.sh` plumbing, and elicitation/socket fixes for that implementation) are dropped. #94 already covers the same "DMG `node_repl` is macOS Mach-O on Linux" failure case by installing the official Linux `node_repl` from the upstream `codex-primary-runtime` archive instead of shipping our own implementation.

## Validation

- `bash tests/scripts_smoke.sh` — all pass, including the new launcher-template marker test
- Cherry-pick conflicts in `launcher/start.sh.template`, `CHANGELOG.md`, and `tests/scripts_smoke.sh` resolved by keeping both sides where they add independent helpers, and dropping CHANGELOG entries / smoke tests that referenced the bundled-runtime scope.


ilysenko:
Closing this. A second-pass code review surfaced that weakening the packaging deps (Recommends/Suggests instead of Depends) breaks bare-system auto-update on Arch (`optdepends` not auto-installed) and on DEB/RPM with weak deps disabled, because `install.sh` and the launcher's CLI auto-install both still need `npm`/`npx`. Restoring hard deps would re-introduce the duplicate-Node issue for nvm users, so we're going the documentation route instead — see #103 for the upfront "Before you install" callout.

The launcher-side `find_compatible_node_runtime` cascade was useful but isn't critical once users follow the README requirement. We can revisit if mixed-Node setups become a recurring report.

PR #102, opened 2026-05-03
[codex] guard Linux quit helper call sites

DhanushSantosh:
## Summary

This is a small follow-up to #99.

#99 fixed the concrete Linux crash by making the quit-state helper injection more robust when the original insertion point does not match the current upstream bundle shape.

This follow-up keeps that helper behavior exactly as it is on `main` and only adds defensive call-site guards around Linux quit-helper usage.

After this change:

- closing the main window with tray enabled still hides Codex to the tray
- explicit quit still exits cleanly
- quit-state-dependent tray and launch-action paths no longer assume the helper functions always exist at runtime

## Root cause

Even with the helper insertion fallback from #99, the tray and launch-action patches still had a few bare calls like:

- `codexLinuxIsQuitInProgress()`
- `codexLinuxMarkQuitInProgress()`

Those direct calls are fine when the helper is present, but they are still a sharper edge than necessary in this patcher file.

## Fix

This change narrows the defensive surface to the call sites only.

It updates the Linux patcher so quit-helper consumers use guarded checks instead of bare calls:

- close-to-tray logic now checks `typeof codexLinuxIsQuitInProgress === 'function'` before calling it
- Linux tray menu refresh and native tray menu open logic do the same
- `before-quit` handlers now check `typeof codexLinuxMarkQuitInProgress === 'function'` before calling it
- launch-action handling uses the same guarded quit checks before suppressing work during shutdown

It does not change helper insertion anchors, helper generation strategy, or the broader patch structure from `main`.

## Why this matters

This keeps the follow-up tightly scoped while still adding a useful layer of runtime safety in the hottest quit-state call sites.

The goal here is not to replace #99, but to make the remaining helper consumers more defensive without broadening the patcher surface area.

## Validation

I validated the follow-up with:

I also added a small smoke regression that checks the Linux close-to-tray path still hides the window even if the quit helper is unavailable in the evaluated close handler.


DhanushSantosh:
@ilysenko This should do it.

PR #107, opened 2026-05-04
Add manual rollback to the last known-good package

pradigmaz:
Adds a manual rollback command for the update manager so a user can return to the last retained known-good native package after a broken rebuilt update.

Changed:
- add `codex-update-manager rollback`
- retain the previous known-good package path when preparing the next update
- add explicit privileged rollback install paths for DEB, RPM, and pacman
- add separate polkit actions for rollback
- block automatic reinstall of the candidate that was just rolled back
- document the manual rollback command in README

Rollback process:
1. Close Codex Desktop.
2. Run `codex-update-manager rollback`.
3. Authenticate when polkit asks.
4. Wait for the native package manager to reinstall the retained previous package.
5. Start Codex Desktop normally.

Validation:
- `cargo fmt --check`
- `cargo test -p codex-update-manager rollback`
- `cargo test -p codex-update-manager -- --skip upstream::tests::downloads_dmg_and_hashes_contents --skip upstream::tests::fetches_remote_metadata_from_head`
- `make rpm PACKAGE_VERSION=2026.05.04.131500.rollbacktest2`
- real Fedora RPM E2E: installed test package, ran `/usr/bin/codex-update-manager rollback`, confirmed package returned to `2026.05.02.120000-trayfix.fc43`, then launched `/opt/codex-desktop/start.sh`
- `git diff --check`

Limits:
- real E2E was done for RPM on Fedora only
- DEB and pacman rollback paths are covered by command construction tests, not live package installs
- this is manual rollback only; no automatic health check or automatic rollback is included

Follow-up refactor suggestion:
- split updater orchestration into focused modules: check cycle, pending install recovery, privileged install dispatch, rollback state, CLI preflight, and status rendering
- keep package-manager-specific install and rollback command construction separate from app orchestration
- keep state compatibility tests close to `state.rs`, especially for fields written before a downgrade

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,16 @@ sed -n '1,160p' ~/.local/state/codex-update-manager/state.json
 sed -n '1,160p' ~/.local/state/codex-update-manager/service.log
 ```
 
+If a rebuilt update installs but the previous retained package was better, close
+Codex Desktop and run:
+
+```bash
+codex-update-manager rollback
+```
+
+Rollback uses the last retained known-good package and refuses to run when no
+rollback package is available.
+
 Runtime files live in standard XDG locations:
 
 ```text
--- a/packaging/linux/com.github.ilysenko.codex-desktop-linux.update.policy
+++ b/packaging/linux/com.github.ilysenko.codex-desktop-linux.update.policy
@@ -42,4 +42,40 @@
     <annotate key="org.freedesktop.policykit.exec.path">/usr/bin/codex-update-manager</annotate>
     <annotate key="org.freedesktop.policykit.exec.argv1">install-pacman</annotate>
   </action>
+
+  <action id="com.github.ilysenko.codex-desktop-linux.update.rollback-deb">
+    <description>Roll back a Codex Desktop update</description>
+    <message>Authentication is required to roll back the Codex Desktop update.</message>
+    <defaults>
+      <allow_any>auth_admin</allow_any>
+      <allow_inactive>auth_admin</allow_inactive>
+      <allow_active>auth_admin</allow_active>
+    </defaults>
+    <annotate key="org.freedesktop.policykit.exec.path">/usr/bin/codex-update-manager</annotate>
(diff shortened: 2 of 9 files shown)


ilysenko:
Thanks for the PR — this is a valuable and well-structured addition. A manual rollback flow is definitely missing today, and you implemented it in a clean, maintainable way.

That said, this still isn’t acceptable for us yet.

1) **Blocking issue:** rollback target bookkeeping appears unsafe after a successful rollback.

After `rollback`, `state.artifact_paths.package_path` is not updated to reflect what was just installed, while `rollback_package_path` is derived from `package_path` during `record_current_package_as_known_good`.
If the next update cycle runs, that function can overwrite `rollback_package_path` with whatever `package_path` currently represents, which is not guaranteed to be the actual last known-good package anymore. This can lead to rollback returning to an unexpected/incorrect package.

2) Please make rollback state transition explicit and durable:
- On successful rollback, set `state.artifact_paths.package_path` (and/or clearly separate “rollback_target”) to the package that was just installed.
- Preserve/refresh `rollback_package_path` deterministically so it always points to the true last known-good package.
- Avoid re-pointing the rollback target based on a stale update artifact.

3) The overall architecture and UX work is good (new `rollback` command, dedicated module, polkit actions), and this blocker is the only piece we need resolved before merging.

Happy to review again once this is fixed.

pradigmaz:
Thanks, agreed. I pushed a focused fix in `f34d887`.

Changed:
- On successful rollback, the updater now sets `artifact_paths.package_path` to the package that was just installed.
- It also refreshes `rollback_package_path` and `last_known_good_version` to the same installed package, so the next update cycle cannot derive the rollback target from a stale update artifact.
- Added a regression test for the stale candidate package path case.

Verified with:
- `cargo fmt --check`
- `cargo test -p codex-update-manager rollback`
- `cargo test -p codex-update-manager -- --skip upstream::tests::downloads_dmg_and_hashes_contents --skip upstream::tests::fetches_remote_metadata_from_head`
- `git diff --check`

PR #111, opened 2026-05-05
fix: force --ozone-platform=x11 under ChromeOS Crostini Sommelier (#95)

### LINKED ISSUE #95: Codex Desktop does not open from ChromeOS/Crostini app launcher
## Summary

Codex Desktop does not open from the ChromeOS/Crostini app launcher in a Sommelier-based Linux container. Direct terminal launches can reach Electron, but the launcher path used by the desktop/app menu still does not result in a usable visible app for the user.

## Environment

- Host/container: ChromeOS Crostini Linux VM (`penguin`)
- User session services observed:
  - `cros-garcon.service` active
  - `sommelier@0.service` active (`wayland-0`)
  - `sommelier-x@0.service` active (`DISPLAY=:0`)
  - `xdg-desktop-portal.service` and `xdg-desktop-portal-gtk.service` active
- Relevant user manager environment:
  - `DISPLAY=:0`
  - `WAYLAND_DISPLAY=wayland-0`
  - `SOMMELIER_VERSION=0.20`
  - `SOMMELIER_VM_IDENTIFIER=faf5e5b1-e773-49ae-a425-f3c496b51a93`
- Codex CLI is installed under XDG-style nvm:
  - `/home/culpen0/.config/nvm/versions/node/v24.15.0/bin/codex`
- Installed package format: `.deb`

## What happens

Launching from the ChromeOS/Crostini app menu does not open a usable Codex Desktop window.

During debugging, `~/.cache/codex-desktop/launcher.log` showed several distinct failure modes depending on launch path:

1. Default Wayland/auto path can reach Electron and `window ready-to-show`, but the window is not visible under Sommelier. Logs include:

```text
Electron launch mode: ozone_platform=default ozone_hint=auto gpu_enabled=1
ERROR:ui/ozone/platform/wayland/common/drm_render_node_path_finder.cc:45] drmGetDevices2() has not found any devices
[window-manager] window ready-to-show appearance=primary ...
```

2. Forcing X11 from a terminal creates a visible X11 window:

```text
/usr/bin/codex-desktop --x11
Electron launch mode: ozone_platform=x11 ozone_hint=none gpu_enabled=1
```

`xwininfo -root -tree` then shows a visible `Codex` window at about `1280x850+320+175`.

3. `--safe-mode` previously failed because it disabled GPU access completely:

```text
Error: GPU access not allowed. Reason: GPU access is disabled through commandline switch --disable-gpu and --disable-software-rasterizer.
```

4. GUI/app-launch-like environments may not find the Codex CLI because the launcher checks `~/.nvm` but not `~/.config/nvm`. The failing path printed:

```text
Codex CLI is required but is not installed...
```

while the actual CLI exists at:

```text
/home/culpen0/.config/nvm/versions/node/v24.15.0/bin/codex
```

5. A `gio launch /usr/share/applications/codex-desktop.desktop` reproduction did not behave the same as direct `/usr/bin/codex-desktop`: it only appended `Starting Codex Desktop launcher` to the launcher log and exited before the normal `packaged_prelaunch` phase. That may or may not match ChromeOS garcon's actual app-launch path, but it suggests desktop-entry launch behavior is still not robust here.

## Expected behavior

Clicking Codex Desktop from the ChromeOS/Crostini app launcher should open the same visible window that `/usr/bin/codex-desktop --x11` opens from a terminal.

## Local debugging / attempted fixes

The following local launcher changes improved terminal and sparse-environment validation, but the user still reports the app menu does not open the app:

- Auto-select X11 on Sommelier sessions when no explicit Ozone flag is supplied.
- Keep `--wayland`, `--x11`, `--ozone-platform=*`, and `--ozone-platform-hint=*` as explicit overrides.
- Change `--safe-mode` to force X11 without disabling GPU access.
- Search XDG nvm locations such as `${XDG_CONFIG_HOME:-$HOME/.config}/nvm` for the Codex CLI.
- Avoid passing an empty `--cli-path` to `codex-update-manager prompt-install-cli` / `cli-preflight`.
- Make `notify-send` best-effort, since this environment reported no notification service.
- Import Sommelier vars from `systemd --user show-environment`, even when `DISPLAY` is already present.

After those changes, a sparse app-menu-like terminal launch reaches Electron successfully:

```text
Using CODEX_CLI_PATH=/home/culpen0/.config/nvm/versions/node/v24.15.0/bin/codex
Electron launch mode: ozone_platform=x11 ozone_hint=none gpu_enabled=1
[window-manager] window ready-to-show appearance=primary ...
```

However, the real app launcher still does not open the app for the user.

## Suggested areas to inspect

- ChromeOS/Crostini `cros-garcon` desktop-entry launch semantics versus direct `/usr/bin/codex-desktop`.
- Whether garcon is using a stale exported desktop entry or cached command/environment after package updates.
- Whether the desktop entry should force `--x11` in known Sommelier/Crostini sessions, or whether the launcher should have a Crostini-specific mode independent of env inheritance.
- Whether the package install/postinst should refresh the ChromeOS app registry or restart/reload the appropriate user service when the `.desktop` file or launcher changes.

techs-targe:
## Summary

Fixes #95. On a ChromeOS Crostini Linux container the default Wayland session is provided by Sommelier. With Electron's default `--ozone-platform-hint=auto` path the window logs `ready-to-show` but never surfaces in the ChromeOS shell — both bare `codex-desktop` and ChromeOS app-menu launches just leave the launcher icon spinning. The reporter and I have both confirmed `--x11` (which Sommelier proxies to ChromeOS via XWayland) renders correctly.

This hooks the existing flag parser in `set_electron_defaults` so that, when the operator did **not** explicitly choose a platform (no `--x11` / `--wayland` / `--ozone-platform=*`), and `SOMMELIER_VERSION` is non-empty, we substitute `--ozone-platform=x11` and log a one-line note for visibility. Explicit user flags still win, so anyone wanting to test the native Wayland path can opt back in with `--wayland`.

## Why detect via `SOMMELIER_VERSION`

`SOMMELIER_VERSION` and `SOMMELIER_VM_IDENTIFIER` are exported by `cros-garcon` for both terminal sessions and app-menu launches inside Crostini, and they are not set in any other Linux session I'm aware of. They are also what the original report uses to identify the environment, so the detection is symmetric with how the issue was filed.

## Verified on

- Crostini Debian 12 (`bookworm`), Sommelier `0.20`
- `XDG_SESSION_TYPE=wayland`, `XDG_CURRENT_DESKTOP=X-Generic`
- codex-desktop `2026.05.04.231437` (built locally from this branch's parent commit)

| Scenario | Before | After |
|---|---|---|
| `codex-desktop` (terminal, no flags) | window not visible | window visible (auto x11) |
| ChromeOS app launcher click | spinner forever, no window | singleton handoff focuses existing window |
| `codex-desktop --x11` | window visible | window visible (no behavior change) |
| `codex-desktop --wayland` | (Wayland, but invisible) | Wayland honored (no auto-override) |
| `codex-desktop --safe-mode` | x11 + sw rendering | unchanged |

`launcher.log` excerpt after the patch:

```
Detected Sommelier (SOMMELIER_VERSION=0.20); forcing --ozone-platform=x11 (Wayland windows are not visible under ChromeOS Crostini)
Electron launch mode: ozone_platform=x11 ozone_hint=none gpu_enabled=1
```

## Smoke-tested

```bash
# synthetic prelude + template syntax
{ printf '#!/bin/bash\nset -euo pipefail\nCODEX_LINUX_APP_ID=codex-desktop\nCODEX_LINUX_APP_DISPLAY_NAME="Codex Desktop"\nCODEX_LINUX_WEBVIEW_PORT=5175\n'; cat launcher/start.sh.template; } > /tmp/start-test.sh
bash -n /tmp/start-test.sh   # OK

# upstream smoke tests
bash tests/scripts_smoke.sh  # All script smoke tests passed
```

## Test plan for the reviewer

- [ ] `bash -n install.sh launcher/start.sh.template` (syntax)
- [ ] `bash tests/scripts_smoke.sh` (existing smoke)
- [ ] On a non-Crostini Wayland host, confirm the auto-substitution is **not** triggered (no `SOMMELIER_VERSION` env)
- [ ] On Crostini, confirm app-menu launch surfaces a window without `--x11`

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -1033,6 +1033,27 @@ set_electron_defaults() {
         esac
         shift
     done
+
+    # ChromeOS / Crostini (Sommelier) auto-fallback to X11.
+    #
+    # On a Crostini Linux container the default Wayland session is provided by
+    # Sommelier (see SOMMELIER_VERSION/SOMMELIER_VM_IDENTIFIER, exported by
+    # cros-garcon for both terminal and app-menu launches). With the default
+    # Electron path (--ozone-platform-hint=auto) Electron picks the Wayland
+    # backend, reaches `window ready-to-show`, but the window never surfaces in
+    # the ChromeOS shell — leaving the launcher icon spinning indefinitely with
+    # no visible app. Forcing X11 (proxied by sommelier-x to ChromeOS) is the
+    # documented workaround. Skip this only when the operator explicitly chose
+    # a platform via --x11 / --wayland / --ozone-platform=*.
+    #
+    # See https://github.com/ilysenko/codex-desktop-linux/issues/95.
+    if [ -z "$ELECTRON_OZONE_PLATFORM" ] \
+        && [ "$ELECTRON_OZONE_HINT" = "auto" ] \
+        && [ -n "${SOMMELIER_VERSION:-}" ]; then
+        ELECTRON_OZONE_PLATFORM="x11"
+        ELECTRON_OZONE_HINT=""
+        echo "Detected Sommelier (SOMMELIER_VERSION=$SOMMELIER_VERSION); forcing --ozone-platform=x11 (Wayland windows are not visible under ChromeOS Crostini)"
+    fi
 }
 
 build_electron_launch_args() {
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -179,6 +179,16 @@ discover_graphical_env_from_processes() {
 
 hydrate_graphical_session_env() {
     if [ -n "${DISPLAY:-}" ] || [ -n "${WAYLAND_DISPLAY:-}" ]; then
+        # Display vars are already present, but the ChromeOS / Crostini
+        # auto-X11 fallback in set_electron_defaults also needs
+        # SOMMELIER_VERSION. On some app-launcher paths the spawned
+        # process inherits DISPLAY / WAYLAND_DISPLAY but not the Sommelier
+        # markers (those live only in the systemd --user manager env). Try
+        # one user-manager import in that case so the whitelisted
+        # SOMMELIER_VERSION / SOMMELIER_VM_IDENTIFIER get hydrated.
+        if [ -z "${SOMMELIER_VERSION:-}" ]; then
+            import_graphical_env_from_systemd_user >/dev/null 2>&1 || true
+        fi
         return 0
     fi
 
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -112,8 +112,13 @@ import_graphical_env_entry() {
 
     [ "$entry" != "$name" ] || return 0
 
+    # SOMMELIER_VERSION / SOMMELIER_VM_IDENTIFIER are imported here so the
+    # ChromeOS / Crostini auto-X11 fallback in set_electron_defaults sees
+    # them on app-launcher paths where they may only be present in the
+    # systemd --user manager environment (and not in the spawned process's
+    # initial env).
     case "$name" in
-        DISPLAY|WAYLAND_DISPLAY|XDG_SESSION_TYPE|XDG_CURRENT_DESKTOP|DBUS_SESSION_BUS_ADDRESS|XDG_RUNTIME_DIR|DESKTOP_SESSION|XAUTHORITY)
+        DISPLAY|WAYLAND_DISPLAY|XDG_SESSION_TYPE|XDG_CURRENT_DESKTOP|DBUS_SESSION_BUS_ADDRESS|XDG_RUNTIME_DIR|DESKTOP_SESSION|XAUTHORITY|SOMMELIER_VERSION|SOMMELIER_VM_IDENTIFIER)
             if [ -z "${!name:-}" ]; then
                 export "$entry"
             fi


ilysenko:
Thanks for the focused fix. The approach makes sense to me: defaulting Crostini/Sommelier launches to X11 matches the confirmed workaround, and keeping explicit `--x11`, `--wayland`, and `--ozone-platform=*` flags authoritative is the right behavior.

One small concern before merging: the new fallback depends on `SOMMELIER_VERSION` being present in the launcher process, but our existing graphical env hydration only imports display/session variables from `systemctl --user show-environment` / `/proc/*/environ`. It does not currently import `SOMMELIER_VERSION` or `SOMMELIER_VM_IDENTIFIER`.

Could you also add those two variables to the whitelist in `import_graphical_env_entry`? Otherwise there is a possible app-launcher path where `DISPLAY` / `WAYLAND_DISPLAY` are present, `SOMMELIER_VERSION` is only available from the user manager environment, and the auto-X11 fallback never triggers.

After that, this looks good to merge.

techs-targe:
Good catch, thanks! Pushed `d0f9360` which adds `SOMMELIER_VERSION` and `SOMMELIER_VM_IDENTIFIER` to the `import_graphical_env_entry` whitelist, right next to `DISPLAY` / `WAYLAND_DISPLAY` / friends, with a short comment pointing back at the auto-X11 fallback.

Re-verified locally:

- `bash -n` on the synthetic prelude+template: OK
- `bash tests/scripts_smoke.sh`: passes
- Fresh `codex-desktop` launch on Crostini Debian 12 / Sommelier 0.20: still hits the fallback path and Electron starts with `--ozone-platform=x11`. App-menu (`gio launch`) likewise still works (singleton handoff focuses the existing window).
- The two existing Sommelier vars stayed untouched in our session because cros-garcon already exports them directly; the new import path is exercised on the corner case you described (vars only present in the systemd user manager env), where it's now reachable.

Existing display/session imports are unchanged. Ready when you are.

ilysenko:
Thanks for the quick follow-up. Adding `SOMMELIER_VERSION` and `SOMMELIER_VM_IDENTIFIER` to the import whitelist is the right direction.

One remaining edge case: `hydrate_graphical_session_env` currently returns immediately when `DISPLAY` or `WAYLAND_DISPLAY` is already present. That means if an app-launcher path provides display variables but omits `SOMMELIER_VERSION`, we still never call `import_graphical_env_from_systemd_user`, so the newly-whitelisted Sommelier vars would not be imported.

Could you adjust `hydrate_graphical_session_env` so it still attempts the user-manager import when `SOMMELIER_VERSION` is missing? For example, if display vars are already present but `SOMMELIER_VERSION` is empty, run `import_graphical_env_from_systemd_user >/dev/null 2>&1 || true` before returning.

After that I think this fully covers the app-launcher environment case.

techs-targe:
Confirmed on a separate Crostini target machine (Debian 12 / Sommelier 0.20, distinct from the dev machine where the patch was authored). I scripted the three env conditions you flagged so this is reproducible:

- **Test 1 — baseline** (`SOMMELIER_*` set in process env): expect fallback to fire → x11
- **Test 2 — edge case** (`SOMMELIER_*` stripped from process env via `env -u …`, kept in `systemctl --user` manager env via `import-environment`): expect fallback to still fire → x11
- **Test 3 — simulated non-Crostini** (`SOMMELIER_*` removed from both process env and manager env via `unset-environment`): expect fallback **not** to fire → default Wayland

Two rounds were useful:

| Build | Test 1 | Test 2 | Test 3 | Notes |
|---|---|---|---|---|
| Pre-`hydrate_graphical_session_env` patch | PASS | **FAIL** (`ozone_platform=default ozone_hint=auto`) | PASS | Confirms your concern was real — the early `return 0` in the original `hydrate_graphical_session_env` skipped the user-manager import entirely, so the whitelisted Sommelier vars were never hydrated. |
| Tip of this branch (`43d1255`) | PASS | **PASS** (`ozone_platform=x11 ozone_hint=none`) | PASS | All three scenarios behave as designed. Test 3 verifies no false-positives on systems where Sommelier markers truly do not exist. |

Sample log from Test 2 on the patched build:

```
Detected Sommelier (SOMMELIER_VERSION=0.20); forcing --ozone-platform=x11
Electron launch mode: ozone_platform=x11 ozone_hint=none gpu_enabled=1
```

CI is green on all four package targets. Ready when you are.

PR #112, opened 2026-05-06
fix(linux): bypass upstream app sunset gate

jonfry1175:


ilysenko:
Thanks for the PR. I agree that bypassing the upstream sunset/update-required gate may be necessary for the Linux wrapper, but I would not merge this as-is because the PR mixes that patch with a couple of unrelated startup changes that can regress supported environments.

Main concern: changing `find` to `/usr/bin/find` can break launcher startup on NixOS and other non-FHS systems where `find` is not guaranteed to exist at that path. Since `start.sh` runs with `set -e`, this can become a hard startup failure. Please keep using `find` from `PATH` (or use `command find`) instead.

Second concern: copying `codex-packaged-runtime.sh` from `install.sh` puts native-package-only behavior into every local/source install. That helper starts/enables the updater user service, triggers update checks, and exports packaged desktop hints. This currently belongs in the native package staging path (`scripts/lib/package-common.sh`), not the generic installer.

Could you please narrow this PR to the app sunset gate patch and its test, or split the startup hardening changes into a separate PR after addressing the two points above?

Once those unrelated changes are removed/fixed, the sunset gate patch looks much easier to review and merge.

PR #114, opened 2026-05-06
Add Hyprland window targeting backend

christadrian:
## Summary
- Add a Hyprland backend for Linux Computer Use window listing via `hyprctl clients -j`
- Support exact Hyprland window focus through `hyprctl dispatch focuswindow address:<addr>`
- Update doctor diagnostics/readiness to report Hyprland window targeting as fully available when `hyprctl` works

## Verification
- `cargo fmt -- --check`
- `cargo run -p codex-computer-use-linux -- windows` reports `backend: hyprland`
- Built app backend doctor reports `can_focus_windows: true` and no blockers on Hyprland

### DIFF
--- a/computer-use-linux/src/diagnostics.rs
+++ b/computer-use-linux/src/diagnostics.rs
@@ -64,6 +64,7 @@ pub struct AccessibilityReport {
 pub struct WindowingReport {
     pub gnome_shell_introspect: Check,
     pub codex_gnome_shell_extension: Check,
+    pub hyprland: Check,
     pub can_list_windows: bool,
     pub can_focus_apps: bool,
     pub can_focus_windows: bool,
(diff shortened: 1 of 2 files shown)


ilysenko (commented):
Pushed commit 575d150 to address the Hyprland review findings.\n\n- Parse Hyprland client pid as signed and normalize negative/sentinel values to None in WindowInfo.\n- Probe hyprctl directly in doctor readiness instead of failing early when desktop env vars are missing.\n- Added test coverage for negative Hyprland pid handling.\n\nLocal verification: cargo fmt -- --check; cargo test -p codex-computer-use-linux.\n\nCI has been rerun for the updated head commit.

ilysenko (approved):
LGTM after 575d150. The Hyprland pid parser now tolerates sentinel negative values, doctor probes hyprctl directly, local targeted tests pass, and the rerun CI is green.

PR #115, opened 2026-05-07
Fix/0.7.1 missing cargo path

Leay15:
This pull request updates the `codex-update-manager` to version 0.7.1 and addresses issues with reliably locating the Rust toolchain (`cargo`) in local auto-update and packaging scripts, especially when `cargo` is installed via `rustup` under `~/.cargo/bin`. It also adds regression test coverage for the new path resolution logic. The most important changes are:

**Build and Packaging Reliability:**

* Both `scripts/lib/package-common.sh` and `scripts/build-rpm.sh` now use a new `find_cargo_command` function to reliably locate `cargo`, falling back to `~/.cargo/bin/cargo` if it's not on the `PATH`. This ensures builds work even in reduced environment contexts. 

**Testing and Quality Assurance:**

* Added a unit test in `updater/src/builder.rs` to verify that `build_command_path` includes `~/.cargo/bin` when present, ensuring regression coverage for the new path logic.

**Documentation and Versioning:**

* Updated documentation and version references in `README.md`, `AGENTS.md`, `CHANGELOG.md`, and `Cargo.toml` to reflect the new version (0.7.1) and the build-environment improvements.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -151,7 +151,7 @@ Do not assume `codex-app/` is pristine. If behavior differs from `install.sh`, p
 
 ## Crate Versioning
 
-- Current updater crate version: `0.7.0`
+- Current updater crate version: `0.7.1`
 - Bump `patch` for fixes, docs, and maintenance-only updates.
 - Bump `minor` for compatible feature additions.
 - Bump `major` for incompatible CLI, persisted-state, or install-flow changes.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+## [0.7.1] - 2026-05-06
+
+### Fixed
+
+- Local auto-update rebuilds and package builds now find the Rust toolchain reliably when `cargo` is installed via `rustup` under `~/.cargo/bin`, even if the `codex-update-manager` user service or packaging scripts inherit a reduced `PATH`.
+
+### Added
+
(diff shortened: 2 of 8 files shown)


ilysenko (commented):
Thanks for the fix, this is pointed in the right direction. I found two issues that should be addressed before merge:

1. `updater/src/builder.rs`: the new `build_command_path_includes_cargo_bin_from_home` test mutates process-wide `HOME` but does not hold `crate::test_util::env_lock()`.

   I reproduced this locally with the normal parallel test runner:

   ```bash
   cargo test -p codex-update-manager
   ```

   It can race with other updater tests that also isolate `HOME`/`PATH`; one run failed in `app::tests::prompt_install_cli_does_not_treat_non_executable_file_as_installed` after it picked up my real `~/.nvm/.../codex`. Running with `-- --test-threads=1` passes, which points to the missing env lock.

   Please add:

   ```rust
   let _env_guard = crate::test_util::env_lock();
   ```

   at the start of the new test.

2. `scripts/build-rpm.sh`: this script still overrides `ensure_updater_binary()` and returns early whenever `target/release/codex-update-manager` exists. That bypasses the new `updater_binary_is_stale` logic in `scripts/lib/package-common.sh`.

   After this version bump, a local RPM build can still package a stale `0.7.0` updater binary if it already exists in `target/release`. The DEB/pacman builders use the shared helper and rebuild when sources/Cargo files are newer, but RPM does not.

   Please either remove the RPM-specific override and use the shared `ensure_updater_binary`, or copy the same stale-binary check into `scripts/build-rpm.sh`.

Local checks I ran:

```bash
bash -n scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh scripts/lib/package-common.sh
cargo fmt -- --check
cargo check -p codex-update-manager
cargo test -p codex-update-manager
cargo test -p codex-update-manager -- --test-threads=1
```

The parallel `cargo test` was the one that exposed the race.

PR #119, opened 2026-05-08
Fix Computer Use patching for current upstream bundle

ReOT20:
## Summary

Updates the ASAR patcher for the current upstream Codex Desktop bundle shape.

The current bundle can use `isAvailable` instead of `isEnabled` for the bundled Computer Use plugin gate. This caused the Linux Computer Use plugin gate patch to fail hard even though the bundle was otherwise recognizable.

This also makes Linux tray startup tolerate a missing `codexLinuxIsTrayEnabled` helper when the optional Linux settings patch is skipped.

## User-Visible Behavior

Linux builds against the current upstream DMG no longer fail with:

```text
Required Linux Computer Use plugin gate patch failed: could not enable bundled Computer Use on Linux
```

The generated app also no longer fails at startup with:

```
codexLinuxIsTrayEnabled is not defined
```

## Root Cause

The Computer Use plugin gate patch only matched isEnabled, but the current upstream bundle uses isAvailable.

The tray startup patch directly called codexLinuxIsTrayEnabled(), but that helper is introduced by another optional patch that can fail-soft when upstream bundle shapes drift.

## Changes

- Recognize both isEnabled and isAvailable in the Computer Use plugin gate patch.
- Preserve the upstream gate key when rebuilding the patched plugin descriptor.
- Make Linux tray startup default-safe if codexLinuxIsTrayEnabled is not defined.
- Add regression coverage for the isAvailable gate and updated tray startup expression.

## Validation

Ran:

```
```

### DIFF
--- a/scripts/patch-linux-window-ui.js
+++ b/scripts/patch-linux-window-ui.js
@@ -1045,9 +1045,12 @@ function applyLinuxTrayPatch(currentSource, iconPathExpression) {
 
   const trayStartupNeedle = "E&&oe();";
   const previousTrayStartupPatch = "(E||process.platform===`linux`)&&oe();";
-  const trayStartupPatch = "(E||process.platform===`linux`&&codexLinuxIsTrayEnabled())&&oe();";
+  const badTrayStartupPatch = "(E||process.platform===`linux`&&codexLinuxIsTrayEnabled())&&oe();";
+  const trayStartupPatch = "(E||process.platform===`linux`&&(typeof codexLinuxIsTrayEnabled!==`function`||codexLinuxIsTrayEnabled()))&&oe();";
   if (patchedSource.includes(trayStartupPatch)) {
     // Already patched.
+  } else if (patchedSource.includes(badTrayStartupPatch)) {
+    patchedSource = patchedSource.replace(badTrayStartupPatch, trayStartupPatch);
   } else if (patchedSource.includes(previousTrayStartupPatch)) {
     patchedSource = patchedSource.replace(previousTrayStartupPatch, trayStartupPatch);
   } else if (patchedSource.includes(trayStartupNeedle)) {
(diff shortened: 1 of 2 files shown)


ilysenko:
Thanks for the fix — the patcher changes look pointed in the right direction, but CI is currently failing in `Rust and Smoke Tests`.

The failure is in `tests/scripts_smoke.sh`, not in the JS patcher unit tests. I reproduced it locally with:

```bash
bash tests/scripts_smoke.sh
```

It fails at the Linux tray patch smoke check because the test still expects the old startup expression:

```text
(E||process.platform===`linux`&&codexLinuxIsTrayEnabled())&&oe();
```

This PR intentionally changes that output to tolerate the settings helper being absent:

```text
(E||process.platform===`linux`&&(typeof codexLinuxIsTrayEnabled!==`function`||codexLinuxIsTrayEnabled()))&&oe();
```

Please update the tray-related smoke assertions in `tests/scripts_smoke.sh` to match the new safe fallback expression. The relevant JS checks pass locally:

```bash
```

After the smoke fixture is updated and CI is green, this should be much easier to merge.

christadrian:
Pushed a follow-up for the failing smoke assertions here:

- branch: https://github.com/christadrian/codex-desktop-linux/tree/pr-119
- commit: https://github.com/christadrian/codex-desktop-linux/commit/e23fb8e

This updates `tests/scripts_smoke.sh` to match the guarded Linux tray startup expression and keeps the existing test file adjustments aligned with the patched output.

ilysenko:
Thanks for the fix here. The direction was right: supporting the upstream `isAvailable` Computer Use gate and guarding the Linux tray helper are both valid changes.

This PR has since been superseded by the changes already merged into `main` through #122 and #118. The current `main` now includes the Computer Use `isAvailable` handling, the guarded tray startup expression, and the matching smoke test updates.

Because this branch is now conflicting and the same functional fixes are already present on `main`, I am going to close this PR as superseded.

Thanks again for the contribution!

PR #124, opened 2026-05-08
Add dynamic Linux open target discovery

### LINKED ISSUE #123: Linux open targets only show file manager and miss installed IDEs
## Context

Linux desktop integration for open targets is currently limited. The menu can fall back to a generic File Manager entry and miss installed IDEs or terminal targets.

## Actual Behavior

- File/folder open targets do not reliably use the installed Linux file manager.
- Installed IDEs are not discovered dynamically from the desktop environment.
- Linux does not expose a Terminal open target like Windows/macOS do.

## Expected Behavior

- Use an available Linux file manager for opening files/folders.
- Show installed IDEs in the open-target menu.
- Show an installed terminal target when available.
- Discover IDEs from common Linux sources such as .desktop entries, Flatpak, Snap, and JetBrains Toolbox.

## Notes

This affects Linux desktop/runtime behavior rather than package installation or updater flow.

brianmsm:
## Summary

Adds Linux open-target support for installed file managers, terminals, and IDEs.

## User-Visible Behavior

The open-target menu can show installed Linux IDEs and a Terminal target instead of only a generic File Manager entry, and file/folder opens use an available Linux file manager.

## Changes

- Opens files/folders through the installed Linux file manager.
- Adds a Linux Terminal target using xdg-terminal-exec, known terminal commands, and TerminalEmulator .desktop entries.
- Adds Linux IDE targets for known commands when present.
- Discovers additional IDEs from .desktop entries, including Flatpak, Snap, and JetBrains Toolbox-style installs.
- Handles TryExec, executable checks, URI placeholders, duplicate IDs, and older patched bundles.

## Validation

Ran:

```bash
docker run --rm -v "$PWD:/repo" -w /repo node:22-bookworm-slim node --test scripts/patch-linux-window-ui.test.js
```

Also ran a Docker integration check with fake .desktop entries and file-manager commands, plus a patched-current-bundle parse check.

Not run: package/updater validation, because this only changes ASAR patching logic and tests.

Closes #123

### DIFF
--- a/scripts/patch-linux-window-ui.js
+++ b/scripts/patch-linux-window-ui.js
@@ -35,7 +35,6 @@ const {
 } = require("./patches/launch-actions.js");
 const {
   applyBrowserUseNodeReplApprovalPatch,
-  applyLinuxFileManagerPatch,
   applyLinuxGitOriginsSourceFallbackPatch,
   applyLinuxMenuPatch,
   applyLinuxOpaqueBackgroundPatch,
@@ -45,6 +44,11 @@ const {
   applyLinuxTrayPatch,
   applyLinuxWindowOptionsPatch,
 } = require("./patches/main-process.js");
+const {
+  applyLinuxFileManagerPatch,
+  applyLinuxIdeOpenTargetPatch,
+  applyLinuxTerminalOpenTargetPatch,
+} = require("./patches/open-targets.js");
 const {
   patchPackageJson,
   resolveDesktopName,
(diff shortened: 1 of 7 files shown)


ilysenko:
Just a quick heads-up: we added a new opt-in `linux-features/` framework for optional Linux-specific integrations. So if you still want to use or maintain this open-target discovery work on top of the latest `main`, you can now move it into a feature and enable it locally through `linux-features/features.json`.

This should make these kinds of additions easier to keep around without making them part of the default build for everyone.

PR #127, opened 2026-05-08
[codex] Fix Copilot reasoning effort defaults for new chats

PinguuSS:
## Summary

Fixes the Linux webview patching path that left the Intelligence / reasoning effort selector stuck for new chats when the app is using Copilot auth.

The visible symptom was project-specific: a new chat could open with the reasoning effort stuck on the last/default effort for that project, for example `medium` in one project and `low` in another. Existing threads were less affected because the conversation-specific path can still call `set-model-and-reasoning-for-next-turn`; the broken path was the new-chat/default-model path.

## Root cause

The current upstream webview bundle has a separate Copilot settings path:

- the Copilot default reader only loads `copilot-default-model`
- that same reader hardcodes `reasoningEffort: "medium"`
- the Copilot save branch writes `copilot-default-model` and returns before saving the selected effort
- the model list code collapses Copilot-supported reasoning efforts down to a single `medium` option
- the UI disables reasoning effort entries, and the `/reasoning` command, when `authMethod === "copilot"`

I also checked the suspected stale-route path. The current route atom returns no conversation target for `home` / `new-thread-panel`, so this does not look like a stale `conversationId` being passed into the composer from that route state.

## What changed

The ASAR patcher now patches the relevant webview chunks instead of modifying generated `codex-app/` output directly:

- `use-model-settings-*.js`
  - read `copilot-default-reasoning-effort` next to `copilot-default-model`
  - include the effort key in loading/memoized state
  - persist `copilot-default-reasoning-effort` when saving Copilot defaults
- `font-settings-*.js`
  - keep the full `supportedReasoningEfforts` list for Copilot models instead of forcing only `medium`
- `index-*.js`
  - keep reasoning effort dropdown entries enabled for Copilot auth
  - keep the `/reasoning` command enabled when the normal model/effort prerequisites are present

The new patches follow the existing fail-soft style in `scripts/patch-linux-window-ui.js`: if the upstream minified bundle shape drifts, the installer logs a warning and keeps going.

## Why this is a PR instead of a separate issue

I checked the contribution guidance and searched existing issues/PRs for `reasoning effort`, `intelligence`, and `copilot`. I did not find a matching report. Since this is a focused bug fix with tests and only touches the source-of-truth patcher plus its test file, opening the PR directly seems more useful than creating an issue first and duplicating the same context.

- `scripts/patch-linux-window-ui.js`

No generated `codex-app/` assets were committed.

## Validation

Ran locally:

```bash
bash -n install.sh scripts/lib/*.sh launcher/start.sh.template scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh
git diff --check
```

The targeted patcher test suite passes with `56` tests.

I did not run full package builds (`build-deb`, `build-rpm`, `build-pacman`) because this change only alters ASAR patch matching and test coverage, not package layout or maintainer scripts.

### DIFF
--- a/scripts/patch-linux-window-ui.js
+++ b/scripts/patch-linux-window-ui.js
@@ -1275,6 +1275,129 @@ function applyLinuxComputerUseInstallFlowPatch(currentSource) {
   return currentSource;
 }
 
+function applyCopilotReasoningEffortSettingsPatch(currentSource) {
+  let patchedSource = currentSource;
+
+  const copilotDefaultsPatchMarker = "copilot-default-reasoning-effort`),codexCopilotModelValue=";
+  const copilotDefaultsRegex =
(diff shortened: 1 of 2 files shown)


ilysenko:
Just a quick heads-up: we added a new opt-in `linux-features/` framework for optional patches/features. So if you still want to use or maintain this Copilot reasoning-effort work on top of the latest `main`, you can now move it into a feature and enable it locally through `linux-features/features.json`.

This should make these kinds of additions easier to keep around without making them part of the default build for everyone.

PR #134, opened 2026-05-09
Add i3 window backend for Computer Use

moxw:
## Summary

- Add an i3 window backend for Linux Computer Use using `i3-msg`
- Parse i3 trees into targetable windows and hydrate X11 window PIDs with `xprop`
- Report i3 readiness in doctor output and update the Linux Computer Use docs
- Bump the bundled Computer Use plugin version so installs use a fresh cache entry

## User-visible behavior

On i3/X11, Linux Computer Use can now list windows, focus apps, and verify exact targeted window focus. The backend also resolves the i3 IPC socket directly from `I3SOCK` or `/run/user/$UID/i3/ipc-socket.*`, so it still works when the MCP process cannot use X auth to discover the socket.

## Files edited

- `computer-use-linux/src/windows.rs`
- `computer-use-linux/src/diagnostics.rs`
- `computer-use-linux/src/server.rs`
- `plugins/openai-bundled/plugins/computer-use/.codex-plugin/plugin.json`
- `README.md`
- `AGENTS.md`

## Validation

- `cargo fmt -p codex-computer-use-linux -- --check`
- `cargo test -p codex-computer-use-linux` - 74 passed
- Live `doctor` on i3/X11 after restarting the plugin reported no readiness blockers, with `can_query_windows`, `can_focus_apps`, and `can_focus_windows` all true

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Anything systemd-based should work for the optional auto-updater service (`syste
 | Linux tray + warm-start handoff | ✅ always | Single-instance lock, second-instance window focus |
 | GUI install prompts (`kdialog` / `zenity`) | ✅ if installed | Falls back to interactive terminal prompt |
 | Linux browser annotations | ✅ always | Stored-anchor screenshots, isolated marker rendering |
-| Linux Computer Use | ⚠️ opt-in | Linux Computer Use backend with screen capture, accessibility, window targeting, and input synthesis. The MCP server registers by default; the in-app UI surface is enabled at your discretion — see "Enabling Computer Use UI" below. Validated on Ubuntu/GNOME and KDE Plasma/KWin. |
+| Linux Computer Use | ⚠️ opt-in | Linux Computer Use backend with screen capture, accessibility, window targeting, and input synthesis. The MCP server registers by default; the in-app UI surface is enabled at your discretion — see "Enabling Computer Use UI" below. Validated on Ubuntu/GNOME, KDE Plasma/KWin, Hyprland, and i3. |
 | Server-gated features (e.g. `gpt-5.5`) | 🟡 server-side | OpenAI rolls per-account, not project-controlled. Building a fresh package does not unlock these. |
 
 ## Before you install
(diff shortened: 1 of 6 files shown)


ilysenko:
Thanks!

Since this PR was opened, we merged a Computer Use windowing refactor that changes the backend architecture quite a bit. Each window manager now lives in its own backend module instead of adding more logic to the old shared `windows.rs` file.

Could you please rebase this PR on top of the latest `main` and move the i3 implementation into the new backend layout?

Suggested shape:

- add an i3 backend module under `computer-use-linux/src/windowing/backends/`
- register it in `computer-use-linux/src/windowing/backends/mod.rs`
- add the backend descriptor/order/dispatch entries in `computer-use-linux/src/windowing/registry.rs`
- keep i3-specific helpers inside the i3 backend rather than `diagnostics.rs`
- bump the bundled Computer Use plugin version in `plugins/openai-bundled/plugins/computer-use/.codex-plugin/plugin.json`

A few things worth double-checking while rebasing:

- the fallback i3 socket discovery should not accidentally pick a stale or wrong i3 session when `I3SOCK` is not set
- PID hydration through `xprop` should remain best-effort, since it may fail when the MCP process does not have a usable X11 environment
- after the rebase, please rerun the Rust and smoke tests against the new backend registry architecture

The feature itself makes sense, but I do not think we should merge it in its current form because it targets the old monolithic windowing code and conflicts with the new per-window-manager backend structure.

moxw:
Thanks for the guidance.

I rebased this PR on latest `main` and moved the i3 support into the new backend layout:

- added `computer-use-linux/src/windowing/backends/i3.rs`
- registered it in `backends/mod.rs`
- added the i3 backend descriptor/order/dispatch entries in `windowing/registry.rs`
- kept i3-specific helpers inside the i3 backend
- bumped the bundled Computer Use plugin version
- updated the docs for i3 support

I also double-checked the rebase concerns:

- `I3SOCK` is preferred when set
- fallback socket discovery only searches the current `XDG_RUNTIME_DIR/i3` and picks the newest socket candidate
- PID hydration through `xprop` remains best-effort

Validation is passing after the rebase:

- `XDG_RUNTIME_DIR=/tmp/codex-smoke-runtime bash tests/scripts_smoke.sh`

PR #135, opened 2026-05-09
[codex] Handle Computer Use UI asset drift

mtb-xt:
## Summary

- rebased on latest `main`
- narrowed this PR to only the remaining `apps-*` Computer Use availability asset scan
- added a focused `patchExtractedApp` regression test for `apps-current.js`

## What Changed

Latest `main` already includes the broader upstream drift fixes from the earlier version of this PR:

- `name:e.kn` bundled-plugin gate handling
- `use-in-app-browser-use-availability-*` scanning
- `plugins-availability-*` install-flow scanning
- current install-flow and availability patch shapes

This branch now only adds `apps-*` to the opt-in Computer Use UI availability asset scan:

```text
/^(use-model-settings|apps|use-in-app-browser-use-availability)-.*\.js$/
```

The availability logic itself is unchanged, including the existing `isHostLocal` behavior.

## Validation

```text
node --check scripts/patches/registry.js
git diff --check
```

Result:

```text
tests 62
pass 62
fail 0
```

Also ran Claude adversarial review in `code-vs-plan` mode. Verdict:

```text
VERDICT: APPROVED
```

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -1134,6 +1134,45 @@ test("missing icon asset skips only icon patches", () => {
   }
 });
 
+test("patchExtractedApp scans apps bundles for Computer Use availability when UI is enabled", () => {
+  withIsolatedHome(() => {
+    process.env[COMPUTER_USE_UI_ENV_VAR] = "1";
+    const tempRoot = fs.mkdtempSync(path.join(os.tmpdir(), "codex-computer-use-apps-assets-test-"));
+    try {
+      const buildDir = path.join(tempRoot, ".vite", "build");
+      const assetsDir = path.join(tempRoot, "webview", "assets");
+      fs.mkdirSync(buildDir, { recursive: true });
+      fs.mkdirSync(assetsDir, { recursive: true });
+      fs.writeFileSync(
+        path.join(buildDir, "main.js"),
+        [
+          mainBundlePrefix,
+          "process.platform===`win32`&&k.removeMenu(),",
+          alreadyOpaqueBackgroundBundle,
+          fileManagerBundle,
+          trayBundleFixture(),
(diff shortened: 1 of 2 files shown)


ilysenko:
Thanks!

This PR was targeting the same upstream Computer Use bundle drift that we hit during the latest `install.sh --fresh` testing. Since then, we merged a fix on `main` that already covers most of this: the `name:e.kn` bundled-plugin gate, the newer `use-in-app-browser-use-availability-*` / `plugins-availability-*` assets, and the current install-flow shape.

So I think this PR is now mostly superseded by the latest `main`.

There may still be one useful small follow-up here: scanning `apps-*` bundles for the Computer Use availability patch. That looks safe and worth considering separately. I would be more careful with broadening the availability logic around `isHostLocal`, because that could expose the UI in cases where the backend is not actually usable.

Could you please rebase on the latest `main` and narrow this PR down to only the remaining delta that is still needed? If the only remaining change is the `apps-*` asset scan plus a regression test, that should be much easier to review and merge.

mtb-xt:
Thanks, that makes sense.

I rebased on latest `main` and narrowed this PR down to the remaining delta you called out:

- kept only the `apps-*` addition to the Computer Use availability asset scan
- added one focused `patchExtractedApp` regression test with an `apps-current.js` fixture
- dropped the now-superseded changes for `name:e.kn`, `use-in-app-browser-use-availability-*`, `plugins-availability-*`, and install-flow shape handling

I also left the availability patch logic itself unchanged, so this does not broaden the `isHostLocal` behavior.

Validation:

```text
node --check scripts/patches/registry.js
git diff --check
```

Result:

```text
tests 62
pass 62
fail 0
```

I also had the narrowed diff reviewed with Claude adversarial review in code-vs-plan mode; it reported no actionable findings and approved the change.

Thank you.

PR #137, opened 2026-05-09
Fix Linux pet overlay input passthrough

MuQY1818:
## Summary

- Add a Linux avatar overlay input-shape patch that uses `BrowserWindow.setShape()` for the pet mascot and tray hit regions.
- Expand the input shape to the full overlay while dragging, then restore the shaped hit regions after drag end, layout updates, size changes, and show events.
- Keep a main-process cursor polling fallback for Linux environments where `setShape()` is unavailable or fails.
- Add regression coverage and document the behavior in `AGENTS.md`.

## Why

The floating pet overlay uses transparent-area mouse passthrough so clicks outside the pet can reach underlying windows. On Linux, Electron's forwarded mouse events are not reliable enough when the Codex window is unfocused, after workspace changes, or after switching/reopening pets. That can leave the overlay stuck in `setIgnoreMouseEvents(true)` and make the pet impossible to click or drag.

Shaping the BrowserWindow input region keeps transparent areas click-through at the window-manager level while preserving input on the actual pet controls, without depending on renderer mousemove forwarding or Codex window focus.

## Validation

Ran:

```text
git diff --cached --check
```

Result:

```text
tests 62
pass 62
fail 0
```

Also rebuilt and installed locally, then verified the pet remains clickable and draggable when Codex is unfocused.


MuQY1818:
Addressed the review feedback in the updated commit:

- Keep the fallback polling path active when `setShape()` exists but fails at runtime.
- Avoid applying an empty input shape before the avatar overlay layout is available.

Validation: `node --test scripts/patch-linux-window-ui.test.js` passes with 62 tests.

Egdod:
Thanks for the earlier pet overlay passthrough work here. I hit a remaining Nobara/KDE case where the cursor was inside the avatar overlay BrowserWindow content bounds, but outside the computed `layout.mascot` / `layout.tray` rectangles, so the overlay stayed click-through and the renderer never saw the drag events.

I opened a follow-up with the diagnostic evidence and the locally tested fix direction here: #464

PR #138, opened 2026-05-09
fix Linux explicit quit and stale runtime recovery

DhanushSantosh:
## Summary

This fixes two linked Linux shutdown issues:

- closing the main window and choosing background mode should keep Codex alive in the system tray
- choosing to fully quit from the tray or the app's explicit quit flow should exit the app completely
- after a failed quit or a forced kill, the next launch should recover cleanly instead of getting stuck behind stale runtime state

## Root cause

There were two separate failure modes on Linux:

1. The close-to-tray guard already knew how to keep the app alive in the background, but explicit quit paths were not reliably marking Linux quit-in-progress early enough. That let the window-close logic win in some full-exit flows, so the app stayed resident instead of exiting.
2. Launcher startup cleanup only removed a stale `app.pid`. If the app died badly, stale runtime markers like the launch-action socket or launcher-owned webview pid file could remain behind and interfere with relaunch.

## Fix

Main-process patcher:

- added a dedicated `linux-explicit-quit` patch immediately after `linux-quit-guard`
- patched the tray `Quit` menu item to call `codexLinuxMarkQuitInProgress()` before `app.quit()`
- patched the `quit-app` IPC path to do the same before `app.quit()`
- kept the existing close-to-tray behavior unchanged for normal background-mode window closes

Launcher:

- replaced the narrow startup `clear_stale_pid_file` call with `reconcile_runtime_state`
- preserve runtime markers when a real Electron main process is still alive
- clear stale `app.pid` when no live app exists
- remove a stale `launch-action.sock` when no live app exists
- clear stale launcher-owned `webview.pid` markers when they no longer point at a valid owned webview server

## Why this matters

On Linux, the expected behavior after this change is:

- closing the main window in background mode keeps Codex alive in the tray
- explicit quit from the tray fully exits the app
- explicit quit from the app's quit flow fully exits the app
- after a forced kill, the next launch starts normally instead of being blocked by stale runtime state

## Validation

I ran:

- `make build-app`

I also verified the rebuilt payload directly:

- generated `codex-app/start.sh` contains the new `reconcile_runtime_state()` startup reconciliation
- rebuilt `app.asar` contains both explicit quit markers before `app.quit()` in the tray quit and `quit-app` paths

## Notes

This keeps the patch narrow in the hot Linux patcher surface:

- explicit quit is patched only at the concrete tray quit and `quit-app` call sites found in the current upstream bundle
- startup reconciliation only removes stale runtime markers when no live Electron app is actually running

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -1082,6 +1082,31 @@ clear_stale_pid_file() {
     fi
 }
 
+reconcile_runtime_state() {
+    local live_app_pid=""
+    local webview_pid=""
+
+    if live_app_pid="$(find_running_app_pid)" || { [ -S "$LAUNCH_ACTION_SOCKET" ] && live_app_pid="$(discover_running_app_pid)"; }; then
+        echo "$live_app_pid" > "$APP_PID_FILE"
+        return 0
+    fi
+
+    clear_stale_pid_file
+
+    if [ -e "$LAUNCH_ACTION_SOCKET" ]; then
+        rm -f "$LAUNCH_ACTION_SOCKET"
+    fi
+
+    if [ ! -f "$WEBVIEW_PID_FILE" ]; then
+        return 0
+    fi
+
+    webview_pid="$(cat "$WEBVIEW_PID_FILE" 2>/dev/null || true)"
+    if [ -z "$webview_pid" ] || { ! pid_is_webview_server "$webview_pid" && ! pid_is_stale_webview_server "$webview_pid"; }; then
+        rm -f "$WEBVIEW_PID_FILE"
+    fi
+}
+
 set_electron_defaults() {
     ELECTRON_OZONE_PLATFORM=""
     ELECTRON_OZONE_HINT="auto"
@@ -1241,7 +1266,7 @@ configure_side_by_side_app_env
 load_packaged_runtime_helper
 prepend_managed_node_runtime_to_path
 register_url_scheme_handlers
-clear_stale_pid_file
+reconcile_runtime_state
 detect_warm_start
 trap cleanup_launcher EXIT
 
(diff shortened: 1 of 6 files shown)


ilysenko:
Thanks, this looks good overall and the behavior this fixes makes sense.

One thing I would like to tighten before merging: `applyLinuxExplicitQuitPatch` currently patches two required quit surfaces, the tray Quit item and the `quit-app` IPC path, but they are reported as a single required patch.

If future upstream drift causes only one of those two call sites to be patched while the other only emits a warning, the registry can still record `linux-explicit-quit` as `applied` because the source changed. That means CI could stay green while one explicit quit path is broken again.

Could you please adjust this so partial drift cannot pass silently? Either approach would work:

- split this into two registry patches, for example `linux-explicit-tray-quit` and `linux-explicit-ipc-quit`
- or make `applyLinuxExplicitQuitPatch` fail hard when it recognizes a required quit surface but cannot patch it

After that I think this should be good to merge.

ilysenko:
Thanks again. We merged #137, which touches the same patcher files, so this PR is now conflicting with the latest `main`.

Could you please rebase this branch on top of the current `main` and keep the explicit-quit changes adjusted after the avatar overlay patch? After the rebase, please rerun:

```text
bash tests/scripts_smoke.sh
```

The remaining review point from the previous comment still applies: please make sure partial drift in the two explicit-quit call sites cannot pass CI silently.

DhanushSantosh:
@ilysenko on it.

DhanushSantosh:
I rebased this branch onto the current `main` and addressed the remaining review point.

I split the explicit quit coverage into two separate required patch surfaces:
- `linux-explicit-tray-quit`
- `linux-explicit-ipc-quit`

That way, if either call site drifts upstream, it will no longer be able to pass CI just because the other one still patches successfully.

I also reran the requested checks after the rebase:

Both pass on the rebased branch.

ilysenko:
Thanks!

PR #136, opened 2026-05-09
[codex] refactor Linux Computer Use window backends

ilysenko:
## Summary

This PR reorganizes the Linux Computer Use windowing layer so each desktop/window-manager backend lives in its own module, while shared routing, target resolution, focus validation, and backend ordering stay in a central registry.

Included changes:

- Split the previous monolithic `windows.rs` implementation into `computer-use-linux/src/windowing/` modules.
- Added backend-specific modules for GNOME, COSMIC, KWin/Plasma, and Hyprland.
- Added a window backend registry that owns backend order, probe metadata, list notes, activation dispatch, and diagnostics mapping.
- Kept `windows.rs` as a compatibility re-export so existing call sites and tests can migrate gradually.
- Updated diagnostics and server output to use registry metadata instead of hardcoded backend branches.
- Updated repository maintenance notes to document where future backend work should go.
- Fixed current upstream Computer Use ASAR patch drift where plugin descriptors can use imported namespace constants such as `name:e.kn`.
- Updated the opt-in Computer Use UI patches for current webview bundle names/shapes.
- Fixed `--inspect --report-dir` handling so relative report paths are anchored to the caller directory instead of the temporary extraction directory.

## Why

The Computer Use backend was becoming difficult to extend because GNOME, COSMIC, KWin, Hyprland, target resolution, diagnostics, and shared types all lived in the same large module. That made independent PRs likely to collide, especially when adding new desktop-manager support.

This structure gives each backend a clear ownership boundary and keeps shared behavior in one place, so future backend PRs can be smaller and less likely to conflict.

The ASAR patch updates are included because a fresh build against the current upstream DMG exposed drift in the Computer Use plugin gate and UI availability/install-flow bundles. Without those fixes, `CODEX_LINUX_ENABLE_COMPUTER_USE_UI=1 ./install.sh --fresh` could fail during patching or install successfully without enabling the expected UI paths.

## Validation

Ran:

```bash
CODEX_LINUX_ENABLE_COMPUTER_USE_UI=1 ./install.sh --inspect --report-dir /tmp/codex-rebuild-debug-fixed
CODEX_LINUX_ENABLE_COMPUTER_USE_UI=1 ./install.sh --fresh --reuse-dmg
node --test scripts/patch-linux-window-ui.test.js
bash tests/scripts_smoke.sh
cargo test -p codex-computer-use-linux
bash -n scripts/lib/rebuild-report.sh
```

Also verified the inspect patch report marks all Computer Use patches as applied:

[...]

PR #139, opened 2026-05-09
Improve Linux Computer Use fallback and input handling

PinguuSS:
Tightens the Linux Computer Use backend after the windowing refactor landed in #136, now focused on fallback and readiness robustness.

What changed:

- The window backend registry keeps trying later backends when an earlier backend returns an empty window list, instead of treating that empty result as final
- The i3 backend from #134 stays in the registry order, diagnostics, activation path, and tests after the rebase on current `main`
- Runtime ydotool setup now only exports `YDOTOOL_SOCKET` when the socket can actually be connected to, so stale socket files are skipped
- `doctor` reports ydotool socket connection failures with useful detail, including permission problems
- `can_send_development_input` now requires `ydotool`, a running `ydotoold`, and a connectable ydotoold socket, matching the current runtime path that still calls the `ydotool` CLI
- `/dev/uinput` stays in diagnostics as useful detail, but it no longer makes input readiness pass by itself
- `press_key` stays on raw `ydotool key` events such as `29:1 25:1 25:0 29:0`
- `type_text` sends literal text through stdin via `ydotool type --file -`, avoiding shell quoting issues without changing the `ydotool key` protocol
- Large KWin-generated `window_id` values tolerate JSON/JavaScript number rounding on the client side, while still preferring exact id matches first
- The flaky update-manager lock test was made less timing-sensitive, and CLI lookup tests no longer depend on a system-installed `codex`

What stayed out after review feedback:

- Symbolic `ydotool key` names are not part of this PR anymore
- The PageUp/PageDown keyboard-scroll fallback was removed from this PR; scroll remains on the existing portal path and ydotool wheel fallback for now

How this fits the #136 refactor:

- Backend ordering, fallback labels, and capability metadata stay in `computer-use-linux/src/windowing/registry.rs`
- Backend-specific i3 behavior remains in `computer-use-linux/src/windowing/backends/i3.rs`
- Target resolution, including the rounded large-window-id fallback, stays in `computer-use-linux/src/windowing/target.rs`
- Shared MCP input handling stays in `computer-use-linux/src/server.rs`, with diagnostics matching the runtime ydotool socket requirement

Validation run locally:

```bash
rtk git diff --check
rtk cargo fmt --check
rtk cargo clippy --workspace --all-targets -- -D warnings
rtk cargo test -p codex-computer-use-linux
rtk cargo test -p codex-update-manager
rtk cargo test --workspace --all-targets
```

All passed locally after the scope reduction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,35 @@ sudo usermod -a -G input "$USER"
 
 On Ubuntu 24.04, the `ydotoold` package may install `/usr/bin/ydotoold` without a systemd unit. In that case, create or install a `ydotoold.service` unit before running `systemctl enable --now ydotoold`.
 
+If `doctor` reports `ydotoold` as running but `ydotool_socket` is `Permission denied`, the daemon is reachable only by root. This unit keeps the daemon privileged enough to open `/dev/uinput`, but makes the socket usable by desktop users in the `input` group:
+
+```ini
+[Unit]
+Description=ydotool daemon
+Documentation=man:ydotool(1)
+After=multi-user.target
+
+[Service]
+Type=simple
+Group=input
+UMask=0007
+ExecStartPre=/usr/bin/rm -f /tmp/.ydotool_socket
+ExecStart=/usr/bin/ydotoold
+ExecStartPost=/bin/sh -c 'for i in $(seq 1 50); do [ -S /tmp/.ydotool_socket ] && chgrp input /tmp/.ydotool_socket && chmod 0660 /tmp/.ydotool_socket && exit 0; sleep 0.1; done; exit 1'
+Restart=on-failure
+RestartSec=2
+
+[Install]
+WantedBy=multi-user.target
+```
+
+Save it as `/etc/systemd/system/ydotoold.service`, then run:
+
+```bash
+sudo systemctl daemon-reload
+sudo systemctl enable --now ydotoold
+```
+
 A working XDG Desktop Portal implementation is needed if you are not on GNOME — `xdg-desktop-portal-kde` for KDE Plasma, `xdg-desktop-portal-wlr` for sway / Hyprland, or your distro's preferred portal backend for i3. GNOME ships a working portal by default.
 
 ### Verifying readiness
@@ -363,6 +392,7 @@ make clean-state
 | Stale install / cached DMG | `./install.sh --fresh` removes the existing install dir and re-downloads |
 | Computer Use plugin invisible in UI | Most likely the OpenAI per-account Statsig rollout (`computerUse` feature flag) hasn't been enabled for your account. Building / reinstalling does not change this |
 | Computer Use `doctor` reports `ydotool not running` | `sudo systemctl enable --now ydotoold` and add your user to the `input` group |
+| Computer Use `doctor` reports `ydotool_socket: Permission denied` | The daemon socket is root-only. Use the `ydotoold.service` example above so `/tmp/.ydotool_socket` becomes `root:input` with `0660` permissions |
 | `ConnectTimeoutError` for `www.electronjs.org` during `@electron/rebuild` | Re-run `make build-app`; the installer now uses `https://artifacts.electronjs.org/headers/dist` for Electron headers by default |
 | Computer Use AT-SPI tree empty | Run `codex-computer-use-linux setup` to flip GNOME accessibility on, then restart the target app |
 | `codex-update-manager` keeps running after package removal | `systemctl --user disable --now codex-update-manager.service` once in the affected session, then confirm `/opt/codex-desktop` is gone |
--- a/computer-use-linux/src/diagnostics.rs
+++ b/computer-use-linux/src/diagnostics.rs
@@ -7,6 +7,8 @@ use serde::Serialize;
 use std::{
     collections::{BTreeMap, HashMap},
     env, fs,
+    fs::OpenOptions,
+    os::unix::net::UnixStream,
     path::{Path, PathBuf},
     process::Command,
 };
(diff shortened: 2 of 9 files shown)


ilysenko:
Thanks for the work here!

After #134 was merged, this PR is now conflicting with the latest `main`, most likely around the windowing registry changes. Could you please rebase this branch on top of the current `main` and resolve the conflicts while keeping the new i3 backend entries intact?

Also, CI currently has one failing check: `Rust and Smoke Tests`. The Computer Use tests appear to pass in the log, but the job fails on:

```text
app::tests::held_check_lock_blocks_second_acquire_until_drop
```

in `codex-update-manager`.

Could you please rerun/fix that after the rebase so the full CI is green again? The package build jobs are passing already.

ilysenko:
Thanks.

This PR is moving in the right direction, especially the backend fallback change and the stronger ydotool diagnostics, but I think there are a few risky parts that should be fixed before merge:

1. Please rebase on top of the latest main after #134, and make sure the new i3 backend is preserved in the windowing backend registry/tests.

2. The diagnostics and runtime ydotool socket handling should use the same readiness logic. Right now diagnostics may report input as ready when /dev/uinput is usable, but the runtime can still pick the first existing YDOTOOL_SOCKET even if that socket is stale or not connectable. That can make Computer Use look ready while click/type still fail at runtime. Please make the runtime only use a socket that can actually be connected to, or avoid setting YDOTOOL_SOCKET when it should fall back to /dev/uinput.

3. The press_key changes are broader than the PR title suggests. Switching from raw ydotool keycodes to symbolic key names affects input handling across all desktop backends, not only Plasma fallback. Please either split that into a separate PR, or add enough validation/tests to show that the symbolic names and combinations work with the ydotool versions we support.

4. Please make CI green after the rebase.

PinguuSS:
Quick update on the items from https://github.com/ilysenko/codex-desktop-linux/pull/139#issuecomment-4412610190:

I rebased this on top of current `main` after #134 and kept the i3 backend in the registry/tests. The ydotool diagnostics and runtime path now use the same socket-readiness rule, so runtime only sets `YDOTOOL_SOCKET` when the socket is actually connectable.

I also kept the `press_key` symbolic-name change in this PR and added tests around the key sequences, navigation keys, `Space`, and stale socket selection. While doing a manual KDE/Kate pass, I noticed large KWin `window_id` values can get rounded by JSON/JS clients, so I added a small target-resolution fallback for that too.

Local validation passed, and the latest CI run is green now. I also filled in the PR description with the full context and manual test notes. This should be ready for review when you have a chance.

ilysenko:
Thanks for the update. The rebase looks good, CI is green, and several parts of this PR look useful: skipping empty backend results, preserving i3 after the rebase, selecting only connectable ydotool sockets at runtime, rounded large window_id handling, and the README note about ydotoold socket permissions.

I think the PR would be much safer and easier to merge if we keep it focused on fallback/readiness robustness and avoid changing the keyboard input protocol in the same change.

The main thing I would like to adjust is the `ydotool key` handling. This PR changes `press_key` from raw keycode events like `29:1 25:1 25:0 29:0` to symbolic names like `ctrl+l`, `pagedown`, `super`, etc. However, `ydotool key --help` documents raw `<keycode>:<pressed>` events, and non-interpretable values are ignored / treated as delays. That means `press_key`, keyboard-scroll, and shortcuts like `Ctrl+L` may silently stop working on the ydotool versions we support.

Could you please keep the raw keycode event generation for `ydotool key` in this PR? If we want symbolic key support later, I think that should be a separate PR with runtime compatibility detection and a raw-keycode fallback.

Related to that, I would also remove the new keyboard-scroll path from this PR. It depends on `PageUp` / `PageDown` going through the new symbolic key path, so it has the same compatibility risk. The existing scroll fallback can stay as-is for now.

One more thing: diagnostics readiness should match the actual runtime behavior. The diagnostics currently treat input as ready when `ydotool` exists and either a connectable `ydotoold` socket exists or `/dev/uinput` is read/write. But the runtime still calls the `ydotool` CLI; it does not directly write to `/dev/uinput`. So without a usable `ydotoold` socket, diagnostics can report input as ready while click/type/press_key still fail. I would make `can_send_development_input` require `ydotool.ok && ydotoold.ok && ydotool_socket.ok` for now, and keep `/dev/uinput` as diagnostic detail only unless a real direct `/dev/uinput` runtime fallback is added.

So the version I would be comfortable merging is roughly:

- keep: skip empty window backend results;
- keep: i3 preserved in the registry/order/tests;
- keep: runtime connectable ydotool socket selection;
- keep: diagnostics for socket permission issues;
- keep: rounded large window_id handling;
- keep: README ydotoold socket permission guidance;
- keep: updater test fixes if they are needed for CI;
- remove/revert for now: symbolic `ydotool key` generation;
- remove/revert for now: keyboard-scroll via `PageUp` / `PageDown`;
- adjust: input readiness should require a connectable ydotoold socket.

I think that would make the PR much smaller, safer, and easier to merge.

PinguuSS:
Thanks, I reduced the scope along the lines you suggested.

The latest push keeps `press_key` on raw `ydotool key` events, so combinations still generate values like `29:1 25:1 25:0 29:0`. I also removed the PageUp/PageDown keyboard-scroll fallback from this PR, since that depended on the symbolic key path.

I adjusted diagnostics readiness to match the current runtime behavior too: `can_send_development_input` now requires `ydotool`, `ydotoold`, and a connectable ydotoold socket. `/dev/uinput` is still reported as diagnostic detail, but it is no longer enough to make readiness pass by itself while the runtime still goes through the `ydotool` CLI.

Kept the narrower pieces from your list: skipping empty backend results, preserving i3, connectable socket selection at runtime, socket permission diagnostics, rounded large `window_id` handling, the README socket guidance, and the updater test fixes.

Local validation passed:

```bash
rtk git diff --check
rtk cargo fmt --check
rtk cargo clippy --workspace --all-targets -- -D warnings
rtk cargo test -p codex-computer-use-linux
rtk cargo test -p codex-update-manager
rtk cargo test --workspace --all-targets
```

PR #140, opened 2026-05-09
Add Linux Chrome extension native host

moxw:
## Summary
- bundle the upstream Chrome plugin into Linux builds alongside Browser Use and Computer Use
- add a Rust native-messaging host that bridges Chrome extension frames to Browser Use Unix-socket clients and emits turnEnded from rollout files
- patch bundled Chrome plugin scripts for Linux and Brave profiles, manifests, diagnostics, and site-status allowlist fallback
- make the Electron settings connected badge detect Chrome/Brave/Chromium Linux profile roots

## User-visible behavior
- Codex Desktop for Linux can install and use the Chrome extension native host on Google Chrome or Brave
- Brave installs can show as connected in Codex settings instead of being reported as missing
- launcher startup keeps the user plugin cache, bundled marketplace link, and native-host manifests in sync

## Validation
- I tested the Chrome/Brave plugin locally and confirmed it works
- cargo fmt -p codex-computer-use-linux --check
- cargo check -p codex-computer-use-linux
- node --check scripts/lib/patch-chrome-plugin.js scripts/patches/main-process.js scripts/patches/registry.js scripts/patch-linux-window-ui.js scripts/patch-linux-window-ui.test.js
- bash -n install.sh scripts/lib/*.sh launcher/start.sh.template scripts/install-deps.sh scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh tests/scripts_smoke.sh
- bash tests/scripts_smoke.sh
- git diff --check

Post-rebase validation also reran node --test scripts/patch-linux-window-ui.test.js, cargo test -p codex-computer-use-linux, bash tests/scripts_smoke.sh, node --check, bash -n, and git diff --check origin/main...HEAD.


ilysenko:
Thanks, this looks good overall and the direction makes sense. I do have a few concerns in these areas before we merge it:

1. Chromium support looks partial right now. The patch scans Chromium profiles / commands, but the launcher only writes native-host manifests for Google Chrome and Brave. Could you either fully support Chromium by installing its native-host manifest too, or remove Chromium from the Linux detection path for now? Otherwise Chromium may look supported while native messaging cannot actually work.

2. The Chrome extension id appears to be hardcoded in the launcher manifest writer. If the upstream bundled Chrome plugin changes its extension id, the Linux native host could silently stop working. Could we derive the extension id from the staged upstream Chrome plugin resources / manifest logic instead?

3. I am not fully sure about the native host client routing. For Chrome-originated requests, the Rust host sends the request to the first client from a HashMap, which is nondeterministic. Could you either enforce a single connected client explicitly, or track the selected client_id in pending requests and clean up pending state on disconnect?

4. Could you add focused tests for these cases: Chromium support vs no Chromium support, derived extension id in the native-host manifest, and native-host behavior with multiple clients / disconnects?

The PR looks promising, but my main concern is that the UI/settings could report the Chrome extension as installed/connected while the actual native messaging bridge does not work for some users.

moxw:
Thanks, addressed these in `58a32c6`.

- Chromium is now fully supported in the Linux path: the launcher writes Chromium native-host manifests, and the plugin patcher covers Chromium profiles, commands, running-process checks, manifest checks, and diagnostics.
- The launcher now derives `extensionId` / `extensionHostName` from the staged plugin metadata (`scripts/extension-id.json`, with `installManifest.mjs` fallback) instead of hardcoding them.
- The Rust native host now rejects Chrome-originated requests unless exactly one browser client is connected, and disconnect cleanup removes both Chrome- and client-originated pending request state.
- Added focused coverage for Chromium staging/manifest sync, derived native-host manifest metadata, and native-host multiple-client/disconnect behavior.

Local validation passed, and the PR checks are green now.

PR #142, opened 2026-05-09
Add Linux Zed opener support

moxw:
## Summary
- add a Linux ASAR patch for the upstream Zed opener block
- detect common Linux Zed CLI names: `zed`, `zeditor`, `zedit`, and `zed-cli`
- register the patch in the main bundle pipeline and cover direct/full-bundle patching in tests

## User-visible behavior
- Linux builds can expose Zed as an editor opener when a supported Zed CLI binary is available on PATH
- Existing macOS and Windows Zed opener behavior is preserved

## Validation
- `node --check scripts/patches/registry.js`
- `git diff --check`
- live ASAR extraction check: `linux-zed-opener` applied once and reported `already-applied` on rerun

Full smoke tests and package builds were not run because this change is scoped to the JS ASAR patcher and its unit coverage.

### DIFF
--- a/linux-features/zed-opener/README.md
+++ b/linux-features/zed-opener/README.md
@@ -1,0 +1,49 @@
+# Zed Opener
+
+Adds Zed as an opt-in Linux editor opener in Codex Desktop. The patch extends
+the upstream Zed opener block with a Linux platform entry and reuses the
+upstream `path:line:column` argument builder.
+
+This feature is opt-in. The loader reads enabled feature ids from the root
+config at `linux-features/features.json`, then loads this feature's manifest
+from `linux-features/zed-opener/feature.json`.
+
+To enable it locally, create the root config if needed:
+
+```bash
+cp linux-features/features.example.json linux-features/features.json
+```
+
+Then list `zed-opener` in `linux-features/features.json`:
+
+```json
+{
+  "enabled": [
+    "zed-opener"
+  ]
+}
+```
+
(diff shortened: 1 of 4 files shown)


ilysenko:
Thanks for working on this.

We added a new opt-in `linux-features/` framework for Linux-specific integrations like this one. Since Zed opener support is useful but not something every Linux build needs by default, I think it would fit better as an optional feature.

The same probably applies to the Chrome extension work from #140 as well: since it is not part of the baseline Linux wrapper behavior, it would be better to move it into `linux-features/` too and make it opt-in. That should give us fewer reasons for future PR conflicts, regressions, and issues around integrations that only some users need.

If you’re interested, feel free to rework this PR so the Zed opener patch lives under `linux-features/` with its own feature manifest, README, and tests.

moxw:
Reworked this based on the linux-features feedback: Zed opener support now lives under linux-features/zed-opener as a disabled-by-default opt-in feature with feature.json, a feature README, and feature-local tests. The baseline ASAR patch registry no longer includes the Zed patch.

Validation run:
- node --test linux-features/zed-opener/test.js
- node --check scripts/patches/main-process.js scripts/patches/registry.js scripts/patch-linux-window-ui.js linux-features/zed-opener/patch.js linux-features/zed-opener/test.js
- git diff --check

CI is green for Debian, RPM, Pacman, and Rust/smoke checks.

PR #170, opened 2026-05-13
Support datagram ydotool sockets

mosesmrima:
## Summary
- Accept Unix datagram sockets when probing ydotoold sockets in diagnostics
- Use the same datagram-aware socket selection for runtime ydotool commands
- Add coverage for datagram sockets while preserving existing stream-socket behavior

## Why
Debian packages ydotoold with a Unix datagram socket at /run/user/UID/.ydotool_socket. The existing UnixStream-only probe reports this as Protocol wrong type for socket, so Linux Computer Use incorrectly reports ydotool input as unavailable even though ydotool itself can use the socket.

## Validation
- cargo fmt --all --check
- git diff --check
- Live Debian 13 check: codex-computer-use-linux doctor reports ydotool_socket connectable and can_send_development_input=true

### DIFF
--- a/computer-use-linux/src/diagnostics.rs
+++ b/computer-use-linux/src/diagnostics.rs
@@ -598,7 +598,7 @@ fn socket_connect_result(path: &Path) -> std::result::Result<(), String> {
     }
 
     match UnixStream::connect(path) {
-        Ok(_) => return Ok(()),
+        Ok(_) => Ok(()),
         Err(stream_error) => {
             match UnixDatagram::unbound().and_then(|socket| socket.connect(path)) {
                 Ok(()) => Ok(()),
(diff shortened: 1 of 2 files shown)


ilysenko:
The following CI checks are failing:

- **Rust and Smoke Tests**: failure — https://github.com/ilysenko/codex-desktop-linux/actions/runs/25800600184/job/75789265812

ilysenko:
Thanks, this looks directionally good. The CI failure is just clippy, not a larger test failure.

Could you please apply the two clippy suggestions?

- In `computer-use-linux/src/diagnostics.rs`, remove the needless `return`:

```rust
Ok(_) => Ok(()),
```

- In `computer-use-linux/src/server.rs`, replace the redundant closure with the function directly:

```rust
.find(ydotool_socket_connects)
```

After that, `cargo clippy --workspace --all-targets -- -D warnings` should get past the current failure.

PR #174, opened 2026-05-13
Fix layout-safe Linux type_text

### LINKED ISSUE #132: type_text produces wrong characters under non-US keyboard layouts
## Environment                                                                                                                                                                                                                                                                          
                                                                  
  * OS: CachyOS (Arch-based)
  * Kernel: 7.0.2-2-cachyos                                                                                                                                                                                                                                                               
  * Desktop: KDE Plasma 6 (Wayland)                                                                                                                                                                                                                                                       
  * Desktop: KDE Plasma 6 (Wayland)
  * Node.js: v22.22.2
  * Electron: v41.2.0
  * Active keyboard layout: German (de)

  ## Problem

  `type_text` writes wrong characters when the active keyboard layout is not US.
  Calling:

  ```text
  mcp__computer_use__.type_text { text: "https://example.com:8080/page#anchor" }
  ```

  produces text with wrong characters at `:`, `#`, and other shifted symbols.

  A clipboard-paste workaround (set clipboard, then send `Ctrl+v`) produces the
  correct text.

  ## Cause

  The Linux Computer Use backend converts each character to a hardware scancode in:

  ```text
  computer-use-linux/src/server.rs
  ```

  ```rust
  fn keycode(key: &str) -> Option<u16> {
      // ...
      value if value.len() == 1 => keycode_for_ascii(value.as_bytes()[0] as char),
      // ...
  }

  fn keycode_for_ascii(value: char) -> Option<u16> { /* ... */ }
  ```

  These scancodes are sent through
  `org.freedesktop.portal.RemoteDesktop.NotifyKeyboardKeycode`. The compositor
  maps the scancode using the currently active keyboard layout. Since
  `keycode_for_ascii` returns codes that assume a US layout, the resulting
  characters are wrong on any other layout.

  ## Workaround

  Inject text via clipboard and send `Ctrl+v` through `press_key`. This bypasses
  the layout-dependent scancode path entirely.

  ## Result

  With the clipboard workaround, all characters are typed correctly regardless
  of active keyboard layout. Without it, `type_text` is unusable on
  non-US layouts for anything beyond plain lowercase letters and digits.

  ## Suggestion

  Switch the keyboard input path from `NotifyKeyboardKeycode` to
  `NotifyKeyboardKeysym`. The keysym variant carries the symbol value directly
  and is interpreted by the compositor without going through the layout
  mapping. The portal exposes both methods on the same `RemoteDesktop`
  interface, so the change is local to
  `computer-use-linux/src/remote_desktop.rs` and the call site in `type_text`.

PinguuSS:
## Summary
- Add a keyboard-capable XDG RemoteDesktop portal session for Linux Computer Use.
- Route literal `type_text` through layout-safe text paths on Wayland before falling back to `ydotool type --file -`.
- Keep `press_key` on the existing raw `ydotool key <keycode>:<pressed>` path.

Fixes #132.

## Why
`ydotool type` is layout-dependent, so non-US layouts can emit the wrong literal characters for strings containing symbols like `:`, `/`, `#`, `?`, and `@`.

The first Wayland path uses `NotifyKeyboardKeysym` so literal text is converted to X11 keysyms before dispatch. Local KDE Plasma validation showed KDE still layout-maps those keysyms under a German layout, so Plasma uses Klipper plus portal-delivered `Ctrl+V` to preserve exact text while still avoiding synthetic layout-dependent typing.

## Source of truth
- `computer-use-linux/src/remote_desktop.rs`
- `computer-use-linux/src/server.rs`
- `computer-use-linux/Cargo.toml`
- `Cargo.lock`

## User-visible behavior
- On Wayland, `type_text` tries a layout-safe portal-backed text path unless `CODEX_COMPUTER_USE_FORCE_YDOTOOL_KEYBOARD=1` is set.
- On KDE Plasma Wayland, `type_text` uses the KDE clipboard integration and restores the previous clipboard contents after paste.
- If portal permission/session setup fails before text is emitted, behavior falls back to the existing `ydotool` text path.
- `press_key`, keyboard scroll behavior, window backends, diagnostics readiness, and clipboard tool behavior are unchanged.

## Validation
- `rtk cargo fmt -p codex-computer-use-linux -- --check`
- `rtk cargo test -p codex-computer-use-linux`
- `rtk cargo clippy -p codex-computer-use-linux --all-targets -- -D warnings`
- `rtk git diff --check`

## Manual validation
KDE Plasma 6 Wayland with active German keyboard layout:
- Opened a focused editable field.
- Verified pure `NotifyKeyboardKeysym` still produced layout-mapped output: `https_@@example.com_8080@page§anchor`.
- Verified the fixed Plasma path returned `Action pasted through KDE clipboard integration.`
- Verified the field contained exactly `https://example.com:8080/page#anchor`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -467,6 +467,7 @@ dependencies = [
  "tokio",
  "wayland-client",
  "wayland-protocols",
+ "xkeysym",
  "zbus",
 ]
 
@@ -3417,6 +3418,12 @@ name = "writeable"
 version = "0.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9edde0db4769d2dc68579893f2306b26c6ecfbe0ef499b013d731b7b9247e0b9"
+
+[[package]]
+name = "xkeysym"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b9cc00251562a284751c9973bace760d86c0276c471b4be569fe6b068ee97a56"
 
 [[package]]
 name = "yoke"
--- a/computer-use-linux/Cargo.toml
+++ b/computer-use-linux/Cargo.toml
@@ -25,4 +25,5 @@ serde_json = "1.0.149"
 tokio = { version = "1.51.1", features = ["macros", "rt", "time"] }
 wayland-client = "0.31.11"
 wayland-protocols = { version = "0.32.9", features = ["client", "staging"] }
+xkeysym = "0.2.1"
 zbus = "5.14.0"
--- a/computer-use-linux/src/remote_desktop.rs
+++ b/computer-use-linux/src/remote_desktop.rs
@@ -2,6 +2,7 @@ use crate::diagnostics::hydrate_session_bus_env;
 use anyhow::{bail, Context, Result};
 use futures_util::StreamExt;
 use std::{collections::HashMap, time::Duration};
+use xkeysym::Keysym;
 use zbus::{
     proxy::SignalStream,
     zvariant::{OwnedObjectPath, OwnedValue, Value},
@@ -15,9 +16,13 @@ const PORTAL_SCREENCAST_INTERFACE: &str = "org.freedesktop.portal.ScreenCast";
 const PORTAL_REQUEST_INTERFACE: &str = "org.freedesktop.portal.Request";
 const REQUEST_TIMEOUT: Duration = Duration::from_secs(120);
 
+const DEVICE_KEYBOARD: u32 = 1;
 const DEVICE_POINTER: u32 = 2;
 const SOURCE_MONITOR: u32 = 1;
 const CURSOR_MODE_HIDDEN: u32 = 1;
+
+const KEY_RELEASED: u32 = 0;
+const KEY_PRESSED: u32 = 1;
 
 const POINTER_BUTTON_RELEASED: u32 = 0;
 const POINTER_BUTTON_PRESSED: u32 = 1;
@@ -38,6 +43,12 @@ pub struct PortalPointerSession {
     connection: Connection,
     session_handle: OwnedObjectPath,
     streams: Vec<PortalStream>,
+}
+
+#[derive(Clone)]
+pub struct PortalKeyboardSession {
+    connection: Connection,
+    session_handle: OwnedObjectPath,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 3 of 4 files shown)


ilysenko (commented):

🔴 `high` · @PinguuSS action needed · 3 open items

**Summary:** Targeted layout-safe type_text fix via portal NotifyKeyboardKeysym plus a KDE clipboard fallback; localized to computer-use-linux but in the locale-sensitive keyboard-input area.

@PinguuSS please address:
- KDE clipboard branch propagates portal session-start errors to the user, but the non-KDE keysym branch silently falls back to ydotool. PR body promises ydotool…
- The KDE path requires `qdbus6` and `org.kde.klipper`, but qdbus6 is not declared in packaging/linux/{control,*.spec,PKGBUILD.template} and Klipper may be disab…
- Replace the magic evdev keycodes `&[29]` (Ctrl) and `47` (V) in the press_keycode_chord call with named constants or a `// KEY_LEFTCTRL / KEY_V` comment so the…

Refs: `computer-use-linux/src/server.rs:691`, `computer-use-linux/src/server.rs:723`, `computer-use-linux/src/server.rs:2079`, `computer-use-linux/src/server.rs:2049`, `computer-use-linux/src/server.rs:2053`, `computer-use-linux/src/remote_desktop.rs:227`, `computer-use-linux/src/remote_desktop.rs:244`, `computer-use-linux/Cargo.toml:28`

</details>

ilysenko (changes requested):
Thanks for tackling this. I re-reviewed the current diff and I agree this is the right direction, but I would not merge it yet.

Blocking issues:

- `computer-use-linux/src/server.rs:667`: the KDE path returns an error when `ensure_portal_keyboard_session()` fails. That happens before any text is emitted, so it should fall back to the existing `ydotool` path just like the non-KDE keysym branch does. As written, Plasma Wayland users who deny portal keyboard access, lack a working portal, or hit a session setup error lose `type_text` even though the old backend could still work.

- `computer-use-linux/src/server.rs:2046`: the KDE path also hard-requires `qdbus6` and `org.kde.klipper` before text is emitted. `qdbus6` is not declared as a package dependency, and Klipper can be disabled or unavailable, so this should either be detected and fall back before mutating the clipboard, or be made an explicit supported dependency with diagnostics. Fallback is probably the safer behavior here.

- `computer-use-linux/src/server.rs:2059`: if the paste succeeds but restoring the previous clipboard fails, the action currently returns `ok: false`. That can make the caller retry and paste the same text twice even though the requested text already reached the target. Please report success once paste succeeds and surface the restore failure as a warning/message/log instead of a failed action.

Small cleanup while touching this: please replace or document the magic keycodes in `press_keycode_chord(session, &[29], 47)` so future changes know these are Ctrl+V evdev codes.

PinguuSS:
Thanks. I pushed `eebf371` addressing the blocking points.

Changes made:
- KDE portal session startup failures now fall back to the existing `ydotool type --file -` path before any text is emitted.
- Missing/unavailable `qdbus6` or `org.kde.klipper` now also falls back before mutating the clipboard, so this does not become a new hard packaging/runtime dependency.
- Once the paste succeeds, failure to restore the previous clipboard contents is reported as a warning in the success message instead of returning `ok: false`, avoiding a caller retry that could duplicate the pasted text.
- Replaced the Ctrl+V magic numbers with named evdev constants (`EVDEV_KEY_LEFTCTRL` / `EVDEV_KEY_V`).

Validation:
- `rtk cargo fmt -p codex-computer-use-linux -- --check`
- `rtk cargo test -p codex-computer-use-linux`
- `rtk cargo clippy -p codex-computer-use-linux --all-targets -- -D warnings`
- `rtk git diff --check`

ilysenko (commented):
## Review Mate: follow-up

🟢 `low`

**Summary:** Follow-up resolves all three prior blocking items: KDE path now falls back to ydotool, restore failure becomes a warning, and Ctrl+V magic numbers are named constants.

✅ Fixed
- KDE clipboard branch now falls back to ydotool on portal keyboard session failure via can_fallback_to_ydotool (server.rs:683-691, 694-695)
- Missing/unavailable qdbus6 or org.kde.klipper now classified as before_text_input → ydotool fallback, avoiding a new hard packaging dep (server.rs:2072-2073, 2…

Refs: `computer-use-linux/src/server.rs:667`, `computer-use-linux/src/server.rs:683`, `computer-use-linux/src/server.rs:2039`, `computer-use-linux/src/server.rs:2068`, `computer-use-linux/src/server.rs:2082`, `computer-use-linux/src/server.rs:2105`

</details>

ilysenko (approved):
Re-reviewed the follow-up commit. The previous blockers are addressed: KDE portal/session and qdbus6/Klipper failures now fall back before text is emitted, clipboard restore failure after a successful paste no longer reports the action as failed, and the Ctrl+V evdev codes are named constants.

I also ran `cargo test -p codex-computer-use-linux` locally against the PR head: all tests passed. Looks good to merge.

PR #183, opened 2026-05-14
Fix Linux Chrome profile metadata matching

moxw:
## Summary
- Update the bundled Chrome plugin patcher so browser-client metadata lookup scans Linux Chrome-family profile roots
- Match profile metadata by the connected extensionInstanceId instead of assuming one Linux browser profile root
- Keep the Chrome plugin staging smoke test deterministic by stubbing the extension host build in the fixture

## User-visible behavior
- Linux Chrome plugin diagnostics can annotate the connected Brave, Chrome, or Chromium profile more accurately when the Codex extension is installed outside Google Chrome

## Source of truth
- scripts/lib/patch-chrome-plugin.js
- tests/scripts_smoke.sh

## Validation
- bash tests/scripts_smoke.sh

## Notes
- Existing generated plugin caches need to be restaged or rebuilt before they pick up this patch
- No generated codex-app or dist artifacts are included


ilysenko (commented):

🟡 `medium` · @moxw action needed · 3 open items

**Summary:** Single-file Chrome-plugin patcher fix to scan Brave/Chrome/Chromium user-data dirs and key profile lookup by extensionInstanceId, with matching smoke fixture updates.

@moxw please address:
- Smoke fixture only exercises the `var Tc=...` variant; consider extending tests/scripts_smoke.sh to also cover the `var Ic=...` mirror so both oldText branches…

Maintainer check:
- Confirm extensionInstanceId is unique enough across browsers that `.find` returning the first match across Brave/Chrome/Chromium is acceptable for diagnostics…
- Anchored expressions still target the minified upstream bundle; the rO/nO/oO/IS identifiers are upstream-mangled and will need a re-anchor on the next upstream…

Refs: `tests/scripts_smoke.sh:1086`, `scripts/lib/patch-chrome-plugin.js:254`, `scripts/lib/patch-chrome-plugin.js:277`, `scripts/lib/patch-chrome-plugin.js:265`, `scripts/lib/patch-chrome-plugin.js:276`, `scripts/lib/patch-chrome-plugin.js:246`, `scripts/lib/patch-chrome-plugin.js:261`, `scripts/lib/patch-chrome-plugin.js:272`

</details>

moxw:
Addressed the review items in d658242:

- Added smoke coverage for both browser-client profile-root minifier variants: `var Tc=...` and `var Ic=...`
- Changed `extensionInstanceId` lookup to require exactly one matching Linux profile; duplicate matches now return null instead of picking the first match
- Left the minified bundle anchors in the existing fail-soft patcher style, with added variant coverage and the idempotency guard updated for the new unique-match code path. If upstream changes the bundle shape again, this will still need the same kind of re-anchor as the other bundle patches

Validation:

ilysenko (commented):
## Review Mate: follow-up

🟢 `low`

**Summary:** Follow-up addresses both prior asks: smoke now exercises Tc and Ic variants, and duplicate extensionInstanceId matches now return null instead of picking the first profile.

✅ Fixed
- Smoke fixture now covers both `var Tc=` and `var Ic=` browser-client variants via test_chrome_browser_client_profile_root_variants (tests/scripts_smoke.sh:1210…
- extensionInstanceId lookup now returns null when more than one Linux profile matches, so cross-browser diagnostics no longer silently pick the first hit (scrip…

Refs: `scripts/lib/patch-chrome-plugin.js:246`, `scripts/lib/patch-chrome-plugin.js:261`, `scripts/lib/patch-chrome-plugin.js:272`, `tests/scripts_smoke.sh:1201`, `tests/scripts_smoke.sh:1210`, `tests/scripts_smoke.sh:2501`

</details>

ilysenko (approved):
Reviewed after updating the branch against current main. I resolved the smoke-test conflict by preserving both the Chrome marketplace fallback test from main and the Chrome profile-root test from this PR. Local validation passed: bash -n tests/scripts_smoke.sh, node --check scripts/lib/patch-chrome-plugin.js, git diff --check, and tests/scripts_smoke.sh. GitHub CI is green for Rust and Smoke Tests, Debian, RPM, and Pacman package jobs.

PR #184, opened 2026-05-14
Fix install when upstream marketplace omits chrome plugin

sabrehagen:
## Summary

- Upstream `Codex.dmg` no longer lists the `chrome` plugin in `plugins/openai-bundled/.agents/plugins/marketplace.json`, but the plugin directory is still shipped under `plugins/openai-bundled/plugins/chrome/`. `install.sh` was staging the directory successfully and then throwing in `write_bundled_plugins_marketplace` (`scripts/lib/bundled-plugins.sh`) because the source registry has no `name === "chrome"` entry, blocking the whole install with `Error: Bundled marketplace does not contain chrome plugin`.
- Mirror the existing `computer-use` handling in the same function: when the upstream marketplace omits chrome, synthesize a local-source entry. `name` and `category` are read from the staged chrome plugin's `.codex-plugin/plugin.json`, with a try/catch falling through to `"chrome"` / `"Productivity"` defaults if that manifest is missing or malformed. `source.path` and `policy` stay hardcoded because they are marketplace-registration concerns (where the plugin sits in the staged tree, and what policy this installer applies) rather than plugin-manifest concerns. The fallback silently goes unused if upstream re-adds the entry, so the change is forward-compatible.

## Test plan

- [x] `TMPDIR="$HOME/.cache/codex-desktop/tmp" ./install.sh` against the current upstream DMG completes — previously failed at this step.
- [x] Generated `codex-app/resources/plugins/openai-bundled/.agents/plugins/marketplace.json` contains all three of `browser-use`, `chrome`, and `computer-use`.
- [x] New `test_chrome_marketplace_fallback_synthesis` in `tests/scripts_smoke.sh`: with the upstream marketplace stripped of its chrome entry and the staged `plugin.json` containing distinctive `name="chrome-fallback-test"` / `category="FallbackCategory"` values, the resulting marketplace.json contains exactly those values — proving the synthesized entry reads `plugin.json` rather than reusing defaults.
- [x] Forward-compat (isolated): when fed a `marketplace.json` that does contain a chrome entry, the function uses the upstream entry verbatim and does not fall through to the synthesized path.
- [x] Defensive fallback (isolated): when the staged `plugin.json` is malformed JSON, the synthesized entry falls through to hardcoded `chrome` / `Productivity` rather than throwing.
- [x] Pre-fix simulation reproduces the original `Error: Bundled marketplace does not contain chrome plugin` throw on the same input, confirming the patch is what removes the failure.
- [ ] Manual launch of the resulting app and confirmation that the Chrome plugin appears in the in-app plugin list (recommend a maintainer with an end-to-end setup verify this).

### DIFF
--- a/scripts/lib/bundled-plugins.sh
+++ b/scripts/lib/bundled-plugins.sh
@@ -682,10 +682,22 @@ if (includeBrowser) {
 
 if (includeChrome) {
   const chrome = sourcePlugins.find((plugin) => plugin.name === "chrome");
-  if (chrome == null) {
-    throw new Error("Bundled marketplace does not contain chrome plugin");
+  if (chrome != null) {
+    plugins.push(chrome);
+  } else {
+    plugins.push({
+      name: "chrome",
+      source: {
+        source: "local",
+        path: "./plugins/chrome",
+      },
+      policy: {
+        installation: "AVAILABLE",
+        authentication: "ON_INSTALL",
+      },
+      category: "Productivity",
+    });
   }
-  plugins.push(chrome);
 }
 
 if (includeComputerUse) {
--- a/scripts/lib/bundled-plugins.sh
+++ b/scripts/lib/bundled-plugins.sh
@@ -685,8 +685,34 @@ if (includeChrome) {
   if (chrome != null) {
     plugins.push(chrome);
   } else {
+    let name = "chrome";
+    let category = "Productivity";
+    const stagedManifestPath = path.join(
+      path.dirname(destinationPath),
+      "..",
+      "..",
+      "plugins",
+      "chrome",
+      ".codex-plugin",
+      "plugin.json",
+    );
+    try {
+      const manifest = JSON.parse(fs.readFileSync(stagedManifestPath, "utf8"));
+      if (typeof manifest.name === "string" && manifest.name.length > 0) {
+        name = manifest.name;
+      }
+      const manifestCategory =
+        manifest && manifest.interface ? manifest.interface.category : undefined;
+      if (typeof manifestCategory === "string" && manifestCategory.length > 0) {
+        category = manifestCategory;
+      }
+    } catch (_err) {
+      // Fall through to defaults when the staged plugin manifest is
+      // missing or malformed — stage_chrome_plugin_from_upstream only
+      // existence-checks plugin.json, so it can still be unparseable here.
+    }
     plugins.push({
-      name: "chrome",
+      name,
       source: {
         source: "local",
         path: "./plugins/chrome",
@@ -695,7 +721,7 @@ if (includeChrome) {
         installation: "AVAILABLE",
         authentication: "ON_INSTALL",
       },
-      category: "Productivity",
+      category,
     });
   }
 }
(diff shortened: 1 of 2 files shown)


ilysenko (commented):

🟢 `low` · @sabrehagen action needed · 2 open items

**Summary:** Small, source-of-truth fix in bundled-plugins.sh that synthesizes a local chrome marketplace entry when upstream omits it, mirroring the existing computer-use fallback.

@sabrehagen please address:
- Add a regression test in tests/scripts_smoke.sh that feeds a marketplace.json without a chrome entry and asserts the synthesized entry is emitted; current fixt…
- PR body says the fallback is built 'using the chrome plugin's own plugin.json', but the code hardcodes name/source/policy/category without reading plugin.json…

Refs: `tests/scripts_smoke.sh:62`, `tests/scripts_smoke.sh:1073`, `scripts/lib/bundled-plugins.sh:687`, `scripts/lib/bundled-plugins.sh:683`, `scripts/lib/bundled-plugins.sh:703`, `scripts/lib/bundled-plugins.sh:536`

</details>

sabrehagen:
Thanks for the review. Pushed ee3be6b addressing both items:

1. The synthesized chrome entry now actually reads `name` and `category` from the staged plugin's `.codex-plugin/plugin.json` (so the PR body claim matches the code). `source.path` and `policy` stay hardcoded because they're marketplace-registration concerns, not plugin-manifest concerns. A `try`/`catch` falls through to `"chrome"` / `"Productivity"` if the manifest is missing or malformed, since `stage_chrome_plugin_from_upstream` only existence-checks `plugin.json`.

2. Added `test_chrome_marketplace_fallback_synthesis` in `tests/scripts_smoke.sh`. The existing fixture in `make_fake_chrome_upstream_app` ships a chrome marketplace entry, so the new test reuses it and overwrites `marketplace.json` to omit chrome, plus writes a `plugin.json` with distinctive `name="chrome-fallback-test"` / `category="FallbackCategory"` so the assertions prove the synthesized entry is actually reading the manifest rather than reusing defaults. Registered in `main()` alongside the other chrome tests.

ilysenko (commented):
## Review Mate: follow-up

🟢 `low`

**Summary:** Both prior items addressed: synthesized chrome entry now reads name/category from staged plugin.json, and a regression test asserts the fallback path.

✅ Fixed
- Synthesized chrome marketplace entry now reads name/category from staged .codex-plugin/plugin.json (scripts/lib/bundled-plugins.sh:699-713) with try/catch fall…
- New test_chrome_marketplace_fallback_synthesis in tests/scripts_smoke.sh uses distinctive name/category to prove the fallback reads plugin.json (tests/scripts_…

Refs: `scripts/lib/bundled-plugins.sh:683`, `scripts/lib/bundled-plugins.sh:699`, `scripts/lib/bundled-plugins.sh:714`, `tests/scripts_smoke.sh:1201`, `tests/scripts_smoke.sh:2515`

</details>

ilysenko (approved):
Reviewed via maintainer check. The fallback only synthesizes a Chrome marketplace entry after the Chrome plugin has been staged successfully, and the smoke test covers the upstream marketplace-without-Chrome case. Looks good to merge.

PR #185, opened 2026-05-15
[codex] Use modern 7zz for Nix builds

colonelpanic8:
## Summary

- use Nixpkgs' modern `7zz` package instead of legacy `p7zip` in the Nix payload build, installer app, and dev shell
- allow the installer dependency check to pass when `7zz` is available without a separate `7z` binary
- refresh the fixed-output payload hash after the Nix source change

## Why

The installer already prefers `7zz` because newer APFS DMGs require a modern 7-Zip. Nixpkgs' `_7zz` package only exposes the `7zz` binary, so the previous dependency check still failed before the installer could select it.

## Validation

- `nix develop .#default -c bash -lc 'set -e; command -v 7zz; if command -v 7z >/dev/null; then echo "unexpected 7z on PATH" >&2; exit 1; fi'`
- `nix build .#installer --no-link`
- `nix shell nixpkgs#{_7zz,python3,curl,unzip,gnutar,gnumake,gcc} --command bash -lc 'set -euo pipefail; tmp=$(mktemp -d); export SCRIPT_DIR=$PWD WORK_DIR=$tmp; source scripts/lib/install-helpers.sh; check_deps; rm -rf "$tmp"'`
- `nix build .#default --no-link`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 - The Chrome native-messaging host now evicts stale browser clients when a newer Codex browser client connects, preventing old Node REPL sessions from repeatedly reattaching CDP and driving extension service-worker CPU.
 - The bundled Chrome plugin is now auto-installed during app startup, matching Browser Use, so the plugin page no longer falls back to an install button after restart when the Linux native host is already staged.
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed locally. No PR-specific blockers found: CI is green, script smoke tests and patcher tests pass, 7zz-only dependency check passes, and Nix flake evaluation accepts pkgs._7zz. Full local Nix build is still blocked by the existing codexDmg hash drift already present on main, not by this PR.

PR #190, opened 2026-05-15
[codex] Fix Linux Computer Use session hydration

colonelpanic8:
## Summary
- hydrate Computer Use desktop environment from systemd user manager, including Hyprland and ydotool variables
- make AT-SPI diagnostics/setup work with busctl when gdbus/gsettings are unavailable
- let the Hyprland backend infer the active instance from XDG runtime sockets when HYPRLAND_INSTANCE_SIGNATURE is missing
- update the fixed-output Nix payload hash so the packaged app actually contains the rebuilt backend

## Root cause
Codex-launched Computer Use processes could miss session variables that were present in the graphical user manager, and the diagnostics path assumed gdbus/gsettings were available on PATH. On this NixOS/Hyprland setup that made AT-SPI, Hyprland window introspection, and ydotool socket checks all report false blockers. The Nix package also uses a fixed-output payload derivation, so the payload hash must change whenever bundled plugin binaries change.

## Validation
- git diff --check
- nix build .#packages.x86_64-linux.default --no-link --print-out-paths --option warn-dirty false
- env -u HYPRLAND_INSTANCE_SIGNATURE -u YDOTOOL_SOCKET PATH=/usr/bin:/bin <packaged>/codex-computer-use-linux doctor


ilysenko:
Thanks, this looks directionally good to me. I think it needs one more pass before merge:

- Please rebase onto current `main`; GitHub now reports conflicts, and the existing CI run is from an older base.
- Please do not set `CODEX_LINUX_ENABLE_COMPUTER_USE_UI=1` in the default Nix payload. The default flake package should follow the same policy as the regular installer: the bundled Computer Use plugin may be present/registered, but the in-app Computer Use UI should remain opt-in.
- If we want a Nix opt-in path for the UI, let us expose it as a separate flake output, e.g. `codex-desktop-computer-use-ui`, rather than changing `packages.default` / `codex-desktop`.
- Please add a short README note showing the exact install URL/flake ref for users who want the Nix opt-in build, e.g. `github:ilysenko/codex-desktop-linux#codex-desktop-computer-use-ui`.
- After that, please refresh the fixed-output hash and rerun the Nix checks.

The session hydration, Hyprland instance detection, and ydotool/session env propagation changes look reasonable to me. I also ran the targeted diagnostics and Hyprland unit tests locally and they pass.

PR #194, opened 2026-05-15
fix(nix): harden upstream hash refresh

colonelpanic8:
## Summary

This hardens the Nix upstream hash refresh path for the current mutable Codex DMG.

- refreshes the pinned Codex.dmg and recursive payload hashes for the current upstream DMG
- forces the recursive fixed-output payload hash to be recomputed instead of letting Nix reuse a stale store path
- restores both hashes if the forced rebuild fails before a fixed-output mismatch is available
- rejects p7zip as APFS-capable in installer dependency checks and install-deps bootstrap logic
- removes a stale Nix substitution for `npx --yes @electron/rebuild`, which no longer exists in the current native-module rebuild path
- invokes the vendored Electron rebuild CLI through `node` so Nix builds do not depend on `/usr/bin/env`
- gives the scheduled hash updater workflow concurrency and a longer timeout

## Root cause

The old hash refresh script could update the DMG hash and then let Nix reuse an already-cached fixed-output `codexDesktopPayload` path with the old recursive output hash. That made the updater report success without actually rebuilding the payload from the new DMG. The latest upstream DMG also requires APFS-capable 7-Zip and exposed a couple of stale assumptions in the Nix native-module path.

## Validation

- `scripts/ci/update-nix-hashes.sh`
- `nix build .#codex-desktop --no-link --print-build-logs` via the hash refresh verification pass
- `tests/scripts_smoke.sh`
- `bash -n install.sh scripts/ci/update-nix-hashes.sh scripts/install-deps.sh scripts/lib/install-helpers.sh scripts/lib/native-modules.sh`

### DIFF
--- a/.github/workflows/update-codex-hash.yml
+++ b/.github/workflows/update-codex-hash.yml
@@ -8,9 +8,14 @@ on:
 permissions:
   contents: write
 
+concurrency:
+  group: update-nix-upstream-hashes
+  cancel-in-progress: false
+
 jobs:
   update-hash:
     runs-on: ubuntu-latest
+    timeout-minutes: 90
     env:
       NIX_CONFIG: experimental-features = nix-command flakes
     steps:
--- a/flake.nix
+++ b/flake.nix
@@ -22,7 +22,7 @@
 
         codexDmg = pkgs.fetchurl {
           url = "https://persistent.oaistatic.com/codex-app-prod/Codex.dmg";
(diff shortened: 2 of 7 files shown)


ilysenko (approved):
Reviewed and validated. Visible GitHub checks are green, and I additionally ran CI_SKIP_PULL=1 ./scripts/ci-local.sh nix locally in Docker on the PR branch; it completed successfully: nix flake check, nix build .#codex-desktop, and nix build .#installer.

PR #193, opened 2026-05-15
[codex] Add experimental Linux mobile remote control feature

atomashevic:
## Summary

Closes #188.

Adds an opt-in `linux-features/remote-mobile-control` feature that lets Linux builds try the upstream Codex mobile remote-control host flow.

## What changed

- Adds an experimental Linux device-key provider for the upstream `remote-control-device-key.node` interface.
- Preserves `remote_control` / `features.remote_control` for local Linux app-server startup instead of letting the upstream main bundle strip it.
- Patches the remote-control Connections visibility chunk so Linux opt-in builds are not hidden by the upstream availability / rollout gates.
- Documents the feature as experimental and disabled by default, including the file-backed key storage risk.

## Why

The upstream macOS app ships the mobile remote-control UI and app-server plumbing, but Linux builds hit local macOS-only blockers before pairing can be attempted:

- the main bundle throws on non-macOS before loading a device-key implementation;
- local host config removes `remote_control` before app-server startup;
- the renderer hides the mobile remote-control Connections UI when upstream marks the feature unavailable or withholds the visibility rollout.

## Validation

- `node --test linux-features/remote-mobile-control/test.js`
- `node --test scripts/patch-linux-window-ui.test.js`
- `node --test linux-features/*/test.js`
- Rebuilt and launched the local Linux app with `remote-mobile-control` enabled, then verified the mobile section appears and phone pairing starts.

## Notes

This is intentionally opt-in. The Linux provider uses file-backed ECDSA P-256 key material protected by ordinary user file permissions, not macOS Secure Enclave-backed storage. Server-side enrollment policy may still reject some Linux hosts/accounts.

PR #195, opened 2026-05-15
[codex] Add Nix remote mobile control package

colonelpanic8:
## Summary

Builds on and acknowledges https://github.com/ilysenko/codex-desktop-linux/pull/193 by adding a Nix flake entrypoint for the experimental `remote-mobile-control` Linux feature.

## What changed

- Factors the Nix package construction so feature-enabled payload variants can be built declaratively.
- Adds `packages.*.codex-desktop-remote-mobile-control` and `apps.*.remote-mobile-control`.
- Passes an explicit `CODEX_LINUX_FEATURES_CONFIG` during the fixed-output payload build so the git-ignored local `linux-features/features.json` file is not required.
- Documents `nix run .#remote-mobile-control` / `nix run github:ilysenko/codex-desktop-linux#remote-mobile-control`.

## Why

PR #193 relies on the normal local opt-in path via `linux-features/features.json`. That file is intentionally gitignored, so a Nix flake consumer cannot enable the feature from a pure flake build without a committed/declarative variant.

## Validation

- `nix flake show --no-write-lock-file`
- `nix build .#codex-desktop --no-link --print-build-logs`
- `nix build .#codex-desktop-remote-mobile-control --no-link --print-build-logs`

## Notes

The default `codex-desktop` package remains feature-disabled. The new package/app is explicitly opt-in and uses the fixed-output payload hash for the remote-control-enabled build.

### DIFF
--- a/.github/workflows/update-codex-hash.yml
+++ b/.github/workflows/update-codex-hash.yml
@@ -36,9 +36,10 @@ jobs:
           fi
 
           CODEX_DMG_HASH="$(scripts/ci/update-nix-hashes.sh read-flake-hash 'codexDmg = pkgs.fetchurl {' 'hash = ')"
-          PAYLOAD_HASH="$(scripts/ci/update-nix-hashes.sh read-flake-hash 'codexDesktopPayload = mkCodexDesktopPayload {' 'outputHash = ')"
-          COMPUTER_USE_UI_PAYLOAD_HASH="$(scripts/ci/update-nix-hashes.sh read-flake-hash 'codexDesktopComputerUseUiPayload = mkCodexDesktopPayload {' 'outputHash = ')"
-          REMOTE_MOBILE_CONTROL_PAYLOAD_HASH="$(scripts/ci/update-nix-hashes.sh read-flake-hash 'codexDesktopRemoteMobileControlPayload = mkCodexDesktopPayload {' 'outputHash = ')"
+          PAYLOAD_HASH="$(scripts/ci/update-nix-hashes.sh read-flake-hash 'codexDesktop = mkCodexDesktop {' 'payloadHash = ')"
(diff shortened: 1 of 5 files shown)


ilysenko:
Thanks for the follow-up here. The combined-output shape looks useful, but I think the hash refresh flow needs one more update before this is mergeable.

After the latest `flake.nix` refactor, the fixed-output hashes are no longer declared under:

- `codexDesktopPayload = mkCodexDesktopPayload {`
- `codexDesktopComputerUseUiPayload = mkCodexDesktopPayload {`
- `codexDesktopRemoteMobileControlPayload = mkCodexDesktopPayload {`

They now live as `payloadHash = ...` inside `mkCodexDesktop` calls. However, `scripts/ci/update-nix-hashes.sh` and `.github/workflows/update-codex-hash.yml` still use the old anchors, so the hash bot will fail when it tries to read or replace those hashes.

Also, the new combined output (`codex-desktop-computer-use-ui-remote-mobile-control`) has its own fixed-output payload hash, but it is not included in the refresh/restore/verify flow.

Could you update the Nix hash refresh script and workflow so all four outputs are handled with the new `payloadHash` layout?

Expected coverage:

- default `codex-desktop`
- `codex-desktop-computer-use-ui`
- `codex-desktop-remote-mobile-control`
- `codex-desktop-computer-use-ui-remote-mobile-control`

If you have access to an LLM/code-review assistant, it might be worth asking it to review the PR as well before the next update. This one touches the Nix fixed-output hash maintenance path, and that tends to break in non-obvious ways.

colonelpanic8:
@ilysenko i made the requested changes, but honestly the way it is being approached right now is not the most idiomatic way of doing nix derivations. #206 is a draft of what I think would be a better approach, but its a slightly more intrusive change.

colonelpanic8:
@ilysenko is there a reason you closed this?

PR #206, opened 2026-05-15
[codex] Refactor Nix package inputs

colonelpanic8:
## Summary
- remove fixed-output payload hashes from the Nix package variants
- model external artifacts as explicit Nix inputs for Electron, native Node modules, Rust plugin binaries, and Browser Use node_repl
- add installer env hooks so the Nix build can inject explicit artifacts while the normal shell installer keeps its existing download/build flow
- simplify the Nix hash refresh script/workflow to update only the upstream DMG hash and verify all package outputs

## Why
The previous Nix build needed a separate fixed-output hash for each package variant because the payload derivation downloaded and built external artifacts during the payload build. This made every feature combination look like its own external artifact. Moving those external pieces into explicit Nix inputs makes the package variants ordinary derivations again and leaves the DMG hash as the main upstream refresh point.

## Validation
- nix flake check --no-build
- bash -n scripts/ci/update-nix-hashes.sh
- scripts/ci/update-nix-hashes.sh read-flake-hash ...
- nix build .#codex-desktop --no-link --print-build-logs
- nix build .#codex-desktop .#codex-desktop-computer-use-ui .#codex-desktop-remote-mobile-control .#codex-desktop-computer-use-ui-remote-mobile-control --no-link --print-build-logs

## Notes
This still pins Electron and native module versions explicitly in the Nix layer. A useful follow-up would be a CI guard that checks those pins against the downloaded DMG metadata during hash refresh.

### DIFF
--- a/.github/workflows/update-codex-hash.yml
+++ b/.github/workflows/update-codex-hash.yml
@@ -36,9 +36,6 @@ jobs:
           fi
 
           CODEX_DMG_HASH="$(scripts/ci/update-nix-hashes.sh read-flake-hash 'codexDmg = pkgs.fetchurl {' 'hash = ')"
-          PAYLOAD_HASH="$(scripts/ci/update-nix-hashes.sh read-flake-hash 'codexDesktopPayload = mkCodexDesktopPayload {' 'outputHash = ')"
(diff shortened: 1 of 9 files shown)


ilysenko:
Thanks for the original Nix refactor. I used it as the basis for a fresh PR rebased on current main and resolved the conflicts here: https://github.com/ilysenko/codex-desktop-linux/pull/221

I’m going to close this PR in favor of that one so we keep the review and CI history in one place.

PR #210, opened 2026-05-16
fix: handle Linux side-by-side edge cases

ryan-mt:
## Summary
- Read the Linux Computer Use UI opt-in from the active app id settings directory, so side-by-side builds such as `codex-cua-lab` do not silently ignore their own settings.
- Let rounded large `window_id` matches be narrowed by `title`, `pid`, `app_id`, or `wm_class`, matching the resolver error message and avoiding dead-end ambiguity.
- Preserve `+commitish` suffixes in Arch/pacman `pkgver` while keeping `pkgrel=1` for package rebuilds.

## Test plan
- `bash tests/webview_probe_equivalence.sh`
- `bash -n install.sh scripts/lib/*.sh launcher/start.sh.template scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh scripts/build-appimage.sh scripts/install-deps.sh`
- `cargo fmt --check`
- `cargo test --workspace`
- `cargo check --workspace`
- `git diff --check`

### DIFF
--- a/computer-use-linux/src/windowing/mod.rs
+++ b/computer-use-linux/src/windowing/mod.rs
@@ -284,6 +284,42 @@ mod tests {
     }
 
     #[test]
+    fn rounded_window_id_can_be_disambiguated_by_title() {
+        let first_window_id = 7_511_476_032_840_641_491;
+        let second_window_id = 7_511_476_032_840_641_999;
+        let rounded_window_id = 7_511_476_032_840_642_000;
+        assert_eq!(first_window_id as f64, rounded_window_id as f64);
+        assert_eq!(second_window_id as f64, rounded_window_id as f64);
+
+        let windows = vec![
+            window(
+                first_window_id,
+                "First — Kate",
+                "org.kde.kate",
+                "org.kde.kate",
+            ),
+            window(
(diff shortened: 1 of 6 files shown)


ilysenko:
Thanks, this looks good overall.

Small test nit: `isComputerUseUiEnabled()` now reads `CODEX_LINUX_APP_ID` too, but `withIsolatedHome()` only resets `CODEX_APP_ID`. Could you reset/restore `CODEX_LINUX_APP_ID` there as well?

After that this looks mergeable.

ryan-mt:
Thanks for catching that — you're right. `isComputerUseUiEnabled()` can now consult `CODEX_LINUX_APP_ID`, so the isolated-home helper should clear and restore it alongside `CODEX_APP_ID`.

I pushed `a223538` to do exactly that. I also verified the tests with an ambient `CODEX_LINUX_APP_ID` set so the helper would have failed if it leaked through:

- `CODEX_LINUX_APP_ID=codex-cua-lab node --test scripts/patch-linux-window-ui.test.js`
- `git diff --check`

GitHub Actions are green again as well.

PR #214, opened 2026-05-16
[codex] Add persistent composer rate limit footer patch

nofihq:
## Summary

Adds an optional Linux webview asset patch that keeps the compact composer rate-limit footer visible across project chats, normal chats, and new chats.

The patcher now detects the current minified rate-limit symbols, injects a `conversationId`-based footer helper, migrates previously broken helper/call shapes, and removes the old home-only gate.

## Safety

This is intentionally fail-closed. If the expected composer needles drift, the optional patch skips instead of blocking launch. If the injected footer hits unexpected account or rate-limit data at runtime, the helper catches the error and returns `null` instead of tripping the renderer error boundary.

The footer group insertion also now requires a conversation-id binding in the matched composer scope before injecting a `conversationId` reference. When inserted, it replaces the old `Kt` memo branch with a recomputed group so the footer cannot retain a stale conversation id while `Ut`/`Wt`/`Gt` stay stable.

## Root Cause

A previous local patch could leave calls to `codexLinuxRateLimitFooter` without a valid helper after an app update, or pass rate-limit variables from the wrong minified scope. That produced renderer errors instead of simply hiding the footer.

## Validation

- `node -c scripts/patches/webview-assets.js`
- `node --test scripts/patch-linux-window-ui.test.js` (`97` tests passing)
- Local smoke check of the served installed composer bundle: one helper, one runtime guard, two footer placements, and no home-only gate

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -57,6 +57,9 @@ const {
 const {
   validateReport,
 } = require("./ci/validate-patch-report.js");
+const {
+  applyPersistentRateLimitFooterPatch,
+} = require("./patches/webview-assets.js");
 
 const mainBundlePrefix =
   "let n=require(`electron`),i=require(`node:path`),o=require(`node:fs`);";
@@ -266,6 +269,7 @@ test("default core patch descriptors are grouped and unique", () => {
     "linux-computer-use-install-flow",
     "linux-app-updater-bridge",
     "browser-annotation-screenshot",
+    "composer-persistent-rate-limit-footer",
     "keybinds-settings",
     "package-desktop-name",
   ];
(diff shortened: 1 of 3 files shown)


nofihq:
Addressed in `94f9ebe`.

The patcher now detects a conversation-id binding in the matched composer scope before injecting the footer group call. If the fixture/shape has no binding, the group is left unchanged, so there is no unbound `z` reference before the helper can fail closed.

Added regression coverage for both cases:
- no conversation-id binding: no unsafe footer insertion
- current normalized binding shape: `z=R??conversationId` still receives the footer

Validation: `node -c scripts/patches/webview-assets.js` and `node --test scripts/patch-linux-window-ui.test.js` (`97` passing).

nofihq:
Addressed in `eabadef`.

The inserted footer group intentionally replaces the old `Kt` memo branch with a recomputed group. That avoids putting a `conversationId`-dependent child inside a cache guard that only tracks `Ut`/`Wt`/`Gt`, so it cannot reuse an old footer element when the active conversation changes.

I added an assertion that the generated patch contains the uncached `Kt=(0,Q.jsxs)(...)` group and does not leave `codexLinuxRateLimitFooter` inside the old `t[131]!==Ut||t[132]!==Wt||t[133]!==Gt` branch.

Validation: `node -c scripts/patches/webview-assets.js` and `node --test scripts/patch-linux-window-ui.test.js` (`97` passing).

PR #220, opened 2026-05-16
fix: patch better-sqlite3 for Electron 42

sanyinchen:
(cherry picked from commit 93fb7d907c00db9e8b33157012b4c57e73397b8d)


ilysenko:
Thanks for the patch. This is superseded by the current `patch_better_sqlite3_for_v8_external_pointer_api` implementation on main, which already handles the Electron 42/V8 external pointer API change and has smoke coverage.

This PR also currently breaks `scripts/lib/native-modules.sh` syntax because the added function is missing its closing brace, so `bash -n scripts/lib/native-modules.sh` fails with `syntax error: unexpected end of file`.

Closing this as superseded.

PR #222, opened 2026-05-17
Fix user-local updates with relative origin URLs

ryan-mt:
## Summary

Fixes the user-local managed checkout flow when the recorded/source Git origin is a relative local path such as `../origin.git`.

The helper previously passed that relative URL to `git ls-remote`/`git clone` from the updater script's current working directory, so Git resolved it against the wrong directory and failed before the managed checkout could refresh.

## Fix

- Resolve relative local origin URLs against the source/managed repository directory before using them outside that repository.
- Keep absolute paths, protocol URLs, and scp-style Git URLs unchanged.
- Run remote queries with `git -C` from the repository context.
- Add a smoke regression that creates a source checkout with `../origin.git`, then verifies `prepare_build_repo` can clone/read the managed checkout.

## Verification

Fail-first repro against `origin/main` with the new smoke case:

```text
RELATIVE_ORIGIN_FAIL_FIRST origin-main rc=128
fatal: repository '../origin.git' does not exist
```

Local gates on this branch:

```text
targeted_relative_origin PASS
shell_syntax PASS
helper_shell_syntax PASS
shell_smoke PASS
diff_check PASS
```

### DIFF
--- a/contrib/user-local-install/files/.local/lib/codex-desktop-linux/common.sh
+++ b/contrib/user-local-install/files/.local/lib/codex-desktop-linux/common.sh
@@ -91,38 +91,63 @@ repo_origin_url_is_relative_local() {
 resolve_repo_origin_url() {
     local origin_url="$1"
     local base_dir="$2"
-    local base_abs target_dir target_name
+    local base_abs candidate target_dir target_name
 
-    if ! repo_origin_url_is_relative_local "$origin_url" || [ ! -d "$base_dir" ]; then
+    if ! repo_origin_url_is_relative_local "$origin_url" || [ -z "$base_dir" ]; then
         printf '%s\n' "$origin_url"
         return 0
     fi
 
-    base_abs="$(cd "$base_dir" && pwd -P)" || {
-        printf '%s\n' "$origin_url"
-        return 0
-    }
-    target_dir="$(dirname "$origin_url")"
-    target_name="$(basename "$origin_url")"
-    if [ -d "$base_abs/$target_dir" ]; then
-        printf '%s/%s\n' "$(cd "$base_abs/$target_dir" && pwd -P)" "$target_name"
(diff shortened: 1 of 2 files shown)


ilysenko:
Thanks, this is a useful fix, but there is one remaining edge case before merge.

If the first managed checkout succeeds from a recorded relative `REPO_ORIGIN_URL` like `../origin.git`, the managed repo gets a usable absolute origin. However, if the original `SOURCE_REPO_DIR` is later moved or deleted, `repo_origin_url()` still prefers the recorded relative URL and now resolves it from `MANAGED_REPO_DIR` via `repo_remote_query_dir()`. That points at a non-existent path under `xdg-data/codex-desktop-linux/` and breaks subsequent updates.

Could you please handle that case as well? A good fix would be to resolve the recorded relative URL against the saved `SOURCE_REPO_DIR` path, or fall back to the managed repo’s configured origin when the source checkout is unavailable. Please also add a smoke regression for: clone once with `../origin.git`, remove/rename the source checkout, then run `prepare_build_repo` again.

ryan-mt:
Yes I can handle it this morning when I wake up thank you for letting me know. Don’t close this pr🫶🏻

ryan-mt:
@ilysenko Fixed now — I updated the relative `REPO_ORIGIN_URL` handling so recorded relative origins resolve against the saved `SOURCE_REPO_DIR`, and when that no longer resolves after the source checkout is moved/deleted it falls back to the managed repo’s configured origin. I also extended the smoke regression to clone once with `../origin.git`, move the source checkout away, advance the origin, then run `prepare_build_repo` again.

Local verification passed:
- `bash -n contrib/user-local-install/files/.local/lib/codex-desktop-linux/common.sh`

ilysenko (approved):
Thanks, this addresses the requested edge case.

PR #226, opened 2026-05-17
Persist user-local X11 fallback preference

### LINKED ISSUE #215: User-local launcher should force X11 for draggable avatar overlay
## Summary
On GNOME Wayland, the user-local Codex Desktop launcher starts Electron with the default `--ozone-platform-hint=auto`, which selects native Wayland. The avatar/pet overlay window is created, but the compositor centers it and ignores subsequent absolute positioning/drag movement. The visible result is that `/pet` appears around the middle of the main window/screen instead of bottom-right, and dragging does not work.

Forcing X11/XWayland via the existing launcher option fixes the issue:

```bash
~/.local/opt/codex-desktop-linux/codex-app/start.sh --x11
```

After launching with `--ozone-platform=x11`, the overlay moved to the persisted bottom-right bounds and dragging worked again.

## Environment
- Ubuntu 25.10
- GNOME Wayland (`WAYLAND_DISPLAY=wayland-0`, `DISPLAY=:0`)
- Display: 1920x1080
- User-local install from `contrib/user-local-install`
- Codex Desktop Electron command before workaround included `--ozone-platform-hint=auto`
- Relaunched command after workaround included `--ozone-platform=x11`

## Repro
1. Install using the user-local integration.
2. Launch `codex-desktop` under GNOME Wayland.
3. Run `/pet`.
4. Observe the pet/overlay appears centered rather than bottom-right.
5. Try dragging the pet.
6. Dragging does not move it reliably.

## Expected
The avatar overlay should open at its persisted/default bottom-right location and be draggable.

## Actual
The overlay window is centered by the compositor and drag movement does not reposition it.

## Suggested fix
For the user-local launcher, pass `--x11` by default, or add a Wayland-session fallback for avatar overlay support similar to the existing Sommelier X11 fallback.

The minimal tested user-local launcher change was:

```diff
- exec "${APP_DIR}/start.sh" ""
+ exec "${APP_DIR}/start.sh" --x11 ""
```

This uses an already-supported launcher flag and avoids changing Codex configs.

Aminkbi:
## Summary
- add user-local installer flags to persistently enable or disable the X11/XWayland fallback
- store the preference in ~/.config/codex-desktop-linux/user-local.env so updater refreshes preserve it
- have the user-local wrapper apply --x11 only when that persisted preference, or an explicit environment override, requests it

Fixes #215.

## Validation
- bash -n contrib/user-local-install/install-user-local.sh contrib/user-local-install/files/.local/bin/codex-desktop tests/scripts_smoke.sh
- bash tests/scripts_smoke.sh

### DIFF
--- a/contrib/user-local-install/README.md
+++ b/contrib/user-local-install/README.md
@@ -53,6 +53,18 @@ To also enable the weekly auto-update timer, pass `--enable-timer`:
 ./contrib/user-local-install/install-user-local.sh --enable-timer
 ```
 
+To persistently force the user-local launcher through X11/XWayland, pass `--force-x11`:
+
+```bash
+./contrib/user-local-install/install-user-local.sh --force-x11
+```
+
+To return to the default generated launcher behavior, pass `--no-force-x11`:
+
+```bash
+./contrib/user-local-install/install-user-local.sh --no-force-x11
+```
+
 The installer:
 
 1. copies standalone helper scripts into `~/.local/opt/codex-desktop-linux`
@@ -80,4 +92,5 @@ codex-desktop-version
 - The icon is not committed as a binary asset here. It is generated locally from `Codex.dmg`.
 - The helper scripts track both upstream wrapper changes and upstream `Codex.dmg` headers.
 - The helper scripts are copied into `~/.local/opt` and do not run from the git checkout directly.
+- The X11/XWayland preference is stored in `~/.config/codex-desktop-linux/user-local.env` and is preserved across updater refreshes.
 - The weekly timer runs `codex-desktop-update --quiet`. It is opt-in: pass `--enable-timer` to `install-user-local.sh` to activate it, or run `systemctl --user enable --now codex-desktop-update.timer` manually after install.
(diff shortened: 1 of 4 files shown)


ilysenko (changes requested):
Thanks for tracking this down. I’d rather not force `--x11` unconditionally from the user-local wrapper.

A current Ubuntu GNOME Wayland setup here works fine without this workaround, so this looks environment-specific rather than a user-local default. If we need this for affected setups, could we make it an install-time opt-in/persisted preference instead?

One possible shape: add a user-local installer option such as `--force-x11` / `--x11-fallback`, store that preference outside the generated wrapper, for example in `~/.config/codex-desktop-linux/user-local.env` or a settings flag, and have the wrapper/launcher apply `--x11` only when that persisted setting is enabled.

The updater should preserve that user preference across refreshes instead of overwriting or removing it. That would still fix #215 for users who need the fallback, while keeping the default launcher behavior unchanged and reversible.

Aminkbi:
Updated in that direction. The installer now supports `--force-x11` / `--no-force-x11`, stores the preference in `~/.config/codex-desktop-linux/user-local.env`, and the wrapper only applies `--x11` when that persisted setting is enabled. `--from-update` leaves the file untouched, with a smoke test covering that.

ilysenko (approved):
Thanks, this addresses the requested changes.

PR #229, opened 2026-05-17
fix multi-instance app lock bypass

avifenesh:
Follow-up to #227.

## What changed
- Mark explicit Linux multi-instance launches with `CODEX_LINUX_MULTI_LAUNCH=1` and export the instance identity into Electron.
- Let only explicit multi-instance launches bypass the Linux app-level `requestSingleInstanceLock()` guard in the main-process patch.
- Add an extracted-app bootstrap patch for the earlier bootstrap-owned single-instance lock, which was still able to exit the second app before the main-process patch mattered.
- Cover both lock paths in patch tests and smoke checks.

## Why
The launcher-side port/state isolation from #227 was necessary but not sufficient: current upstream bundles also have app-level single-instance locks. A second `--new-instance` launch could allocate a separate port and user data dir, then Electron would exit from the bootstrap/main lock path.

## Validation
- `git diff --check`
- `node --check scripts/patches/bootstrap.js`
- `bash -n launcher/start.sh.template && bash -n tests/scripts_smoke.sh`
- Local rebuilt install smoke: original app stayed on `127.0.0.1:5175`; `--new-instance` stayed up on `127.0.0.1:5176` with isolated Electron user data under `~/.local/state/codex-desktop/instances/port-5176/electron-user-data`.

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -36,6 +36,9 @@ MULTI_LAUNCH_ACTIVE=0
 CODEX_LINUX_INSTANCE_ID=""
 LAUNCHER_ARGS=()
 
+# Internal marker consumed by patched Electron bundles; never trust inheritance.
+unset CODEX_LINUX_MULTI_LAUNCH
+
 early_truthy_env_value() {
     case "${1:-}" in
         1|true|TRUE|yes|YES|on|ON) return 0 ;;
@@ -156,6 +159,7 @@ configure_multi_launch_instance() {
     CODEX_LINUX_WEBVIEW_PORT="$selected_port"
     CODEX_LINUX_INSTANCE_ID="port-$CODEX_LINUX_WEBVIEW_PORT"
     MULTI_LAUNCH_ACTIVE=1
+    CODEX_LINUX_MULTI_LAUNCH=1
 
     APP_STATE_DIR="$base_state_dir/instances/$CODEX_LINUX_INSTANCE_ID"
     APP_PID_FILE="$APP_STATE_DIR/app.pid"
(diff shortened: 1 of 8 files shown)


ilysenko (changes requested):
Thanks, this is the right direction, but there is one blocker around the new internal env marker.

`CODEX_LINUX_MULTI_LAUNCH=1` is exported into Electron for explicit `--new-instance` launches, but the launcher does not clear an inherited value on normal launches. That means a child process or shell that inherits `CODEX_LINUX_MULTI_LAUNCH=1` can run `start.sh` without `--new-instance`; the launcher will use the default shared port/profile path, while the patched bootstrap/main-process code will still bypass `requestSingleInstanceLock()`.

A minimal repro is a normal launcher run with `CODEX_LINUX_MULTI_LAUNCH=1` already in the environment: Electron still receives that value even though no multi-instance mode was requested.

Could you reset this internal marker at launcher startup, for example by setting it to `0` or unsetting it before argument parsing, and only exporting `CODEX_LINUX_MULTI_LAUNCH=1` from `configure_multi_launch_instance()` when `--new-instance` or `CODEX_MULTI_LAUNCH=1` actually requested a multi-instance launch? A smoke assertion for the inherited-env case would be good too.

avifenesh:
Addressed in `07631af`.

I now clear `CODEX_LINUX_MULTI_LAUNCH` at launcher startup before parsing args, so an inherited internal marker cannot affect a normal launch. Only explicit `--new-instance` / `CODEX_MULTI_LAUNCH=1` sets it back to `1` inside `configure_multi_launch_instance()`.

Also added a smoke assertion for this inherited-env case. Local checks passed:
- `git diff --check`

avifenesh:
Follow-up pushed in `a7034a5` for the CI smoke failure.

The failed job was not the env-marker assertion; it was the webview probe equivalence harness. Both the old Python reference and new curl probe failed the same ok-marker scenarios because the fixture server could disappear between the TCP-open probe phase and the HTTP marker phase.

I hardened the harness by isolating those phases: TCP verdicts run first, then HTTP marker checks get a fresh threaded fixture server. Also guarded cleanup so helper subshells cannot tear down the parent fixture. Local validation now passes:
- `bash tests/webview_probe_equivalence.sh` x100

ilysenko (approved):
Thanks, this addresses the requested changes.

PR #231, opened 2026-05-17
fix: harden Linux patch helper regressions

ryan-mt:
## Summary

Fixes three small Linux patch-helper regressions found during a focused bug hunt:

- Persist side-by-side Linux settings under the effective app id/settings file instead of hard-coding `codex-desktop`.
- Reset global regex state when scanning webview assets, so global filename patterns match every candidate file.
- Tighten Linux opaque-window idempotency detection so unrelated Linux `userAgent` checks do not suppress required patch warnings.

## Verification

- `git diff --check`
- Added-line static security scan: no findings
- Independent diff review: passed with no security concerns or logic errors


ilysenko (approved):
Thanks, this looks good.

PR #233, opened 2026-05-18
fix: avoid unsupported app-server feature sync during connector login

mingshi2333:
## Problem

On Linux, Codex Desktop's webview syncs its Statsig default feature enablement to the local Codex CLI app-server. Current CLI app servers reject some feature keys present in the bundled webview assets, including `auth_elicitation`, `enable_mcp_apps`, and the default-list entry that resolves to `workspace_dependencies`.

This shows up during connector authentication flows such as Google/Gmail/Drive/Calendar login. After the browser OAuth flow returns, the app can log `unsupported feature enablement ...` and `Failed to sync experimental feature enablement`, leaving the connector auth state looking like it did not make it back into Codex Desktop.

## Root Cause

The Linux package reuses upstream webview assets whose default feature list assumes those keys can be sent to `set-experimental-feature-enablement-for-host`. The current local Codex CLI app-server only accepts a smaller allowlist for that method.

## Changes

- Adds a core Linux webview-asset patch for app-server feature enablement compatibility.
- Filters the default app-server feature sync array down to currently supported feature keys before the webview calls `set-experimental-feature-enablement-for-host`.
- Keeps the broader Statsig mapping data intact; only the default sync payload is filtered.
- Adds patch tests for the current minified bundle shape and drift warning behavior.
- Updates the changelog.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` passes: 112/112.
- `nix build .#codex-desktop --print-out-paths -L` succeeds.
- Inspected the packaged webview asset and confirmed the default sync array is reduced to `apps`, `memories`, `plugins`, `tool_call_mcp_elicitation`, `tool_search`, and `tool_suggest`.
- Launched the patched build side-by-side with the current local Codex CLI app-server and confirmed `experimentalFeature/enablement/set` returns `errorCode=null`.
- Manually tested Google connector login locally with the patched build; the login flow returned successfully and the logs no longer show `unsupported feature enablement` or `Failed to sync experimental feature enablement` for the connector flow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 - The bundled Chrome plugin is now auto-installed during app startup, matching Browser Use, so the plugin page no longer falls back to an install button after restart when the Linux native host is already staged.
 - Nix builds, installer apps, and dev shells now use modern `7zz`, and the installer dependency check accepts `7zz` without requiring a separate legacy `7z` binary.
 - Codex Desktop no longer removes user-enabled `remote_control = true` from the local Linux config before starting the app server.
+- Linux webview bundles no longer ask current Codex CLI app servers to enable unsupported feature flags, avoiding connector authentication sync errors.
 
 ## [0.8.0] - 2026-05-16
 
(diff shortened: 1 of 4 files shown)


mingshi2333:
Good catch. The patch now replaces the feature array at the regex match index instead of calling string replace on the first identical literal. I also added a regression test where an identical unsupported feature array appears earlier in the bundle, so the patch only mutates the array associated with the statsig_default_enable_features / set-experimental-feature-enablement-for-host sync path.

PR #240, opened 2026-05-19
[codex] Fix Linux remote mobile control auth

avifenesh:
## Summary

Fix Linux remote mobile control so a Linux desktop can both enroll as a controller and load remote-control environments after authorization.

This carries three narrow compatibility fixes in the `remote-mobile-control` Linux feature:

- Accept the server-returned account user id when the auth token uses the account-scoped user id shape, and store the enrollment under the server-returned id.
- Treat a revoked remote-control client as recoverable, deleting the stale local enrollment/key and re-enrolling instead of leaving the UI stuck.
- Enable the remote-control environment loader gate on Linux. The settings UI was visible on Linux, but the app-main bridge still only loaded environments when the upstream rollout flag was true.

## Why

On Linux the UI could reach the authorization flow, but the client enrollment could fail on account id shape mismatches. After reconnect/revoke flows, a stale revoked local client was reused. Once authorization succeeded, the Linux renderer still did not ask the backend for environments because the loader gate remained tied to the upstream rollout flag.

## Validation

### DIFF
--- a/linux-features/remote-mobile-control/patch.js
+++ b/linux-features/remote-mobile-control/patch.js
@@ -339,13 +339,16 @@ function applyLinuxRemoteControlClientAccountCompatibilityPatch(source) {
 }
 
 function applyLinuxRemoteControlClientRevocationRecoveryPatch(source) {
-  if (source.includes("Remote-control client has been revoked")) {
+  if (
+    source.includes("e.message===`Remote-control client key material missing`") &&
+    source.includes("e.message===`Remote-control client has been revoked`")
+  ) {
     return source;
   }
 
   const recoverableErrorNeedle =
-    "e.message===`Remote control request failed (403): Remote-control client key material missing`:!1";
-  if (!source.includes(recoverableErrorNeedle)) {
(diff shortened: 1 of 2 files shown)


avifenesh:
In simple non-codex words, I couldn't control my codex on my Mac from my PC. This fixes it and lets me authorize and control other non-SSH machines with account auth.

ilysenko:
Thanks, I rechecked this against a fresh upstream DMG. The current bundle throws 403 remote-control errors as the bare server detail, via `new Error(await response.detail)`, so the recoverable predicate also needs to handle:

`Remote-control client key material missing`

Right now the PR only keeps the prefixed form for that case:

`Remote control request failed (403): Remote-control client key material missing`

Could you add the bare `Remote-control client key material missing` case and a test for it?

avifenesh:
Thanks, that was the missing upstream shape. I added the bare `Remote-control client key material missing` predicate alongside the existing prefixed 403 form and kept the revoked-client recovery path idempotent. Also added a focused test for the bare error. Local validation: `node --test linux-features/remote-mobile-control/test.js` and `node --test linux-features/*/test.js` both pass.

PR #244, opened 2026-05-19
Add guided native setup helper

PinguuSS:
## Summary

- Add make setup-native as an optional guided native setup entrypoint that calls scripts/bootstrap-wizard.sh.
- Add a conservative wizard that reports distro/package/session readiness, discovers linux-features/*/feature.json, and can update the git-ignored feature config in non-interactive mode.
- Document guided setup, non-interactive feature edits, manual-update package mode, and safe opt-out behavior.
- Cover feature enable/disable, invalid feature ids, non-destructive opt-out guidance, and target composition in script smoke tests.

## Notes

This keeps make bootstrap-native, make install-native, make package, and make install non-interactive. The wizard does not install packages, change groups, start or stop services, or delete device keys, Read Aloud runtimes/models, plugin caches, or other user data.

## Test Plan

- bash -n scripts/bootstrap-wizard.sh
- git diff --check
- bash tests/scripts_smoke.sh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ if [ -z "$$format" ]; then \
 fi; \
 printf '%s\n' "$$format"
 
-.PHONY: help check test build-updater maybe-build-updater update rebuild rebuild-install inspect-upstream build-app build-app-fresh bootstrap-native install-native update-native rebuild-next run-app build-dev-app run-dev-app deb rpm pacman appimage package install service-enable service-status clean-dist clean-state
+.PHONY: help check test build-updater maybe-build-updater update rebuild rebuild-install inspect-upstream build-app build-app-fresh setup-native bootstrap-native install-native update-native rebuild-next run-app build-dev-app run-dev-app deb rpm pacman appimage package install service-enable service-status clean-dist clean-state
 
 help:
 	@printf '\nCodex Desktop Linux Make Targets\n\n'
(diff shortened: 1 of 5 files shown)


ilysenko:
Thanks, this looks useful overall and I like that it stays separate from the non-interactive install/package flows.

One thing to fix before merge: `portal_summary()` can report a false negative with `set -o pipefail`. On my GNOME Wayland system `xdg-desktop-portal` is running and `busctl --user --list` contains `org.freedesktop.portal.Desktop`, but the wizard prints `portal=not detected`. The likely cause is `busctl --user --list | grep -q ...`: `grep -q` can close the pipe early, `busctl` exits with SIGPIPE, and the pipeline fails.

Could you change that check to avoid the early-closing pipe, for example by capturing the `busctl` output first or using a non-`-q` grep redirected to `/dev/null`? It may also be worth avoiding `pgrep -x xdg-desktop-portal` because long process names can be truncated.

Small optional improvement: if `conversation-mode` is enabled without `read-aloud`, the wizard could print a warning, since the conversation-mode README says it needs Read Aloud for speech output.

PinguuSS:
Thanks. Fixed in 4f48fbd.

Changes:
- `portal_summary()` now captures `busctl --user --list` output before checking for `org.freedesktop.portal.Desktop`, so it no longer depends on a `grep -q` pipeline under `pipefail`.
- Removed the primary `pgrep -x xdg-desktop-portal` path; process detection is now only a fallback and uses `pgrep -f` to avoid long-name truncation issues.
- Added a warning when `conversation-mode` is enabled without `read-aloud`.

Validation:
- Added smoke coverage for the simulated `busctl` SIGPIPE/exit-141 case.
- Added smoke coverage for the `conversation-mode` without `read-aloud` warning.
- Ran:
  - `bash -n scripts/bootstrap-wizard.sh`
  - `git diff --check`

PR #246, opened 2026-05-19
Expand guided native setup orchestration

PinguuSS:
## Summary

- extend `make setup-native` with explicit opt-in orchestration for dependency bootstrap and native install commands, including dry-run output
- add deeper read-only readiness guidance for Computer Use and Read Aloud paths, packages, services, and doctor commands
- add a separate interactive cleanup path that requires exact `DELETE <path>` confirmation before removing feature-owned local data

## Notes

This builds on #244 and keeps `make bootstrap-native`, `make install-native`, `make package`, and `make install` non-interactive. The wizard still avoids changing system services, groups, key files, model files, Python runtimes, or plugin caches unless the user explicitly opts into the relevant step.

## Test plan

- `bash -n scripts/bootstrap-wizard.sh`

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ make install-native
 PACKAGE_WITH_UPDATER=0 make install-native
 ```
 
-The wizard is conservative with opt-outs. Removing a feature id from `features.json` does not delete local device keys, Read Aloud model files, Python runtimes, plugin caches, or system services. Cleanup is a separate interactive path through `CODEX_BOOTSTRAP_CLEANUP_FEATURES=remote-mobile-control,read-aloud make setup-native`; each deletion requires typing `DELETE <exact path>`. It prints the relevant paths and tells you when a rebuild/reinstall, `sudo` / `pkexec`, logout/login, input-group membership, ydotoold service work, or portal package install needs explicit user action.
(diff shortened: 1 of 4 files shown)


ilysenko (changes requested):
Thanks, this is useful, but I found two issues that should be fixed before merge.

First, `make setup-native` can hang while printing Computer Use readiness. On my machine the wizard stops in `uinput_summary()` because `stat -c '%A %U:%G' /dev/uinput` does not return (`timeout 3 stat ...` exits 124). Since this is only diagnostic output, please make it bounded/fail-soft, for example by wrapping it with a short timeout or skipping the extra stat metadata when it cannot be read quickly.\n\nSecond, dry-run cleanup still requires an interactive TTY. The docs say `CODEX_BOOTSTRAP_DRY_RUN=1` previews cleanup targets without deleting them, but `run_feature_cleanup()` exits with “Cleanup requires an interactive terminal...” before `confirm_and_delete_path()` can print `Would delete`. Please allow non-interactive cleanup when dry-run is enabled, and keep the TTY/exact confirmation requirement only for real deletion. A smoke test for non-interactive dry-run cleanup would be good too.

ilysenko (approved):
Thanks, the requested fixes are in place now.

PR #253, opened 2026-05-20
Use Linux release assets for updater

avifenesh:
## Summary
- Make `codex-update-manager` check this repository's GitHub Releases instead of the upstream DMG URL.
- Select the matching native package asset for the current system (`.deb`, `.rpm`, or `.pkg.tar.*`) and install that release asset directly.
- Treat an empty releases feed as a clean idle state so the updater is safe before the first release is published.

## Why
The updater should follow the Linux project releases as the source of truth. This keeps user updates aligned with the packages published by `codex-desktop-linux`, rather than rebuilding from OpenAI's macOS artifact on every client machine.

## Scope
This is intentionally narrow: no feature-selection UI changes, no appcast support, and no release automation changes.

## Validation
- `git diff --check`
- Smoke: `codex-update-manager check-now` against the live repo releases API returns `idle` when no releases are published yet.

### DIFF
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -577,8 +577,6 @@ async fn run_check_cycle(
         );
     }
 
-    let retrying_failed_update = state.status == UpdateStatus::Failed;
-
     let Some(_check_lock) = try_acquire_check_lock(paths)? else {
         return Ok(());
     };
@@ -610,21 +608,21 @@ async fn run_check_cycle(
             }
             Err(error) => return Err(error),
         };
-        let previous_headers_fingerprint = state.remote_headers_fingerprint.clone();
         state.remote_headers_fingerprint = Some(metadata.headers_fingerprint.clone());
         state.last_successful_check_at = Some(Utc::now());
 
-        if previous_headers_fingerprint.as_deref() == Some(metadata.headers_fingerprint.as_str())
-            && !retrying_failed_update
-            && installed_version_satisfies_candidate(&state.installed_version, &metadata.candidate_version)
-        {
-            set_status(state, paths, UpdateStatus::Idle)?;
+        if release_candidate_is_already_installed_or_superseded(
+            &state.installed_version,
+            &metadata.candidate_version,
+        ) {
+            state.status = UpdateStatus::Idle;
             state.candidate_version = None;
             state.remote_package_download_url = None;
             state.remote_package_asset_name = None;
             state.artifact_paths.package_path = None;
+            state.error_message = None;
             persist_state(paths, state)?;
-            info!("Linux release fingerprint unchanged and installed version is current");
+            info!("Linux release candidate is already installed or superseded");
             return Ok(());
         }
 
@@ -898,7 +896,7 @@ async fn prepare_update_package_if_needed(
     set_status(state, paths, UpdateStatus::DownloadingDmg)?;
 
     let downloads_dir = config.workspace_root.join("downloads");
-    let downloaded = upstream::download_dmg(
+    let downloaded = upstream::download_asset(
         &client,
         &config.dmg_url,
         "Codex.dmg",
@@ -985,7 +983,7 @@ async fn download_ready_release_package(
 
     set_status(state, paths, UpdateStatus::DownloadingDmg)?;
     let downloads_dir = config.workspace_root.join("downloads");
-    let downloaded = upstream::download_dmg(
+    let downloaded = upstream::download_asset(
         &client,
         &download_url,
         &asset_name,
@@ -1104,6 +1102,10 @@ fn installed_version_satisfies_candidate(installed: &str, candidate: &str) -> bo
         Some(_) => true,
         None => installed == candidate,
     }
+}
+
+fn release_candidate_is_already_installed_or_superseded(installed: &str, candidate: &str) -> bool {
+    installed_version_satisfies_candidate(installed, candidate)
 }
 
 fn installed_version_matches_candidate(installed: &str, candidate: &str) -> bool {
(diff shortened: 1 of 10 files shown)


Leay15:
@avifenesh You read my mind. I was actually thinking about publishing distro packages so it can be installed from there, and letting `codex-update-manager` handle all the required actions on each user's machine.

That way, we could avoid distributing OpenAI's proprietary DMG.

avifenesh:
Screenshot From 2026-05-20 04-33-38
Screenshot From 2026-05-20 04-34-47
Screenshot From 2026-05-20 05-15-11

avifenesh:
> Screenshot From 2026-05-20 04-33-38 Screenshot From 2026-05-20 04-34-47 Screenshot From 2026-05-20 05-15-11

@Leay15 its also looks nice

ilysenko (changes requested):
Thanks for the work here, but I do not think we should merge this direction.

This changes the updater source of truth from "each user downloads the official OpenAI Codex DMG and rebuilds locally" to "install Linux packages from this repository's GitHub Releases". If those release packages bundle the Codex Desktop app/resources, we would effectively be redistributing OpenAI's proprietary desktop product ourselves.

I think this repo should continue distributing only the Linux adaptation/build tooling, while the upstream Codex app is obtained from OpenAI by each user during install/update. The updater UX and feature-selection pieces may still be useful, but the public Linux release-asset install path should not land unless the release assets contain only our own code/metadata or we have explicit permission to redistribute Codex Desktop.

avifenesh:
> Thanks for the work here, but I do not think we should merge this direction.
> 
> This changes the updater source of truth from "each user downloads the official OpenAI Codex DMG and rebuilds locally" to "install Linux packages from this repository's GitHub Releases". If those release packages bundle the Codex Desktop app/resources, we would effectively be redistributing OpenAI's proprietary desktop product ourselves.
> 
> I think this repo should continue distributing only the Linux adaptation/build tooling, while the upstream Codex app is obtained from OpenAI by each user during install/update. The updater UX and feature-selection pieces may still be useful, but the public Linux release-asset install path should not land unless the release assets contain only our own code/metadata or we have explicit permission to redistribute Codex Desktop.

@ilysenko 
I actually built it to follow Codex upstream updates, then changed to follow the repo so I won't break the versioning.
I can change it back to follow the codex upstream and update on the user machine, but the PR will be much larger, it requires actually doing what today the repo does or the installer does.

I must say that at this point we are not really distributing even if you pull the DMG, you just align the wrapper with the version an give us an easier life. Give me a green light and I'm moving the heavy electron into Rust + GPUI, and you get almost all the advanced capabilities self-implemented anyway, while the backend interacts with Codex CLI like the real Codex app. 

Screenshot From 2026-05-20 09-52-01

avifenesh:
@ilysenko With a lot of respect of course, your repo, your rule, im adjusting myself. LMK if you want it to follow the upstream anyway. I can have two update button if it matter.

ilysenko:
Thanks, that makes sense. I opened a discussion so we can think through the updater direction with the wider community before changing the model: https://github.com/ilysenko/codex-desktop-linux/discussions/257

avifenesh (commented):
Two technical findings on the updater state-machine changes, independent of the design-direction concern already raised.

---

### 1. `auto_install_on_app_exit` is silently broken (HIGH)

The new `run_check_cycle` transitions to `ReadyToInstall` with `artifact_paths.package_path` set to `None` (line 657). The existing `reconcile_pending_install` (unchanged by this PR) handles `ReadyToInstall` by cloning `package_path` and returning early if it is `None` (app.rs:700–702). That early return skips the `auto_install_on_app_exit` check entirely, so the daemon never auto-installs updates detected through the new flow.

`auto_install_on_app_exit` defaults to `true` (`config.rs:89`), so this regression affects most users.

The old flow worked because `run_check_cycle` downloaded the DMG and built the package during the check itself, so `package_path` was always set before the daemon's reconcile loop ran. The new flow defers package acquisition to `prepare_update_package_if_needed`, which is only called from `run_install_ready` (user-triggered), never from `reconcile_pending_install`.

If the intent is to keep auto-install working for the no-features case, the reconcile path needs to call `prepare_update_package_if_needed` (which already handles headless/no-features gracefully) instead of returning early.

### 2. GitHub releases API pagination not handled (MEDIUM)

`fetch_remote_metadata` calls `GET /repos/{owner}/{repo}/releases` without `?per_page=N`. GitHub returns at most 30 releases per page. If the repository ever accumulates more than 30 releases, `select_latest_release` only sees the first page and could miss the actual latest non-prerelease.

The GitHub API returns releases newest-first by default, so the practical risk is low in the near term, but it's a latent correctness gap. A simple fix is to append `?per_page=100` to the URL (the maximum allowed), or use a higher-level strategy if the repo expects many releases.



---

updater/src/app.rs
```diff
@@ -603,14 +664,11 @@ async fn run_check_cycle(
...
-            "Preparing a local Linux package from the new upstream DMG.",
+            "Choose Update in Codex Desktop to select optional Linux features and prepare a local package.",
         )?;
 
-        let candidate_version = state
-            .candidate_version
-            .clone()
-            .expect("candidate version should be set before local build");
-        builder::build_update(config, state, paths, &candidate_version, &downloaded.path).await?;
+        state.status = UpdateStatus::ReadyToInstall;
```
avifenesh, line 670: The check cycle now transitions to `ReadyToInstall` with `package_path = None` (set on line 657). The existing `reconcile_pending_install` handles `ReadyToInstall` at app.rs:699–702 by cloning `package_path` and returning early when it is `None`:

```rust
UpdateStatus::ReadyToInstall => {
    let Some(package_path) = state.artifact_paths.package_path.clone() else {
        return Ok(());  // ← always taken now
    };
    // auto_install_on_app_exit check never reached
}
```

This silently disables `auto_install_on_app_exit` for the new update flow. Since `auto_install_on_app_exit` defaults to `true`, most existing users will stop receiving auto-installed updates. The daemon's reconcile loop will spin every `RECONCILE_INTERVAL_SECONDS` doing nothing.

If the intent is to keep auto-install working (at least for the no-features case where `prepare_update_package_if_needed` can download the release package without a GUI prompt), the reconcile path should call `prepare_update_package_if_needed` instead of returning early.



---

updater/src/upstream.rs
```diff
@@ -1,60 +1,134 @@
...
+
+/// Fetches the latest published Linux repo release and package asset for this system.
+pub async fn fetch_remote_metadata(
+    client: &Client,
+    releases_api_url: &str,
+    package_kind: PackageKind,
+) -> Result<RemoteMetadata> {
     let response = client
-        .head(dmg_url)
+        .get(releases_api_url)
```
avifenesh, line 89: This GET request fetches the default page (30 items) from the GitHub releases API. If the repository accumulates more than 30 releases, `select_latest_release` may not see the actual latest non-prerelease.

Consider appending `?per_page=100` to `DEFAULT_RELEASES_API_URL` (or passing `per_page` as a query parameter here) to reduce the risk. The GitHub API maximum is 100 per page, which should be sufficient for the foreseeable lifetime of this repo.



---

PR #256, opened 2026-05-20
[codex] Fix bundled backend plugin discovery

colonelpanic8:
## Summary

- refresh runtime marketplace plugin links for bundled Computer Use and Read Aloud backends
- keep refreshing `latest` and marketplace symlinks even when plugin cache contents are already current
- add a launcher smoke invariant so backend plugin cache syncs must expose marketplace plugin links

## Root Cause

The launcher copied bundled backend plugins into `~/.codex/plugins/cache/openai-bundled/<plugin>/<version>`, but Computer Use and Read Aloud did not create the `latest` symlink or `~/.codex/.tmp/bundled-marketplaces/openai-bundled/plugins/<plugin>` link used by runtime plugin discovery. Remote mobile control could therefore show UI/enrollment surfaces while app-server could not resolve the actual Computer Use backend plugin.

## Validation

- `bash -n launcher/start.sh.template tests/scripts_smoke.sh`
- `git diff --check`
- `node --test linux-features/remote-mobile-control/test.js`
- `node --test scripts/patch-linux-window-ui.test.js`
- `nix build .#codex-desktop-remote-mobile-control` and inspected generated launcher/backend binary
- `nix build .#codex-desktop-computer-use-ui-remote-mobile-control` and inspected generated launcher/backend binary

Note: full `bash tests/scripts_smoke.sh` did not complete locally; plain PATH lacked `awk`, and rerunning with `nix shell nixpkgs#gawk` hit an unrelated existing packaging-log assertion before the launcher sanity block.

PR #260, opened 2026-05-20
Fix Linux remote mobile daemon runtime startup

### LINKED ISSUE #254: Mobile device no longer able to connect to machines running codex-desktop-linux
I'm seeing
"Connect to Codex Remote to load recent threads for this host."

on all of my machines now. Is anyone else seeing this?

I'm running the nix version of the package.

joshyorko:
## Summary

Fixes #254.

Remote mobile control can pass the Linux enrollment/UI patches but still fail to reconnect from Android if the upstream managed standalone Codex runtime is missing at `~/.codex/packages/standalone/current/codex`.

This PR makes the opt-in `remote-mobile-control` feature stage a runtime marker, then has the Linux launcher cold-start the remote-control app-server daemon from that managed standalone runtime.

## Details

- Adds a `remote-mobile-control` stage hook that writes `.codex-linux/remote-mobile-control-enabled` into the generated app.
- On cold start, when that marker exists, the launcher provisions the standalone runtime under `~/.codex/packages/standalone` if it is missing.
- The installer is run with `CODEX_INSTALL_DIR` pointed at `~/.codex/packages/standalone/.bin`, a sanitized system PATH, and `setsid` so it does not prompt, does not see user-managed `codex`, does not write `~/.local/bin/codex`, does not edit shell profiles, and does not mutate `CODEX_CLI_PATH`.
- Starts `app-server daemon start --enable remote_control` from the managed standalone binary.
- Documents the split between the user's interactive CLI, such as Homebrew `codex`, and the daemon-only standalone runtime.
- Keeps the Chrome marketplace fallback smoke fixture isolated by stubbing the Chrome host builder in that test.

## Validation

From the upstream-main PR branch:

- `bash -n launcher/start.sh.template tests/scripts_smoke.sh linux-features/remote-mobile-control/stage.sh`
- `git diff --check`

All passed locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,31 @@ Feature-specific Nix outputs are additive. To enable both the Computer Use UI an
 nix run github:ilysenko/codex-desktop-linux#computer-use-ui-remote-mobile-control
 ```
 
+For a declarative NixOS/Home Manager install with the mobile remote-control
+app-server managed by systemd instead of the Desktop launcher, import the flake
+module:
+
+```nix
+{
+  imports = [
+    inputs.codex-desktop-linux.homeManagerModules.default
+  ];
+
+  programs.codexDesktopLinux = {
+    enable = true;
+    computerUseUi.enable = true;
+    remoteMobileControl.enable = true;
+    remoteControl.enable = true;
+  };
+}
+```
+
+This installs the selected Codex Desktop package variant and starts a user
+`codex-remote-control.service` with
+`codex app-server --remote-control --listen unix://`. A
+`nixosModules.default` export is also available for system-level configurations
+that prefer a global user unit.
+
 `nix develop github:ilysenko/codex-desktop-linux` enters a dev shell with the required tooling.
 
 ### Cachix binary cache
--- a/flake.nix
+++ b/flake.nix
@@ -609,5 +609,15 @@ PY
           ];
         };
       }
-    );
+    ) // {
+      homeManagerModules = rec {
+        default = import ./nix/home-manager-module.nix { inherit self; };
+        codex-desktop-linux = default;
+      };
+
+      nixosModules = rec {
+        default = import ./nix/nixos-module.nix { inherit self; };
+        codex-desktop-linux = default;
+      };
+    };
 }
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -28,7 +28,7 @@ WEBVIEW_PID_FILE="$APP_STATE_DIR/webview.pid"
 LAUNCH_ACTION_RUNTIME_DIR="${XDG_RUNTIME_DIR:-$APP_STATE_DIR}/$CODEX_LINUX_APP_ID"
 LAUNCH_ACTION_SOCKET="$LAUNCH_ACTION_RUNTIME_DIR/launch-action.sock"
 PACKAGED_RUNTIME_HELPER="$SCRIPT_DIR/.codex-linux/codex-packaged-runtime.sh"
-REMOTE_MOBILE_CONTROL_MARKER="$SCRIPT_DIR/.codex-linux/remote-mobile-control-enabled"
+COLD_START_HOOK_DIR="$SCRIPT_DIR/.codex-linux/cold-start.d"
 MANAGED_NODE_BIN_DIR="$SCRIPT_DIR/resources/node-runtime/bin"
 APP_NOTIFICATION_ICON_NAME="$CODEX_LINUX_APP_ID"
 APP_NOTIFICATION_ICON_BUNDLE="$SCRIPT_DIR/.codex-linux/$APP_NOTIFICATION_ICON_NAME.png"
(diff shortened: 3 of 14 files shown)


ilysenko (changes requested):
Thanks, this is useful and the separation from the user's interactive `CODEX_CLI_PATH` looks good.

Could you please keep the remote-mobile-specific runtime logic inside the `linux-features/remote-mobile-control` feature instead of hardcoding it directly into the main launcher?

Right now the feature is opt-in via a staged marker, but the actual standalone runtime install/start logic lives in `launcher/start.sh.template`. It would be cleaner if the launcher only had a small generic best-effort hook runner, and `remote-mobile-control` staged its own cold-start hook from `linux-features/remote-mobile-control`.

Also, please make that hook non-blocking or strictly time-bounded. The installer fetch/install path and `app-server daemon start` should not block Desktop cold start if the network, GitHub, or the installer stalls.

joshyorko:
> Thanks, this is useful and the separation from the user's interactive `CODEX_CLI_PATH` looks good.
> 
> Could you please keep the remote-mobile-specific runtime logic inside the `linux-features/remote-mobile-control` feature instead of hardcoding it directly into the main launcher?
> 
> Right now the feature is opt-in via a staged marker, but the actual standalone runtime install/start logic lives in `launcher/start.sh.template`. It would be cleaner if the launcher only had a small generic best-effort hook runner, and `remote-mobile-control` staged its own cold-start hook from `linux-features/remote-mobile-control`.
> 
> Also, please make that hook non-blocking or strictly time-bounded. The installer fetch/install path and `app-server daemon start` should not block Desktop cold start if the network, GitHub, or the installer stalls.

done

colonelpanic8:
> * Starts `app-server daemon start --enable remote_control` from the managed standalone binary.

This is fine, but I think some users may prefer to run this as a systemd service or something like that.

Ah, I see there is an environment variable to control this.

colonelpanic8:
I took a look at this from the Nix package/NixOS angle. I don't think this will work there as-is, even though the feature hook itself should be staged by the Nix remote-mobile package.

Two issues stand out:

1. The hook's sanitized runtime PATH is not NixOS-compatible. It only searches `/usr/bin:/bin:/usr/sbin:/sbin` for `setsid`, `curl`/`wget`, and `tar`, but on NixOS those are typically under `/run/current-system/sw/bin` and/or the user's profile. The Nix wrapper intentionally adds those paths, but the hook overrides PATH before lookup, so first-run standalone provisioning will fail on a normal NixOS host unless `~/.codex/packages/standalone/current/codex` already exists.

2. `codex app-server daemon start --enable remote_control` looks like it won't actually make the daemon-managed app-server start with remote control. In current upstream Codex, `--enable` is folded into root config overrides, but the `app-server daemon start` path doesn't pass those overrides into `codex_app_server_daemon::run(Start)`. The daemon decides whether to pass `--remote-control` from its persisted daemon settings (`remoteControlEnabled`), and `daemon start` only loads those settings. The commands that appear to set that state are `codex remote-control start`, `codex app-server daemon bootstrap --remote-control`, or `codex app-server daemon enable-remote-control`.

For NixOS specifically, I think a declarative user service is still the better fit: it can run the Nix-managed `codex` directly as `codex app-server --remote-control --listen unix://`, avoids mutable curl-installed binaries under `~/.codex`, starts at login instead of only Desktop cold start, and gives systemd restart/logging semantics.

Separately, this PR also does not include the app-server feature sync fix from #256, so it addresses daemon startup but not the `remote_control` feature missing from the Desktop -> app-server feature sync list.

colonelpanic8:
I opened a stacked draft PR against this branch with the NixOS/Home Manager service approach and the two runtime fixes discussed above:

https://github.com/joshyorko/codex-desktop-linux/pull/21

The intent is to keep #260 useful for non-declarative Linux installs, while giving NixOS users a declarative `codex-remote-control.service` path and avoiding a second launcher-managed daemon when that service is active.

joshyorko:
Updated the branch after the review feedback:

- merged the stacked NixOS/Home Manager service work from joshyorko/codex-desktop-linux#21
- kept launcher behavior to a generic best-effort cold-start hook runner; remote-mobile runtime startup lives in linux-features/remote-mobile-control
- switched the hook to use `codex remote-control start`, added the NixOS PATH segment, and skips autostart when `codex-remote-control.service` is already active
- included the `remote_control` app-server feature sync fix
- tightened the no-`timeout` fallback so the hook continues best-effort in the background

Current head `a6dfb8f` has all PR checks green, including Nix Package Builds.

colonelpanic8:
Small follow-up to the Nix module layer: https://github.com/joshyorko/codex-desktop-linux/pull/22\n\nIt exports CODEX_REMOTE_CONTROL_DAEMON_AUTOSTART_DISABLED via Home Manager's systemd.user.sessionVariables as well as home.sessionVariables, so GUI/systemd-launched Codex Desktop processes see the declarative-service disable flag too. The hook already skips when codex-remote-control.service is active, so this is a robustness improvement rather than a blocker.

colonelpanic8:
@joshyorko can you add that this fixes #254 to the description.

ilysenko (approved):
Thanks, the requested changes are addressed.

PR #261, opened 2026-05-20
fix(launcher): keep GPU compositing on

moxw:
## Summary

- Keep native Linux launches from adding `--disable-gpu-compositing` by default
- Preserve `CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=1` as the opt-in flicker workaround
- Update launcher smoke coverage, troubleshooting docs, and changelog notes for the proposed default

## Context

The current launcher history looks like this:

- `8f8f150` added `--disable-gpu-compositing` as a default Linux launcher flag to work around Ubuntu window flickering
- `8c52fa4` added the WSLg rendering profile and carved WSLg out of that default, so WSLg keeps GPU compositing enabled while native/default Linux still disables it
- This PR proposes making native/default Linux match WSLg: keep GPU compositing enabled unless the user explicitly opts into the old workaround

The reason for proposing the flip is that on my native Linux desktop, the live Electron GPU process was pinned around 100% CPU with `--disable-gpu-compositing`. A side-by-side launch with `CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=0` idled normally around 0-1% GPU-process CPU. That suggests the old flicker workaround can be much more expensive than the default compositor path on at least some native Linux setups.

## Maintainer question

@ilysenko does this default flip match how you want to handle the tradeoff? My read is that GPU compositing is the more performant/default Electron path, and `--disable-gpu-compositing` should be a fallback for users who still see the original flickering issue. If you think this should stay default-on for a narrower class of Linux sessions, I can rework this into detection or a more targeted launcher profile instead.

## User-visible behavior

- Native Linux/X11 users keep Electron GPU compositing enabled unless they explicitly opt into the legacy workaround
- This avoids sustained Electron GPU-process CPU usage seen when `--disable-gpu-compositing` is forced on some X11/NVIDIA desktops
- Users who still hit flickering can start with `CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=1`

## Validation

- Isolated `test_launcher_template_sanity` smoke function passed
- `git diff --check`
- `bash tests/scripts_smoke.sh` currently fails before the launcher coverage on fresh `main` at `setup-native-read-aloud-defaults`: expected `Settings file: .../settings.json (file)` in the generated output log

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 - Nix builds, installer apps, and dev shells now use modern `7zz`, and the installer dependency check accepts `7zz` without requiring a separate legacy `7z` binary.
 - Codex Desktop no longer removes user-enabled `remote_control = true` from the local Linux config before starting the app server.
 - Linux webview bundles no longer ask current Codex CLI app servers to enable unsupported feature flags, avoiding connector authentication sync errors.
+- Native Linux launches now keep GPU compositing enabled by default, avoiding sustained Electron GPU-process CPU usage on some X11/NVIDIA desktops. Users who still need the old flicker workaround can opt in with `CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=1`.
 
 ## [0.8.0] - 2026-05-16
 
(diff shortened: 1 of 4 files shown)


ilysenko (approved):
Thanks, this default flip makes sense to me. Keeping the old workaround available behind CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=1 is a good tradeoff.

PR #262, opened 2026-05-20
Stop forcing the in-app browser visible

moxw:
## Summary
- Remove the Linux in-app browser auto-visible core patch
- Keep browser automation itself, node_repl approval, and Chrome integration patches unchanged
- Add regression coverage that main bundle patching leaves in-app browser tab creation untouched

## Context
This removes a Linux-only convenience patch that forced the in-app browser panel visible during browser automation tab creation/reuse by injecting setBrowserVisibleForBrowserUse(true, turnId) into upstream createTabForBrowserUse.

I observed an Electron SIGSEGV on Linux immediately after the in-app browser webview detach/dispose path, so this PR avoids changing that upstream webContents lifecycle from the Linux patch layer.

## Validation

### DIFF
--- a/scripts/patch-linux-window-ui.js
+++ b/scripts/patch-linux-window-ui.js
@@ -54,7 +54,6 @@ const {
 } = require("./patches/launch-actions.js");
 const {
   applyBrowserUseNodeReplApprovalPatch,
-  applyLinuxBrowserUseIabVisibleOnCreatePatch,
   applyLinuxChromeExtensionStatusPatch,
   applyLinuxExplicitIpcQuitPatch,
   applyLinuxExplicitQuitPromptBypassPatch,
@@ -134,6 +133,11 @@ function main() {
 
 if (require.main === module) {
   main();
+}
+
+function applyLinuxBrowserUseIabVisibleOnCreatePatch(currentSource) {
+  // Compatibility shim for old callers after the runtime patch was removed.
+  return currentSource;
 }
 
 module.exports = {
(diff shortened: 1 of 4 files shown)


ilysenko:
Thanks for looking into this. I agree with the direction of not forcing the in-app browser visible if that is contributing to the Linux webContents detach/dispose crash.

One cleanup issue before this is ready: the PR removes `applyLinuxBrowserUseIabVisibleOnCreatePatch` from `main-process.js` and from the core descriptor registry, but `scripts/patch-linux-window-ui.js` still imports and re-exports it through the compatibility surface. That leaves the exported value as `undefined`. The CLI patching path probably still works, but existing tests/helpers or downstream users that import the old compatibility export and call it would now fail with `TypeError`.

Could you either remove it cleanly from the wrapper export surface too, or keep a no-op compatibility function so existing callers do not break? Since this PR is still Draft, I would also wait to merge until it is marked ready for review.

moxw:
Addressed, thanks. I rebased this onto current main, kept the removed IAB auto-visible patch out of the runtime patch registry, and retained a no-op compatibility export for applyLinuxBrowserUseIabVisibleOnCreatePatch so older helper imports can still call it without throwing. I also added regression coverage for that compatibility surface. Validation: node --test scripts/patch-linux-window-ui.test.js passes.

ilysenko (approved):
Reviewed current head 0012a079. The previous compatibility-surface concern is addressed with a no-op export, the runtime patch descriptor is removed cleanly, and regression coverage verifies both the compatibility export and that patching no longer forces the IAB visible. Local focused tests and GitHub checks pass.

PR #263, opened 2026-05-21
[codex] Treat read aloud button clicks as explicit speech

### LINKED ISSUE #234: Read aloud interest check
I have a personal patch that reads aloud the response, sometimes when I'm too lazy to read it all. For normal voice it requires some ONNX model, like Kokoro, 200 MB, running on CPU. Otherwise a robotic unpleasent one that was built in the distro. 
Any interest?
Can add it as opt-in, with a comfortable download model from HF for those who want it.
Will probably make it MCP for live conversation next.

avifenesh:
## Summary

Fixes the message-level Read Aloud button path so a visible speaker click is treated as the explicit user action it already is.

The backend and setup flow can be healthy while the button still returns to idle if the app-side runtime opt-in gate or settings path drifts. This keeps direct/internal speech calls guarded by the runtime setting, but lets the explicit message button call the same backend with `requireEnabled:false`.

This also keeps conversation mode compatible with the new explicit-button handler shape, so `source:"conversation"` continues to route through Read Aloud after both features are enabled.

Refs #234.

## Validation

- `git diff --check`
- `node --test linux-features/conversation-mode/test.js`


avifenesh:
@Leay15 can you check if this build  solve your solution?

Leay15:
@avifenesh now everything works as expected, I saw a little bug but it isn't a blocker

When the conversation follows, the button miss its state so the button begins to say the wrong hint

Leay15 (approved):

avifenesh:
@Leay15 thanks, good catch. I pushed a small follow-up in `8952856` that keeps the composer voice button hint tied to `codexLinuxConversationActive`, so it now shows the stop hint while the conversation is active and repairs older/static patched bundles too.

Validated locally:
- `node --test linux-features/conversation-mode/test.js`
- `git diff --check`

avifenesh:
I reproduced another rare conversation-mode ordering issue before patching it:

- section one starts speaking
- an older completed assistant message re-renders during the tool gap / while the next section is pending
- before this fix, that older message could become the next queued speech item

Pushed `f8ee7e0` to keep the speech cursor on the pending assistant turn. Completed text that arrives while another section is still being spoken is held at the speech boundary; if the real next live turn appears before that boundary, it replaces the old deferred text. That keeps normal multi-turn speech working while preventing the jump back.

Added regression coverage for both orderings:
- old completed message appears after the newer live section starts
- old completed message appears during the tool gap before the newer live section starts

Validated locally:
- `node --test linux-features/conversation-mode/test.js`
- `git diff --check`

avifenesh:
Added one more regression guard in `baf78c9` for the exact same-message-after-finish case:

- speak a completed assistant message
- run its speech completion timer
- re-render the same completed message with both the same key and a different key
- assert it is still spoken only once

This passed on the PR branch and also passes on the commit before `f8ee7e0`, so the simple same-final-rerender case was already guarded. The repeat reported in the live older app is therefore more likely the multi-section/tool-gap selection path covered by `f8ee7e0`, where an older assistant item can be selected as the next speech item while the response is still active.

Validation after the added guard:
- `node --test linux-features/conversation-mode/test.js`
- `git diff --check`

PR #264, opened 2026-05-21
Enable Linux Desktop app-server as mobile Remote host

HanCheg0812:
## Summary
- start the Linux Desktop-managed local app-server with `--remote-control` by default
- keep custom `hostConfig.codex_cli_command` launches untouched
- render the full mobile setup flow inline for the Linux `Control this Linux desktop` setup state

## Why
On Linux, Codex Desktop starts its own child app-server for the local Electron host. The previous default launch args were only:

```text
app-server --analytics-default-enabled
```

That left the Desktop-managed host out of mobile Remote host mode, even when a separate official CLI Remote daemon was working. The phone could connect to the CLI identity, while the Desktop identity appeared grey/unusable.

After this patch, the packaged Linux Desktop child process starts as:

```text
app-server --analytics-default-enabled --remote-control
```

Local validation confirmed the Desktop child app-server established its own Remote 443 connection and the Android app could connect to the Desktop entry.

## Tests
- `git diff --check -- linux-features/remote-mobile-control/patch.js linux-features/remote-mobile-control/test.js`
- rebuilt and installed local `.deb`; verified `/opt/codex-desktop/resources/app.asar` contains the patched default app-server args

### DIFF
--- a/linux-features/remote-mobile-control/patch.js
+++ b/linux-features/remote-mobile-control/patch.js
@@ -1,5 +1,8 @@
 "use strict";
 
+const fs = require("node:fs");
+const path = require("node:path");
+
 function requireName(source, moduleName) {
   const escaped = moduleName.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
   const match = source.match(new RegExp(`([A-Za-z_$][\\w$]*)=require\\(\`${escaped}\`\\)`));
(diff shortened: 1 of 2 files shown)


colonelpanic8:
I believe that #260 accomplished the same thing.

HanCheg0812:
Yes, I agree. When I opened this PR, #260 had not yet landed on main, and the desktop-managed app-server was still launching without remote control locally. I see #260 later added `fix(remote): enable desktop app-server remote control` and is now merged, which covers the same root cause.

I'll close this PR as superseded by #260. Thanks for pointing it out.

PR #265, opened 2026-05-21
[codex] Harden launcher child environment cleanup

avifenesh:
## Summary
- clear the launcher EXIT trap inside the webview TCP probe and watchdog subshells
- clear inherited `ELECTRON_RUN_AS_NODE` before both warm-start and cold-start Electron launches
- add launcher smoke-test invariants for both cleanup guards

## Why
The launcher installs cleanup logic for the long-lived app process, but its short-lived helper subshells should never run that cleanup on exit. The Electron binary also needs to be launched as the app even if the parent environment exported `ELECTRON_RUN_AS_NODE`.

Together these guards make close/reopen and second-window launches more idempotent instead of leaving the app stuck behind stale helper state or an Electron-as-Node launch.

## Validation
- `bash tests/webview_probe_equivalence.sh`
- `git diff --check`


ilysenko (approved):
Thanks, this looks good. I updated the branch against current main and kept both smoke-test invariants.

PR #267, opened 2026-05-21
Fix Linux remote mobile Desktop-owned host cleanup

### LINKED ISSUE #266: Track Linux remote mobile host cleanup after Desktop app-server migration
Follow-up to #260.

## Problem

Linux remote mobile support now has two possible runtime owners:

- the standalone `codex app-server daemon start --enable remote_control` path staged by the remote-mobile feature
- the Desktop app-server when the app bundle is patched to launch with `--remote-control`

When Desktop app-server owns remote-control, the cold-start hook should not also launch or manage a standalone daemon. Otherwise rebuilds/reinstalls can leave stale `~/.codex/app-server-daemon/*.pid` records, duplicate Android-visible hosts, or a leaked `~/.local/bin/codex` symlink pointing at the feature-owned standalone runtime instead of the user interactive CLI.

There is also a UI grouping edge case: remote/mobile tasks without owner/repo metadata hit `No owner repo found for remote task` and get dropped from the sidebar, so they only appear after later refresh/navigation or not as expected in live Desktop views.

## Desired behavior

- If patch-report shows the Desktop app-server remote-control patch applied, skip the standalone daemon cold-start path.
- Remove dead standalone daemon pid files before handing ownership to Desktop app-server.
- Remove only leaked standalone `~/.local/bin/codex` symlinks, preserving user-managed Codex CLI links outside the feature runtime.
- Keep projectless remote/mobile tasks visible by grouping them under synthetic remote task groups instead of dropping them.

## Local validation context

This came from testing Linux Desktop with Android remote-control on Bluefin/Homebrew self-hosted builds. The stable target behavior is one `CODEX_DESKTOP_APP` host online, no stale standalone socket/pid ownership, Android avatar/host registration intact, and live remote tasks visible without waiting on a later sidebar refresh.

joshyorko:
Fixes #266.

## Summary

Follow-up to #260. Once the Linux Desktop app-server is patched to launch with `--remote-control`, the remote-mobile cold-start hook should not also launch a standalone remote-control daemon. This keeps Android from seeing duplicate/stale hosts and avoids stale local daemon ownership after rebuilds or reinstalls.

## Changes

- Stage an installed `.codex-linux/desktop-app-server-remote-control-enabled` marker when the patched app layout contains the Desktop app-server remote-control marker.
- Skip the cold-start standalone daemon from that installed marker instead of depending on a build-only `patch-report.json` path.
- Remove dead `app-server.pid` and `app-server-updater.pid` files before handing remote-control ownership to Desktop app-server.
- Remove leaked `~/.local/bin/codex` symlinks that point into the remote-mobile standalone runtime, while preserving user-managed Codex symlinks elsewhere.
- Keep remote/mobile tasks without owner/repo metadata visible as synthetic projectless remote task groups instead of dropping them after `No owner repo found for remote task`.

## Validation

- `bash -n launcher/start.sh.template tests/scripts_smoke.sh linux-features/remote-mobile-control/cold-start-hook.sh linux-features/remote-mobile-control/stage.sh`
- `git diff --check`
- `env -u CODEX_LINUX_SETTINGS_FILE bash tests/scripts_smoke.sh`

The last command unsets a local Codex Desktop runtime override from this development session; CI should not inherit that Desktop-specific `CODEX_LINUX_SETTINGS_FILE` value.

### DIFF
--- a/linux-features/remote-mobile-control/cold-start-hook.sh
+++ b/linux-features/remote-mobile-control/cold-start-hook.sh
@@ -8,6 +8,30 @@ truthy_env_value() {
     esac
 }
 
+cleanup_remote_mobile_control_interactive_symlink() {
+    local codex_home="$1"
+    local home_dir="${HOME:-}"
+    local user_codex=""
+    local resolved_user_codex=""
+    local standalone_root=""
+
+    [ -n "$home_dir" ] || return 0
+    user_codex="$home_dir/.local/bin/codex"
+    [ -L "$user_codex" ] || return 0
+    resolved_user_codex="$(readlink -f "$user_codex" 2>/dev/null || true)"
+    [ -n "$resolved_user_codex" ] || return 0
+    standalone_root="$(readlink -f "$codex_home/packages/standalone" 2>/dev/null || true)"
+    [ -n "$standalone_root" ] || standalone_root="$codex_home/packages/standalone"
+
+    case "$resolved_user_codex" in
+        "$standalone_root"/*)
+            if rm -f "$user_codex"; then
+                echo "Removed remote mobile control standalone symlink from interactive PATH: $user_codex -> $resolved_user_codex"
+            fi
+            ;;
+    esac
+}
+
 install_remote_mobile_control_runtime() {
     local codex_home="$1"
     local private_bin="$codex_home/packages/standalone/.bin"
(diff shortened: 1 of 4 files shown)


ilysenko (changes requested):
Thanks for the follow-up here. I think this needs one production-layout fix before merge.

The cold-start hook now decides that Desktop owns remote-control by reading `CODEX_LINUX_APP_DIR/../patch-report.json`, but that file is not reliably present in normal installs. The launcher passes `CODEX_LINUX_APP_DIR="$SCRIPT_DIR"`, package staging copies the generated app directory into `/opt/codex-desktop`, and the ordinary `make build-app` path does not set `CODEX_PATCH_REPORT_JSON`, so there may be no parent `patch-report.json` beside the installed app. In that layout `desktop_app_server_remote_control_enabled` returns false and the hook can still start the standalone daemon, which is the duplicate-host case this PR is trying to prevent.

The tests create exactly that parent patch-report file, so they validate the intended branch but not the real installed layout. Could we make this use a signal that is staged with the installed app, or stage the relevant patch report/marker explicitly during install/package creation and cover that layout in the test?

joshyorko:
valid. I corrected.

ilysenko (approved):
Reviewed current head 9cb5238. The production-layout marker fix is now staged into the installed app dir, the cold-start hook reads that installed marker, and the added projectless remote-task fallback is covered by tests. Local validation and GitHub checks pass.

PR #269, opened 2026-05-21
Fix Chrome plugin auto-install patch drift

multinomi:
## Summary
- tolerate upstream bundles where `externalBrowserUseAllowed` appears outside the Chrome plugin gate
- add a bounded fallback matcher for nearby Chrome plugin `name`/availability pairs
- add a regression test for feature-default occurrences that should not fail the patcher

## Root cause
The Chrome plugin patch treated any remaining `externalBrowserUseAllowed` string as proof that the Chrome auto-install gate was recognizable but unpatched. The current upstream Codex Desktop bundle also contains that string in feature defaults, so the Linux rebuild failed even when the plugin gate shape had simply drifted.

## Validation
- local updater rebuild produced and installed `codex-desktop 2026.05.21.204000-1` after this patch

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -22,20 +22,20 @@
 
         codexDmg = pkgs.fetchurl {
           url = "https://persistent.oaistatic.com/codex-app-prod/Codex.dmg";
-          hash = "sha256-bUQMcTN3GTXIYKVUa81gP4ubZbN+m4K9sAGdT9DIW2o=";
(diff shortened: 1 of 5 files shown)


colonelpanic8:
@multinomi conflicts

colonelpanic8:
Obviated now that #271 closed due to merger of #286 @ilysenko @multinomi close?

Issue #271, opened 2026-05-21
Required Linux Chrome plugin auto-install patch failed


LajnaLegenden:
Hey, getting this error whenever I try to rebuild with the remote control features enabled. I am not sure if I am missing some dependency or something but would love to get those extra features working. Running on hyprland, wayland.

Love the project!

```
~/setup/codex-desktop-linux main 
❯ make install-native > log.log
============================================
  Codex Desktop for Linux — Installer
============================================

[INFO] All system dependencies found (using 7z)
[INFO] Removing existing install directory: /home/lajna/setup/codex-desktop-linux/codex-app
[INFO] Removing cached DMG: /home/lajna/setup/codex-desktop-linux/Codex.dmg
[INFO] Using cached managed Node.js runtime: /home/lajna/.cache/codex-desktop/node-runtime/node-v22.22.2-linux-x64.tar.xz
[INFO] Managed Node.js runtime ready: /home/lajna/setup/codex-desktop-linux/codex-app/resources/node-runtime
[INFO] Downloading Codex Desktop DMG...
[INFO] URL: https://persistent.oaistatic.com/codex-app-prod/Codex.dmg
############################################################# 100.0%
[INFO] Saved: /home/lajna/setup/codex-desktop-linux/Codex.dmg (454M)
[INFO] Extracting DMG with 7z...
[INFO] Found: Codex.app
[INFO] Detected Electron version from DMG: 42.1.0
[INFO] Extracting app.asar...
[INFO] Native modules: better-sqlite3@12.9.0, node-pty@1.1.0
[INFO] Installing fresh sources from npm...
[INFO] Patched better-sqlite3 source for V8 external pointer API
[INFO] Compiling for Electron v42.1.0 (this takes ~1 min)...
[INFO] Using Electron headers: https://artifacts.electronjs.org/headers/dist
[INFO] Native modules built successfully
[INFO] Patching Linux window and shell behavior...
WARN: Could not find avatar overlay layout application — skipping Linux avatar overlay layout sync patch
WARN: Could not find close-to-tray condition — skipping Linux close-to-tray patch
/home/lajna/setup/codex-desktop-linux/scripts/patches/chrome-plugin.js:62
    throw new Error("Required Linux Chrome plugin auto-install patch failed: could not enable bundled Chrome auto-install");
    ^

Error: Required Linux Chrome plugin auto-install patch failed: could not enable bundled Chrome auto-install
[...]

PR #273, opened 2026-05-22
Fix/chrome plugin patch

Leay15:
This pull request works issue #271 and also related to PR #269 
It includes improvements to the Linux updater bridge patch and Chrome plugin auto-install patch logic, focusing on making the updater helpers more robust and ensuring Chrome plugin gates are consistently auto-installed and synchronized. The changes enhance how Electron is resolved at runtime, improve migration of previously patched code, and expand support for Chrome plugin variants. Tests are updated to cover these scenarios.

**Linux updater bridge improvements:**

* Refactored updater helpers (`codexLinuxShowUpdateMessage` and `codexLinuxQuitForUpdate`) to resolve the Electron module at runtime instead of relying on captured aliases, making the updater more resilient to bundling and code movement. (`scripts/lib/linux-update-bridge-patch.js`, 
* Improved migration logic to update already-patched code to use the new Electron resolution approach, ensuring consistent behavior after multiple patch applications. (`scripts/lib/linux-update-bridge-patch.js`, 
* Updated and added tests to verify that the new helper migration works as intended and that the patch is idempotent. (`scripts/patch-linux-window-ui.test.js`, 

**Chrome plugin gate improvements:**

* Enhanced the auto-install patch for Chrome plugin gates to support additional Chrome variants (including `chrome-dev` and `chrome-internal`) and ensure the `syncInstallStateWithChromeExtension` property is set for all relevant gates. 
* Updated the patch logic to preserve any extra fields between `name` and `isAvailable`/`isEnabled` in the gate object, preventing accidental field removal.

### DIFF
--- a/scripts/lib/linux-update-bridge-patch.js
+++ b/scripts/lib/linux-update-bridge-patch.js
@@ -28,22 +28,21 @@ function replaceAfter(source, anchor, search, replacement) {
   return source.slice(0, matchIndex) + replacement + source.slice(matchIndex + search.length);
 }
 
+function buildElectronResolverSource() {
+  return "function codexLinuxGetElectronModule(){try{return require(`electron`)}catch{try{return require(`node:electron`)}catch{return null}}}";
+}
+
 function buildQuitForUpdateSource(electronVar, callInstallAfterQuit) {
-  if (electronVar == null) {
-    return "function codexLinuxQuitForUpdate(){codexLinuxInstallAfterQuit()}";
-  }
   const prefix = callInstallAfterQuit ? "codexLinuxInstallAfterQuit();" : "";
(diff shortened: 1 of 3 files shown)


avifenesh (approved):
@Leay15, this is a clean, well-targeted fix 🙏

A couple of tiny, totally optional cleanups — none of them block merge:

- `electronVar` is now unused in `buildQuitForUpdateSource` / `buildBridgeSource` since the runtime resolver replaced it — could be dropped to avoid a confusing dead param.
- The `require(`node:electron`)` fallback can't ever resolve (there's no such builtin), so a plain `require(`electron`)` + `catch` would say the same thing more clearly.

Both are minor — fix them in a quick follow-up if you feel like it, otherwise good to merge as-is. CI is green apart from the Nix job.

Approving 🚀

PR #275, opened 2026-05-22
fix(nix): appcast-gate the upstream pin refresh bot so version bumps self-heal

### LINKED ISSUE #272: Codex app version pin mismatch in workflow causes nix failure
The 'Update Nix upstream hashes' job fails due to a version pin mismatch between the expected Codex app version and the version found in the downloaded Codex.dmg. See https://github.com/ilysenko/codex-desktop-linux/actions/runs/26256774890/job/77281189822
This means that nix can't build the package, e.g.:

```
~ ❯ nix run github:ilysenko/codex-desktop-linux                                                                       
error: hash mismatch in fixed-output derivation '/nix/store/k1dnbmdqwb77g1y2xg1g534bqvixhh28-Codex.dmg.drv':
         specified: sha256-bUQMcTN3GTXIYKVUa81gP4ubZbN+m4K9sAGdT9DIW2o=
            got:    sha256-f4/zAyMCeW9hDeIy6Hrac5VdGJyq1a0UAlSGXLrXtyk=
error: Cannot build '/nix/store/d8hx42crgrfg9pmrzfw3zjxkp2hwra5n-codex-desktop-payload-26.513.31313.drv'.
       Reason: 1 dependency failed.
       Output paths:
         /nix/store/78yijblqd38h8j5cr7xg01axhiiymgxy-codex-desktop-payload-26.513.31313
error: Build failed due to failed dependency
error: Cannot build '/nix/store/7vz98bflvll63bl2q4y20p31wkn8j27i-codex-desktop-26.513.31313.drv'.
       Reason: 1 dependency failed.
       Output paths:
         /nix/store/90b8ajrndxlklkp2j2fkbnmx30w9k12c-codex-desktop-26.513.31313
error: Build failed due to failed dependency
```

avifenesh:
Fixes #272.

## The error
The `Update Nix upstream hashes` bot wedges on every upstream Codex release, and the stale pin then breaks `nix run`/`nix build` for everyone:

```
ERROR: Codex app version pin mismatch: expected '26.519.31651', got '26.513.31313'
```
```
error: hash mismatch in fixed-output derivation '…-Codex.dmg.drv':
         specified: sha256-bUQMcTN3GTXIYKVUa81gP4ubZbN+m4K9sAGdT9DIW2o=
            got:    sha256-f4/zAyMCeW9hDeIy6Hrac5VdGJyq1a0UAlSGXLrXtyk=
```

## Root cause
`update-nix-hashes.sh` refreshes only the **Codex.dmg SRI hash** — it never updates the **version strings** (`codexVersion`/`electronVersion` in `flake.nix`, plus the native-module pins). When upstream bumps the version it refreshes the hash, then `validate-nix-pins.sh` fails on the now-stale version pins, so the bot exits before committing. It's been failing every run since the last upstream release, and meanwhile every PR's *Nix Package Builds* job fails identically on the stale pin.

## Fix — appcast-gated, version-self-healing refresh
- **`validate-nix-pins.sh`**: opt-in `WRITE_PINS` mode rewrites all five version pins from the DMG before the assertions (so they confirm the write). PR-CI path is unchanged (neither env var set → same strict behavior). With `APPCAST_URL` set, the write is gated on the upstream **Sparkle appcast** (`appcast-x64.xml`): the moving `Codex.dmg` must already match the appcast's advertised latest version, otherwise it's mid-rollout and the script exits `75` (skip) instead of pinning a transient build.
- **`update-nix-hashes.sh`**: runs validate in `WRITE+APPCAST` mode, treats exit `75` as a clean no-op skip, and only writes the DMG hash + runs the verify build once the pins are aligned.
- **`update-codex-hash.yml`**: stages and diff-guards `nix/native-modules/package.json` alongside `flake.nix`; commit message names the version. Cron tightened 2h → 1h (each run is cheap now).

## Why appcast-gated
The bot previously hashed the moving `Codex.dmg` blindly. That URL lags/rolls (we observed `26.519.22136` and `26.519.31651` minutes apart), so even a naive version-bump could pin a transient artifact. The appcast is the canonical release feed (exact `sparkle:shortVersionString`, newest-first); gating on it makes the refresh release-accurate and race-free. The feed is pull-only (no upstream webhook), so it stays a poll — just precise.

## Validation
`bash -n` both scripts; workflow YAML parses; live appcast parse → `26.519.31651`; `write_nix_string`/`write_json_dep` exercised on temp copies. Full e2e (`nix build`) runs in the workflow itself.

### DIFF
--- a/.github/workflows/update-codex-hash.yml
+++ b/.github/workflows/update-codex-hash.yml
@@ -2,7 +2,7 @@ name: Update Nix upstream hashes
 
 on:
   schedule:
-    - cron: '0 */2 * * *'
+    - cron: '0 * * * *'
   workflow_dispatch: {}
 
 permissions:
--- a/.github/workflows/update-codex-hash.yml
+++ b/.github/workflows/update-codex-hash.yml
@@ -49,19 +49,20 @@ jobs:
         run: |
           set -euo pipefail
 
-          if git diff --quiet -- flake.nix; then
-            echo "Nix hashes unchanged, nothing to do."
+          if git diff --quiet -- flake.nix nix/native-modules/package.json; then
+            echo "Nix pins unchanged, nothing to do."
             exit 0
           fi
 
           CODEX_DMG_HASH="$(scripts/ci/update-nix-hashes.sh read-flake-hash 'codexDmg = pkgs.fetchurl {' 'hash = ')"
+          CODEX_VERSION="$(grep -m1 'codexVersion = ' flake.nix | sed 's/.*"\(.*\)".*/\1/')"
 
           git config user.name  "codex-dmg-hash-bot"
           git config user.email "actions@github.com"
-          git add flake.nix
+          git add flake.nix nix/native-modules/package.json
           git commit \
-            -m "fix(nix): update upstream Nix hashes" \
-            -m "Refreshed Codex.dmg SRI hash to $CODEX_DMG_HASH." \
+            -m "fix(nix): refresh upstream Nix pins${CODEX_VERSION:+ for $CODEX_VERSION}" \
+            -m "Refreshed Codex.dmg SRI hash to $CODEX_DMG_HASH and synced codexVersion / electronVersion / native-module pins to the upstream Sparkle appcast latest." \
             -m "Verified all Codex Desktop Nix package outputs against the refreshed DMG." \
             -m "[skip ci]"
           git push origin main
--- a/.github/workflows/update-codex-hash.yml
+++ b/.github/workflows/update-codex-hash.yml
@@ -49,7 +49,7 @@ jobs:
         run: |
           set -euo pipefail
 
-          if git diff --quiet -- flake.nix nix/native-modules/package.json; then
+          if git diff --quiet -- flake.nix nix/native-modules/package.json nix/native-modules/package-lock.json; then
             echo "Nix pins unchanged, nothing to do."
             exit 0
           fi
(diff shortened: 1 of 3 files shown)


ilysenko (changes requested):
Thanks for taking this on. I do not think this is safe to merge yet.

I ran the new WRITE_PINS path against the live appcast/current DMG. The gate itself works: appcast and DMG both report 26.519.31651, and the script rewrites codexVersion/electronVersion to 26.519.31651 / 42.1.0. But that exposes two remaining blockers:

1. Electron fixed-output hashes are not refreshed. validate-nix-pins.sh now writes electronVersion, but update-nix-hashes.sh only rewrites the Codex.dmg hash before running the verify build. flake.nix still has pinned electronPlatform hashes and electronHeaders hash for the previous Electron, so an Electron bump changes the URL to v42.1.0 while keeping the v42.0.1 hashes. The bot should update those hashes too, or the verification nix build will fail after the current version write.

2. nix/native-modules/package-lock.json is left stale and is not staged. After WRITE_PINS, package.json requires electron@42.1.0 while the lockfile still pins electron@42.0.1. I confirmed npm ci --dry-run fails with: lock file's electron@42.0.1 does not satisfy electron@42.1.0. The workflow currently diff-guards/adds only flake.nix and nix/native-modules/package.json, so the committed refresh would not be reproducible for importNpmLock.

The current PR Nix failure is expected for the read-only PR path, but the post-merge self-heal path is still incomplete. Please update/stage the lockfile and Electron zip/header hashes as part of the refresh path, then rerun.

avifenesh:
Note for review: this PR's own **Nix Package Builds** job will be red — it fixes the *bot mechanism* but intentionally does not bump the stale `26.513` pin, so it fails on the same #272 condition every open PR currently hits. It's not a regression from this change.

After merge, the fix takes effect by running the bot:
- trigger **Update Nix upstream hashes** via `workflow_dispatch` (or wait for the next hourly run), and it self-heals `codexVersion`/`electronVersion` + native-module pins to the appcast latest (`26.519.31651`) and refreshes the DMG hash.

That bot commit then turns the Nix job green for everything.

ilysenko (approved):
Reviewed current head f8f9c75 after the previous requested changes. The post-merge refresh path now addresses the earlier blockers: Electron zip/header hashes are refreshed when electronVersion moves, nix/native-modules/package-lock.json is regenerated and staged, and the PR CI path remains strict/read-only unless WRITE_PINS is set. I also verified the live WRITE_PINS+appcast path against the current DMG and confirmed npm ci --dry-run succeeds after the generated lockfile refresh. The red Nix PR check is the expected stale-pin strict path and should be fixed by running the hash-refresh workflow after merge.

PR #280, opened 2026-05-22
Fix Linux Computer Use UI availability drift

moxw:
Summary
- Include the drifted `use-is-plugins-enabled-*` webview chunk in the Linux Computer Use UI availability patch
- Make the availability rewrite tolerate the current minified hook variable names
- Add regression coverage for the new upstream chunk shape

User-visible behavior
- Restores the Linux Computer Use "Any App" control when the opt-in UI flag is enabled and upstream moves the availability gate into the shared plugins-enabled chunk

Validation
- `git diff --check`
- `git diff --cached --check`

Notes
- Source-of-truth changes are limited to the patcher registry, Computer Use patch helper, and patcher tests
- Generated `codex-app/` output is not included in this PR

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -2038,6 +2038,20 @@ test("shows current Computer Use plugin UI on Linux without the upstream rollout
   );
 });
 
+test("shows current use-is-plugins-enabled Computer Use UI on Linux", () => {
+  const source =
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed the current head against the latest main. The Computer Use UI availability drift patch is scoped to the opt-in webview asset patch, has regression coverage for the new chunk shape, and local patcher/smoke tests pass.

PR #281, opened 2026-05-22
Guard Chrome browser client env lookup

moxw:
## Summary
- Patch the 1.1.15 Browser Use browser-client helper so nodeRepl.env is optional before reading feature env keys
- Apply the guard while staging both Chrome and browser bundled plugin copies
- Add smoke coverage for the guarded helper in Chrome and browser staging fixtures

## User-visible behavior
- @chrome no longer crashes before listing/opening tabs when the Node REPL runtime exposes nodeRepl without an env object

## Validation
- bash -n scripts/lib/bundled-plugins.sh
- bash -n scripts/lib/*.sh
- targeted smoke functions: test_browser_use_node_repl_fallback_runtime, test_browser_plugin_renamed_upstream_staging, test_chrome_plugin_staging
- Applied patch helper to a copy of the live cached Chrome 1.1.15 browser-client.mjs and verified the unsafe lookup was removed
- git diff --check
- git diff --cached --check

## Not run
- Full bash tests/scripts_smoke.sh currently stops before the plugin tests on an unrelated setup wizard assertion: Expected 'Window backend hint: KDE/Plasma -> KWin' in setup-native-readiness/output.log

### DIFF
--- a/scripts/lib/bundled-plugins.sh
+++ b/scripts/lib/bundled-plugins.sh
@@ -580,6 +580,7 @@ stage_chrome_plugin_from_upstream() {
     cp -R "$source_plugin" "$target_plugin"
     remove_macos_sidecar_files "$target_plugin"
     patch_chrome_plugin_for_linux "$target_plugin"
+    patch_browser_use_node_repl_env_guard "$target_plugin/scripts/browser-client.mjs"
     patch_browser_use_site_status_allowlist_fallback "$target_plugin/scripts/browser-client.mjs"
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed current head 39e147b54b554e9bcb03c905a632a750b4764f5e. The guard is scoped to the bundled browser-client helper used by both Browser and Chrome staging, remains idempotent, and leaves drift as a warning. Verified against current main with a no-commit merge, bash syntax, diff check, and full scripts_smoke.sh.

PR #285, opened 2026-05-22
fix(chrome): use Linux native pipe

moxw:
## Summary
- patch the 1.1.15 Browser and Chrome browser clients to read Linux native pipe access from `import.meta.__codexNativePipe`
- preserve the newer upstream `globalThis.nodeRepl.nativePipe` path when the trusted runtime provides it
- avoid the raw `node:net` socket fallback while keeping the Linux packaged runtime compatible

## User-visible behavior
- Linux `@chrome` / Browser Use connection attempts can use the existing native pipe bridge instead of failing with `privileged native pipe bridge is not available; browser-client is not trusted`

## Validation

### DIFF
--- a/scripts/lib/bundled-plugins.sh
+++ b/scripts/lib/bundled-plugins.sh
@@ -581,6 +581,7 @@ stage_chrome_plugin_from_upstream() {
     remove_macos_sidecar_files "$target_plugin"
     patch_chrome_plugin_for_linux "$target_plugin"
     patch_browser_use_node_repl_env_guard "$target_plugin/scripts/browser-client.mjs"
+    patch_browser_use_native_pipe_import_meta_bridge "$target_plugin/scripts/browser-client.mjs"
     patch_browser_use_site_status_allowlist_fallback "$target_plugin/scripts/browser-client.mjs"
     if ! install_chrome_extension_host_resource "$target_plugin"; then
         rm -rf "$target_plugin"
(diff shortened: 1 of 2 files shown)


ilysenko (changes requested):
Thanks for the fix. I found one blocker before merge:  currently only matches helpers shaped like . If the upstream/browser client already uses , the patch warns and leaves the helper unchanged, so the staged client never gains the intended  compatibility path. Please extend the patcher to handle the import-meta-only helper shape as well, and add smoke coverage for that case. It would also be good to assert the native-pipe bridge patch on the Browser plugin staging fixture, not only the Chrome fixture.

ilysenko:
Clarifying the blocker from my review above, because the inline code examples were stripped by shell quoting:

`patch_browser_use_native_pipe_import_meta_bridge` currently only matches helpers shaped like:

```js
function th(){let e=globalThis.nodeRepl?.nativePipe;return e==null||typeof e.createConnection!="function"?null:e}
```

If the upstream/browser client already uses an import-meta-only helper such as:

```js
function th(){let e=import.meta.__codexNativePipe;return e==null||typeof e.createConnection!="function"?null:e}
```

the patch warns and leaves the helper unchanged. The staged client then does not get the intended compatibility path:

```js
globalThis.nodeRepl?.nativePipe ?? import.meta.__codexNativePipe
```

Please extend the patcher to handle the import-meta-only helper shape too, and add smoke coverage for that case. It would also be useful to assert this native-pipe bridge patch on the Browser plugin staging fixture, not only the Chrome fixture.

moxw:
Addressed in e84dcb8.

- Extended `patch_browser_use_native_pipe_import_meta_bridge` to match helpers starting from either `globalThis.nodeRepl?.nativePipe` or `import.meta.__codexNativePipe`
- Updated the Browser staging fixture to start from the import-meta-only helper and assert the combined compatibility path
- Kept the Chrome staging fixture coverage for the global nativePipe helper shape

Validation:

ilysenko (approved):
Reviewed current head e84dcb8 after the requested fix. The nativePipe bridge patch now handles both globalThis.nodeRepl?.nativePipe and import.meta.__codexNativePipe helper shapes, and smoke coverage asserts the Browser import-meta-only fixture plus the Chrome global nativePipe fixture. Local bash syntax, git diff --check, and tests/scripts_smoke.sh passed; GitHub checks are green.

PR #286, opened 2026-05-22
[codex] Fix remote mobile patch drift for latest DMG

colonelpanic8:
## Summary
- Replays #270 remote-mobile-control patch drift fixes onto current `main`.
- Keeps the current upstream minified enrollment/settings/refresh/hydration matchers from that PR.
- Adds conflict-resolution test hardening so cold-start hook tests do not inherit host remote-control launcher state or an active user `systemctl` service.

## Root Cause
PR #270 applied cleanly, but its cold-start hook tests were not isolated from the local launcher environment on the current base. A machine with `CODEX_REMOTE_CONTROL_DAEMON_AUTOSTART_DISABLED=1` or an active `codex-remote-control.service` exits before the Desktop-owned app-server fixture branch, so the suite failed despite the fixture setup being correct.

## Validation
- `git diff --check origin/main..HEAD`

Supersedes / conflict-resolves #270.


ilysenko (approved):
Reviewed current head against main. The remote-mobile-control drift patch keeps fail-soft behavior, has idempotency/regression coverage for the current upstream shapes, and local focused plus smoke tests pass.

PR #287, opened 2026-05-22
Fix Browser Use bundled plugin staging on Linux

ohnicholas93:
## Summary

- Support both current upstream `browser-use` and legacy `browser` bundled plugin layouts.
- Make launcher-side Browser Use cache sync follow the generated bundled marketplace metadata.
- Remove stale Browser Use sibling directories during non-fresh rebuilds so marketplace entries and staged files cannot drift.
- Guard bundled marketplace local paths so Browser Use discovery cannot escape the `openai-bundled/` root.

## User-Visible Behavior

Browser Use is now discovered and installed from the bundled Linux resources again.

Before this fix, builds using the current upstream plugin layout could stage `plugins/browser-use/` but still generate/sync Linux metadata that expected `plugins/browser/`, causing Browser Use to be missing or reported as unavailable.

After this fix, the bundled marketplace exposes `browser-use`, the launcher syncs the matching plugin cache, and Browser Use is available in the Linux app.

Left is before, right is after. 

image

## Source Files

  - discovers `browser-use` first and falls back to legacy `browser`
  - removes obsolete sibling directories on non-fresh rebuilds
  - writes marketplace entries only for usable local plugin paths
  - rejects marketplace paths outside `openai-bundled/`
  - resolves Browser Use from bundled marketplace metadata before filesystem fallback
  - keeps runtime plugin cache and temporary bundled marketplace links aligned
- `tests/scripts_smoke.sh`
  - covers current `browser-use`, legacy `browser`, stale sibling cleanup, and unsafe/malformed marketplace paths

## Validation

```bash
bash -n install.sh
bash -n scripts/lib/install-helpers.sh scripts/lib/node-runtime.sh scripts/lib/process-detection.sh scripts/lib/dmg.sh scripts/lib/native-modules.sh scripts/lib/asar-patch.sh scripts/lib/webview-install.sh scripts/lib/bundled-plugins.sh scripts/lib/linux-features.sh scripts/lib/package-common.sh
bash -n launcher/start.sh.template
bash -n scripts/install-deps.sh scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh
```

## Notes

- This does not change the Browser Use runtime itself.
- A separate authenticated fetch error, `Codex auth token is unavailable`, can still occur for external URL safety checks, but local Browser Use flows such as `localhost:3000` work after the plugin staging fix.

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -534,7 +534,7 @@ PY
 }
 
 sync_browser_use_bundled_plugin_cache() {
-    local source_plugin="$SCRIPT_DIR/resources/plugins/openai-bundled/plugins/browser"
+    local source_plugin=""
     local source_marketplace="$SCRIPT_DIR/resources/plugins/openai-bundled/.agents/plugins/marketplace.json"
     local plugin_json=""
     local source_client=""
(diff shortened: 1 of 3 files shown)


moxw:
fyi: "browser-use" is deprecated and removed from upstream

ohnicholas93:
Oh, sorry, it turns out I was using a stale DMG and the PR is no longer relevant. Closing, thank you.

PR #288, opened 2026-05-22
Refactor updater bridge and fix main-process lifecycle issues

Leay15:
This PR addresses some comments made by @avifenesh and also fixes warnings that appear when running `make inspect-upstream`, due to the recent upstream DMG change.

A full build review will be required, as well as verification that all patch gates are functional. 

Since there is a considerable number of warnings, the review should be handled accordingly.

The smoke tests and separate verifications are green.

### DIFF
--- a/linux-features/remote-control-ui/patch.js
+++ b/linux-features/remote-control-ui/patch.js
@@ -41,6 +41,13 @@ function applyRemoteControlConnectionsVisibilityPatch(source) {
   if (patched !== source || source.includes(`||${LINUX_GATE}`)) {
     return patched;
   }
+  if (
+    source.includes("function p(){let") &&
+    source.includes("remote_control_connections") &&
+    source.includes("addedRemoteControlEnvIds")
+  ) {
+    return source;
+  }
   warn(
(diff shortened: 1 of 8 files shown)


ilysenko (changes requested):
Thanks for the patch. I found one blocker before merge: scripts/patches/webview-assets.js:327-331 captures drifted comment-preload minified names, but the replacement hard-codes Wd and S. If the regex fallback matches a bundle where the highlight helper or viewport object has any other minified name, the patch injects undefined Wd/S identifiers into comment-preload.js. Please capture the highlight helper and size object in currentCommentPreloadElementRegex and use those captured names in the replacement, then add a test with non-Wd/non-S names so this fallback branch is covered.

ilysenko (approved):
Reviewed the current head ebf123b after the follow-up fix. The previous comment-preload blocker is fixed: the patch now reuses captured helper/viewport identifiers and has regression coverage for non-Wd/non-S names. I rechecked the broader updater/lifecycle/remote-control patch drift changes and did not find remaining blockers. Local smoke and focused tests passed, and all GitHub checks are green.

PR #290, opened 2026-05-22
[codex] Expose Linux build provenance in app

### LINKED ISSUE #289: Make it easy to see the versions of the dmg and this codebase that were used to build the app from inside the app

colonelpanic8:
## Summary
- write Linux build provenance metadata into generated app installs, including upstream DMG hash/version, Electron version, source revision, detected Linux target, package profile, and enabled Linux features
- preserve source revision metadata inside the packaged update-builder so future local rebuilds can report which Linux codebase produced them
- add a Linux tray menu item that opens an in-app build information dialog

Fixes #289.

## Validation
- bash -n install.sh scripts/lib/*.sh scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh scripts/build-appimage.sh scripts/install-deps.sh
- bash tests/scripts_smoke.sh


ilysenko (changes requested):
Thanks for the provenance work. I found two issues that should be fixed before merging:

1. `scripts/lib/build-info.js:50`, `scripts/lib/build-info.js:57`, `scripts/lib/build-info.js:191`, and `scripts/lib/package-common.sh:447` persist local absolute paths and unsanitized git remotes into installed app/package metadata. The package builders copy these JSON files into `/opt/...` and the packaged update-builder, so a locally built `.deb`/`.rpm`/pacman artifact can expose `/home/<builder>/...`, the Linux features config path, and credential-bearing remotes such as `https://token@...`. Please omit `path`/`configPath`/`sourceInfoPath` from packaged metadata and sanitize or omit `remote` (strip credentials at minimum). `fileName`, `sizeBytes`, `sha256`, `appVersion`, `commit`, `branch`, `describe`, and `dirty` should cover the in-app provenance goal.

2. `generatedAt` and `capturedAt` use wall-clock time even when `SOURCE_DATE_EPOCH` is set (`flake.nix:381`). That makes the Nix payload and native packages less reproducible. The Nix build can also report an unknown Linux source revision because the cleaned source copied into `$TMPDIR/codex-source` has no `.git` and the flake does not pass `CODEX_LINUX_SOURCE_COMMIT`. Please derive timestamps from `SOURCE_DATE_EPOCH` when present and pass flake revision metadata (`self.rev` / `self.dirtyRev` / `self.lastModified`) into the install phase.

Local verification on the current merge result after #285: `git diff --cached --check`, `node --check scripts/lib/build-info.js`, `node --check scripts/patches/main-process.js`, `bash -n` for the touched shell scripts, `node --test scripts/patch-linux-window-ui.test.js`, and `bash tests/scripts_smoke.sh` all pass.

ilysenko (approved):
Re-reviewed current head 80721b2 after the requested changes. The build metadata no longer persists the local DMG path, Linux features config path, or staged sourceInfoPath; git remotes are sanitized for credential-bearing HTTPS URLs and local paths are omitted; generatedAt/capturedAt now honor SOURCE_DATE_EPOCH; and the Nix build passes flake revision/timestamp metadata into install.sh. I also reviewed the merge result on top of current main after #293. Local checks passed: git diff --cached --check, node --check for build-info and main-process patch code, bash -n for touched shell scripts, node --test scripts/patch-linux-window-ui.test.js, a focused sanitization/timestamp probe, and bash tests/scripts_smoke.sh. GitHub checks are green.

PR #301, opened 2026-05-23
Add Thorium support for the bundled Chrome plugin on Linux

ohnicholas93:
## Summary
- add Thorium native-messaging manifest paths, browser detection, running-process detection, and launch command support for the bundled Chrome plugin on Linux
- include Thorium in the Linux Chrome extension status/profile checks used by the app and plugin scripts
- add a small native-host compatibility fallback for older Chromium-family runtimes where the extension's `getInfo` handler calls `chrome.runtime.getVersion()`

Note: `chrome.runtime.getVersion()` is a Chrome/Chromium 143+ API. The native-host fallback is generic for older Chromium-family runtimes that can otherwise run the extension; it may become unnecessary for Thorium once Thorium moves to its next Chromium LTS base.

## User-visible behavior
Thorium users can install and use the bundled Codex Chrome plugin on Linux instead of being routed through Chrome/Brave/Chromium-only setup paths. Existing Chrome, Brave, and Chromium behavior is preserved.

## Validation:
- `cargo test -p codex-computer-use-linux --bin codex-chrome-extension-host`
- `node --test linux-features/thorium-chrome-plugin/test.js`
- `tests/scripts_smoke.sh`

### DIFF
--- a/computer-use-linux/src/bin/codex-chrome-extension-host.rs
+++ b/computer-use-linux/src/bin/codex-chrome-extension-host.rs
@@ -37,6 +37,7 @@ struct Client {
 struct PendingChromeRequest {
     client_id: usize,
     client_request_id: Value,
+    fallback_extension_info: bool,
 }
 
 #[derive(Clone)]
@@ -65,6 +66,7 @@ impl ChromeClientRouteError {
 struct HostState {
     stdout: Arc<Mutex<io::Stdout>>,
     rollout_tracker: RolloutTracker,
+    extension_id: Option<String>,
     clients: HashMap<usize, Client>,
     pending_chrome_requests: HashMap<String, PendingChromeRequest>,
     pending_client_requests: HashMap<String, PendingClientRequest>,
@@ -74,10 +76,15 @@ struct HostState {
 }
 
 impl HostState {
-    fn new(stdout: Arc<Mutex<io::Stdout>>, rollout_tracker: RolloutTracker) -> Self {
+    fn new(
+        stdout: Arc<Mutex<io::Stdout>>,
+        rollout_tracker: RolloutTracker,
+        extension_id: Option<String>,
+    ) -> Self {
         Self {
             stdout,
             rollout_tracker,
+            extension_id,
             clients: HashMap::new(),
             pending_chrome_requests: HashMap::new(),
             pending_client_requests: HashMap::new(),
@@ -305,7 +312,12 @@ fn main() -> Result<()> {
 
     let stdout = Arc::new(Mutex::new(io::stdout()));
     let rollout_tracker = RolloutTracker::new(Arc::clone(&stdout));
-    let state = Arc::new(Mutex::new(HostState::new(stdout, rollout_tracker)));
+    let extension_id = extension_id_from_args();
+    let state = Arc::new(Mutex::new(HostState::new(
+        stdout,
+        rollout_tracker,
+        extension_id,
+    )));
 
     log(&format!("listening on {}", socket_path.display()));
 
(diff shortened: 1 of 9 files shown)


ilysenko (changes requested):
Thanks for the Thorium support work. I did a stricter pass over the native host, launcher integration, patcher idempotency, and tests. CI is green and the fresh happy path looks close, but I found a couple of migration/idempotency issues that should be fixed before merge.

Findings:

1. `scripts/lib/patch-chrome-plugin.js`: the Thorium open-window command patch can falsely skip itself on an already Linux-patched Chrome/Brave/Chromium script. The new default-browser fallback inserts `thorium-browser-avx2.desktop`, while the later command patch uses `alreadyText: "thorium-browser-avx2"`. That makes the command patch report “already patched” even when the `commandPath("thorium-browser-avx2") || commandPath("thorium-browser") || commandPath("thorium")` branch is still missing. I reproduced this with a fixture that has the previous Linux open-window command block: after running the patcher, only the desktop-id fallback was added and the actual Thorium launch command was absent. Please use a command-specific idempotency marker, for example `commandPath("thorium-browser-avx2")`, and add a regression test for the already-patched Chrome/Brave/Chromium migration case.

2. `scripts/lib/patch-chrome-plugin.js`: `check-native-host-manifest.js` does not cover the previous multi-browser Linux fallback shape when upgrading from Chrome/Brave/Chromium support to Chrome/Brave/Chromium/Thorium. The current `oldText` cases cover mac/win-only and google-only Linux fallback, but not the existing `manifestPaths` array with google-chrome, BraveSoftware/Brave-Browser, and chromium. I reproduced that shape and the patcher only warned, leaving Thorium out of diagnostics. Please add that migration target and a smoke test so existing Linux-patched script shapes are upgraded to include `~/.config/thorium/NativeMessagingHosts`.

Non-blocking residual risks I noticed:
- The native host fallback for Thorium is scoped correctly to `getInfo`, but it depends on the exact `error.message` containing `chrome.runtime.getVersion is not a function`; if the extension wraps this differently, it will not trigger.
- `extension_id_from_args()` is not directly tested, so argv shape drift would silently drop `metadata.extensionId`.
- There is no live Thorium native-messaging handshake in CI, which is understandable, but the patcher tests should cover the runtime script shapes as much as possible.

Validation I ran locally:
- `node --check scripts/lib/patch-chrome-plugin.js`
- `bash -n tests/scripts_smoke.sh launcher/start.sh.template`
- `cargo test -p codex-computer-use-linux --bin codex-chrome-extension-host`
- `tests/scripts_smoke.sh`

ilysenko:
Thanks for working on Thorium support. After thinking about the maintenance boundary, I don’t think we should add Thorium-specific browser support to the always-on core path.

Thorium is a narrower browser variant, and supporting its profile paths, desktop IDs, process names, launch commands, plugin-script patches, and native-host compatibility in core would mean we own ongoing maintenance for a browser we do not regularly test.

This repo has `linux-features/` for opt-in Linux integrations like this. If you want to continue, please rework this as a disabled-by-default feature, e.g. `linux-features/thorium-chrome-plugin`, with its own `feature.json`, README, stage hook/patches, and focused tests.

If a small generic native-host compatibility fix is still needed, we can discuss that separately, but the Thorium-specific support should live as an opt-in Linux feature rather than in the default build.

ohnicholas93:
Thank you for your feedback. I agree that Thorium is not as widely used as other browser flavors, and I will indeed move it to `linux-features/` if that's okay with you.

ilysenko (approved):
Approved the updated version. Thorium support is now isolated as a disabled-by-default linux-features integration, and the remaining core changes are generic extension points/compatibility fallback.

PR #302, opened 2026-05-24
Add Wayland GPU rendering profile

VrtxOmega:
## Summary

- Add `CODEX_LINUX_RENDERING_MODE=wayland-gpu` to force native Wayland while keeping GPU compositing enabled.
- Skip forced renderer accessibility by default for that profile, while preserving the `CODEX_FORCE_RENDERER_ACCESSIBILITY=1` override.
- Document the profile for Wayland desktops where XWayland or software rendering is unstable, and add launcher smoke coverage.

## Context

On Pop!_OS 24.04 COSMIC with NVIDIA, the safe-mode/software-rendering path failed with `GPU access not allowed`, and the X11/Xwayland path could hang. The stable launch profile was native Wayland with GPU enabled, GPU compositing enabled, and forced renderer accessibility disabled. This PR keeps the existing default Linux and WSLg behavior intact while making that working profile first-class.

## Validation

- `git diff --check`
- `env -i HOME="$HOME" PATH="$PATH" USER="$USER" SHELL="$SHELL" TERM="${TERM:-xterm}" bash tests/scripts_smoke.sh`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Added
 
+- Launcher rendering mode `CODEX_LINUX_RENDERING_MODE=wayland-gpu`, which
+  forces native Wayland with GPU compositing enabled and skips forced renderer
+  accessibility by default for Wayland desktops where XWayland or software
+  rendering is unstable.
 - New opt-in Linux feature `read-aloud-mcp` that stages a standalone Rust Read
   Aloud MCP plugin with `doctor`, `read_aloud`, and `stop` tools. The MCP server
   reuses the Kokoro runner/model configuration from the Read Aloud UI feature
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Optional Linux-only additions live in `linux-features/`. Use them for integratio
 | Atomic desktops / other Linux distros | none | `.AppImage` | Local self-build only; no bundled auto-updater |
 | NixOS / Nix | flake | runnable directly | `nix run github:ilysenko/codex-desktop-linux` |
 
-Anything systemd-based should work for the optional auto-updater service (`systemd --user`). The launcher targets Wayland with `XWayland` first (better Electron popup positioning); pure Wayland sessions fall through to `--ozone-platform-hint=auto`. X11 is fully supported.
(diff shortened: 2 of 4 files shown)


ilysenko (changes requested):
Thanks for adding the Wayland GPU profile. I re-reviewed the current head and confirmed the existing renderer-accessibility concern is still present, so I cannot merge this yet.

Finding:

- `launcher/start.sh.template`: `should_force_renderer_accessibility()` now keys off `ELECTRON_RENDERING_MODE=wayland-gpu`, not the effective Electron platform. When a user has `CODEX_LINUX_RENDERING_MODE=wayland-gpu` exported globally but launches an explicit X11 fallback (`--x11` or `--safe-mode`), the launcher still suppresses `--force-renderer-accessibility` even though it is actually launching with `--ozone-platform=x11`. I reproduced this locally: `CODEX_LINUX_RENDERING_MODE=wayland-gpu ... --x11` produced `ozone_platform=x11 renderer_accessibility=0` and no `--force-renderer-accessibility`. The `--safe-mode` path has the same issue while also disabling GPU.

Please base the default accessibility decision on the effective platform/explicit overrides, or add a guard so explicit X11 fallback keeps the prior default accessibility behavior. Also add smoke coverage for `CODEX_LINUX_RENDERING_MODE=wayland-gpu --x11` and `CODEX_LINUX_RENDERING_MODE=wayland-gpu --safe-mode`.

Validation I ran:
- targeted launcher probe for `wayland-gpu --x11` and `wayland-gpu --safe-mode`

VrtxOmega:
Thanks for the clear review. I pushed `c7e7e27` to address this.

The follow-up changes `should_force_renderer_accessibility()` so the `wayland-gpu` default is based on the effective ozone platform rather than only the requested rendering profile. Native Wayland still skips `--force-renderer-accessibility`, but explicit X11 fallbacks now preserve the prior default behavior:

- `CODEX_LINUX_RENDERING_MODE=wayland-gpu --x11` keeps `--force-renderer-accessibility`
- `CODEX_LINUX_RENDERING_MODE=wayland-gpu --safe-mode` keeps `--force-renderer-accessibility`
- pass-through `--ozone-platform=x11` also keeps the X11 accessibility default

I also added smoke coverage for those cases and reran validation:

- `git diff --check`
- targeted launcher sanity coverage
- `env -i HOME="$HOME" PATH="$PATH" USER="$USER" SHELL="$SHELL" TERM="${TERM:-xterm}" bash tests/scripts_smoke.sh`

ilysenko (approved):
Re-reviewed current head c7e7e27. The previous renderer-accessibility concern is addressed: wayland-gpu still skips forced renderer accessibility on native Wayland, while explicit X11 fallback paths (--x11, --safe-mode, and pass-through --ozone-platform=x11) preserve the existing --force-renderer-accessibility default.

Validation checked:
- GitHub Actions are green
- bash -n launcher/start.sh.template tests/scripts_smoke.sh
- targeted launcher probe for native Wayland, --x11, --safe-mode, and pass-through X11
- full tests/scripts_smoke.sh

PR #303, opened 2026-05-24
fix(build): add GCC 16 std::nullptr_t compatibility workaround

Vikyek:
GCC 16.1.1 no longer exposes `nullptr_t` in the global namespace. Electron 42.1.0's V8 headers (`v8-internal.h` and others) use it unqualified, which causes compilation failures of native modules (like `keytar`, `cpu-features`, etc.) during the npm/electron-rebuild stage when building on systems with GCC 16+.

This change wraps the C++ compiler (`CXX`) to force-include a local header containing `using std::nullptr_t;` to pull `nullptr_t` back into the global namespace, but only if the C++ compiler actually requires this workaround. This ensures compatibility on newer toolchains without modifying Electron's headers directly.

### DIFF
--- a/scripts/lib/native-modules.sh
+++ b/scripts/lib/native-modules.sh
@@ -168,6 +168,26 @@ build_native_modules() {
     info "Compiling for Electron v$ELECTRON_VERSION (this takes ~1 min)..."
     info "Using Electron headers: $ELECTRON_HEADERS_URL"
     [ -f "$build_dir/node_modules/@electron/rebuild/lib/cli.js" ] || error "electron-rebuild CLI not found in native build toolchain"
+
+    # GCC 16+ no longer exposes nullptr_t in the global namespace; Electron's V8
+    # headers use it unqualified, breaking native module compilation.  Detect the
+    # issue and wrap the C++ compiler to force-include a fixup header if needed.
+    local _cxx="${CXX:-c++}"
+    if ! echo 'nullptr_t x = nullptr;' | "$_cxx" -x c++ -std=c++20 -fsyntax-only - 2>/dev/null; then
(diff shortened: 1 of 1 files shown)


ilysenko (changes requested):
Thanks for the GCC 16 compatibility fix. I found a blocker in the detection guard that should be fixed before this lands.

Finding:

- `scripts/lib/native-modules.sh:176`: the probe checks `nullptr_t` without including `<cstddef>`. On the current GCC 15 toolchain this already fails (`nullptr_t` is defined in `<cstddef>`), while `#include <cstddef>
nullptr_t x = nullptr;` passes. That means the wrapper is applied on toolchains that do not need the GCC 16 workaround, so this path always replaces `CXX` during native module rebuilds instead of only doing so when Electron/V8 headers need it. Please make the probe mirror the actual failure mode, e.g. include `<cstddef>` before checking unqualified `nullptr_t`, and add a smoke/regression test proving the wrapper is not installed on a compiler where the included probe succeeds.

Non-blocking follow-up: the wrapper path also assumes `CXX` is a single executable name/path, so it would not preserve common wrappers like `CXX="ccache g++"`. That is existing-adjacent shell complexity, but if this code is touched it would be good to make the failure mode explicit.

Validation I ran:
- `bash -n scripts/lib/native-modules.sh tests/scripts_smoke.sh`
- local compiler probe with and without `#include <cstddef>`

ilysenko:
Thanks again for the GCC 16 nullptr_t compatibility report and initial fix. I landed a scoped version in #310 that keeps the workaround gated behind a <cstddef>-based probe, applies it to both the regular native-module rebuild path and the Nix native-modules build, and adds smoke coverage for both the no-wrapper and wrapper-needed cases. Closing this PR as superseded by #310.

PR #305, opened 2026-05-24
Allow Browser plugin file URLs on Linux

### LINKED ISSUE #268: Browser plugin file:// support is blocked by URL policy on Linux desktop build
## Summary
In `codex-desktop-linux`, the bundled OpenAI Browser plugin advertises support for `file://` URLs, but Browser Use rejects `file://` before origin allowlisting or user approval can happen.

I am not sure yet whether this is specific to the Linux wrapper/metadata path or an upstream OpenAI Browser plugin mismatch, but I wanted to report it here first because this install is using the Linux community desktop build.

## Environment
- App: `ilysenko/codex-desktop-linux`
- Codex Desktop shown in logs: `26.513.31313`
- Browser plugin: `browser@openai-bundled` `0.1.0-alpha2`
- OS: Fedora Linux 44
- Browser backend: in-app browser (`iab`)
- Installed app path includes: `/opt/codex-desktop`

## Observed behavior
Using the Browser plugin, localhost HTTP works correctly after approval/allowlisting:

```text
http://127.0.0.1:8766/
```

However, navigating to a local `file://...` URL fails with:

> Browser Use cannot visit the requested page because its URL is blocked by the Browser Use URL policy.

`data:` URLs are blocked the same way.

## Expected behavior
Either:

1. `file://` should work, because the Browser plugin description says it supports local `file://` targets, or
2. the plugin manifest/docs should not advertise `file://` support.

## Local evidence
The Browser plugin manifest/skill description says Browser can verify local targets including `file://`.

But the bundled `browser-client.mjs` appears to gate URLs with logic equivalent to:

```js
var Z6 = new Set(["about:blank"])

function Qv(url) {
  if (Z6.has(url)) return true
  const parsed = new URL(url)
  return parsed.protocol === "http:" || parsed.protocol === "https:"
}
```

So `file:` and `data:` are rejected before any origin allowlist/user approval can apply.

I also observed an internal override key in the bundled client:

```text
x-codex-browser-use-security-mode = disabled-for-local-testing
```

but normal Browser plugin request metadata did not include it in my session, and it does not appear user-configurable.

## Workaround
Serving the same local files over localhost HTTP works:

```bash
python3 -m http.server 8766 --bind 127.0.0.1
```

Then Browser can use:

```text
http://127.0.0.1:8766/
```

## Question
Is this expected for the Linux desktop build, or is the Linux wrapper missing a metadata/security-mode path that upstream Codex Desktop uses for local file browsing? If this is purely upstream, I can also report it to `openai/codex` with a link back here.

VrtxOmega:
## Summary

- Patch the staged Browser plugin `browser-client.mjs` URL policy so advertised local `file://` targets are allowed on Linux.
- Keep `data:` URLs blocked by the upstream policy.
- Add smoke coverage to the Browser plugin staging fixtures.

## Context

Fixes #268. The Browser plugin description advertises local `file://` support, but the bundled Browser client only allowed `http:` and `https:` after `about:blank`. This keeps the policy narrow by adding `file:` only for the staged Browser plugin path.

## Validation

- `git diff --check`
- targeted Browser staging smoke: `test_browser_use_node_repl_fallback_runtime; test_browser_plugin_renamed_upstream_staging`
- `env -i HOME="$HOME" PATH="$PATH" USER="$USER" SHELL="$SHELL" TERM="${TERM:-xterm}" bash tests/scripts_smoke.sh`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- Bundled Browser plugin staging now preserves `file://` URL support advertised
+  by the Browser plugin while keeping `data:` URLs blocked by the upstream URL
+  policy.
 - `codex-update-manager` now prunes unreferenced updater workspaces under `~/.cache/codex-update-manager/workspaces`, removing heavy build artifacts (`builder/`, `codex-app/`, `dist/`) while preserving lightweight diagnostics such as `logs/` and rebuild reports.
 - The Chrome native-messaging host now evicts stale browser clients when a newer Codex browser client connects, preventing old Node REPL sessions from repeatedly reattaching CDP and driving extension service-worker CPU.
 - The bundled Chrome plugin is now auto-installed during app startup, matching Browser Use, so the plugin page no longer falls back to an install button after restart when the Linux native host is already staged.
--- a/scripts/lib/bundled-plugins.sh
+++ b/scripts/lib/bundled-plugins.sh
@@ -643,6 +643,51 @@ path.write_text(source[:match.start()] + replacement + source[match.end():], enc
 PY
 }
 
+patch_browser_use_file_url_policy() {
+    local client="$1"
+
+    if grep -q "codexLinuxFileUrlPolicy" "$client"; then
+        return 0
+    fi
+
+    python3 - "$client" <<'PY'
+from pathlib import Path
+import re
+import sys
+
+path = Path(sys.argv[1])
+source = path.read_text(encoding="utf-8")
+pattern = re.compile(
+    r'function\s+(?P<helper>[A-Za-z_$][\w$]*)\((?P<url>[A-Za-z_$][\w$]*)\)\{'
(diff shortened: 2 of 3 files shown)


ilysenko:
Thanks for working on this. I am going to close this PR because we fixed the issue in #306 with a narrower patch and tests against the current upstream bundle shape.

The main problems with this PR were:

- The URL-policy needle does not match the current Browser plugin `browser-client.mjs`; fresh upstream `26.519.41501` still uses a `try { new URL(...) } catch { return false }` helper shape, so this patch leaves the real staged client unchanged.
- The smoke fixture used a simpler helper shape than the current upstream bundle, so CI could pass without proving that the real Browser client was fixed.
- The policy allowed all `file:` URLs. The replacement fix keeps this narrower by allowing local `file:///...` and `file://localhost/...`, while keeping remote `file://host/...`, `data:`, `javascript:`, `ftp:`, and invalid URLs blocked.
- The tests only checked for inserted strings, not the actual URL-policy behavior. #306 adds a behavior test for the helper true/false decisions.

Closing this in favor of #306. Thanks again for identifying the issue and pushing the first fix.

PR #307, opened 2026-05-24
Add Linux AppShots support

moxw:
## Summary
- enable upstream AppShots availability, capture, hotkey, and settings paths on Linux
- add a Linux AppShot backend path that captures the focused window screenshot plus AT-SPI context through bundled Computer Use
- add screenshot CLI fallbacks for grim, gnome-screenshot, spectacle, and ImageMagick import
- document Linux AppShots usage and backend commands

## User-visible behavior
- Linux builds expose AppShots and attach the focused Linux window with screenshot and accessibility context
- the default Linux AppShots hotkey is both Shift keys, with both Alt keys and Ctrl+Alt+A available in settings

## Validation
- node --check scripts/patches/appshots.js
- cargo fmt --check
- cargo check -p codex-computer-use-linux
- git diff --check
- rebuilt a side-by-side app at /tmp/codex-appshots-lab-app and verified the generated app.asar includes the AppShots preview payload


ilysenko:
Thanks for working on this. The idea is useful, but I do not think AppShots should land as an always-on core patch right now.

Please move the Linux AppShots UI exposure/settings/hotkey patches into an opt-in `linux-features/appshots` feature, disabled by default. This fits the repo policy better: AppShots is a useful Linux-specific integration, but not mandatory compatibility for every Linux build.

Main reasons:

- The availability patch currently bypasses the upstream gate by changing `macOS && flag` into `linux || macOS && flag`. That should not be core/default behavior.
- The screenshot path can fall back to CLI tools after portal denial/cancel, which may bypass the user/compositor consent boundary.
- AppShot can degrade from “focused window capture” into a full-screen screenshot if focused-window/crop fails.
- The default `DoubleShift` hotkey depends on X11/XInput tooling (`xmodmap`, `xinput`) and is not reliable enough as a default Linux-wide behavior.
- The new AppShots descriptors should not be `required-upstream` if this is optional; otherwise upstream chunk drift can break fresh builds for a non-essential feature.

I would be happy with this as an optional `linux-features/appshots` integration after the privacy/fail-closed cases are tightened.

moxw:
Thanks, addressed in the latest push.

What changed:
- Moved the AppShots main-process, availability, settings, and hotkey patch descriptors out of core and into disabled-by-default linux-features/appshots
- Left the backend appshot command available, but made AppShot screenshot capture fail closed when focused-window bounds are missing, bounds do not intersect the screenshot, or GNOME Shell / XDG Portal capture fails
- AppShot capture no longer uses the CLI screenshot fallback path
- Removed the Linux default global hotkey; opt-in users can configure a hotkey from settings after enabling the feature
- Updated README / CHANGELOG and added feature-level tests

Local validation passed:
- cargo fmt --check
- cargo clippy --workspace --all-targets -- -D warnings
- cargo check --workspace --all-targets
- cargo test --workspace --all-targets
- tests/scripts_smoke.sh
- git diff --check

moxw:
Hi @ilysenko, could you share why this PR was closed? I pushed the opt-in linux-features/appshots changes you requested and the checks are passing. If there is still a blocker or a different direction you prefer, I would appreciate the context so I can adjust it.

PR #311, opened 2026-05-25
fix(chrome): prefer running Linux profiles

moxw:
## Summary
- Prefer currently running Chrome-family profile directories before Linux Local State fallbacks in extension setup/status and window opening
- Patch current upstream browser-client profile metadata minifier symbols so Linux profile metadata is restored
- Rank Linux extension backends with open user tabs ahead of idle profile sockets before agent selection
- Extend smoke coverage for running-profile routing, current browser-client variants, and backend ordering

## Validation
- node --check scripts/lib/patch-chrome-plugin.js
- bash tests/scripts_smoke.sh
- Patched a temporary copy of the current cached Chrome plugin, verified profile markers and backend ordering marker, and ran node --check on the patched browser-client.mjs

### DIFF
--- a/scripts/lib/patch-chrome-plugin.js
+++ b/scripts/lib/patch-chrome-plugin.js
@@ -48,14 +48,18 @@ function patchFileFirstMatch(filePath, { label, oldTexts, newText, alreadyText =
     return;
   }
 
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head. No blockers found; CI and focused local checks passed.

PR #315, opened 2026-05-26
[codex] Fix build info tray patch for current bundle

colonelpanic8:
## Summary
- restore the Linux Build Information tray item for current Codex Desktop bundles
- add the same Build Information dialog to the app Help menu on Linux
- relax the tray menu matcher so it handles method-local setup before `return[`
- add regression tests for both the current tray menu shape and the Help menu insertion

## Root Cause
The build-info tray patch was narrowed to match only `getNativeTrayMenuItems(){return[`. Current upstream bundles initialize local variables before the returned menu array, so the patch skipped with `WARN: Could not find tray menu items method` even though build metadata was present.

## Impact
Linux users can reach package/build provenance from the tray menu and from the app Help menu. Both surfaces use the same `codex-linux-build-info.json` metadata and dialog renderer.

## Validation

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -1326,6 +1326,7 @@ test("adds Linux build information to the app Help menu", () => {
   const patched = applyPatchTwice(applyLinuxBuildInfoTrayPatch, source);
 
   assert.match(patched, /function codexLinuxShowBuildInfo\(\)/);
+  assert.doesNotThrow(() => new Function(patched));
   assert.match(
     patched,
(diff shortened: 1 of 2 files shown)


ilysenko:
Thanks, I re-tested this against a fresh current upstream DMG (`26.519.41501`, Electron `42.1.0`), and the real generated bundle parses correctly. The Build Information item is added to both the tray menu and the Help menu there.

I still see one future-drift risk before merging: in the Help-menu-only fallback path, if the tray class is not matched but the Help menu is, the helper insertion point falls back to `helpMenuMatch.index`. That can insert `function codexLinuxShowBuildInfo()` inside the menu template array and make the patched bundle syntactically invalid.

Could you please make that path fail soft instead? If there is no safe helper insertion point outside the menu array/object expression, skip the Help menu insertion and warn, rather than patching partially. It would also be good to add a `new Function(patched)` assertion for the Help-menu-only fixture so the test catches invalid patched JS.

ilysenko (approved):
Reviewed the updated head. The Help-menu-only fallback now inserts helpers before the menu-template statement instead of inside the array, the added syntax assertion covers the previous invalid-JS case, and the focused patch tests pass locally.

PR #316, opened 2026-05-26
fix(launcher): close successful webview tcp probes

DhanushSantosh:
## Summary
- make the launcher webview TCP probe close its successful `/dev/tcp` file descriptor and exit cleanly before the watchdog fires
- preserve the closed-port failure path by exiting early when the TCP connect itself fails
- update the launcher smoke assertion so the probe-equivalence harness covers the fixed subshell shape

## Validation
- `git diff --check`
- verified the fixed probe against the live packaged webview server on `/opt/codex-desktop`

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -1382,7 +1382,9 @@ webview_port_is_open() {
     # the 200 ms cap it would enforce.
     local probe_pid kill_pid rc=1 self_pid=$$
     ( trap - EXIT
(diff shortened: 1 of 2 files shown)


DhanushSantosh:
@ilysenko This is just a small  fix, where the tcp ports were not closing properly resulting the app not opening after first use.

ilysenko (approved):
Reviewed the launcher TCP probe change and smoke coverage. The successful probe now closes fd 3 explicitly, the closed-port failure path is preserved, and the launcher smoke suite still verifies the probe equivalence/open-closed behavior.

PR #317, opened 2026-05-26
fix(webview): guard fast-mode model tiers

LiukScot:
## Summary
- add a required Linux webview patch for the fast-mode availability hook
- guard missing model serviceTiers with optional chaining while preserving additionalSpeedTiers detection
- cover the patch helper and core descriptor registration in tests

## Tests
- node --check scripts/patches/webview-assets.js
- node --check scripts/patch-linux-window-ui.js
- node --check scripts/patch-linux-window-ui.test.js


ilysenko (approved):
Reviewed the fast-mode model guard patch. The webview asset patch is focused, idempotent for exact and drifted minified shapes, keeps additionalSpeedTiers detection intact, and the focused patch tests pass locally.

PR #318, opened 2026-05-27
Fix mobile control gate and add build thread knob

mkaoudis:
## Summary
- patch the remote-mobile-control visibility gate for the current use-plugin-install-flow bundle shape while preserving accessRequired checks
- add one MAX_BUILD_THREADS setting across Cargo, native module rebuilds, Debian/RPM/pacman packaging, and Make targets
- propose a conservative default of MAX_BUILD_THREADS=4 so builds use more than one worker by default without saturating small systems; MAX_BUILD_THREADS=0 still preserves each tool's automatic/default behavior
- document the setting and add smoke coverage for the thread propagation paths

## Existing configurability check
- origin/main had no first-class Make/script setting for native build parallelism across app rebuilds and package builders
- there are tool-specific escape hatches (for example Cargo jobs, RPM payload flags, dpkg-deb threads, and makepkg config), but they were not wired into a single repo-level knob for this flow

## Test plan
- bash tests/scripts_smoke.sh
- git diff --check origin/main..HEAD
- make -n PACKAGE_WITH_UPDATER=0 rpm

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,9 @@ NEXT_APP_DIR := $(CURDIR)/codex-app-next
 REBUILD_REPORT_DIR := $(CURDIR)/dist-next/rebuild
 PACKAGE_NAME := codex-desktop
 PACKAGE_WITH_UPDATER ?= 1
+MAX_BUILD_THREADS ?= 4
+RPM_BINARY_PAYLOAD ?= w19T$(MAX_BUILD_THREADS).zstdio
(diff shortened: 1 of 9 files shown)


mkaoudis:
Superseded by focused draft PRs #319 (remote mobile control visibility fix) and #320 (MAX_BUILD_THREADS build setting) to match the contributor guidelines' one-logical-change preference.

PR #319, opened 2026-05-27
Fix remote mobile control visibility gate

mkaoudis:
## Summary
- patch the current use-plugin-install-flow visibility bundle for remote mobile control
- preserve the upstream accessRequired guard while allowing Linux hosts to expose the connection UI
- add coverage for the current bundle name/pattern and gate shape

## Review order
This PR should be reviewed after #322 lands or main otherwise gets the same PR-only Nix rollout handling. The code change is independent, but its current Nix check is red only because upstream Codex.dmg metadata is ahead of the committed flake pins.

## Why / user-visible behavior
- The current upstream bundle moved the remote-control visibility gate into use-plugin-install-flow, so the opt-in Linux mobile remote-control feature can miss the connection UI.
- With linux-features/remote-mobile-control enabled, Linux hosts can expose the mobile remote-control connection UI while still respecting upstream accessRequired state.

## Scope and limitations
- This stays inside the opt-in linux-features/remote-mobile-control feature and does not enable the feature by default.
- No committed features.example.json, distro packaging, installer, or updater behavior changes are included.
- Unrelated upstream/Nix pin drift is left out of scope for separate maintenance changes.

## Source of truth
- linux-features/remote-mobile-control/patch.js
- linux-features/remote-mobile-control/test.js

## Test plan
- git diff --check origin/main..HEAD

## CI note
- Current CI has Debian/RPM/pacman packaging and Rust/smoke tests passing.
- The remaining Nix Package Builds failure is from validate-nix-pins.sh reporting an upstream DMG app-version pin mismatch: expected 26.519.81530, got 26.519.41501. That happens before this feature patch is exercised and is addressed by #322.

### DIFF
--- a/linux-features/remote-mobile-control/patch.js
+++ b/linux-features/remote-mobile-control/patch.js
@@ -23,7 +23,7 @@ const REMOTE_CONTROL_VISIBILITY_REPLACEMENT =
 const REMOTE_CONTROL_VISIBILITY_OLD_REPLACEMENT =
   "function a({remoteControlConnectionsState:e,slingshotEnabled:t}){let n=typeof navigator!=`undefined`&&navigator.userAgent.includes(`Linux`);return t&&(n||(e?.available??!0))&&e?.accessRequired!==!0}";
 const REMOTE_CONTROL_SETTINGS_VISIBILITY_NEEDLE =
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the updated head after rebasing onto current main. The remote-mobile-control patch stays scoped to the opt-in linux-feature, local tests pass, and the fresh CI run is green.

PR #322, opened 2026-05-27
fix(ci): tolerate upstream DMG rollout in PR Nix checks

mkaoudis:
## Summary

- keep Nix package CI strict for main-branch and non-DMG failures
- allow pull-request CI to skip package-output builds when upstream Codex.dmg metadata or fixed-output hash drift is detected during rollout
- serialize Nix package-output builds and collect garbage between outputs to reduce runner store pressure

## Review order

This is the first PR in the small CI-fix stack and is ready for maintainer review from my side. Follow-up #323 depends on this behavior to get its otherwise unrelated Nix check green during the current upstream DMG rollout.

## User-visible behavior

No shipped package behavior changes. This only changes GitHub Actions behavior for PR-only upstream DMG rollout windows.

## Source of truth

- .github/workflows/ci.yml

## Validation

Local validation:

- python3 workflow YAML parse for .github/workflows/*.yml
- bash -n scripts/ci/validate-nix-pins.sh
- git diff --check origin/main...HEAD

GitHub validation on this PR:

- Rust and Smoke Tests: pass
- Build Debian Package: pass
- Build RPM Package: pass
- Build Pacman Package: pass
- Nix Package Builds: pass

## Known scope

This intentionally does not refresh committed Nix pins. Main-branch validation still fails on drift so the scheduled hash-refresh workflow remains responsible for updating pins.

PR #323, opened 2026-05-27
fix(webview): handle fast-mode guard bundle drift

mkaoudis:
## Summary

- teach the required fast-mode model guard patch to recognize the current upstream Array.isArray serviceTiers shape
- keep the existing null-safe Linux guard replacement so missing serviceTiers cannot crash the fast-mode availability hook
- add targeted patcher coverage for the drifted upstream bundle shape

## Review order

This PR should be reviewed after #322. The code change itself is independent, but its Nix check is currently red only because main lacks #322's PR-only upstream DMG rollout handling. Once #322 lands, this PR should go green after rebase or rerun.

## User-visible behavior

Linux app builds from the current upstream DMG should no longer fail required upstream patch validation because of the fast-mode guard shape.

## Source of truth

- scripts/patches/webview-assets.js

## Validation

Local validation:

- node --check scripts/patches/webview-assets.js

GitHub validation on this draft PR:

- Build App Against Upstream DMG: pass
- Rust and Smoke Tests: pass
- Build Debian Package: pass
- Build RPM Package: pass
- Build Pacman Package: pass
- Nix Package Builds: fail only on the existing upstream Codex app version pin mismatch, which is addressed by #322

## Known scope

This only fixes the fast-mode required patch drift. It intentionally leaves unrelated remote-mobile and subagent drift handling to separate focused changes.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -1118,6 +1118,17 @@ test("guards fast-mode model tier lookup when serviceTiers is missing", () => {
   assert.doesNotMatch(patched, /e\.serviceTiers\.length/);
 });
 
+test("guards current fast-mode model tier lookup with an array precheck", () => {
+  const source =
+    "function m(e){return Array.isArray(e.serviceTiers)&&e.serviceTiers.length>0||e.additionalSpeedTiers?.includes(u)===!0}";
+
+  const patched = applyPatchTwice(applyLinuxFastModeModelGuardPatch, source);
+
+  assert.match(patched, /\(e\?\.serviceTiers\?\.length\?\?0\)>0/);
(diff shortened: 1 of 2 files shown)


ilysenko:
Thanks for putting this together. We are going to close this PR because the fast-mode guard drift has already been fixed on `main` by #336 with a broader matcher and additional coverage.

The failing Nix check here was not caused by this patch; it was the older upstream app version pin mismatch. However, the branch is now draft/conflicting and the functional fix has already landed, so keeping this PR open would only duplicate the current `main` implementation.

Thanks again for catching the upstream bundle drift.

mkaoudis:
No problem! Happy to contribute when I run into stuff, thanks for reviewing :)

PR #313, opened 2026-05-26
Add desktop service and installed doctor

remybanks77:
## Summary

This PR adds a safer installed-Linux surface around the unofficial Codex Desktop Linux package:

- installs an opt-in `codex-desktop.service` plus `codex-desktop-doctor` for local health checks
- expands parity validation with `parity-smoke`, `parity-full`, strict installed-profile checks, browser matrix smoke, opt-in live browser/profile and native-host bridge checks, opt-in live Secret Service canary, static service lifecycle coverage, fake user-systemd packaged-runtime prelaunch coverage, fixture-covered Sway window backend coverage, isolated plugin install/uninstall coverage, isolated plugin detail-read and installed-plugin lifecycle coverage, isolated skill enable/disable config-write coverage, external-agent detection plus empty no-op import coverage, temporary filesystem metadata/directory/file-read plus isolated create/write/copy/remove/watch/unwatch coverage, read-only model-provider/features/permissions/collaboration surface coverage, empty no-op feature enablement-set coverage, and no-op MCP tool-call plus text/blob resource-read fixtures through an ephemeral thread
- keeps update rebuilds behind the app-server schema guard and redacts command logs for bearer/API tokens, URLs, JSON-ish remote/browser identifiers, thread IDs, MCP resource text/blob payloads, plugin IDs, plugin skill contents, skill config paths, filesystem paths, file contents, model names, feature names, permission profile IDs, collaboration mode names, and private conversation content
- makes PR-only Nix CI tolerate transient upstream `Codex.dmg` fixed-output hash or pin drift while preserving hard failures for non-DMG Nix errors and main-branch builds
- hardens managed Node runtime probing so version-only stubs are rejected
- stores experimental remote-mobile private keys in Secret Service when available, with app-id scoped fallback metadata, legacy fallback for side-by-side installs, sanitized canary failure classes, and key-file doctor output limited to mode/counts/booleans
- validates Chrome/Brave/Chromium native-messaging manifest coverage, Flatpak Chrome routing, Linux native-host binary presence, and an isolated native-host bridge loopback
- aligns Computer Use doctor readiness with the real Linux input/window backend order, including Sway and i3 capability reporting
- refreshes required upstream bundle patches for current upstream drift in fast-mode model tier guards, remote-mobile account compatibility, and subagent nickname metadata detection
- documents the desktop-environment matrix and keeps locked Computer Use explicitly research-only, with no Linux unlock/bypass implementation
[...]

PR #325, opened 2026-05-27
Reject stub managed Node runtimes

remybanks77:
## Summary

First focused split from the larger closed #313 branch, following maintainer guidance to break the work into smaller PRs.

- Changes the managed Node runtime probe from `node -v` to a real `node -e` execution check.
- Rejects version-only stub binaries that print a version but cannot execute JavaScript.
- Adds smoke coverage proving an existing stub runtime is replaced by a real managed runtime source.

## Validation

Local on `9e6ef10`:

- `bash tests/scripts_smoke.sh` -> all script smoke tests passed

GitHub on `9e6ef10`:

- `CI` run `26493942582` passed.

## Notes

This is intentionally narrow: `scripts/lib/node-runtime.sh` plus its smoke tests only.

### DIFF
--- a/scripts/lib/node-runtime.sh
+++ b/scripts/lib/node-runtime.sh
@@ -44,14 +44,20 @@ Set CODEX_MANAGED_NODE_SHA256 when overriding CODEX_MANAGED_NODE_VERSION or CODE
 
 node_version_parts() {
     local node_path="$1"
+    local probe
     local version
     local major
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head and verified the managed Node runtime probe change plus smoke coverage. This is narrow and safe to merge.

PR #327, opened 2026-05-27
Improve open-target-discovery TryExec handling

ohnicholas93:
## Summary

Focused follow-up to the closed #309 branch, following guidance to keep this work inside the existing opt-in `linux-features/open-target-discovery` feature.

This keeps the feature disabled unless explicitly listed in `linux-features/features.json`, and narrows the contribution to safer Linux desktop-entry discovery behavior.

## Changes

- Improves `TryExec` evaluation for Linux desktop entries discovered by `open-target-discovery`.
- Handles shell wrappers, env wrappers, Flatpak `info` / `run` probes, path expansion, and wrapper-only false positives more conservatively.
- Keeps Flatpak and shell probe handling fail-closed when the result cannot be statically verified.
- Adds focused discovery-level and parser-level regression coverage without a large desktop-entry matrix.


ilysenko (approved):
Reviewed the current head carefully. The change stays inside the disabled-by-default open-target-discovery feature, hardens TryExec handling fail-closed, and has focused regression coverage.

PR #328, opened 2026-05-27
fix: Update hash in flake.nix

cngongit:
the hash was out of date

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -24,7 +24,7 @@
 
         codexDmg = pkgs.fetchurl {
           url = "https://persistent.oaistatic.com/codex-app-prod/Codex.dmg";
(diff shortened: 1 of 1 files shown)


ilysenko:
Thanks for jumping on this. We opened #330 with the full fix instead of merging this hash-only update. This PR updates only the Codex.dmg hash, but the current DMG contains version 26.519.81530 while flake.nix still pins codexVersion to 26.519.41501, so the Nix validation still fails. #330 changes the refresh action so the bot follows the verified current DMG payload even when the Sparkle appcast lags, and it refreshes the hash/version pins together.

ilysenko:
Thanks for the quick hash update. This is no longer needed because #330 has been merged with the same Codex.dmg hash plus the matching codexVersion update to 26.519.81530 and the workflow fix that keeps future Nix pins refreshed from the current upstream DMG. Leaving this hash-only PR open would now reintroduce the stale version pin, so I am closing it as superseded.

PR #331, opened 2026-05-27
Add niri Computer Use window backend

mgher668:
## Summary
- add a niri window backend using `niri msg -j windows` and `niri msg action focus-window --id`
- wire niri into Computer Use backend discovery, diagnostics, focused-window lookup, and permission hints
- update setup wizard and README backend documentation

## Testing
- `cargo fmt --check`
- `cargo test -p codex-computer-use-linux windowing::`
- targeted bootstrap wizard niri hint check

Note: full `tests/scripts_smoke.sh` was attempted but the sandbox denied Python localhost socket creation in the webview probe harness, unrelated to this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ Linux Computer Use is an **opt-in** plugin that lets Codex inspect and control d
 
 - app listing and accessibility trees via AT-SPI
 - screenshots through GNOME Shell DBus or XDG Desktop Portal
-- window listing and focusing on GNOME, KWin/Plasma, Hyprland, and i3
+- window listing and focusing on GNOME, KWin/Plasma, Hyprland, niri, and i3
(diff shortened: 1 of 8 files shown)


ilysenko:
Thank you for the work on this. After reviewing it again, we are going to close this PR rather than merge it.

The Niri integration is useful for Niri users, but it adds another compositor-specific backend that we would need to keep working across Niri IPC changes, service-launched environments, portal setups, and window geometry edge cases. Our current strategy is to keep the core Linux Computer Use surface small and maintainable: core support should work for most users with as little long-term maintenance and issue volume as possible, and we cannot reasonably support every compositor-specific workflow in core.

There are also a few technical concerns in the current version: the backend does not robustly handle `NIRI_SOCKET` when launched from a service environment, some IPC type and geometry mapping details need more validation, and the portal guidance looks risky for Niri setups.

This is not a rejection of the idea, but we do not want to take it into core right now. A smaller optional integration or an externally maintained feature would be a better fit. Thanks again for contributing it.

PR #332, opened 2026-05-27
Validate app-server schemas during update rebuilds

remybanks77:
Second focused split from the closed oversized #313 branch.

Scope:
- Adds `scripts/app-server-schema-guard.js` to generate/read Codex app-server JSON schemas and verify the request/notification surface needed by Linux desktop parity probes.
- Stages the schema guard and `scripts/ci/validate-patch-report.js` into packaged update-builder bundles.
- Runs the schema guard during local update rebuilds after `install.sh` and before native package creation.
- Redacts sensitive tokens/private app fields from update-builder command logs, including common remote-control/app-server token and key field variants.
- Adds CI/package smoke assertions, schema-dir fixture coverage, and updater unit coverage.

Local validation:
- `node --check scripts/app-server-schema-guard.js`
- `bash -n scripts/ci/container-entrypoint.sh`
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p codex-update-manager builder`
- `node scripts/app-server-schema-guard.js --json --timeout-ms 30000` passed locally with `ok=true` (summary only; no raw schema output included here).

GitHub validation:
- `CI` run `26516478578` passed on `ad577fb`:
  - Rust and Smoke Tests
  - Build Debian Package
  - Build RPM Package
  - Build Pacman Package
  - Nix Package Builds

### DIFF
--- a/scripts/app-server-schema-guard.js
+++ b/scripts/app-server-schema-guard.js
@@ -1,0 +1,393 @@
+#!/usr/bin/env node
+"use strict";
+
+const { spawnSync } = require("node:child_process");
+const fs = require("node:fs");
+const os = require("node:os");
+const path = require("node:path");
+
+const REQUIRED_CLIENT_REQUEST_METHODS = [
+  "initialize",
+  "thread/list",
+  "thread/loaded/list",
+  "plugin/list",
+  "plugin/installed",
+  "plugin/read",
+  "plugin/install",
+  "plugin/uninstall",
+  "app/list",
+  "mcpServerStatus/list",
+  "mcpServer/resource/read",
+  "mcpServer/tool/call",
+  "skills/list",
+  "skills/config/write",
+  "model/list",
+  "modelProvider/capabilities/read",
+  "experimentalFeature/list",
+  "experimentalFeature/enablement/set",
+  "permissionProfile/list",
+  "collaborationMode/list",
+  "fs/createDirectory",
+  "fs/copy",
+  "fs/getMetadata",
+  "fs/readDirectory",
+  "fs/readFile",
+  "fs/remove",
+  "fs/unwatch",
+  "fs/watch",
+  "fs/writeFile",
+  "config/read",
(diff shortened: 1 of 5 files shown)


ilysenko:
Thank you for splitting this out and for adding the tests around it. The goal is useful: catching app-server protocol drift before Linux-specific parity paths silently break would be valuable. I am closing this PR because the current implementation puts that check in the wrong place and makes local updates too fragile.

The main issue is that the schema guard becomes a hard blocker in the packaged update-builder path. `build_update` runs it after `install.sh` and before package creation, so a local auto-update rebuild now fails completely if the guard cannot resolve or run a Codex CLI. That is risky for normal users: the Desktop package update path should not depend on a user CLI install being present, healthy, and able to generate schemas at that exact moment. The CLI preflight/update logic is intentionally best-effort around app startup; making it mandatory for rebuilding a Desktop package changes that contract.

The second issue is that the guard validates the user-installed `codex` CLI from persisted state / PATH, not necessarily the app-server that belongs to the candidate DMG being rebuilt. That means it can pass while the candidate Desktop bundle has drifted, or fail because the local CLI is older/newer/broken even though the candidate package would be fine. For an update-builder gate, this needs to validate the candidate bundle being packaged, not an unrelated user CLI binary.

There is also a scope concern: this hard-fails every local rebuild for a protocol surface that is mainly relevant to optional/experimental remote-control and parity probes. If we keep this kind of check, I would prefer one of these shapes:

- run it in CI / upstream-build validation rather than in every user auto-update rebuild; or
- make it a soft warning recorded in the rebuild report/logs, without blocking package creation; or
- if it must be a hard gate, make it validate the candidate app-server from the rebuilt bundle and make the required method set feature-aware.

So the direction has promise, but this version would make the updater less reliable for users. Smaller follow-up PRs are welcome if they move the guard into CI, make the local updater path non-blocking, or validate the actual candidate app-server instead of the user CLI.

PR #335, opened 2026-05-27
Install codex desktop doctor

remybanks77:
## Summary

Fourth focused split from the closed oversized #313 branch.

This PR installs a package-rendered `codex-desktop-doctor` helper for native packages. It intentionally excludes the optional `codex-desktop.service` app unit and app-service Makefile targets, which should remain a separate follow-up split.

## Changes

- Adds `packaging/linux/codex-desktop-doctor.py` as a safe installed-state healthcheck template.
- Renders `/usr/bin/<package>-doctor` for DEB, RPM, and pacman package payloads, including side-by-side package identities.
- Stages the doctor template into the installed update-builder bundle so future native package rebuilds can keep packaging the helper.
- Adds `make doctor` as a thin convenience target for installed native packages.
- Extends package-builder smoke coverage and CI package-content assertions for default and `PACKAGE_WITH_UPDATER=0` native packages.

## Safety notes

- The doctor reports sanitized statuses and file paths only.
- It does not print Chrome helper raw output, browser contents, cookies, screenshots, QR/pairing data, remote-control key material, key ids, or remote key-file paths/modes.
- The smoke suite now feeds fake secret-like helper output through the Chrome checks and asserts the doctor report does not include it.

## Validation

- `python3 -m py_compile packaging/linux/codex-desktop-doctor.py`
- `bash -n tests/scripts_smoke.sh scripts/lib/package-common.sh scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh scripts/ci/container-entrypoint.sh`
- `git diff --check`
- `node --test scripts/patch-linux-window-ui.test.js` -> 147 pass / 0 fail
- Read-only subagent review found initial issues; fixes were applied before opening this PR.


remybanks77:
Follow-up: the upstream-build failure here is the fast-mode guard drift fixed in #336. The doctor-specific native package/smoke jobs are passing; after #336 lands, this PR should only need a rerun/rebase for the upstream DMG job.

ilysenko:
Thank you for putting this together. I am closing this PR because installing a new desktop doctor command would add a new supported diagnostic surface to every package, and right now we are trying to keep the default distribution focused on the core install/update path and broadly needed fixes. Smaller targeted fixes from this work can still be proposed separately.

PR #336, opened 2026-05-27
Handle fast-mode guard bundle drift

remybanks77:
## Summary
- generalize the fast-mode model guard patch to handle drifted minified tier lookup shapes
- keep required-upstream validation failing when a recognizable unsafe lookup cannot be patched
- add coverage for old, drifted, and unpatchable fast-mode guard shapes

## Validation
- git diff --check
- bash tests/scripts_smoke.sh


ilysenko (approved):
Reviewed current head 5789499e29f65a215191e347591972d39d84b9a0. No blocking issues found; GitHub CI and local patch/smoke checks are green.

PR #337, opened 2026-05-27
Add optional Codex Desktop app service

remybanks77:
## Summary
- install an opt-in codex-desktop systemd user service in native DEB/RPM/pacman packages
- add make targets and docs for enabling, disabling, and checking the app service
- keep AppImage excluded and keep no-updater packages limited to removing updater-specific units
- add packaging smoke coverage for service inclusion and systemd shell-variable escaping

## Validation
- bash -n tests/scripts_smoke.sh scripts/lib/package-common.sh scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh scripts/ci/container-entrypoint.sh
- systemd-analyze verify --user packaging/linux/codex-desktop.service
- git diff --check
- bash tests/scripts_smoke.sh


remybanks77:
Note on CI: the upstream DMG job is currently failing on the same fast-mode guard drift fixed separately in #336. Native package/smoke jobs for this service split are passing. Once #336 lands, this branch should be rerun/rebased so the upstream-build job can validate against the drift fix without mixing the two changesets.

ilysenko:
Thank you for the contribution. I am closing this PR because an optional Codex Desktop app service adds another lifecycle mode that we would need to maintain across distros and package formats. For now we want to keep the supported runtime model smaller and focused on the existing launcher and updater behavior.

PR #320, opened 2026-05-27
Add MAX_BUILD_THREADS setting

mkaoudis:
## Summary
- add one opt-in MAX_BUILD_THREADS setting across Make targets, Cargo updater builds/tests, native module rebuilds, and Debian/RPM/pacman packaging
- keep the default at MAX_BUILD_THREADS=0 so existing tool defaults, MAKEFLAGS, makepkg config, and RPM macros continue to apply unless a user opts in
- avoid multiplying native-module concurrency by running electron-rebuild sequentially when MAX_BUILD_THREADS is nonzero while passing the requested job count to each native build
- make MAX_BUILD_THREADS authoritative for native-module MAKEFLAGS so inherited GNU make metadata cannot turn -jN into a bogus make target
- document the setting and add smoke coverage for thread propagation, default behavior, and override precedence

## TL;DR
I want one repo-level knob for local build parallelism so a native rebuild/package does not spend several extra minutes in single-threaded defaults, and so users do not have to remember separate Cargo, node-gyp, dpkg, rpm, and makepkg controls for each Make target.

## User-visible behavior
- Existing builds keep their current behavior by default because MAX_BUILD_THREADS defaults to 0.
- Users can opt in with commands such as MAX_BUILD_THREADS=8 make build-app-fresh, MAX_BUILD_THREADS=8 make package, or MAX_BUILD_THREADS=8 make install-native.
- For supported phases, a nonzero MAX_BUILD_THREADS value is the requested per-phase worker count. It also overrides inherited MAKEFLAGS for controlled make-backed subprocesses.
- RPM builders can still set RPM_BINARY_PAYLOAD directly when they need exact payload flags instead of the MAX_BUILD_THREADS-derived zstd thread setting.

## Existing configurability check
- origin/main had no first-class Make/script setting for native build parallelism across app rebuilds and package builders
- tool-specific escape hatches exist, including Cargo jobs, RPM payload flags, dpkg-deb threads, makepkg config, and MAKEFLAGS, but they were not wired into one repo-level knob for this flow
- MAX_BUILD_THREADS=0 preserves those existing escape hatches; nonzero MAX_BUILD_THREADS opts into the new repo-level behavior

## Source of truth
- Makefile
- scripts/lib/native-modules.sh
- scripts/build-deb.sh
- scripts/build-rpm.sh
- scripts/build-pacman.sh
- README.md
- tests/scripts_smoke.sh

## Local benchmark
[...]

PR #340, opened 2026-05-27
Improve Linux file manager reveal

remybanks77:
## Summary
- promote the reveal-aware Linux file-manager opener into the core main-process patch
- prefer Dolphin/Nautilus select-file behavior before folder open fallbacks and Electron shell fallback
- update open-target feature and smoke coverage for the new core baseline

## Validation
- node --check scripts/patches/main-process.js
- node --test linux-features/open-target-discovery/test.js
- bash tests/scripts_smoke.sh
- git diff --check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 - Linux webview bundles no longer ask current Codex CLI app servers to enable unsupported feature flags, avoiding connector authentication sync errors.
 - Native Linux launches now keep GPU compositing enabled by default, avoiding sustained Electron GPU-process CPU usage on some X11/NVIDIA desktops. Users who still need the old flicker workaround can opt in with `CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=1`.
 - Linux Keybinds settings now show current upstream shortcut defaults for Quick Chat, alternate New Chat, search, terminal, browser, and thread actions, and no longer lists non-dispatchable runtime rows.
+- Linux `Open in File Manager` now prefers desktop file managers directly, selecting files in Dolphin and Nautilus where supported before falling back to Electron's shell opener.
 
 ## [0.8.0] - 2026-05-16
 
--- a/linux-features/open-target-discovery/test.js
+++ b/linux-features/open-target-discovery/test.js
@@ -795,16 +795,16 @@ test("open-target discovery upgrades the baseline file manager target", async ()
   });
 });
 
-test("open-target discovery stays disabled until listed in features.json", () => {
+test("open-target discovery leaves terminal and IDE disabled until listed in features.json", () => {
   withTempFeatureConfig([], (root) => {
     assert.deepEqual(enabledLinuxFeatureIds({ featuresRoot: root }), []);
     assert.deepEqual(loadLinuxFeatureMainBundlePatches({ featuresRoot: root }), []);
 
     withLinuxFeatureRootEnv(root, () => {
       const patched = captureWarns(() => patchMainBundleSource(openTargetsBundle, null)).value;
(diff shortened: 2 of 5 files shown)


remybanks77:
CI triage:

- Debian/RPM/Pacman and Rust+Smoke passed.
- Nix failed while fetching `crate-atspi-0.29.0.tar.gz` from crates.io with HTTP 403, before building this patch.
- Upstream DMG validation failed on `linux-fast-mode-model-guard` drift; #336 updates that guard and is green.

The file-manager reveal changes are covered locally by `node --check scripts/patches/main-process.js`, `node --test scripts/patch-linux-window-ui.test.js`, `node --test linux-features/open-target-discovery/test.js`, `bash tests/scripts_smoke.sh`, and `git diff --check`.

remybanks77:
Rebased onto current `main` after #320/#338/#344 merged and force-pushed latest head `a85833a`.

Fresh local validation on `a85833a` passed:

- `git diff --check origin/main..HEAD`
- `node --test scripts/patch-linux-window-ui.test.js` -> 150 pass / 0 fail
- `node --test linux-features/open-target-discovery/test.js` -> 20 pass / 0 fail

Fresh GitHub validation passed on `a85833a`:

- CI / Rust and Smoke Tests
- CI / Build Debian Package
- CI / Build RPM Package
- CI / Build Pacman Package
- CI / Nix Package Builds
- Upstream Build App / Build App Against Upstream DMG

ilysenko (commented):
Looks good from a narrow test-coverage perspective, but I would not merge this into core as-is.

Blockers:
- This promotes the richer file-manager reveal behavior into the default main-process patch. We already have  for this kind of optional desktop integration, so the core path should stay minimal and predictable unless we first decide this behavior is required for almost everyone.
- The new core helper searches the process  and spawns the first matching , , , etc. That is a larger default runtime surface than : terminal-launched sessions or customized environments can have user/project directories ahead of system paths, so a default core action could execute an unexpected binary.
- The PR duplicates a lot of logic that already exists in the opt-in Linux feature. That makes future maintenance harder because file-manager behavior would now have to stay consistent in two places, or the feature and core patches can drift.

Non-blocking improvements if this is reworked:
- Keep the reveal-aware behavior inside , or introduce a small generic extension point in core and let the Linux feature register the richer opener.
- If any part is kept in core, avoid launching arbitrary PATH-discovered desktop apps by default; prefer a conservative  fallback there.
- Keep the default disabled-feature test asserting that optional terminal/IDE/file-manager enhancements do not activate until the feature is enabled.

ilysenko:
Corrected review comment, with code terms restored:

Looks good from a narrow test-coverage perspective, but I would not merge this into core as-is.

Blockers:
- This promotes the richer file-manager reveal behavior into the default main-process patch. We already have `linux-features/open-target-discovery` for this kind of optional desktop integration, so the core path should stay minimal and predictable unless we first decide this behavior is required for almost everyone.
- The new core helper searches the process `PATH` and spawns the first matching `dolphin`, `nautilus`, `nemo`, etc. That is a larger default runtime surface than `shell.openPath`: terminal-launched sessions or customized environments can have user/project directories ahead of system paths, so a default core action could execute an unexpected binary.
- The PR duplicates a lot of logic that already exists in the opt-in Linux feature. That makes future maintenance harder because file-manager behavior would now have to stay consistent in two places, or the feature and core patches can drift.

Non-blocking improvements if this is reworked:
- Keep the reveal-aware behavior inside `linux-features/open-target-discovery`, or introduce a small generic extension point in core and let the Linux feature register the richer opener.
- If any part is kept in core, avoid launching arbitrary PATH-discovered desktop apps by default; prefer a conservative `shell.openPath` fallback there.
- Keep the default disabled-feature test asserting that optional terminal/IDE/file-manager enhancements do not activate until the feature is enabled.

PR #341, opened 2026-05-27
Promote Linux Zed opener to core

remybanks77:
## Summary
- add a core Linux Zed opener patch for `zed`, `zeditor`, `zedit`, and `zed-cli`
- keep the opt-in `zed-opener` feature as a compatibility no-op once core has patched the bundle
- update Zed/open-target docs and patch report tests for the new core baseline

## Validation
- `node --test linux-features/zed-opener/test.js`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   bundled plugin registry so the app keeps `read-aloud` installed, and the
   launcher syncs the plugin cache so new Codex windows expose the MCP tools
   through the same auto-install path as Computer Use.
+- Linux Zed editor opener support is now enabled by default when the `zed`,
+  `zeditor`, `zedit`, or `zed-cli` command is available.
 
 ### Fixed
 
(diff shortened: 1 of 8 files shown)


remybanks77:
CI triage after the current `origin/main` rebase:

- `Build App Against Upstream DMG`: passed
- `Rust and Smoke Tests`: passed
- Debian/RPM/Pacman package builds: passed
- `Nix Package Builds`: failed while fetching `crate-atspi-0.29.0.tar.gz` from `https://crates.io/api/v1/crates/atspi/0.29.0/download`; curl got HTTP 403 on all retries and Nix reported it could not download the crate from any mirror.

That Nix failure matches the external crates.io fetch failure seen on the adjacent split PRs and is not from the Zed opener patch.

ilysenko:
Thank you for the work here. I am closing this PR because promoting a Zed-specific opener into core would make an editor-specific integration part of the default supported surface. That kind of integration fits better as an opt-in Linux feature, or as a separate discussion if there is a broader editor-opening problem we should solve generally.

PR #349, opened 2026-05-28
Support XDG nvm CLI discovery

tbezman:
## Summary
- detect nvm installs under `${XDG_CONFIG_HOME:-$HOME/.config}/nvm` in the generated launcher
- include XDG nvm installs in updater Codex CLI discovery and preflight PATH construction
- add regression coverage for GUI-style environments that do not inherit shell `NVM_DIR` or `PATH`

## Why
Some Linux setups install nvm under `$XDG_CONFIG_HOME/nvm` rather than `~/.nvm`. GUI launchers may not inherit the user shell environment, so Codex Desktop can report that the Codex CLI is missing even though `codex` works from an interactive terminal.

## Source of truth files edited
- `updater/src/codex_cli.rs`
- `tests/scripts_smoke.sh`

## Validation
- `cargo test -p codex-update-manager xdg_nvm_install_is_discovered_without_shell_env`
- `bash tests/scripts_smoke.sh launcher`

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -1170,8 +1170,16 @@ find_codex_cli() {
         return 0
     fi
 
-    if [ -s "${NVM_DIR:-$HOME/.nvm}/nvm.sh" ]; then
-        export NVM_DIR="${NVM_DIR:-$HOME/.nvm}"
+    local nvm_dir="${NVM_DIR:-}"
+    if [ -z "$nvm_dir" ]; then
+        if [ -s "${XDG_CONFIG_HOME:-$HOME/.config}/nvm/nvm.sh" ]; then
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed the current head. This is a small compatibility fix for XDG nvm installs in GUI-style environments where NVM_DIR/PATH are not inherited. The launcher keeps explicit NVM_DIR first, then prefers XDG nvm when present, and the updater mirrors that discovery in both CLI path candidates and preflight PATH construction.

Local validation passed:
- cargo test -p codex-update-manager xdg_nvm_install_is_discovered_without_shell_env
- bash tests/scripts_smoke.sh

GitHub checks are green.

PR #351, opened 2026-05-28
Expand opt-in Linux features framework

ilysenko:
## Summary

This PR turns `linux-features/` into a more complete opt-in extension framework for Linux-specific integrations that should not live in the core patch flow.

The goal is to make optional Linux features easy to add, test, package, and carry through updater rebuilds while keeping the default app minimal. By default, no optional features are enabled. A feature becomes active only when its id is listed in the git-ignored `linux-features/features.json` file.

## How it works now

- Feature manifests are discovered from tracked `linux-features/<feature-id>/feature.json` directories.
- User-private features can live in git-ignored `linux-features/local/<feature-id>/feature.json` directories, using the same manifest contract.
- Every feature directory must include a neighboring `README.md` that documents what the feature does, how to test it, and known support risks.
- Enabled features can contribute ASAR patch descriptors, staged resources, runtime launcher hooks, package hooks, or legacy `stage.sh` hooks.
- Runtime hooks are consumed by the generated Linux launcher:
  - `runtimeHooks.env` exports environment variables before launch.
  - `runtimeHooks.prelaunch` runs before webview setup on cold start.
  - `runtimeHooks.electronArgs` appends Electron launch arguments.
  - `runtimeHooks.coldStart` runs after bundled plugin cache sync.
- Runtime hooks receive `CODEX_HOME`, `CODEX_LINUX_APP_DIR`, `CODEX_LINUX_APP_STATE_DIR`, `CODEX_LINUX_FEATURES_DIR`, and `CODEX_LINUX_LAUNCHER_LOG`, so feature-owned user artifacts can be installed from app-local staged resources at real user runtime.
- Package hooks run during `.deb`, `.rpm`, and pacman staging, so optional features can adjust native package payloads when needed.
- Native package update-builder bundles preserve the enabled feature ids, so local auto-update rebuilds keep the same opt-in Linux feature set.

## Safety and maintenance details

- Features are disabled by default; `defaultEnabled: true` is rejected.
- Repository and local features share one id namespace, so local features cannot shadow tracked ones.
- Feature dependencies and conflicts are validated before staging.
- Declarative resources and runtime hooks are tracked in `.codex-linux/linux-features-staged.json` and removed on the next install when the owning feature is disabled.
- Staged file modes must be quoted octal strings and are preserved after native package permission normalization.
- Resource targets must stay inside the app directory and cannot target the app root itself, preventing cleanup from deleting the install payload.
- The top-level README now links to `docs/linux-features-architecture.md`, which documents the extension contract and intended boundaries.

## Tests

- `node --check scripts/lib/linux-features.js`
- `bash -n scripts/lib/package-common.sh scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh scripts/bootstrap-wizard.sh tests/scripts_smoke.sh launcher/start.sh.template`
- `node --test linux-features/example-feature/test.js`
- `git diff --check`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ target/
 dist/
 dist-next/
 linux-features/features.json
+linux-features/local/
 .idea/
 .claude/
 .codex/
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Before opening a pull request, please read [CONTRIBUTING.md](CONTRIBUTING.md).
 
 ## Linux features
 
-Optional Linux-only additions live in `linux-features/`. Use them for integrations that are useful for some users but should not become mandatory core patches. Copy `linux-features/features.example.json` to the git-ignored `linux-features/features.json` before building; enabled features are applied during the install/build pipeline. See [`linux-features/README.md`](linux-features/README.md) for the feature contract.
+Optional Linux-only additions live in `linux-features/`. Use them for integrations that are useful for some users but should not become mandatory core patches. Copy `linux-features/features.example.json` to the git-ignored `linux-features/features.json` before building; enabled features are applied during the install/build pipeline. User-local features can live under the git-ignored `linux-features/local/<feature-id>/` directory. See [`linux-features/README.md`](linux-features/README.md) and [`docs/linux-features-architecture.md`](docs/linux-features-architecture.md) for the feature contract.
 
 ## Supported platforms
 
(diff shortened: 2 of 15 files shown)


avifenesh:
I reviewed the code path in this PR and found one packaging edge case that looks worth fixing in a follow-up.

`resources` can point at a directory (`fs.cpSync(..., { recursive: true })`), and the initial staging path handles restrictive modes by adding traverse bits to directories in `chmodRecursive()` (`scripts/lib/linux-features.js`). However, package permission restore later records only the raw manifest mode and applies it directly to the top-level target in `scripts/lib/package-common.sh`:

```js
fs.chmodSync(target, Number.parseInt(entry.mode, 8));
```

I reproduced this with a directory resource using `mode: "0600"`: initial staging produced traversable directories (`0700`) and a private file (`0600`), but after package permission normalization + restore, the top-level payload directory became `0600`, making descendants inaccessible with `Permission denied`. With `0750`, the top-level directory is restored, but descendants remain package-normalized instead of preserving the recursive restrictive mode.

Suggested fix: either reject directory resources with non-traversable/restrictive modes, store and restore adjusted directory modes recursively, or reuse the staging `chmodRecursive` semantics in the package restore path. A smoke test with a directory resource and a restrictive mode (`0600` or `0750`) would catch this.

PR #352, opened 2026-05-28
Fix Computer Use availability and add devcontainer

joshyorko:
Fixes Linux Computer Use UI availability after upstream renderer bundle drifted to a new object-helper shape.

Adds a generic devcontainer so Rust/Node validation works cleanly on immutable Linux hosts. No Codex auth, profile, or server state included.

Validation:
- devcontainer Docker build
- Rust fmt, diagnostics tests, and clippy inside devcontainer

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,62 @@
+FROM rust:1-bookworm
+
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+ARG NODE_VERSION=22.22.2
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV PATH="/usr/local/cargo/bin:${PATH}"
+
+RUN apt-get update -qq \
+    && apt-get install -y --no-install-recommends \
+        bash \
+        ca-certificates \
+        curl \
+        file \
+        g++ \
+        gcc \
+        git \
+        jq \
+        libssl-dev \
+        make \
+        p7zip-full \
+        pkg-config \
+        python3 \
+        ripgrep \
+        shellcheck \
+        sudo \
+        tar \
+        unzip \
+        xz-utils \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN set -eux; \
+    arch="$(dpkg --print-architecture)"; \
+    case "$arch" in \
+      amd64) node_arch="x64" ;; \
+      arm64) node_arch="arm64" ;; \
+      armhf) node_arch="armv7l" ;; \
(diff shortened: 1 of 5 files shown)


ilysenko:
Thanks for the PR. The Computer Use availability fix looks useful, and I am okay with keeping the devcontainer in this PR, but I do not want to merge it as-is yet.

Main blockers:

1. The new object-helper availability patch can report success while Linux may still remain incompatible.

In `scripts/patches/computer-use.js`, the new branch patches:

```js
isComputerUseFeatureEnabled
isComputerUseFeatureLoading
isComputerUseGateEnabled
```

but it leaves:

```js
isHostCompatiblePlatform: platformPredicate(platform)
```

If the upstream platform predicate shape drifts and the earlier predicate patch does not match, the patch can still mark the availability object as changed, while `isHostCompatiblePlatform` continues to reject Linux.

Please either patch this field directly, for example by making Linux explicitly compatible there, or make the object-helper patch fail/skip unless the platform predicate was also successfully patched or already includes Linux. Please add a regression test for this case.

2. The devcontainer currently is not Docker-compatible.

`.devcontainer/devcontainer.json` uses:

```json
"--userns=keep-id"
```

That is Podman-specific and fails with Docker. Since this is presented as a generic devcontainer, please remove that run arg or make the Docker/Podman behavior explicit and safe.

3. The devcontainer installs Node 18.

Because it uses `rust:1-bookworm` plus Debian `nodejs/npm`, the container gets Node 18. This repo expects Node 20+ for the host toolchain, and the default/bootstrap path uses Node 22. Please install Node 22 in the devcontainer instead of Debian's Node 18 package.

For future PRs, please keep the scope to one feature/fix per PR when possible. I am fine with keeping both pieces here now, but separate PRs are usually much easier to review.

joshyorko:
> 5\. ls Node 18.
> 
> Because it uses `rust:1-bookworm` plus Debian `nodejs/npm`, the container gets Node 18. This repo expects Node 20+ for the host toolchain, and the default/bootstrap path uses Node 22. Please install Node 22 in the devcontainer instead of Debian's Node 18 package.
> 
> For future PRs, please keep the scope to one feature/fix per PR when possible. I am fine with keeping both pieces here now, but separate PRs are usually much easier to review.

understood, yeah i just really wanted that devcontainer in here,  thought it would be much easier for us to keep it moving if we have something disposable for the agents to use,  multiple builds etc.  updating based on your comments and guidance now. ty

ilysenko (approved):
Reviewed the updated head. The previous blockers are fixed: the object-helper availability patch now keeps Linux host compatibility explicit, the devcontainer no longer uses the Docker-incompatible user namespace flag, and it installs Node 22. Local focused tests and Docker validation passed.

PR #354, opened 2026-05-28
fix(launcher): harden cold-start webview readiness

DhanushSantosh:
## Summary
- harden launcher cold-start readiness by waiting on the actual localhost webview origin instead of only the raw TCP bind
- tolerate a concurrent verified webview server winning the bind race by adopting it instead of failing before Electron launch
- update launcher smoke coverage for the HTTP readiness path and concurrent-server race
- ignore local graphify artifacts so repo-wide graph indexing stays out of commits

## Validation
- launched `/usr/bin/codex-desktop` and verified the packaged launcher reached `Webview origin verified.`
- launched the local desktop-wrapper path and verified it reused an existing verified webview server instead of looping on startup


DhanushSantosh:
@ilysenko merging main branch to this PR?

ilysenko:
I merged current main into this PR and resolved the test conflict.

Please don't merge yet. I found one remaining runtime risk: `webview_origin_is_reachable_fast` uses `--max-time 0.2`, so a valid local webview server that answers a little slowly during cold start can be treated as not ready. I reproduced this with a valid response delayed by 300ms: the fast check fails, while the full `verify_webview_origin` check passes.

Could you please adjust the readiness path so it does not fail valid but slower local responses? For example, keep the fast probe for the quick path, but fall back to the full 2s verification before reporting startup failure. It would also be good to add a small behavioral test for this slow-valid-response case.

DhanushSantosh:
@ilysenko Sure, I'll do it in a few minutes.

DhanushSantosh:
Done @ilysenko — addressed the slow-response case.

`wait_for_webview_server` still uses the 0.2s fast probe in the loop, but if that exhausts without success it now falls back to the full 2s `verify_webview_origin` check before reporting failure. Added a behavioral test with a 300ms delayed valid response (fast probe fails, full verify passes).

Pushed in the latest commit on this branch.

ilysenko (approved):
Reviewed the latest head. The slow valid webview response fallback is now bounded and covered by the behavioral probe test; local smoke/probe tests and GitHub checks pass.

PR #355, opened 2026-05-28
fix: avoid stale config write versions on Linux

LiukScot:
## Summary
- Add a Linux webview asset patch that drops stale `expectedVersion` fields from config write requests.
- Register the patch for settings/config-writing bundles, including agent settings, apps/plugins availability, MCP settings, personalization, and personality assets.
- Add tests for expectedVersion rewriting and descriptor discovery.

## Why
Some settings writes can be rejected after startup changes the local config version first. On Linux we can avoid that stale version guard by letting the app-server write against the current config state.

## Validation
- `node --test scripts/patch-linux-window-ui.test.js` (153/153 pass)
- Manually verified by user: settings persisted after quit/reopen

No linked issue.

### DIFF
--- a/scripts/patch-linux-window-ui.js
+++ b/scripts/patch-linux-window-ui.js
@@ -97,6 +97,7 @@ const {
 const {
   applyBrowserAnnotationScreenshotPatch,
   applyLinuxAppSunsetPatch,
+  applyLinuxConfigWriteVersionConflictPatch,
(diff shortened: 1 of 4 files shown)


ilysenko (approved):
Reviewed the current head. No blocking issues found; checks and local smoke tests pass.

PR #357, opened 2026-05-29
feat(updater): add "install on close" toggle for auto-updates

avifenesh:
## Summary
- Add an "Install updates when you close Codex" toggle to the Keybinds settings page, under a new "Updates" section.
- When on (the default, preserving current behavior), a downloaded update installs automatically after you quit Codex. When off, updates wait until you explicitly click Update.
- Teach `codex-update-manager` to read the toggle from the app `settings.json` as an overlay over `config.toml`, so the in-app preference wins over the built-in default.

## Why
Today `auto_install_on_app_exit` is effectively hardcoded on, and nothing ever writes the updater `config.toml`, so a ready update always installs silently on app close with no user control. This adds a user-facing switch without changing the default behavior.

## Source-of-truth files
- `updater/src/config.rs` — `settings_auto_install_override()` reads `codex-linux-auto-update-on-exit` from `settings.json` (app-id + `CODEX_LINUX_SETTINGS_FILE` resolution mirrors the launcher/main-bundle), with the same bool/number/string coercion as `start.sh.template`.
- `updater/src/app.rs` — applies the overlay over the loaded config at the single `run()` source point.
- `scripts/patches/shared.js` — allowlists the new key in `linuxSettingsKeys` (auto-picked up by the existing settings persistence helper).
- `scripts/patches/keybinds-settings.js` — adds the "Updates" section + toggle row.

## Behavior / polarity
Absent setting ⇒ ON (matches the prior default). `LinuxToggle` defaults to `true`; the Rust overlay returns `None` when absent and falls through to the config/default. The explicit Sparkle "Update" click still installs regardless of the toggle.

## Validation
- `cargo test -p codex-update-manager` — pass (incl. new `config.rs` overlay tests: present true/false, string `off`, numeric, missing file, malformed JSON, absent key)
- `node --check scripts/patch-linux-window-ui.js` — OK
- `bash -n install.sh launcher/start.sh.template`; `bash tests/scripts_smoke.sh` — pass

## Versioning
Backward-compatible feature addition to the updater crate. I left the `updater/Cargo.toml` version bump out because this is part of a small stack of updater PRs and the final version depends on merge order — happy to bump (minor) on request, or in whichever PR lands last.

## Notes
No linked issue. Part of a focused stack of independent Linux update-UX PRs (auto-install toggle, GUI feature picker, wrapper-update axis); this one stands alone.


ilysenko (commented):
Looks good as a small UI/config idea, but I would not merge this yet.

Blockers:
- The branch currently conflicts with `main`, so it needs a rebase before it can be merged.
- The new toggle description says a downloaded update installs automatically after Codex quits, but the daemon path does not actually move a `ReadyToInstall` update into `WaitingForAppExit`. In `reconcile_pending_install`, a ready update with `auto_install_on_app_exit=true` only notifies while the app is running and then returns. It still waits for an explicit `install-ready` action before the "install after close" flow starts.
- The updater reads `settings_auto_install_override()` only once during `run()`. The daemon is long-lived, so changing the in-app toggle while the updater service is already running will not affect the running reconcile loop until the service restarts.

Other risks:
- The tests add a private env-var lock in `config.rs`. Other updater tests already share process-global env, so this should use the existing shared test env lock pattern to avoid future parallel-test races.
- The settings key and UI placement look fine, but the user-facing copy should match the real state-machine behavior after the daemon path is fixed.

Suggested path:
- Rebase on current `main`.
- Either make `ReadyToInstall + auto_install_on_app_exit` transition into `WaitingForAppExit` when the app is running, or rename the toggle/copy so it only describes the manual in-app install-ready flow.
- Reload the setting during daemon reconciliation, or document that it requires a service restart.

avifenesh:
@ilysenko 
Thanks, addressed in the latest push.

What changed:
- Rebasing conflict with current `main` is resolved.
- `ReadyToInstall` now transitions to `WaitingForAppExit` when `auto_install_on_app_exit` is enabled and Codex is still running.
- The running daemon now rereads the persisted auto-install setting during reconciliation, so toggling it in the app does not require a service restart.
- The settings copy now matches the actual state-machine behavior.
- Tests now use the shared updater env lock instead of a private `config.rs` lock.

Validation:
- Local `cargo clippy --workspace --all-targets -- -D warnings`
- Local `cargo test -p codex-update-manager`
- Local patcher test suite
- GitHub CI is green, including Rust/smoke, upstream DMG build, Debian, RPM, Pacman, and Nix.

ilysenko (approved):
Reviewed the current head, including the follow-up fix for cancelling queued auto-installs when the setting is turned off. Local tests and GitHub CI are green.

PR #359, opened 2026-05-29
feat(updater): track wrapper releases and surface a changelog

avifenesh:
## Summary
- Teach `codex-update-manager` a second update axis: detect when the *wrapper* itself (this repository's own Linux features and fixes) has advanced upstream, in addition to the existing Codex DMG axis.
- Add a `check-wrapper` subcommand and extend `status --json` to report the detected wrapper commit, its date, and a changelog of what changed.
- Surface a curated changelog: prefer `CHANGELOG.md` sections newer than the installed build, falling back to git commit subjects when versions don't map cleanly.

## Why
The Rust updater (the daily driver) tracked only OpenAI's DMG. Wrapper releases ship real Linux features/fixes that users currently only get via a manual rebuild, with no changelog. This adds the detection + changelog plumbing as the foundation for surfacing wrapper updates to the user.

## Design / safety
- Detection is **git-only and read-only** (no GitHub API, no `gh`/`curl`): a shallow fetch into a cache dir under the updater workspace; it never mutates the user's working tree.
- Installed identity is the `source.commit` recorded in `<app_dir>/.codex-linux/build-info.json` (exact for both packaged and user-local installs), with a package-version build-timestamp fallback for older installs.
- **Off by default** (`enable_wrapper_updates`): the updater touches the wrapper repo nowhere today, so opt-in keeps existing DMG-only behavior byte-for-byte. Packaged frozen bundles without a git checkout degrade gracefully.
- All new `state.json` / `config.toml` fields are `#[serde(default)]`, so existing on-disk files parse unchanged.

## Source-of-truth files
- `updater/src/wrapper.rs` *(new)* — git-based detection.
- `updater/src/changelog.rs` *(new)* — pure `CHANGELOG.md` parser + semver + git-log fallback.
- `updater/src/{state.rs,config.rs}` — new `#[serde(default)]` fields.
- `updater/src/{cli.rs,app.rs,main.rs,builder.rs}` — `check-wrapper` wiring + a `build_update_from` seam.
- `CHANGELOG.md` — `[Unreleased]` entry.

## Validation
- `cargo test -p codex-update-manager` — pass (133; incl. pure changelog/semver tests + git-based detection tests using a local temp origin)
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `bash tests/scripts_smoke.sh` — pass
- Existing config/state/builder/app tests untouched (new fields default).

## Versioning
[...]

PR #358, opened 2026-05-29
feat(setup): GUI checklist for install-time Linux feature selection

avifenesh:
## Summary
- Add an optional GUI checklist (zenity/kdialog) to the `make setup-native` Linux feature picker, pre-checked with the currently-enabled features.
- Falls back to the existing terminal numbered picker when no dialog tool or display is available (headless/SSH/CI) or when `CODEX_BOOTSTRAP_NO_GUI=1` is set.
- Reuses the existing Python discovery/validation/write path unchanged; the GUI layer only chooses GUI-vs-terminal and translates the checklist result into the same enable/disable selectors.

## Why
The install-time feature picker was terminal-only (a numbered list). On a graphical desktop a checklist is friendlier, while keeping the terminal path intact for headless installs. The change is additive and gated, so default non-graphical behavior is unchanged.

## Source-of-truth files
- `scripts/bootstrap-wizard.sh` — `gui_feature_picker_available()` gate (interactive TTY + display + dialog tool + `python3`, minus `CODEX_BOOTSTRAP_NO_GUI`), `prompt_for_feature_changes_gui()`, a backward-compatible `tsv` discovery mode added as a 4th positional arg to the existing Python helper (all prior 3-arg callers behave identically), and a zenity 4.x-safe checklist (`--print-column=2 --separator`, no removed `--separate-output`).
- `tests/scripts_smoke.sh` — pin the interactive PTY-driven wizard smoke tests to the terminal picker (see below).

## Test change rationale
The GUI path engages when an interactive TTY + display + zenity/kdialog are all present. Several smoke tests drive the terminal numbered/blank-id selection over a PTY, so on a graphical dev box they were being diverted into the GUI checklist (which expects dialog input, not piped stdin). Those PTY tests are pinned to the terminal picker with `CODEX_BOOTSTRAP_NO_GUI=1` so they stay deterministic everywhere. Headless CI is unaffected (no display ⇒ terminal path already).

## Validation
- `bash -n scripts/bootstrap-wizard.sh` — OK
- `bash tests/scripts_smoke.sh` — pass, both headless (`DISPLAY` unset) and with `DISPLAY` + zenity present
- Backward-compat: the 3-arg `run_feature_config_python` output is unchanged vs `main`; the 4th `tsv` arg emits clean `id\ttitle\tflag`.

## Notes
[...]

PR #360, opened 2026-05-29
feat(updater): in-app wrapper update button

avifenesh:
## Summary
- Add a separate in-app **Update** button for wrapper updates, shipped as the opt-in `codex-wrapper-updater` Linux feature (distinct from the upstream DMG Sparkle button, which is unchanged).
- The button is invisible unless a wrapper update is pending, shows the changelog as a tooltip, applies the update after the app exits, and relaunches into the freshly built version; it hides again once the install is aligned with upstream.
- Universal git-only detection that works for both packaged (.deb/.rpm/pacman) and user-local (`install.sh`) installs.

## Why
The wrapper-update axis (detection + changelog) only fired a notification / CLI output — there was no way to "open Codex, see there's an update, click, continue." This adds that surface as an opt-in feature, mirroring the DMG button's invisible-unless-pending behavior.

## Stacked on
This PR builds on two others in the same stack and is **diffed against them** until they merge:
- #357 (auto-install toggle) — provides the "Updates" settings section this PR adds a toggle row to.
- #359 (wrapper-update axis + changelog) — provides the detection/state this button reads.

Please merge #357 and #359 first; this PR's diff will then reduce to just the button + apply path. I'll rebase after they land.

## Design / safety
- **Opt-in feature**, disabled by default (`linux-features/features.example.json` stays empty; no `defaultEnabled`). Enable via the "Check for Codex Desktop Linux updates" toggle (`codex-linux-wrapper-updates-enabled`).
- Detection is git-only (shallow fetch, no GitHub API / `gh` / `curl`); offline simply shows no button.
- **Packaged** apply rebuilds + installs via `pkexec`, and degrades to a desktop notification when the build toolchain (cargo / DMG extractor) is missing — never fails mid-rebuild. **User-local** apply reuses `~/.local/bin/codex-desktop-update` (no privilege escalation).
- The button uses the custom-DOM pattern with a distinct handler id, marker, and screen position so it never collides with the DMG button.

## Source-of-truth files
- `linux-features/codex-wrapper-updater/{feature.json,README.md,stage.sh,patch.js}` *(new opt-in feature)*.
- `updater/src/wrapper_apply.rs` *(new)* — apply per install type.
- `updater/src/{wrapper.rs,app.rs,cli.rs,builder.rs}` — universal detection rework, `ApplyWrapperUpdate`, `build_update_from` seam.
- `launcher/start.sh.template` — consume the `wrapper-update-pending` marker, apply, and exec-relaunch into the new build.
- `scripts/patches/{shared.js,keybinds-settings.js}` — wrapper-enable toggle row.

## Validation
- `cargo test -p codex-update-manager` — pass (139)
- `bash -n launcher/start.sh.template`; `node -e "require('./linux-features/codex-wrapper-updater/patch.js')"`; `bash tests/scripts_smoke.sh` — pass
- Live: clicked the button in an isolated desktop; the app exited, the launcher rebuilt + relaunched into the new build, and the button hid once aligned.

## Versioning
Backward-compatible feature addition to the updater crate; left the version bump out pending merge order of the stack (see #357/#359).

## Notes
No linked issue. The auto-reopen watcher and hide-when-aligned behavior are covered; a follow-up PR adds a feature picker on reinstall and a dev-mode/SHA chip on top of this.


avifenesh:
This is part of a focused stack of Linux update-UX PRs, split per the "one logical change per PR" guidance. Merge order (independents first):

- **#357** auto-install-on-exit toggle — independent (off `main`)
- **#358** GUI install-time feature picker — independent (off `main`)
- **#359** wrapper-update axis + changelog — independent (off `main`)
- **#360** in-app wrapper Update button — needs #357 + #359 *(this PR)*
- **#361** feature picker on reinstall — needs #358 + #360
- **#362** dev-mode detection + SHA/dev-mode chips — needs #360

Each PR is diffed against its dependencies until they land; I'll rebase the dependents as the bases merge. The three independents (#357/#358/#359) can be reviewed and merged in any order. Happy to reshape the stack or squash if you'd prefer fewer PRs.

ilysenko (commented):
Looks good as an opt-in Linux feature concept, but I would not merge this stack yet.

Blockers:
- The branch conflicts with `main` and is stacked on unresolved wrapper-update work. The foundation issues from #359 still matter here: commit inequality is treated as an update without an ancestry check, and stale wrapper candidate/changelog state is not cleared when a later check finds no update.
- The PR also carries the auto-install-on-exit settings overlay from #357, which has its own unresolved behavior mismatch. Please do not stack this feature on blocked changes.
- `launcher/start.sh.template` calls `apply_pending_wrapper_update "$@"` before exporting `CODEX_LINUX_APP_DIR` and `CODEX_LINUX_LAUNCHER_CMD`. `wrapper_apply.rs` says those launcher env vars are used to choose the correct install type and relaunch path, but the manual-relaunch marker path will run without them. That can mis-detect user-local vs packaged installs or fail the user-local fallback.
- The marker paths are hardcoded under `codex-desktop` instead of using the active app id. Side-by-side identities can write/read the wrong marker or miss their own pending update.

Other risks:
- The packaged path rebuilds and installs a native package from a GitHub-fetched wrapper checkout. That is a large runtime/update surface for an in-app button, so it needs much stronger tests around failure, retry, install type detection, and marker cleanup.
- The webview runtime injects a floating DOM button by heuristically finding headers. Since this is an opt-in Linux feature that may be acceptable, but it should be documented as best-effort and should fail closed without visual overlap.
- `missing_build_dependency()` only checks a small subset of tools. A packaged rebuild can still fail later for missing package tooling or build dependencies; the UI should make that failure visible and recoverable.

Suggested path:
- First land a fixed, rebased wrapper-detection foundation.
- Then keep this PR focused on the opt-in Linux feature and launcher handoff only.
- Export the launcher env needed by `apply-wrapper-update` before `apply_pending_wrapper_update`, and make marker paths app-id aware.
- Add tests for the marker/app-id path and install-type detection.

avifenesh:
> Looks good as an opt-in Linux feature concept, but I would not merge this stack yet.
> 
> Blockers:
> 
> * The branch conflicts with `main` and is stacked on unresolved wrapper-update work. The foundation issues from #359 still matter here: commit inequality is treated as an update without an ancestry check, and stale wrapper candidate/changelog state is not cleared when a later check finds no update.
> * The PR also carries the auto-install-on-exit settings overlay from #357, which has its own unresolved behavior mismatch. Please do not stack this feature on blocked changes.
> * `launcher/start.sh.template` calls `apply_pending_wrapper_update "$@"` before exporting `CODEX_LINUX_APP_DIR` and `CODEX_LINUX_LAUNCHER_CMD`. `wrapper_apply.rs` says those launcher env vars are used to choose the correct install type and relaunch path, but the manual-relaunch marker path will run without them. That can mis-detect user-local vs packaged installs or fail the user-local fallback.
> * The marker paths are hardcoded under `codex-desktop` instead of using the active app id. Side-by-side identities can write/read the wrong marker or miss their own pending update.
> 
> Other risks:
> 
> * The packaged path rebuilds and installs a native package from a GitHub-fetched wrapper checkout. That is a large runtime/update surface for an in-app button, so it needs much stronger tests around failure, retry, install type detection, and marker cleanup.
> * The webview runtime injects a floating DOM button by heuristically finding headers. Since this is an opt-in Linux feature that may be acceptable, but it should be documented as best-effort and should fail closed without visual overlap.
> * `missing_build_dependency()` only checks a small subset of tools. A packaged rebuild can still fail later for missing package tooling or build dependencies; the UI should make that failure visible and recoverable.
> 
> Suggested path:
> 
> * First land a fixed, rebased wrapper-detection foundation.
> * Then keep this PR focused on the opt-in Linux feature and launcher handoff only.
> * Export the launcher env needed by `apply-wrapper-update` before `apply_pending_wrapper_update`, and make marker paths app-id aware.
> * Add tests for the marker/app-id path and install-type detection.

@ilysenko if its ok, it will be much easier if we first land the non stacked that you agree on, then the diff will be more focused. my intent, and i might missed something is to install the default of the user (.deb/..../git), mine is git because im working on the repo and breaking my app daily.
Are we fine with first merging the ones that has no dispute?

avifenesh:
@ilysenko i see that you did, great, will iterate one by one.

ilysenko:
@avifenesh I think I’ll wait for #357 and #359 first. Don’t worry too much about rebasing every dependent PR right now; if conflicts remain after the fixes, I can help resolve/rebase them on my side before merging.

ilysenko:
My current thinking is that this should live almost entirely as a Linux feature, disabled by default.

The reason is maintenance: this is useful for some workflows, but it is a fairly large update/apply surface. If too much of it becomes part of the core launcher/updater path, we will likely create long-term maintenance headaches for users who never enabled this feature.

Suggested target shape:

- Rebase on current `main` first. #359 is merged now, and this branch still carries old wrapper-state/app/changelog files and conflicts with current main.
- Keep `codex-wrapper-updater` as a self-contained optional Linux feature, disabled by default.
- Core should stay minimal and generic. At most, add small generic hook points that Linux features can use.
- Core should not hardcode `codex-wrapper-updater` marker paths, apply flow, or UI behavior.
- The feature should own the webview button, marker handling, docs, staged hooks/scripts, and best-effort UI behavior under `linux-features/codex-wrapper-updater/`.
- If a launcher hook is needed, make it generic for Linux features, app-id aware, and run only after the launcher exports the env needed by the updater.
- Marker/state paths must use the active app id / XDG paths, not hardcoded `codex-desktop`.
- The apply path must fail closed: if deps are missing, git/network fails, rebuild fails, install fails, or relaunch cannot be resolved, normal app launch must continue and the state must not be marked as installed.
- Do not clear the wrapper candidate or pending marker unless the update was actually installed.
- Use the bounded/non-interactive git behavior from the merged wrapper detection code.
- Add tests for disabled feature no-op, app-id scoped markers, launcher env order, install type detection, missing dependency behavior, failed apply retry/marker cleanup, and packaged vs user-local paths.
- The webview patch should be optional, idempotent, and avoid duplicate or overlapping buttons.

avifenesh:
@ilysenko I think that making the button an optional feature will kill the cause.
You have it in Codex, nothing that I invented, but instead of the user manually checking or having a transparent cron, he gets a small notifier. I can make it opt-out if someone doesn't want it, but making opt-in a feature that's supposed to help users stop marking things opt in, is a little overkill of the feature.
For the tests, sure, we'll make it more extensive.

But as always, your repo, your rules. lmk.

avifenesh:
@ilysenko I'm still coding :)
Hands off?

avifenesh:
Pushed a follow-up that is tests-only: it only expands coverage for the opt-in wrapper updater feature hooks and marker handling. No behavioral/runtime changes in this push.

ilysenko:
@avifenesh I had some time, so I pushed my current take on isolating this into a Linux feature. If you prefer to remove or rewrite my commit while you keep iterating, no problem.

My main concern is the default path. You and I are developers, so wrapper self-updates are useful for us, but most users installing this repo just want Codex to launch and update the upstream app. They will never think about wrapper branches, local rebuilds, package rebuilds, or git-based update flows.

If this is enabled by default, the updater surface becomes much larger: git/network checks, markers, rebuilds, package installs, relaunch flow, and failure recovery. Even if the check/apply flow is fail-closed, a bad wrapper merge on `main` can still become an update candidate; if it builds and installs successfully but breaks at runtime, users who never asked for this channel can end up with a broken Codex install and we get a wave of issues.

That is exactly why we have `linux-features/`, and now also local/private Linux features: advanced users can enable or build whatever workflow they want without putting that risk on everyone else.

So my preference is: ship it as an optional Linux feature, disabled by default. It can be easy to enable, but it should not affect the default install/update path.

ilysenko:
@avifenesh One more maintainer-cost point: if this ever becomes part of the default path, it has to work like clockwork. I watch this repo constantly, review PRs continuously, and often spend time running Codex reviews on high effort - not just occasionally when I feel like it.

I think part of the difference here is perspective: you are looking at this as a useful playground for advanced update workflows, while I have to look at the default build as something regular users can simply download, launch, and trust.

We already found several issues very early in this feature and spent a lot of time tightening the design. That is normal for a complex updater/apply flow, but it also tells me the future maintenance cost will be real. If you are busy, away, or on vacation, that support burden still lands on me and on the repo.

That is another reason I want this to stay optional and disabled by default unless/until it proves extremely stable over time.

P.S. A possible path forward: let's keep it optional first and test it there, both you and me. If bugs shake out and it proves stable, we can revisit making it default later. If/when we do that, I would not want it to track arbitrary `main` commits directly. We should tie updates to an explicit wrapper release/version channel, so only intentionally published versions become update candidates.

avifenesh:
@ilysenko I absolutely understand! Was my take only. 
I'll just add that Linux users who use Codex are most certainly devs, but I genuinely understand your point. 
I aligned my changes with your decision, as well as the stacked PRs.

To maintainer note:
I'm kind of “working” in OSS.
At work I have to be “on call” since it’s work. 
At my private repos and as a belief, although I don't always manage to follow it, you are working for free. If something breaks, it's ok; you don't owe anything. If somebody is not happy he can find another solution or contribute a fix. The expectation from owners and maintainers to work for free and be on the clock to make everything work and if something breaks, not to sleep until it's fixed, kills open source.

ilysenko (approved):
Reviewed current head 7f9eb85510b16e33002c1533dfa9ca57002b4bd4. The wrapper updater is isolated as an opt-in Linux feature, the runtime setting now persists through the Linux global-state path, and the app-id-aware marker/retry flow is covered by tests. Local focused tests and GitHub Actions are green.

PR #361, opened 2026-05-29
feat(updater): feature picker on wrapper-update reinstall

avifenesh:
## Summary
- The in-app wrapper Update button now opens a feature picker (zenity/kdialog checklist, pre-checked with the currently-enabled features) on each update, so you can re-choose which optional Linux features to enable before the rebuild.
- A "(Don't ask again on future updates)" row, or the "Ask which features to enable on update" toggle in Keybinds → Updates (`codex-linux-feature-picker-on-update`), suppresses the prompt.
- The selection is saved to `~/.config/<appId>/linux-features.json` and the rebuild honors it via `CODEX_LINUX_FEATURES_CONFIG`.

## Why
The wrapper Update button rebuilt with whatever feature set was already enabled. This lets each update revisit that choice, with a clear opt-out, so the rebuild doesn't silently lock in a stale selection.

## Stacked on
Diffed against two PRs in the same stack until they merge:
- #358 (GUI feature picker) — shares the zenity/kdialog approach and the same `linux-features.js` catalog access.
- #360 (in-app wrapper update button) — provides the button + apply path this hooks into.

Please merge #358 and #360 first; I'll rebase after they land.

## Design / safety
- The picker runs at **click time** (a new `codex-update-manager pick-features` subcommand), while the display is still alive — the detached apply runs headless after the app exits, so the dialog cannot run there.
- The installed app bundle ships no feature catalog, so the picker reads it from a wrapper source checkout (builder bundle, then a warm cache). A stale bundle whose `linux-features.js` predates `--features-json` is transparently skipped in favor of a current checkout.
- Every failure mode (no display, no dialog tool, no catalog, cancel) is a graceful skip that leaves the current feature set unchanged — the picker never blocks or fails the update.
- The saved config lives outside any wrapper-src checkout so a fresh clone can't clobber it; the contrib `codex-desktop-update` helper's `CODEX_LINUX_FEATURES_CONFIG` export is made conditional so the picked path wins.

## Source-of-truth files
- `updater/src/feature_picker.rs` *(new)* — `pick-features` subcommand (catalog read, dialog, write).
- `updater/src/{config.rs,cli.rs,app.rs,main.rs,wrapper_apply.rs}` — `feature_config_path()` + picker setting, `PickFeatures` wiring, and threading `CODEX_LINUX_FEATURES_CONFIG` into both apply paths.
- `linux-features/codex-wrapper-updater/patch.js` — invoke the picker on click before marker + exit.
- `scripts/patches/{shared.js,keybinds-settings.js}` — picker toggle.
- `contrib/.../codex-desktop-update` — conditional feature-config export.

## Validation
- `cargo test -p codex-update-manager` — pass (144; incl. write-config, path-outside-wrapper-src, fake-zenity selection/cancel, no-display skip, don't-ask-again-writes-key)
- `node -e "require('./linux-features/codex-wrapper-updater/patch.js')"`; `bash tests/scripts_smoke.sh` — pass
- Live: clicked Update → checklist appeared pre-filled → picked → the rebuild used the selection; the "(Don't ask again)" row wrote the setting and suppressed the next prompt.

## Versioning
Backward-compatible feature addition to the updater crate (new `pick-features` subcommand + `#[serde(default)]` settings key); version bump deferred to the stack's merge order.

## Notes
No linked issue. The repo already serialises env-mutating updater unit tests via a shared lock; the new `feature_picker` tests follow that same pattern.

### DIFF
--- a/updater/src/feature_picker.rs
+++ b/updater/src/feature_picker.rs
@@ -253,11 +253,12 @@ fn read_catalog(config: &RuntimeConfig, source: &Path) -> Result<Vec<CatalogEntr
     Ok(entries)
 }
 
-/// Reads the currently-enabled feature ids. Prefers the saved feature config
-/// (the picker's own prior answer), else `linux-features.js --enabled` (which
-/// reads the source `features.json`). Errors degrade to an empty set.
+/// Reads the currently-enabled feature ids. Prefers the saved picker config,
+/// then the installed builder bundle's preserved feature config, then
+/// `linux-features.js --enabled` from the selected source. Errors degrade to an
+/// empty set.
 fn read_enabled(config: &RuntimeConfig, source: &Path) -> std::collections::HashSet<String> {
-    if let Some(path) = config::feature_config_path() {
+    if let Some(path) = config::effective_feature_config_path(config) {
         if let Ok(content) = std::fs::read_to_string(&path) {
             if let Ok(value) = serde_json::from_str::<serde_json::Value>(&content) {
                 if let Some(array) = value.get("enabled").and_then(|v| v.as_array()) {
@@ -583,6 +584,28 @@ if (arg === "--features-json") {
             sources.is_empty(),
             "recorded candidate failures must not use the installed catalog"
         );
+    }
+
+    #[test]
+    fn enabled_reads_builder_feature_config_when_saved_picker_config_is_absent() {
+        let _g = env_lock();
+        let root = tempdir().unwrap();
+        let settings = tempdir().unwrap();
+        write_fake_catalog_script(root.path());
+        let builder_config = root.path().join("linux-features/features.json");
+        std::fs::create_dir_all(builder_config.parent().unwrap()).unwrap();
+        std::fs::write(&builder_config, r#"{"enabled":["alpha"]}"#).unwrap();
+
+        let settings_file = settings.path().join("settings.json");
+        std::env::set_var("CODEX_LINUX_SETTINGS_FILE", &settings_file);
+        let config = base_config(root.path());
+
+        assert_eq!(
+            read_enabled(&config, root.path()),
+            std::collections::HashSet::from(["alpha".to_string()])
+        );
+
+        std::env::remove_var("CODEX_LINUX_SETTINGS_FILE");
     }
 
     #[test]
--- a/updater/src/feature_picker.rs
+++ b/updater/src/feature_picker.rs
@@ -1,0 +1,748 @@
+//! Interactive feature picker for the in-app wrapper Update button.
+//!
+//! When the user clicks the wrapper "Update" button, the renderer shells out to
+//! `codex-update-manager pick-features` *while the display is still alive* (the
+//! detached `apply-wrapper-update` runs after the app exits → headless, so the
+//! dialog must run here, at click time). This subcommand:
+//!
+//! 1. Locates a wrapper source checkout that ships the feature catalog
+//!    (`scripts/lib/linux-features.js` + the `linux-features/<id>/feature.json`
+//!    set). When an update candidate is recorded, it prepares that candidate
+//!    source first so the picker reflects the build that will actually run. A
+//!    manual `pick-features` invocation without a candidate falls back to the
+//!    installed builder bundle.
+//! 2. Reads the catalog (`--features-json`) and the currently-enabled set
+//!    (the saved `linux-features.json`, else `--enabled`).
+//! 3. Shows a zenity/kdialog checklist pre-checked with the enabled set, plus a
+//!    sentinel "(Don't ask again …)" row.
+//! 4. Writes the chosen `{"enabled":[…]}` to [`config::feature_config_path`] so
+//!    the rebuild (which points `CODEX_LINUX_FEATURES_CONFIG` at that path) uses
+//!    the selection. If the sentinel row was checked, persists
+//!    `codex-linux-feature-picker-on-update=false` so future updates skip the
+//!    prompt.
+//!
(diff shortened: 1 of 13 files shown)


ilysenko (commented):
Looks good as a UX idea, but I would not merge this yet.

Blockers:
- The branch conflicts with `main` and is stacked on #360/#359, which still have unresolved wrapper-update correctness issues.
- The picker reads the feature catalog primarily from the currently installed builder bundle. For a wrapper update, that is the old wrapper catalog, not necessarily the candidate wrapper catalog. On the first update click, `wrapper-src` usually has not been fetched yet, so newly-added features in the update cannot be selected.
- The picker filters the saved selection to ids present in the repo catalog and writes a fresh `{"enabled":[...]}` file. That can silently drop user-local/private feature ids that are already enabled but are not present in the bundled catalog. One of the goals of Linux features is to allow local features outside the repo, so the picker must preserve unknown/local enabled ids or explicitly display them.

Other risks:
- A cancelled dialog correctly leaves the file unchanged, but an accepted dialog rewrites the whole feature config. That makes the "drop unknown ids" behavior especially risky.
- The click-time picker uses GUI dialogs from the Electron session, which is the right place to do it, but it needs clearer fallback behavior in the UI when no display/dialog tool/catalog exists.
- `which()` only checks `is_file`, not executable permissions, for `zenity`, `kdialog`, and `node` fallback.

Suggested path:
- Fix and land the wrapper-update foundation first.
- Fetch or otherwise inspect the candidate wrapper source before showing the picker, so the catalog matches the build that will actually be installed.
- Preserve enabled feature ids that are unknown to the candidate catalog, especially local/private features.
- Add tests for preserving unknown/local feature ids and for using the candidate wrapper catalog rather than the installed one.

updater/src/feature_picker.rs
```diff
@@ -0,0 +1,608 @@
...
+    match show_picker(&tool, &catalog, &enabled)? {
+        None => {
+            info!("feature picker cancelled; feature set unchanged");
+            Ok(PickOutcome::Skipped("cancelled"))
+        }
+        Some(selection) => {
+            let dont_ask = selection.iter().any(|id| id == DONT_ASK_SENTINEL);
+            let catalog_ids: std::collections::HashSet<&str> =
+                catalog.iter().map(|entry| entry.id.as_str()).collect();
+            let mut picked: Vec<String> = selection
```
avifenesh, line 120: The picker rewrites the entire feature config with only IDs present in the *catalog*. Any previously-enabled feature IDs that are not in the catalog — notably **user-local / private features** discovered from `linux-features/local/<id>/` (see `linux-features/README.md` lines 29–32) — are silently dropped.

**Scenario:** A user has `my-custom-feature` enabled in `linux-features.json`. The picker pre-checks only catalog features (line 287–291 only iterates `catalog`), so `my-custom-feature` never appears in the dialog. On OK, `write_feature_config(&picked)` replaces the entire file with only catalog IDs → `my-custom-feature` is gone. This happens even if the user opens the picker and clicks OK without changing anything.

**Suggested fix:** Before writing, merge the picked catalog IDs with any previously-enabled IDs that are *not* in the catalog (i.e., preserve unknown/local IDs):

```rust
let existing = read_enabled(config, &source);
let mut merged: Vec<String> = picked;
merged.extend(
    existing.iter().filter(|id| !catalog_ids.contains(id.as_str())),
);
merged.sort();
merged.dedup();
write_feature_config(&merged)?;
```

The existing maintainer review flagged this same concern. Adding the source-line trace and a concrete fix approach.



---

updater/src/feature_picker.rs
```diff
@@ -0,0 +1,608 @@
...
+                    "TRUE"
+                } else {
+                    "FALSE"
+                });
+                cmd.arg(&entry.id);
+                cmd.arg(&entry.title);
+            }
+            // Sentinel "don't ask again" row, unchecked by default.
+            cmd.arg("FALSE").arg(DONT_ASK_SENTINEL).arg(DONT_ASK_LABEL);
+            cmd.output().context("Failed to run zenity")?
```
avifenesh: A zenity/kdialog launch failure (non-executable file, transient display disconnect, etc.) propagates as `Err` through `show_picker(..)?` at line 105, which contradicts the module's stated guarantee (lines 14–17):

> Every failure mode (no display, no dialog tool, no catalog, cancelled dialog) is a graceful skip

In practice the Node.js caller ignores the exit code, so the update proceeds, but the `pick-features` subcommand exits with a non-zero status and the guarantee is broken in the Rust layer.

**Suggested fix:** Wrap the dialog call to degrade to a skip on launch failure, matching the same graceful pattern used for `has_display` and `dialog_tool`:

```rust
match show_picker(&tool, &catalog, &enabled) {
    Ok(result) => result,
    Err(error) => {
        warn!(?error, "feature picker dialog failed; skipping");
        return Ok(PickOutcome::Skipped("dialog-error"));
    }
};
```

This keeps the "every failure mode is a graceful skip" contract.



---

avifenesh (commented):
Reviewed the feature-picker-on-update implementation. Two findings below — one correctness issue (silent data loss of non-catalog feature IDs) and one graceful-degradation gap (dialog launch failure not caught as a skip).

Both were partially noted in the existing maintainer review; the inline comments add source-line evidence and suggest concrete fixes.



---

ilysenko (approved):
Reviewed the rebased head ac51e66. The previous blockers are addressed: the branch is on current main, the picker uses the recorded candidate catalog, preserves unknown enabled feature ids, requires executable dialog helpers, and keeps the existing feature config as a fallback when the picker is skipped or cancelled. Local updater/patcher tests and GitHub Actions are green.

PR #362, opened 2026-05-29
fix(updater): dev-mode detection + SHA/dev-mode chips for wrapper updates

avifenesh:
## Summary
- Fix a downgrade bug in the in-app wrapper Update button: when the installed build is *ahead* of upstream (carrying unmerged local work), commit-inequality detection used to offer a phantom "update" that was actually a downgrade, and clicking it rebuilt the app down to the older upstream commit.
- Detection now distinguishes ahead-of-upstream (the installed commit is not reachable on the remote — a local/dev build) from genuinely-behind (reachable, older), and reports the former as "dev mode" with updates disabled.
- Add an always-on **SHA chip** showing the installed short commit, and a distinct color/glyph-coded action chip: green ↓ Update, amber ⚙ dev mode (non-clickable), spinner while updating.

## Why
Without this, a contributor running a build ahead of upstream gets the Update button rebuilding their build *down* to upstream and overwriting it — including the launcher. The SHA chip also makes it obvious which build is actually running.

## Stacked on
Diffed against #360 (in-app wrapper update button) until it merges — this fixes/extends that button's detection and toolbar. Please merge #360 first; I'll rebase after it lands.

## Design / safety
- Reachability is decided **git-only**: `git fetch --depth 1 <remote> <full-sha>` succeeds iff the commit is reachable on the remote (`not our ref` ⇒ genuinely absent). A control-probe of the known-reachable HEAD sha distinguishes "server blocks sha-fetch" from "commit absent".
- **Fail-open to no-downgrade**: offline, or reachability unprovable while the installed commit differs from HEAD ⇒ dev mode (suppress the update). The clobber can no longer be triggered even when the network/server misbehaves.
- **Apply-side guard**: `apply-wrapper-update` hard-refuses (silent `Ok(())`, since it runs detached post-exit) while `wrapper_dev_mode` is recorded — defense-in-depth even if detection mis-records or state is hand-edited. It reads the recorded verdict and does not re-probe (apply may be offline).
- New `state.json` fields are `#[serde(default)]`; old state parses unchanged. The SHA chip text is set via `textContent` (no injection surface); chip glyphs are static literals.

## Source-of-truth files
- `updater/src/wrapper.rs` — `WrapperDetectionState` enum, `remote_has_commit` / `remote_supports_sha_fetch`, `detect_wrapper_update` returns `(state, Option<WrapperUpdate>)`.
- `updater/src/{state.rs,app.rs}` — `installed_wrapper_commit` + `wrapper_dev_mode` fields; record/clear per detection state.
- `updater/src/wrapper_apply.rs` — dev-mode downgrade guard.
- `linux-features/codex-wrapper-updater/patch.js` — three chip states + always-on SHA chip.

## Validation
- `cargo test -p codex-update-manager` — pass (150; incl. dev-mode-when-not-reachable, update-when-older-reachable, upstream-older-aligned, offline-unknown, `remote_has_commit` true/false, apply-guard refusal)
- `node -e "require + parse patch.js runtime"`; `bash tests/scripts_smoke.sh` — pass
- Live: against the real upstream, an ahead build records `wrapper_dev_mode=true` with the candidate cleared; the toolbar shows the amber "dev mode" pill + SHA chip (no Update offered), and the green Update + SHA chip in the genuinely-behind case.

## Versioning
Backward-compatible updater change (fix + new `#[serde(default)]` state fields); version bump deferred to the stack's merge order.

## Notes
No linked issue. Builds on #360; the dev-mode verdict is sticky until a later online detection proves reachability, which is the conservative no-downgrade choice.


ilysenko (commented):
Looks good as a partial fix for the downgrade/dev-mode problem, but I would not merge this stack yet.

Blockers:
- The branch conflicts with `main` and still includes the unresolved #360 wrapper-update button/apply stack.
- The launcher still calls `apply_pending_wrapper_update "$@"` before exporting `CODEX_LINUX_APP_DIR` and `CODEX_LINUX_LAUNCHER_CMD`. The detached in-app path has those env vars, but the manual-relaunch marker path does not, so install-type detection and relaunch behavior can still be wrong.
- Marker paths are still hardcoded under `codex-desktop` instead of the active app id. Alternate identities can read/write/clear the wrong wrapper-update marker.
- When `enable_wrapper_updates` is false, `detect_and_record_wrapper_update()` returns without clearing an existing `candidate_wrapper_commit`. The webview feature reads `state.json` directly, so a previously detected candidate can still make the button visible after the user disables wrapper-update tracking.

Other risks:
- The new dev-mode guard is conservative: if the remote/server cannot prove reachability of the installed commit, it chooses `DevMode`. That is safer than downgrading, but it may disable wrapper updates for remotes that do not support fetching arbitrary SHA wants. The UI should make that state understandable.
- `UnknownOffline` intentionally preserves the last candidate. That may be okay for a previously valid candidate, but it should not preserve stale UI if tracking was disabled or the candidate was later proven absent.
- This still carries the large packaged self-rebuild/apply surface from #360, so it needs end-to-end tests around marker handling, app-id handling, install-type detection, and disabled tracking.

Suggested path:
- Rebase and split this fix from the broader in-app wrapper update button if possible.
- Keep the dev-mode detection/guard, but also clear candidate state when wrapper tracking is disabled.
- Move launcher exports before marker handling and make marker paths app-id aware.
- Add tests for disabled-tracking clearing and manual-relaunch marker application.

launcher/start.sh.template
```diff
@@ -2233,6 +2272,7 @@ prepend_managed_node_runtime_to_path
 source_feature_env_files
 register_url_scheme_handlers
 reconcile_runtime_state
+apply_pending_wrapper_update "$@"
```
avifenesh: **Ordering bug:** `apply_pending_wrapper_update` is invoked here, but `CODEX_LINUX_APP_DIR` and `CODEX_LINUX_LAUNCHER_CMD` are not exported until lines 2370–2371 (after `launch_electron`). When this function calls `run_update_manager apply-wrapper-update`, the Rust side calls `detect_install_type` which reads `CODEX_LINUX_APP_DIR` to decide between `Packaged` and `UserLocal`, and `user_local_app_dir` which reads it to resolve the app directory. Both will see an absent/unset variable.

For user-local installs, `user_local_app_dir()` returns `None`, causing `apply_user_local` to bail with `"could not resolve user-local app dir (CODEX_LINUX_APP_DIR)"`. For packaged installs, `detect_install_type` falls back to the `builder_bundle_root == /opt/codex-desktop/update-builder` heuristic, which may work but is less reliable than the authoritative launcher hint.

The `CODEX_UPDATE_MANAGER_PATH` export at line 2364 also hasn't happened yet, but `apply_pending_wrapper_update` resolves the path independently via `has_update_manager` → `resolve_update_manager_path`, so that part is fine.

Fix: move the `CODEX_LINUX_APP_DIR`, `CODEX_LINUX_LAUNCHER_CMD`, and (if feasible) `CODEX_UPDATE_MANAGER_PATH` exports above this line, before `apply_pending_wrapper_update` is called.



---

launcher/start.sh.template
```diff
@@ -1331,6 +1332,44 @@ run_update_manager() {
     "$CODEX_UPDATE_MANAGER_PATH" "$@"
 }
 
+# Consume the wrapper-update marker written by the in-app wrapper Update button
+# and apply the pending wrapper (repo) update before the app starts. On success,
+# re-exec the launcher so the user lands in the freshly installed build. On
+# failure the marker is left in place to retry next launch, and startup
+# continues with the current build. Skipped when the app is already running.
+apply_pending_wrapper_update() {
```
avifenesh: **Marker path not app-id aware:** The marker is hardcoded to `codex-desktop`. For side-by-side installs (P31), alternate app identities need their own marker under `$XDG_STATE_HOME/<app-id>/wrapper-update-pending`. The same hardcoded path appears in `wrapper_apply.rs:clear_pending_marker` and in `patch.js:codexLinuxWrapMarkerPath`. All three should use the active `CODEX_LINUX_APP_ID` (with `codex-desktop` as the default when unset) to prevent one install from consuming or clearing another's marker.



---

updater/src/app.rs
```diff
@@ -293,6 +308,138 @@ async fn run_check_now(
...
+/// `Ok(false)`) when wrapper tracking is disabled, offline, or already current.
+/// Detection is git-only and works for all install types; never mutates the
+/// user's working tree.
+fn detect_and_record_wrapper_update(
+    config: &RuntimeConfig,
+    state: &mut PersistedState,
+    paths: &RuntimePaths,
+) -> Result<bool> {
+    if !config.enable_wrapper_updates {
+        return Ok(false);
```
avifenesh, line 374: **Stale candidate not cleared when tracking is disabled:** When `enable_wrapper_updates` is `false`, this early return preserves any existing `candidate_wrapper_commit` (and `wrapper_changelog`) in state. The webview JS reads `state.json` directly via `codexLinuxWrapShouldShow`, which checks `candidate_wrapper_commit` presence but **not** whether wrapper tracking is enabled. So after the user disables the "Check for Codex Desktop Linux updates" toggle, the Update button persists in the UI from the last-known candidate.

Consider clearing `candidate_wrapper_commit`, `candidate_wrapper_date`, `wrapper_changelog`, and setting `wrapper_dev_mode = None` before returning here, so the UI button disappears when tracking is turned off.



---

avifenesh (commented):
Reviewed the wrapper dev-mode detection and SHA/dev-mode chips implementation. The core detection logic in `wrapper.rs` (reachability via sha-in-want, safe `DevMode` fallback) and the apply-side guard in `wrapper_apply.rs` are well-structured. I'm flagging three concrete issues below — two of which the repo owner already identified and I've confirmed against the source.

**1. Launcher ordering: `apply_pending_wrapper_update` runs before `CODEX_LINUX_APP_DIR` is exported** (HIGH)
The Rust `detect_install_type` and `user_local_app_dir` both read `CODEX_LINUX_APP_DIR`. At apply time, it's not yet in the environment. This causes the apply to misidentify the install type or fail outright for user-local installs.

**2. Wrapper-update marker path hardcodes `codex-desktop` instead of the active app ID** (MEDIUM)
Three locations (launcher, Rust apply, JS patch) all use a hardcoded `codex-desktop` subdirectory. Per P31, side-by-side installs require app-id-specific paths via `CODEX_LINUX_APP_ID`.

**3. Disabling wrapper-update tracking does not clear an existing candidate** (MEDIUM)
`detect_and_record_wrapper_update` returns `Ok(false)` at the top without clearing `candidate_wrapper_commit` when tracking is off. The JS `codexLinuxWrapShouldShow` checks candidate presence but not the tracking setting, so the Update button persists in the UI after the user disables tracking.



---

ilysenko (approved):
Reviewed current head 41135c74. No blockers found: previous conflicts and clippy issue are fixed, disabled tracking clears stale wrapper candidates, dev-mode/apply guard prevents downgrade, and CI plus local updater/feature tests passed.

Issue #329, opened 2026-05-27
Clarify path forward for Linux AppShots support


moxw:
PR #307 added Linux AppShots support. After review feedback, I moved the UI, settings, and hotkey pieces into disabled-by-default linux-features/appshots, removed the Linux default global hotkey, and tightened capture so AppShot screenshots fail closed instead of falling back to full-screen or CLI screenshot capture.

The PR was later closed without a follow-up reason. I asked on the closed PR, but there has not been a response yet.

Could you clarify the preferred path?

- Is Linux AppShots out of scope for this repo even as an opt-in feature?
- Is there another privacy or implementation concern still blocking it?
- Would a smaller PR be preferable, for example backend capture only, feature docs only, or no UI exposure?

Relevant PR: https://github.com/ilysenko/codex-desktop-linux/pull/307

ilysenko:
Thanks for opening this, and sorry for closing the PR without a clearer final note. I can see why that was frustrating after you had already moved the work into `linux-features/appshots`.

The short answer is that I do not want to take Linux AppShots support into this repo right now, even as a disabled-by-default feature. It is a useful experiment, but it is a separate product surface rather than core Linux compatibility, and it would add a maintenance burden that is too high for how narrow the feature is.

A few concerns remained for me:

- AppShots is not the same thing as the Linux Computer Use backend. It depends on screenshot/window/accessibility primitives, but exposing the upstream AppShots UI on Linux is a separate integration.
- The PR still touched several broad areas: Rust backend code, screenshot capture behavior, bare-modifier hotkey monitoring, ASAR patches, docs, and tests.
- Screenshot/privacy behavior has to stay very conservative. Even if AppShots itself avoids CLI fallback, adding or relying on non-portal screenshot paths in shared code is something I do not want to expand casually.
- There were still correctness risks around focused-window cropping, especially partially offscreen or negative window bounds.
- The AppShots ASAR patches need to be very robust against upstream minified-bundle drift; partial application would be hard to support.
- Global/bare-modifier hotkey support is desktop- and session-dependent on Linux and would likely create support issues.

So the preferred path for now is not to continue this PR as Linux AppShots enablement. If you want to salvage pieces, smaller PRs that improve existing Computer Use primitives without exposing the AppShots UI would be much easier to consider, for example a focused-window screenshot/cropping correctness fix with tests.

Thanks again for the work here. The closure was about scope and long-term maintenance, not about dismissing the effort.

[...]

PR #363, opened 2026-05-29
Add self-contained Linux AppShots feature

moxw:
## Summary
- Add AppShots as a disabled-by-default Linux feature under `linux-features/appshots/`
- Keep the AppShots patches, helper resource, docs, and tests self-contained in the feature directory
- Use existing Computer Use `windows` / `state` commands for metadata and AT-SPI text
- Add an explicit opt-in `Shift + Shift` helper for X11 bare-modifier hotkeys
- Prefer deterministic ImageMagick crops for AppShot previews and fail closed when capture/cropping is unavailable

## Validation
- `git diff --check`
- `make build-dev-app`
- Manual dev-app test on Linux/i3 with `Shift + Shift` AppShot capture and preview attachment

Addresses #329


ilysenko (approved):
Reviewed the current head. This is isolated under linux-features/appshots, disabled by default, and the focused local checks plus GitHub Actions pass.

PR #364, opened 2026-05-29
fix: keep run action modal inputs editable

LiukScot:
## Summary
- keep add-action modal input draft state inside the stored modal content
- add a required upstream webview asset patch descriptor for the local conversation thread bundle
- cover the regression so name, command, icon, and save payload use the editable draft

## Tests
- node scripts/patch-linux-window-ui.test.js

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -75,6 +75,7 @@ const {
 } = require("./lib/build-info.js");
 const {
   applyBrowserAnnotationScreenshotPatch,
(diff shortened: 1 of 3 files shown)


ilysenko:
Please make this patch `optional` instead of `required-upstream`.

This is a targeted UX fix for the local environment action modal, not a core Linux compatibility requirement. If upstream changes this bundle, we should skip the patch with a warning rather than block the whole Linux build.

One thing I would still like to keep strict inside the patch itself: avoid partial patching. Please verify that all critical replacements were applied, and if any required needle is missing, return the original source with a warning instead of leaving a half-patched bundle.

PR #365, opened 2026-05-29
[codex] fix Nix pin validation for latest DMG

colonelpanic8:
## Summary

- refreshes `flake.nix` to the current upstream `Codex.dmg` (`26.527.30818`)

- teaches `validate-nix-pins.sh` to fall back to the Electron version in `app.asar` package metadata when upstream no longer ships `Electron Framework.framework/Versions/A/Resources/Info.plist`

## Validation
- `bash -n scripts/ci/validate-nix-pins.sh scripts/ci/update-nix-hashes.sh`

- `nix shell nixpkgs#{curl,p7zip,python3,nodejs_24} --command bash -lc 'bash scripts/ci/validate-nix-pins.sh /tmp/Codex.dmg'`

- `scripts/ci/update-nix-hashes.sh` via a Nix shell with curl/p7zip/python3/nodejs_24, including all configured Nix package outputs

## Root cause

The latest upstream DMG layout renamed the embedded framework path, so CI failed before it could compare the actual app and native-module pins. The root package metadata still carries the Electron dependency and is already used as the installer fallback.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -66,10 +66,10 @@
 
         codexDmg = pkgs.fetchurl {
           url = "https://persistent.oaistatic.com/codex-app-prod/Codex.dmg";
-          hash = "sha256-yanuEZhqD4gBWLgeRZtktoRkI5b0nq9/oOAY9KjDe0I=";
+          hash = "sha256-L5fZniPJ9N1mwOAr2tpebGa/Z3cs2J+IaoG1udfrSHw=";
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Looks good. The change is narrowly scoped to the Nix pin refresh and the Electron-version fallback in validate-nix-pins, and CI is green.

PR #368, opened 2026-05-29
[codex] narrow reusable Nix build inputs

colonelpanic8:
## Summary

- stack on #367 so the read-only Cachix setup can consume these more stable derivations

- build `codexComputerUseBinaries` from a minimal generated Cargo workspace containing only `Cargo.lock` and `computer-use-linux/`

- build Electron native modules against a tiny helper source containing only `scripts/lib/native-modules.sh` instead of the full repo source
## Why

The regular app payload still depends on the upstream DMG, but the Computer Use Rust binaries and Electron native-module rebuilds do not need to be invalidated by launcher, docs, patcher, feature, or DMG-only changes. Narrowing their source inputs should make their Cachix store paths survive more ordinary churn.

## Validation

- `nix build .#codex-desktop --no-link --print-build-logs --max-jobs 1`

- `nix build .#codex-desktop-computer-use-ui-remote-mobile-control --no-link --dry-run`

- `git diff --check`


ilysenko (approved):
Reviewed after updating the branch with the current `main`. No blockers found: the PR now only changes `flake.nix`, narrows the Nix source inputs for reusable derivations, and CI is green including the Nix package build outputs.

PR #370, opened 2026-05-30
Fix open-target startup crash from path alias shadowing

moxw:
## Summary
- Added stable `codexLinuxNodeFs()` / `codexLinuxNodePath()` accessors for injected open-target helpers instead of reusing upstream minified module aliases.
- Added a regression test that reproduces a `node:path` alias collision with helper-local variables.

## Why
The current upstream bundle can minify `node:path` to `o`. `codexLinuxDesktopTryExecAvailable()` also declared local `var o`, which hoisted over the path module alias and made startup throw `Cannot read properties of undefined (reading 'basename')` while scanning Linux desktop entries.

## Source of truth
- `linux-features/open-target-discovery/patch.js`
- `linux-features/open-target-discovery/test.js`

## Validation
- `./install.sh --reuse-dmg`
- `CODEX_MULTI_LAUNCH=1 ./codex-app/start.sh` reached normal renderer/app-server startup instead of the bootstrap `basename` crash.
- `make pacman` built `dist/codex-desktop-2026.05.30.080638-1-x86_64.pkg.tar.zst`.

## Risks / follow-up
- No generated artifacts are included in this PR.
- Scope is limited to the opt-in `open-target-discovery` Linux feature.

### DIFF
--- a/linux-features/open-target-discovery/patch.js
+++ b/linux-features/open-target-discovery/patch.js
@@ -130,6 +130,8 @@ function insertOpenTargetHelpers(currentSource, insertionIndex, { fsVar, pathVar
   }
 
   const helpers =
+    `function codexLinuxNodeFs(){return require(\`node:fs\`)}` +
+    `function codexLinuxNodePath(){return require(\`node:path\`)}` +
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed current head 7662983a0c168447928101da28b69f95d0f7664b. The change is scoped to the opt-in open-target-discovery feature, avoids minified fs/path alias shadowing via stable accessors, and adds regression coverage for the alias collision. GitHub checks are green; local tests passed on the current main merge result.

PR #372, opened 2026-05-30
fix(linux): handle current integration drift

joshyorko:
## Summary
- handle current Computer Use availability helper when `isHostLocal` is absent
- let Chrome native-host sync resolve Linux runtime paths from launcher env and `resources/node-runtime/bin/node`
- keep Linux Chrome extension settings opener using Chrome, Brave, or Chromium from `PATH`

## Validation
- `node --test --test-name-pattern "Chrome" scripts/patch-linux-window-ui.test.js`


scripts/patches/core/all-linux/extracted-app/browser-integrations/patch.js
```diff
@@ -0,0 +1,26 @@
...
+    apply: patchLinuxChromeNativeHostRuntimeAssets,
+    status: (result, warnings) => {
+      if (result?.matched === 0) {
+        return {
+          status: "failed-required",
+          reason: result?.reason ?? warnings[0] ?? "Chrome native host runtime resolver not found",
+        };
+      }
+
+      return {
```
ilysenko, line 27: This status path treats `changed === 0` as `already-applied` even when `applyLinuxChromeNativeHostRuntimePatch()` emitted a warning. Because this descriptor is `required-upstream`, a matched bundle whose resolver shape drifted will warn and return unchanged from `chrome-plugin.js`, but the report records `linux-chrome-native-host-runtime` as `already-applied` with only a reason. That lets upstream drift pass the required patch gate while the Chrome native-host runtime fallback was not installed. Please fail this case, for example when `warnings.length > 0 && result.changed === 0`, unless the helper marker proves it was already patched.

ilysenko (changes requested):
Thanks for the drift fixes. I found one required-upstream reporting issue that should be fixed before this leaves draft: a failed runtime-shape patch can currently be reported as already applied.

ilysenko (approved):
Reviewed current head 67f7d5f. The required-upstream drift reporting issue is fixed and covered by a regression test; local patcher suites pass.

PR #373, opened 2026-05-31
fix(patches): re-anchor 26.527.x Linux webview drift

avifenesh:
## Summary
- Re-anchor shared Linux webview patches for the Codex Desktop 26.527.x bundle shape.
- Strip the new dynamic `remote_plugin` app-server feature extra while keeping already-sanitized builders quiet.
- Re-anchor the opt-in remote-mobile feature for the current app-server/browser-client/settings chunks.

## Why
The 26.527.x upstream bundle moved several minified hooks. Without this, optional Linux remote-mobile behavior either warns on valid current shapes or misses the current app-server/browser-client paths.

## Validation
- `node --test scripts/patch-linux-window-ui.test.js linux-features/*/test.js` passed, 399 tests on the final stack.
- `git diff --check` passed.
- Rebuilt the app with all real optional Linux features enabled.


ilysenko (approved):
Reviewed current head b2bc077. Local focused patch suites and GitHub checks pass; no blockers found.

PR #374, opened 2026-05-31
fix(read-aloud): route settings to dedicated page

avifenesh:
## Summary
- Route the Read Aloud settings section to a dedicated generated settings page instead of aliasing General Settings.
- Patch the current inner `general-settings` chunk to export `ReadAloudSettings` before the wrapper imports it.
- Add regression coverage for the current exported inner-chunk shape so Settings does not crash on missing named exports.

## Why
The Read Aloud nav/route existed, but the generated wrapper previously exported `GeneralSettings` under the `ReadAloudSettings` name. A later correction made the wrapper import `ReadAloudSettings`, but the current inner chunk did not export it, causing the Settings route to show the app error screen.

## Validation
- `node --test linux-features/read-aloud/test.js` passed, 32 tests.
- `node --test scripts/patch-linux-window-ui.test.js linux-features/*/test.js` passed, 399 tests on the final stack.
- `git diff --check` passed.
- Rebuilt and synced the local user app; verified the restarted app opens Settings without the error screen.

### DIFF
--- a/linux-features/read-aloud/patch.js
+++ b/linux-features/read-aloud/patch.js
@@ -211,6 +211,24 @@ function generalSettingsReadAloudBlockSource() {
   return `${generalSettingsReadAloudRowSource()}${generalSettingsReadAloudPageSource()}`;
 }
 
+function currentGeneralSettingsReadAloudBlockSource() {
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed current head 1957be1 after merging #373. Scope is limited to linux-features/read-aloud, local read-aloud tests pass, and GitHub checks are green.

PR #376, opened 2026-06-01
Fix Upstream Patch Drifts

Leay15:
**Linux Remote Control Client Compatibility and Testing:**

* Improved detection in `applyLinuxRemoteControlClientAccountCompatibilityPatch` to recognize and skip patching if the source is already compatible with the latest upstream multi-account enrollment logic. This prevents unnecessary or redundant modifications.
* Updated the enrollment patch regex to be more robust against upstream code changes, allowing for flexible matching of code structure and variable spacing.
* Added a new test bundle and test case to verify that the patcher leaves already-upstream-compatible code untouched, ensuring correct no-op behavior. 

**Linux App Updater and Footer Patching:**

* Enhanced the updater bridge patch logic to handle cases where the message dispatcher is farther from the handler wiring, increasing patch robustness for different code layouts.
* Updated the persistent rate limit footer patch to detect and adapt to the current composer permissions footer shape, including new symbol extraction and insertion logic. This ensures the footer patch remains compatible with recent upstream UI changes.

### DIFF
--- a/linux-features/remote-mobile-control/patch.js
+++ b/linux-features/remote-mobile-control/patch.js
@@ -205,6 +205,13 @@ function applyLinuxRemoteControlClientAccountCompatibilityPatch(source) {
   ) {
     return source;
   }
+  if (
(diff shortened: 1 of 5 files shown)


ilysenko (approved):
Reviewed current head 436ff30910fbbe308c95d4610a922d3344420f43. The changes are focused upstream drift fixes, GitHub checks are green, and focused local patch tests pass.

PR #378, opened 2026-06-01
Fix AppShots bare modifier repeat detection

moxw:
## Summary
- require AppShots bare-modifier hotkeys to use two distinct physical keycodes
- keep Shift + Shift and Alt + Alt available while preventing same-side double taps from firing
- document the X11 helper behavior and cover it in the AppShots feature test

## User-visible behavior
- pressing Left Shift twice or the same Alt key twice quickly no longer opens AppShots
- pressing opposite-side modifiers within the timeout still triggers the configured bare-modifier shortcut

## Source of truth
- linux-features/appshots/bin/bare-modifier-monitor
- linux-features/appshots/test.js
- linux-features/appshots/README.md
- CHANGELOG.md

## Validation
- node --test linux-features/appshots/test.js
- git diff --check
- env npm_config_cache=/home/mohit/.cache/npm make build-dev-app
- launched /home/mohit/Github/codex-desktop-linux/bin/codex-cua-lab for manual testing

## Notes
- The helper remains X11-oriented because it still depends on xinput/xmodmap and fails closed elsewhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- The opt-in Linux AppShots bare-modifier shortcuts now require left and right
+  modifier keycodes, preventing a fast double-tap on one physical Alt or Shift
+  key from opening AppShots.
 - The wrapper updater no longer offers a "downgrade as update" when the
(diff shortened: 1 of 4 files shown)


ilysenko (approved):
Reviewed the AppShots feature-only change. The behavior stays isolated to the opt-in linux-features/appshots path, with tests covering the updated bare-modifier helper expectations.

PR #383, opened 2026-06-02
fix(remote-mobile-control): skip unsupported SSH status reads

avifenesh:
## Summary

- add a Linux remote-mobile-control webview guard that skips `remoteControl/status/read` for `remote-ssh*` app-server hosts
- mark skipped SSH hosts as disabled for remote-control status while preserving local host status reads
- cover the guard in the feature unit test and ASAR patch/report fixture

## Why

On Linux with the experimental remote-mobile-control feature enabled, the renderer was asking SSH app-server hosts for `remoteControl/status/read`. Those remote SSH app-servers do not currently expose `remoteControl/*`, so the request returns an unknown-method error after a long wait and makes startup feel slow and unstable.

## Source of truth

- `linux-features/remote-mobile-control/patch.js`
- `linux-features/remote-mobile-control/test.js`

## Validation

- `node --check linux-features/remote-mobile-control/test.js`
- `git diff --check`

## Known risk

This assumes Linux `remote-ssh*` app-server hosts do not support `remoteControl/*`. That matches the current runtime behavior. If upstream later adds remote-control support to SSH app-server hosts, this guard should be revisited.

### DIFF
--- a/linux-features/remote-mobile-control/patch.js
+++ b/linux-features/remote-mobile-control/patch.js
@@ -42,6 +42,7 @@ const REMOTE_CONTROL_ENABLEMENT_BRIDGE_MARKER = "codexLinuxRemoteControlEnableme
 const REMOTE_CONTROL_AUTO_CONNECT_CLEANUP_MARKER = "codexLinuxRemoteControlAutoConnectCleanup";
 const REMOTE_CONTROL_SELF_AUTO_CONNECT_MARKER = "codexLinuxRemoteControlSelfAutoConnect";
 const REMOTE_MOBILE_ACTIVE_STATUS_MARKER = "codexLinuxRemoteMobileActiveStatus";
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head. The guard is scoped to the opt-in remote-mobile-control feature, skips only Linux remote-ssh status reads, preserves local host status reads, and local/GitHub checks pass.

ilysenko:
@avifenesh Sorry, Codex removed the draft status and merged this before I caught that it was intentionally left as draft. I will not do that again; for your future draft PRs I will leave them as draft unless you explicitly say they are ready to merge.

avifenesh:
@ilysenko hi, thats ok, i simply tend to draft before i perform more rigorous reviews or enhancements. 
So the main intentions are merged, its just extra testing and maybe more findings. I cross review in general and its more comf in draft open.

Issue #389, opened 2026-06-02
Proposal: optional X11/EWMH-focused Computer Use backend for Cinnamon/X11


AlekseiSeleznev:
Hi! I’ve been working on a standalone Linux/X11 Computer Use plugin and wanted to ask whether you would be open to an optional integration path in this project.

The plugin is here:
https://github.com/AlekseiSeleznev/codex-computer-use-x11

Current release:
https://github.com/AlekseiSeleznev/codex-computer-use-x11/releases/tag/v0.1.2

The short version: it is a Rust MCP plugin focused specifically on the X11/EWMH desktop path, with Linux Mint Cinnamon on X11 as the main validated baseline. It exposes namespaced `x11_*` tools for:

- doctor/readiness checks
- X11 window listing and focus verification
- targeted keyboard input
- click / scroll / drag
- AT-SPI accessibility tree correlation
- app-state style inspection
- target-window context

I know this repository already has a broader Linux Computer Use backend, so I’m not proposing to replace it. My thought is that this could be useful as an optional X11/EWMH-focused feature or compatibility path, especially for users on Cinnamon/X11 where EWMH window control is stable and where Wayland / RemoteDesktop portal support is not the right baseline.

A few design points:

- explicit `x11-ewmh` backend identity
- X11-only readiness baseline, so missing Wayland / RemoteDesktop portal support does not degrade doctor readiness
- exact focus verification before targeted input
- user-local plugin install path
- rollback-friendly installer/uninstaller
- tests around doctor readiness, window targeting, input safety, and packaging docs
- currently validated on Linux Mint Cinnamon/X11

Would this be something you would consider as:

1. an optional `linux-features/` integration,
2. a side-by-side bundled MCP plugin,
3. documentation/linking only,
4. or not a good fit for this repo?

ilysenko:
Thanks for checking first. I think the right path for this is an optional `linux-features/` integration, not a core Computer Use rewrite and not a default bundled plugin.

Please keep it fully opt-in and disabled by default, ideally under something like `linux-features/x11-ewmh-computer-use/`, with its own `feature.json`, README, tests, and X11/EWMH-only readiness checks. Users who do not enable the feature should get exactly the same launcher/updater/Computer Use behavior as before.

[...]

PR #393, opened 2026-06-03
Add optional X11/EWMH Computer Use feature

AlekseiSeleznev:
## What changed

Adds `linux-features/x11-ewmh-computer-use/` as a fully opt-in adapter for the standalone [`codex-computer-use-x11`](https://github.com/AlekseiSeleznev/codex-computer-use-x11) MCP plugin.

The feature:

- stays disabled by default and only activates when listed in `linux-features/features.json`;
- stages the v0.1.3 plugin artifact from a pinned release URL with SHA-256 verification;
- supports local source, local tarball, direct binary, and custom download override modes;
- adds a narrow plugin-gate patch descriptor for Linux only;
- keeps the existing bundled `computer-use` plugin and global doctor behavior unchanged;
- avoids submodules and keeps `codex-computer-use-x11` as the source of truth.

Release used by the default staging path:

- https://github.com/AlekseiSeleznev/codex-computer-use-x11/releases/tag/v0.1.3

## Why

This follows the direction from #389: use an optional `linux-features/` integration instead of rewriting the main Computer Use backend or adding a default bundled plugin.

## Validation

Ran locally from a clean `origin/main` worktree:

```bash
bash -n linux-features/x11-ewmh-computer-use/stage.sh
node --test linux-features/x11-ewmh-computer-use/test.js
node --test linux-features/*/test.js
```

All feature tests passed locally (`246` tests total for `linux-features/*/test.js`).

### DIFF
--- a/linux-features/x11-ewmh-computer-use/README.md
+++ b/linux-features/x11-ewmh-computer-use/README.md
@@ -1,0 +1,79 @@
+# X11/EWMH Computer Use Linux Feature
+
+This optional Linux Feature stages the standalone `codex-computer-use-x11` MCP plugin into Codex Desktop Linux. It stays disabled by default and is enabled only when listed in `linux-features/features.json`.
+
+## Enable
+
+Enable through the git-ignored upstream file `linux-features/features.json`:
+
+```json
+{ "enabled": ["x11-ewmh-computer-use"] }
+```
+
+## Baseline
+
+Supported baseline: Linux Mint Cinnamon on X11 / `x11-ewmh`.
+
+## Tools exposed
+
+The staged plugin exposes the standalone namespaced tool surface:
+
+- `x11_doctor`
+- `x11_list_windows`
+- `x11_focused_window`
+- `x11_focus_window`
+- `x11_accessibility_tree`
+- `x11_type_text`
+- `x11_press_key`
+- `x11_click`
+- `x11_scroll`
+- `x11_drag`
+- `x11_get_app_state`
+- `x11_target_window`
+- `x11_target_context`
+- `x11_release_window`
+
+## Staging modes
+
+Pinned local artifact mode:
+
+```bash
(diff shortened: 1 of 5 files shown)


ilysenko:
Looks reasonable overall, but a few things should be fixed before merging:

1. The default staging path always downloads the x86_64 artifact and ignores `ARCH`. On aarch64 this would stage a non-working binary. Please either add per-arch URL/SHA handling or fail fast for unsupported architectures.

2. Please validate tarball entries before extraction: reject absolute paths, `..` paths, symlinks, and hardlinks before running `tar -xzf`.

3. If the plugin-gate patch cannot find the expected upstream `.computerUse` marker, it should surface that drift instead of silently no-oping as already applied.

4. Since this is an optional Linux Feature, please keep the patch descriptor CI policy optional unless there is a strong reason for `required-upstream`.

AlekseiSeleznev:
Thanks, fixed these in `bff42b1`.

Changes made:

- default release staging now checks `ARCH` and only uses the pinned x86_64 artifact for `x86_64`/`x64`/`amd64`; unsupported arches fail fast with an explicit override message;
- tarball mode now validates entries before extraction and rejects absolute paths, `..` paths, symlinks, hardlinks, and other non-file/non-directory entry types;
- plugin-gate patch now surfaces upstream `.computerUse` marker drift instead of silently no-oping;
- changed the feature patch descriptor `ciPolicy` to `optional`;
- added tests for all of the above.

Validation run locally:

```bash
bash -n linux-features/x11-ewmh-computer-use/stage.sh
node --test linux-features/x11-ewmh-computer-use/test.js
node --test linux-features/*/test.js
```

`linux-features/*/test.js`: 251 passed.

ilysenko (approved):
Reviewed current head bff42b1. The previous blockers are addressed and CI is green.

PR #394, opened 2026-06-03
[codex] Fix Read Aloud settings icon injection

avifenesh:
## Summary
- define the injected Read Aloud settings icon before current `var`-declared settings icon maps reference it
- broaden the Settings page JSX runtime detection to handle current minified declaration shapes
- add a regression test for the `var qe={..."read-aloud-settings":codexLinuxReadAloudSettingsIcon...}` bundle shape

## Why
Opening Settings can crash when the opt-in Read Aloud Linux feature patches a newer Settings page bundle that already contains the `read-aloud-settings` icon-map entry but does not have the injected icon definition. The renderer then hits `ReferenceError: codexLinuxReadAloudSettingsIcon is not defined`.

## Source of truth

## Duplicate check
Searched existing upstream issues and PRs for Read Aloud settings/icon crashes. I found related older merged Read Aloud work and older closed Settings reports, but not this exact undefined icon regression.

## Validation
- `node --check linux-features/read-aloud/patch.js && node --check linux-features/read-aloud/test.js`
- `git diff --check`

## Limitations / risk
This is scoped to the opt-in Read Aloud Linux feature patcher and does not edit generated app assets. It keeps the existing fallback matcher for older bundle shapes.

### DIFF
--- a/linux-features/read-aloud/patch.js
+++ b/linux-features/read-aloud/patch.js
@@ -348,7 +348,9 @@ function applySettingsSharedNavPatch(source) {
 }
 
 function detectSettingsPageJsxRuntime(source) {
-  const iconMatch = source.match(/[,(]([A-Za-z_$][\w$]*)=[A-Za-z_$][\w$]*=>\(0,([A-Za-z_$][\w$]*)\.jsxs\)\(\`svg\`,/);
+  const iconMatch = source.match(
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed current head 2f1f9534eb8142cceec268a53b291af43d365fda. The change is scoped to the opt-in Read Aloud feature, local focused tests passed, and GitHub Actions are green.

PR #396, opened 2026-06-03
Handle missing polkit auth agent in updater

moxw:
## Summary
- Detect missing graphical polkit authentication agents before the updater enters the pkexec install path.
- Keep rebuilt packages ready and report a terminal sudo install command when GUI auth is unavailable.
- Surface the manual-install message through the current Linux package updater bridge.
- Bump codex-update-manager to 0.8.2 and document the behavior.

## User-visible behavior
Clicking the in-app update button in minimal window-manager sessions without a graphical polkit agent no longer quits Codex into a failed pkexec path. The updater leaves the package ready and shows/reports the manual terminal command instead.

## Validation
- git diff --cached --check


ilysenko:
@moxw Two blockers before this can be merged:

1. The class-style updater bridge still does not handle `Manual install required:`. The bootstrap bridge now shows the manual sudo command, but the class-style path in `scripts/lib/linux-update-bridge-patch.js` still only handles `already installed` / `No Codex Desktop update is ready`. Please add the same branch there and cover it in the class-style updater bridge test.

2. `docs/updater.md` still says the final install uses `pkexec` only. Please update it to describe the new manual/sudo fallback when no graphical polkit agent is available.

moxw:
Thanks. I updated docs/updater.md in 94f27a3 to describe the pkexec/manual sudo fallback.

On the class-style updater bridge: I am intentionally leaving that path unchanged in this PR. The current upstream Electron app uses the bootstrap updater bridge, and this PR is scoped to the current upstream bundle shape. The class-style bridge is retained compatibility for older bundle shapes, and expanding this fix there would broaden the patch/test surface beyond the current app path we need to support.

ilysenko (approved):
Reviewed current head. The docs blocker is fixed, and I am accepting the class-style updater bridge as legacy compatibility outside the current upstream bundle path. Current upstream/bootstrap path is covered and CI is green.

Issue #395, opened 2026-06-03
Nix not working


phiwan-dev:
When i run `nix run github:ilysenko/codex-desktop-linux` in a terminal nothing happens. same with "nix shell ..." and then call "codex-desktop" manually. The flake does get downloaded and cached but I don't see any output. No window, no errors, nothing. I'm on Nixos 25.05 with Hyprland on an Intel iGPU. I tried adding "CODEX_LINUX_RENDERING_MODE=wayland-gpu" in front of the commands but it didn't help. Am I missing something?

ilysenko:
This should open a window, so this is worth debugging.

One important detail: after startup the launcher writes output to `~/.cache/codex-desktop/launcher.log`, so no terminal output can be normal. Could you please check:

```bash
codex --version || true
nix run github:ilysenko/codex-desktop-linux -- --help
nix run github:ilysenko/codex-desktop-linux; echo "exit=0"
sed -n '1,220p' ~/.cache/codex-desktop/launcher.log
```

My first suspicion is that the Codex CLI is not found. The Nix flake does not currently include the native package updater/CLI install prompt path, so a missing CLI may fail through the launcher log instead of showing a useful terminal message.

phiwan-dev:
You are right, the logs mention a missing codex cli.
```
[2026-06-03T12:45:27+02:00] Starting Codex Desktop launcher
[2026-06-03T12:45:27+02:00] launcher_phase=feature_prelaunch elapsedMs=91
[2026-06-03T12:45:27+02:00] launcher_phase=packaged_prelaunch elapsedMs=103
Started webview server pid=60798 dir=/nix/store/jmkbsnr8jqhif8nvi99nkkj49zy2vdpv-codex-desktop-26.601.21317/opt/codex-desktop/content/webview
Waiting for webview server on :5175
127.0.0.1 - - [03/Jun/2026 12:45:27] "GET /index.html HTTP/1.1" 200 -
Webview server is ready
127.0.0.1 - - [03/Jun/2026 12:45:27] "GET /index.html HTTP/1.1" 200 -
Webview origin verified.
[2026-06-03T12:45:27+02:00] launcher_phase=webview_ready elapsedMs=380
[2026-06-03T12:45:27+02:00] launcher_phase=cli_lookup elapsedMs=395
[...]

PR #399, opened 2026-06-03
Document Nix Codex CLI requirement

ilysenko:
## What changed

- Added a Nix docs section explaining that the flake does not install the Codex CLI.
- Documented that users should make `codex` available themselves or set `CODEX_CLI_PATH`.
- Added a troubleshooting row for `nix run` appearing to do nothing while the launcher writes the real error to `~/.cache/codex-desktop/launcher.log`.

## Why

Issue #395 shows that Nix users can hit a silent-looking launch failure when the app starts successfully but stops at CLI lookup because the Codex CLI is missing. This was not the upstream patch drift fixed in #397; it is a Nix docs/UX gap.

## Validation

- `git diff --check`

### DIFF
--- a/docs/nix.md
+++ b/docs/nix.md
@@ -11,6 +11,24 @@ bot refreshes the upstream `Codex.dmg` hash and verifies the Nix package outputs
 in `main`. If you hit a hash mismatch right after an upstream release, wait for
 the next bot run and retry.
 
+## Codex CLI Requirement
+
+The Nix package does not install the Codex CLI for you. Before first launch,
+make sure the `codex` command is available in your user environment, or set
+`CODEX_CLI_PATH` to the binary you want Codex Desktop to use.
+
+One option is the upstream npm package:
+
+```bash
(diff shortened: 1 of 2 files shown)


docs/nix.md
```diff
@@ -11,6 +11,24 @@ bot refreshes the upstream `Codex.dmg` hash and verifies the Nix package outputs
...
+## Codex CLI Requirement
+
+The Nix package does not install the Codex CLI for you. Before first launch,
+make sure the `codex` command is available in your user environment, or set
+`CODEX_CLI_PATH` to the binary you want Codex Desktop to use.
+
+One option is the upstream npm package:
+
+```bash
+npm i -g @openai/codex
```
colonelpanic8, line 23: I don't think this is a good recommendation for nix users.

colonelpanic8: would be better to provide a home-manager and nixos module that installs the codex cli package alongside

PR #401, opened 2026-06-03
Fix locale override and settings menu for API key auth on Linux

losese:
## Summary

This PR fixes two Linux webview compatibility issues that affect users who do not sign in with an official ChatGPT account and instead authenticate with an API key, including third-party API-compatible endpoints/keys.

- Respect an explicit `localeOverride` even when the upstream `enable_i18n` gate is disabled.
- Show the existing Settings menu item in the profile dropdown on Linux.

## User-visible behavior

Before this change, API key authenticated users could set a locale override, but the webview i18n provider and language settings row could still stay behind the upstream `enable_i18n` gate. In that state, Chinese and other localized UI strings were not shown.

The profile dropdown also hid the Settings item behind an upstream feature gate, even though the Settings route itself was reachable through other UI paths, such as the archived conversation toast link.

After this change:

- explicit locale overrides work for API key auth on Linux
- the Settings item is visible from the profile dropdown
- the locale patch preserves React hook order by reusing the existing unconditional locale override hook result

## Validation

- `codex-app/resources/node-runtime/bin/node --test scripts/patch-linux-window-ui.test.js`
- Fresh temporary build:
  - `CODEX_INSTALL_DIR=/tmp/codex-app-pr-locale-settings ./install.sh --fresh ""`
- Checked generated webview bundles for the patched i18n/settings shapes.
- Ran `node --check` on the generated patched webview bundles.


ilysenko (approved):
Reviewed current head. The webview patches are scoped, fail-soft, covered by focused tests, and CI is green.

Issue #404, opened 2026-06-03
Linux desktop settings toggles missing from UI on Codex builds with the native Keyboard Shortcuts page


csmashe:
## Summary

On upstream Codex builds that ship a **native "Keyboard shortcuts" settings page**, the Linux-only settings toggles (**System tray**, **Warm start**, **Compact prompt window**, **Install updates when you close Codex**) are missing from the UI entirely. The underlying behavior still works — only the controls are gone, so there is no in-app way to change them.

Observed on Codex `26.601.21317`, repo at `c826e04` (built via the packaged update builder).

## Root cause

The Linux toggles only exist inside the port's injected **"Keybinds"** settings page (`scripts/patches/keybinds-settings.js`, `buildKeybindsSettingsSource` — the page titled `"Keybinds"` / subtitle `"App shortcuts and Linux desktop behavior."`).

`patchKeybindsSettingsAssets()` now bails out before injecting that page when upstream already provides its own shortcuts settings:

```js
// scripts/patches/keybinds-settings.js:245
if (hasNativeKeyboardShortcutsSettings(extractedDir)) {
  return {
    matched: true,
    changed: 0,
    reason: "upstream keyboard shortcuts settings are present",
  };
}
```

`hasNativeKeyboardShortcutsSettings()` returns true when the webview bundle contains both a `settings-sections-*.js` with `slug:\`keyboard-shortcuts\`` and a `keyboard-shortcuts-settings-*.js` asset. The guard was introduced in `d6ac778 "Fix Linux patch drift for current Codex bundle"`.

The guard correctly avoids clobbering/duplicating upstream's new native page, **but the four Linux toggles were never re-homed into that native page** — so they silently disappear. The wrapper records this as `status: "already-applied"` (`scripts/patches/core/all-linux/extracted-app/keybinds/patch.js`), so nothing flags it during the build.

Note the close-to-tray *behavior* patch still applies (`canHideLastLocalWindowToTray` gating reads `codex-linux-system-tray-enabled` in `scripts/patches/launch-actions.js`), which is why closing the last window still hides to tray with no way to turn it off from the UI.

## Confirmation on an affected install

```
content/webview/assets/keyboard-shortcuts-settings-CEkPM3DX.js   # native page present
content/webview/assets/settings-sections-BUO2MNq2.js             # contains slug:`keyboard-shortcuts`
content/webview/assets/keybinds-settings-linux.js                # ABSENT -> injection skipped
```

## Impact

[...]

PR #406, opened 2026-06-03
Make settings section-title patch tolerant of minified identifier drift

csmashe:
## What changed

`injectSettingsSectionTitle()` replaces the literal `titleForSection` section-title needle in both `applyKeybindsSettingsSharedPatch` and `applyLinuxDesktopSettingsSharedPatch`. Instead of matching a hardcoded minified snippet, it matches the upstream `general-settings` case with a regex, captures the JSX factory and message-component identifiers, and reuses them in the injected `keybinds` / `linux-desktop` case.

## Why

#405 added the Linux desktop settings page (System tray, Warm start, Compact prompt window, install-on-close) by injecting a new `titleForSection` switch case into the `settings-shared` bundle. The anchor `general-settings` case was matched with a literal needle that hardcoded the minified JSX message-component identifier (`n`) and memo-cache slot (`t[2]`).

Those identifiers are assigned arbitrarily by the upstream minifier and drift between builds. On Codex `26.601.21317` the component minifies to `r` (and the cache slot to `o[5]`), so the literal needle never matched. Because the patch is transactional, `applyLinuxDesktopSettingsSharedPatch` threw `could not add Linux desktop section title`, the whole optional patch was caught and skipped, and the Linux settings page silently disappeared from the UI — the behavior patches still applied (e.g. close-to-tray), but there was no longer any toggle to control them.

This mirrors the identifier-tolerant regex approach already used by the route/navigation patches in the same file.

- `scripts/patches/keybinds-settings.js` — the fix (real patch source, not generated output)
- `scripts/patch-linux-window-ui.test.js` — regression tests

## How it was tested

- Added a failing-first regression test: a `settings-shared` fixture whose identifiers drift (`r` / `o[5]`, mirroring `26.601.21317`). It reproduces the `could not add ... section title` throw on the old literal needle and passes with the regex. Covers both the keybinds and linux-desktop shared patches.
- `node --test scripts/patch-linux-window-ui.test.js` → 178/178 pass (was 176 before the 2 new tests; existing `n`-identifier tests still pass, so backward compatible).
- `node --check` on both edited files.
- End-to-end on real hardware (Pop!_OS 22.04, x11, `.deb`, Codex `26.601.21317`): rebuilt from the upstream DMG with no `Keybinds settings patch skipped` warning; `linux-desktop-settings-linux.js` is emitted; the injected case uses the bundle's real identifier `(0,d.jsx)(r,{id:\`settings.section.linux-desktop\`...})`; `settings-sections` registers `slug:\`linux-desktop\``. Built a `.deb`, installed it, and confirmed Settings → **Linux desktop** → Desktop integration → **System tray** appears and that toggling it off + relaunch makes closing the window quit the app instead of hiding to the tray.

## Environment limitations

- Verified only on a `deb` / Pop!_OS x11 build against Codex `26.601.21317`. The regex stays backward compatible with the previously-targeted `n`/`t[2]` minification, but I did not separately rebuild `.rpm`/pacman/Nix; the change is in shared patch logic exercised by all package paths, and the JS test suite covers both identifier variants.
- No updater crate (`updater/`) changes, so no version bump per the versioning rules.
- No documentation change — this restores the already-intended #405 behavior rather than changing documented behavior.

## Risks / follow-ups

- Low risk: the change only broadens an existing match; the existing literal-`n` test path is still green.
- Possible follow-up: the same literal-minified-snippet style is used in a few other needles in this file (route/nav already use tolerant regexes); if further drift surfaces they could get the same treatment, but that's out of scope here.

Fixes the regression for bundles described in #404 / #405.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -71,6 +71,8 @@ const {
 const {
   keybindsSettingsAsset,
   linuxDesktopSettingsAsset,
+  applyKeybindsSettingsSharedPatch,
+  applyLinuxDesktopSettingsSharedPatch,
 } = require("./patches/keybinds-settings.js");
 const {
   validateReport,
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head and ran the focused settings patch tests plus the full patcher suite locally. No blockers found.

PR #407, opened 2026-06-04
fix(read-aloud): repair settings route patch

joshyorko:
## Summary
- re-export the Read Aloud settings route from split general-settings wrappers
- follow the current General settings export map instead of stale minifier aliases
- persist Read Aloud custom keys through global-state IPC and mirror them into launcher settings

## Tests
- node --test linux-features/read-aloud/test.js
- node --test --test-name-pattern "persists Linux settings|migrates already-patched Linux settings" scripts/patch-linux-window-ui.test.js
- git diff --check

### DIFF
--- a/linux-features/read-aloud/patch.js
+++ b/linux-features/read-aloud/patch.js
@@ -212,7 +212,7 @@ function generalSettingsReadAloudBlockSource() {
 }
 
 function currentGeneralSettingsReadAloudBlockSource() {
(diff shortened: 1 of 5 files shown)


ilysenko (approved):
Reviewed the current head, including the interaction with the freshly merged settings patch. Ran read-aloud feature tests, focused settings persistence tests, and the full patcher suite on the combined state. No blockers found.

PR #409, opened 2026-06-04
[codex] Invalidate stale DMG cache

avifenesh:
## Summary

- Store upstream `Codex.dmg` HEAD metadata and the upstream URL hash next to the installer cache, then refresh the cached DMG when that fingerprint changes.
- Preserve the previous cached DMG if a refresh download fails, redact override URLs in logs/metadata, and remove the metadata sidecar during `--fresh` rebuilds.
- Let `scripts/rebuild-candidate.sh` use the installer-managed cache by default, while preserving explicit `DMG=/path/to/Codex.dmg` behavior.

## Why

The default rebuild path could silently reuse an old `Codex.dmg`. `rebuild-candidate.sh` also resolved and passed the local `Codex.dmg` explicitly, which bypassed any installer-side chance to refresh it. That made it possible to rebuild a Linux app while believing it was based on the current upstream DMG, even when the local cache was stale.

## Source Of Truth

- `scripts/lib/dmg.sh`
- `scripts/lib/install-helpers.sh`
- `scripts/rebuild-candidate.sh`
- `tests/scripts_smoke.sh`
- `docs/build-and-packaging.md`

## Validation

- `bash -n install.sh scripts/lib/*.sh scripts/rebuild-candidate.sh launcher/start.sh.template tests/scripts_smoke.sh`
- `git diff --check`

No matching open issue or PR was found for `DMG cache`, `cached DMG`, or `Codex.dmg`.

### DIFF
--- a/scripts/lib/dmg.sh
+++ b/scripts/lib/dmg.sh
@@ -54,6 +54,26 @@ dmg_url_cache_key() {
     printf '%s' "$1" | sha256sum | awk '{print $1}'
 }
 
+cached_dmg_metadata_url_sha256() {
+    local metadata_path="$1"
+
+    awk -F= '$1 == "url_sha256" { print $2; exit }' "$metadata_path" 2>/dev/null || true
+}
+
+cached_dmg_metadata_matches_url() {
+    local metadata_path="$1"
+    local dmg_url="$2"
+    local cached_url_sha256
+    local expected_url_sha256
+
+    [ -s "$metadata_path" ] || return 1
+
+    cached_url_sha256="$(cached_dmg_metadata_url_sha256 "$metadata_path")"
+    expected_url_sha256="$(dmg_url_cache_key "$dmg_url")"
+
+    [ -n "$cached_url_sha256" ] && [ "$cached_url_sha256" = "$expected_url_sha256" ]
+}
+
 fetch_dmg_remote_fingerprint() {
     local dmg_url="$1"
     local headers_file="$WORK_DIR/dmg-headers.txt"
@@ -112,8 +132,13 @@ cached_dmg_is_fresh() {
     local remote_fingerprint
 
     if ! remote_fingerprint="$(fetch_dmg_remote_fingerprint "$dmg_url")"; then
-        warn "Could not check upstream DMG metadata; using cached DMG"
-        return 0
+        if cached_dmg_metadata_matches_url "$metadata_path" "$dmg_url"; then
+            warn "Could not check upstream DMG metadata; using cached DMG for matching URL"
+            return 0
+        fi
+
+        warn "Could not check upstream DMG metadata; cached DMG URL metadata does not match current URL"
+        return 1
     fi
     DMG_REMOTE_FINGERPRINT="$remote_fingerprint"
 
--- a/scripts/lib/dmg.sh
+++ b/scripts/lib/dmg.sh
@@ -5,31 +5,186 @@
 # shellcheck shell=bash
 
 # ---- Download or find Codex DMG ----
+DEFAULT_DMG_URL="https://persistent.oaistatic.com/codex-app-prod/Codex.dmg"
+DMG_URL="${CODEX_UPSTREAM_DMG_URL:-$DEFAULT_DMG_URL}"
+DMG_REMOTE_FINGERPRINT=""
+
+redact_dmg_url() {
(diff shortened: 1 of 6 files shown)


scripts/lib/dmg.sh
```diff
@@ -5,31 +5,186 @@
...
 # ---- Download or find Codex DMG ----
+DEFAULT_DMG_URL="https://persistent.oaistatic.com/codex-app-prod/Codex.dmg"
+DMG_URL="${CODEX_UPSTREAM_DMG_URL:-$DEFAULT_DMG_URL}"
+DMG_REMOTE_FINGERPRINT=""
+
+redact_dmg_url() {
+    local dmg_url="$1"
+    local scheme="${dmg_url%%://*}"
+    local remainder="${dmg_url#*://}"
+    local authority="${remainder%%/*}"
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This redaction parser only separates the authority from the path on `/`. A valid override like `https://example.com?token=secret` has no slash after the host, so `authority` becomes `example.com?token=secret`, `path` stays empty, and the later query/fragment redaction never runs before the URL is logged. Since override URLs can carry credentials in query strings, please split/redact on `?`/`#` even when there is no path component.

avifenesh (commented):

---

Found one redaction edge case in the new DMG URL handling.

scripts/lib/dmg.sh
avifenesh: Fixed in the latest push. The authority is now split on the first `/`, `?`, or `#`, so a credential-bearing override like `https://example.com?token=secret` redacts to `https://example.com?REDACTED` even with no path component.

avifenesh (commented):

---

Found one cache-invalidation edge case with URL overrides when the HEAD probe cannot produce a fingerprint.

scripts/lib/dmg.sh
```diff
@@ -5,31 +5,188 @@
...
+}
+
+cached_dmg_is_fresh() {
+    local dmg_dest="$1"
+    local metadata_path="$2"
+    local dmg_url="$3"
+    local remote_fingerprint
+
+    if ! remote_fingerprint="$(fetch_dmg_remote_fingerprint "$dmg_url")"; then
+        warn "Could not check upstream DMG metadata; using cached DMG"
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This fallback reuses the existing `Codex.dmg` without first checking whether the sidecar's `url_sha256` matches the current `DMG_URL`. That means a cache created for the default URL can be reused after `CODEX_UPSTREAM_DMG_URL` is changed if the new endpoint's HEAD request fails or omits all tracked headers, even though the URL hash is specifically part of the fingerprint. In that case the installer never attempts the GET for the override and silently builds from the old upstream DMG. Please gate this fallback on the stored URL hash matching `dmg_url_cache_key "$dmg_url"`; if it differs, refresh/download or fail rather than treating the old cache as fresh.

ilysenko:
One blocker is still reproducible.

When CODEX_UPSTREAM_DMG_URL points to a different URL and the HEAD request fails, the installer can still reuse an existing cached Codex.dmg whose metadata url_sha256 belongs to the default upstream URL. In that case it builds from the wrong DMG instead of downloading/failing.

Before falling back to the cached DMG on HEAD failure, please verify that the cached metadata url_sha256 matches the current DMG_URL. If it does not match, the cache must not be reused.

The redaction issue looks fixed now.

avifenesh:
Addressed the HEAD-failure/cache-reuse blocker in 901d742. The cached DMG is now reused on HEAD failure only when the sidecar url_sha256 matches the current DMG_URL; mismatches force the refresh/download path instead of silently building from the old URL cache. Added a smoke regression for the mismatched-URL HEAD-failure case.

Validation:
- `bash -n scripts/lib/dmg.sh tests/scripts_smoke.sh\n- bash tests/scripts_smoke.sh`
- GitHub CI run 27014393902 is green

ilysenko (approved):
Reviewed current head 901d742. The HEAD-failure URL-mismatch blocker is fixed with metadata url_sha256 validation, the redaction edge case is covered, CI is green, local smoke tests pass, and the branch merges cleanly with current main.

PR #411, opened 2026-06-04
[codex] Fix Read Aloud aplay underrun on busy CPU

avifenesh:
## Problem

`kokoro_stdin.py` spawned `aplay` up front, then synthesized chunk 0 — seconds of work on a loaded CPU. aplay underran its empty ALSA buffer immediately, heard as the last fragment looping followed by a gap when real audio finally arrived.

## Fix

- Pull the first synthesized PCM chunk **before** opening aplay so it starts with audio in hand.
- Deepen the synthesis queue (`maxsize` 2 → 32) so rendering runs ahead of playback.
- Give ALSA a real cushion: `--buffer-time=500000 --period-time=100000` (0.5s buffer / 0.1s period) to ride out scheduler jitter.

## Test

Adds a test asserting first-chunk synthesis completes before aplay launches and that the buffer/period flags are passed. `node --test`: 38/38 pass.

### DIFF
--- a/linux-features/read-aloud/bin/kokoro_stdin.py
+++ b/linux-features/read-aloud/bin/kokoro_stdin.py
@@ -116,14 +116,11 @@ def main() -> int:
     lang = os.environ.get("CODEX_LINUX_READ_ALOUD_KOKORO_LANG", "en-us")
 
     sample_rate = 24000
-    player = subprocess.Popen(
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head. The first PCM chunk is consumed before launching aplay, the chunk is preserved for playback, and the new test covers ordering plus ALSA buffering flags. Local read-aloud tests passed. No blockers found.

PR #413, opened 2026-06-05
fix(linux): use native titlebar overlay colors

jojomini1231-cloud:
## Summary

- keep Linux primary windows frameless so Codex Desktop can use its native/custom titlebar instead of the system titlebar
- avoid applying the Windows `autoHideMenuBar` window-option branch to Linux
- give Linux `titleBarOverlay` explicit theme-aware background and symbol colors so GNOME/Wayland does not render the caption button area as a black invisible block
- add regression coverage for the Linux titlebar/window option patching behavior

## Motivation

On Linux/GNOME Wayland, using the system titlebar leaves a dark/black strip at the top of the Codex Desktop window. Switching Linux to the frameless/titleBarOverlay path fixes that, but reusing the Windows overlay object can make the minimize/maximize/close caption button area render as a black block with invisible symbols.

This patch gives Linux its own overlay colors:

- light theme: light overlay background with dark symbols
- dark theme: dark overlay background with light symbols

## Test plan

```bash
```

Result locally: 180 passing, 0 failing.


scripts/patches/main-process.js
```diff
@@ -64,24 +64,53 @@ function applyLinuxWindowOptionsPatch(currentSource, iconAsset) {
...
+
+  if (currentSource.includes(replacement)) {
+    return currentSource;
+  }
+
+  if (currentSource.includes(linuxTransparentOverlayNeedle)) {
+    return currentSource.split(linuxTransparentOverlayNeedle).join(replacement);
+  }
+
+  if (currentSource.includes(legacyNeedle)) {
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This new patch hardcodes the minified aliases from one upstream bundle (`n`, `a`, `y2`, `b2`, `a2`, `o2`, `v2`, `_2`, `g2`). The rest of this file generally infers aliases or matches structural regexes because upstream rebuilds can rename these identifiers, and the new descriptor is `ciPolicy: "optional"`; if any alias drifts, the helper just warns and CI can still pass with `linux-native-titlebar` skipped. Since this patch is what keeps Linux on the frameless `titleBarOverlay` path, that would silently revert future builds to `{titleBarStyle:default}`/system titlebar. Please either derive the local aliases from the matched titlebar/overlay/color-helper code or make this descriptor required so upstream drift fails the upstream-build report instead of shipping without the patch.

avifenesh (commented):

---

Found one ASAR patching reliability issue in the new titlebar patch.

jojomini1231-cloud:
Thanks for the review. I updated the titlebar patch to derive the minified aliases from the matched bundle code instead of hardcoding the current upstream names.

The patch now:

- matches the primary titlebar branch structurally and captures the platform/titlebar helper aliases
- derives the `titleBarOverlay` helper aliases from the matched overlay helper function
- derives the Linux background color aliases from the existing Linux opaque-background branch
- keeps the already-patched Linux overlay path idempotent

I also updated the regression fixture so it includes the relevant titlebar, overlay-helper, and Linux background code paths.

Tested locally:

```bash
```

Result: 180 passing, 0 failing.

scripts/patches/main-process.js
```diff
@@ -64,24 +64,68 @@ function applyLinuxWindowOptionsPatch(currentSource, iconAsset) {
...
+    console.warn("WARN: Could not find primary BrowserWindow titlebar snippet — skipping Linux native titlebar patch");
+    return currentSource;
+  }
+
+  const [, platformAlias, opaqueWindowsAlias, trafficLightAlias, zoomAlias, , overlayHelperAlias] = primaryTitlebarMatch;
+  const overlayHelperRegex = new RegExp(
+    `function ${escapeRegExp(overlayHelperAlias)}\\([^)]*\\)\\{return\\{color:[A-Za-z_$][\\w$]*,symbolColor:([A-Za-z_$][\\w$]*)\\.nativeTheme\\.shouldUseDarkColors\\?([A-Za-z_$][\\w$]*):([A-Za-z_$][\\w$]*),height:Math\\.round\\(([A-Za-z_$][\\w$]*)\\*[^)]*\\)\\}\\}`,
+  );
+  const overlayHelperMatch = currentSource.match(overlayHelperRegex);
+  const linuxBackgroundMatch = currentSource.match(
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This matcher only succeeds after the Linux background branch has already been inserted, but the new descriptor runs at order 55 while `linux-opaque-background` (the patch that adds `===\`linux\`&&!...{backgroundColor:...}`) still runs later at order 80. On a fresh upstream bundle with the current background helper shape, this leaves `linuxBackgroundMatch` null, hits the warning path below, and returns the source unchanged—so the native titlebar overlay never gets applied. Please either derive the dark/light color aliases from the pre-existing darwin/win32 background branch, or move this descriptor after the opaque-background patch.

avifenesh (commented):

---

Found one remaining correctness issue: the new native-titlebar patch is ordered before the background helper patch it depends on, so it can skip on a fresh bundle.

jojomini1231-cloud:
Thanks again for the review. I updated the PR to address both points:

- The native titlebar patch now derives the minified aliases from the matched titlebar branch, overlay helper, and Linux background branch instead of hardcoding the current upstream bundle identifiers.
- `linux-native-titlebar` now runs after `linux-opaque-background` (`order: 85` vs `80`) because it reuses the inserted Linux background branch aliases. I also added a descriptor-order regression assertion for this.
- The Linux titlebar overlay sync path now handles Linux explicitly on `nativeTheme.updated`, using the same derived aliases and a fixed dark overlay color of `#111111` to match the surrounding dark titlebar area.
- Removed the temporary renderer/background-read approach; the installed ASAR no longer contains the `data-codex-window-type` titlebar probing path.

Tested locally:

```bash
node -c scripts/patches/main-process.js
```

Result: 181 passing, 0 failing.

I also rebuilt and installed the local Debian package, then extracted `/opt/codex-desktop/resources/app.asar` to verify the shipped bundle contains both the initial Linux titlebar overlay and the `nativeTheme.updated` sync path with `#111111`.

ilysenko (approved):
Reviewed current head ea3411f. GitHub checks are green, local patcher tests pass, and the branch merges cleanly with current main.

PR #414, opened 2026-06-05
Add AUR packaging workflow

### LINKED ISSUE #402: Make it available on arch aur
Self explanatory would make it simple and easy to keep the codex workaround up to date with fixes easily

akshaynexus:
## Summary

Adds an AUR packaging path alongside the existing local pacman package builder.

- Adds an AUR PKGBUILD template and install hook for `codex-desktop-linux`
- Adds `scripts/aur/render-aur-package.sh` plus `make aur` and `make aur-srcinfo`
- Adds a manual GitHub Actions workflow to render, validate, and optionally publish AUR metadata
- Declares the managed Node, Electron, and Browser Use runtime archives as AUR sources and passes them into the existing installer through local-source overrides
- Documents local AUR testing and post-merge publishing setup
- Extends CI shell syntax checks to cover the new renderer

Closes https://github.com/ilysenko/codex-desktop-linux/issues/402

The AUR package builds from the upstream repository source URL, `https://github.com/ilysenko/codex-desktop-linux`. This PR does not change `codex-update-manager`; it only omits the updater from the AUR package payload so Arch users update through pacman/AUR helpers instead.

## How maintainers publish to AUR after merge

Publishing is not automatic until the workflow is run manually with `publish=true`.

1. Create the AUR package repository once, if it does not already exist:

   ```bash
   git clone ssh://aur@aur.archlinux.org/codex-desktop-linux.git
   ```

   The first successful workflow push can populate the empty AUR Git repository with `PKGBUILD`, `.SRCINFO`, and `codex-desktop-linux.install`.

2. Add an SSH private key that can push to the AUR package as the `AUR_SSH_PRIVATE_KEY` GitHub Actions secret. The matching public key must be registered on the maintainer's AUR account.

3. Optionally add `AUR_KNOWN_HOSTS`. If omitted, the workflow runs `ssh-keyscan aur.archlinux.org` during setup.

4. Open GitHub Actions in this repository and run `Publish AUR Package` manually with `publish` set to `true`.

5. Optionally set `pkgver` for a specific package version. If omitted, the workflow uses a UTC timestamp in `YYYY.MM.DD.HHMMSS` format.

6. Optionally set `source_ref` for a specific commit, tag, or branch. If omitted, the workflow uses the workflow run's commit SHA. The workflow resolves this value to a full commit SHA before rendering `PKGBUILD`.

The workflow keeps the GitHub source archive and DMG checksums as `SKIP`, while computing checksums for declared runtime archives when `publish=true`. GitHub-generated source archives can be regenerated with different compressed bytes, and the Codex DMG is consumed from the stable upstream download URL. Package freshness is handled by rerunning the publish workflow for a new AUR revision.

The AUR build now declares the large runtime archives that already have installer override hooks: managed Node, Electron, and Browser Use `node_repl` on x86_64. The remaining networked part is npm rebuild inputs for native modules; making those source-declared would require a larger npm/cache mode outside this PR.

## Validation

- `bash -n install.sh`
- `bash -n scripts/aur/render-aur-package.sh`
- `make aur-srcinfo AUR_PKGVER=2026.06.05`
- `AUR_SOURCE_REF=refs/tags/v1 AUR_PKGVER=2026.06.05 ./scripts/aur/render-aur-package.sh` fails with the expected full-SHA validation error
- Full local AUR-style build with `makepkg -si --needed --noconfirm`
- Installed resulting package with `pacman -U dist/aur/codex-desktop-linux-2026.06.05-1-x86_64.pkg.tar.zst` after removing the previous `codex-desktop` package

### DIFF
--- a/packaging/aur/PKGBUILD.template
+++ b/packaging/aur/PKGBUILD.template
@@ -54,6 +54,9 @@ install="${pkgname}.install"
 _source_ref="__SOURCE_REF__"
 _dmg_url="__DMG_URL__"
 _repo_dir="codex-desktop-linux-${_source_ref}"
+_node_version="__NODE_VERSION__"
+_electron_version="__ELECTRON_VERSION__"
+_browser_runtime_version="__BROWSER_RUNTIME_VERSION__"
 
 source=(
     "${_repo_dir}.tar.gz::__SOURCE_REPO__/archive/${_source_ref}.tar.gz"
@@ -63,6 +66,49 @@ sha256sums=(
     '__SOURCE_SHA256__'
     '__DMG_SHA256__'
 )
+source_x86_64=(
+    "node-${_node_version}-linux-x64.tar.xz::https://nodejs.org/dist/${_node_version}/node-${_node_version}-linux-x64.tar.xz"
+    "electron-v${_electron_version}-linux-x64.zip::https://github.com/electron/electron/releases/download/v${_electron_version}/electron-v${_electron_version}-linux-x64.zip"
+    "codex-primary-runtime-linux-x64-${_browser_runtime_version}.tar.xz::https://persistent.oaistatic.com/codex-primary-runtime/${_browser_runtime_version}/codex-primary-runtime-linux-x64-${_browser_runtime_version}.tar.xz"
+)
+sha256sums_x86_64=(
+    '__NODE_X64_SHA256__'
+    '__ELECTRON_X64_SHA256__'
+    '__BROWSER_RUNTIME_X64_SHA256__'
+)
+source_aarch64=(
+    "node-${_node_version}-linux-arm64.tar.xz::https://nodejs.org/dist/${_node_version}/node-${_node_version}-linux-arm64.tar.xz"
+    "electron-v${_electron_version}-linux-arm64.zip::https://github.com/electron/electron/releases/download/v${_electron_version}/electron-v${_electron_version}-linux-arm64.zip"
+)
+sha256sums_aarch64=(
+    '__NODE_ARM64_SHA256__'
+    '__ELECTRON_ARM64_SHA256__'
+)
+
+prepare() {
+    rm -rf "${srcdir}/aur-node-runtime"
+    mkdir -p "${srcdir}/aur-node-runtime"
+
+    case "${CARCH}" in
+        x86_64)
+            tar -xJf "${srcdir}/node-${_node_version}-linux-x64.tar.xz" \
+                -C "${srcdir}/aur-node-runtime" --strip-components=1
+            rm -rf "${srcdir}/aur-browser-use-runtime"
+            mkdir -p "${srcdir}/aur-browser-use-runtime"
+            tar -xJf "${srcdir}/codex-primary-runtime-linux-x64-${_browser_runtime_version}.tar.xz" \
+                -C "${srcdir}/aur-browser-use-runtime" \
+                codex-primary-runtime/dependencies/bin/node_repl
+            ;;
+        aarch64)
+            tar -xJf "${srcdir}/node-${_node_version}-linux-arm64.tar.xz" \
+                -C "${srcdir}/aur-node-runtime" --strip-components=1
+            ;;
+        *)
+            echo "Unsupported architecture: ${CARCH}" >&2
+            return 1
+            ;;
+    esac
+}
 
 build() {
     cd "${srcdir}/${_repo_dir}"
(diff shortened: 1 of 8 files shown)


scripts/aur/render-aur-package.sh
```diff
@@ -0,0 +1,102 @@
...
+AUR_DMG_SHA256="${AUR_DMG_SHA256:-SKIP}"
+AUR_PKGVER="${AUR_PKGVER:-}"
+
+error() {
+    echo "[ERROR] $*" >&2
+    exit 1
+}
+
+sed_escape_replacement() {
+    printf '%s' "$1" | sed -e 's/[\/&]/\\&/g'
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

The renderer uses `|` as the sed delimiter for `__SOURCE_REPO__` and `__DMG_URL__` below, but `sed_escape_replacement` only escapes `/` and `&`. If either URL contains a literal `|` (allowed in a configured `AUR_SOURCE_REPO`/`AUR_DMG_URL` override), the generated sed expression is split and rendering fails. This repo has the same packaging invariant for PKGBUILD substitutions: escape the delimiter you use (or use a delimiter not present in the replacement).

packaging/aur/PKGBUILD.template
```diff
@@ -0,0 +1,98 @@
...
+    export PACKAGE_VERSION="${pkgver}-${pkgrel}"
+    export PACKAGE_WITH_UPDATER=0
+    export DESKTOP_TEMPLATE="$PWD/packaging/linux/codex-desktop.desktop"
+    export ICON_SOURCE="$PWD/assets/codex.png"
+    export PACKAGED_RUNTIME_SOURCE="$PWD/packaging/linux/codex-packaged-runtime.sh"
+
+    # shellcheck source=/dev/null
+    . "$PWD/scripts/lib/package-common.sh"
+
+    stage_common_package_files "$pkgdir"
```
avifenesh, line 152: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`stage_common_package_files` copies the staged app from `$APP_DIR`, but the AUR package never runs `ensure_app_layout` before calling it. If `install.sh` exits without producing `codex-app/start.sh` (for example after an upstream layout drift that still leaves the checkout present), `cp -aT "$APP_DIR"` will be the first failure and makepkg reports a low-level copy error rather than the repository's explicit "Run ./install.sh first" invariant. Please call `ensure_app_layout` before staging, matching the local pacman builder's precondition check.

avifenesh (commented):

ilysenko:
I would not merge this yet. The overall AUR direction can be useful, but there are a couple of concrete blockers to fix first:

1. `render-aur-package.sh` uses `|` as the sed delimiter for URL substitutions, but `sed_escape_replacement` does not escape `|`. I reproduced this locally: an override URL containing `|` silently renders a corrupted PKGBUILD instead of failing. Please escape the delimiter as well, ideally along with backslashes.

2. The AUR `package()` path should call `ensure_app_layout` before `stage_common_package_files`, matching the native package builders. Otherwise a failed or partial `install.sh` run can turn into a low-level copy error instead of the repo's explicit app layout check.

I’m also cautious about the support cost here: this AUR package intentionally uses the normal networked `install.sh` build flow, so it may download Node runtime, Electron, npm inputs, and fallback resources during `build()`. That may be acceptable as a best-effort AUR package, but it should be very clear in the docs because it will likely create more AUR-specific build support requests.

akshaynexus:
Yeah makes sense will fix that shortly

akshaynexus:
@codex review

akshaynexus:
@codex review

scripts/aur/render-aur-package.sh
```diff
@@ -0,0 +1,128 @@
...
+    exit 1
+}
+
+sed_escape_replacement() {
+    printf '%s' "$1" | sed -e 's/[\/&|]/\\&/g'
+}
+
+validate_pkgver() {
+    case "$1" in
+        ""|*[-:[:space:]]*)
```
avifenesh, line 36: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This validation still allows shell metacharacters even though the value is later emitted as an unquoted `pkgver=__PKGVER__` assignment in the PKGBUILD and `makepkg --printsrcinfo` sources that file. For example, a workflow-dispatch `pkgver` like `1$(id>/tmp/aur-pkgver)` contains no hyphen, colon, or whitespace, so it passes here; the command substitution is then executed when makepkg reads the rendered PKGBUILD. Please switch this to an allowlist for Arch pkgver characters (for example alnum plus `._+`) or otherwise render the value so it cannot be evaluated as shell syntax.

avifenesh (commented):

---

Found one issue in the AUR renderer input validation that can still turn workflow/package metadata input into shell syntax when `makepkg` sources the rendered PKGBUILD.

ilysenko:
Thanks again for working on this and for being willing to address the review feedback. Sorry for closing the PR without a clearer note after you had already put time into it.\n\nFor now we decided not to merge the AUR packaging workflow, mostly because we want to keep the supported package/update paths narrower until the AUR support and its maintenance expectations are clearer. This does not mean the idea is rejected permanently; we may come back to AUR support later, and your work here will be useful context if we do.\n\nI appreciate the contribution and the validation you did.

PR #415, opened 2026-06-05
fix(browser): enable Browser Use gates on Linux

losese:
## Summary

Fix Browser Use availability on Linux when the upstream webview Statsig gates are disabled.

This adds Linux webview patches for two Browser Use gates:

- enable the in-app Browser Use availability path when only the Browser Use Statsig gate is disabled
- allow Browser Use to navigate to non-local URLs by syncing `browser-use-non-local-sites-allowed-changed` as enabled on Linux

Without this, Browser settings can show “Disabled by your organization or unavailable in your region”, and Browser Use navigation can fail with:

`Blocked browser navigation to non-local URL: https://www.baidu.com/`

## Local Environment

This was reproduced in a local Linux build with:

1. third-party API key authentication
2. third-party API base URL

The same account/API setup works in the macOS Codex GUI, so this appears to be a Linux wrapper/webview gating issue rather than an account, organization, or region restriction.

## Validation

```bash
codex-app/resources/node-runtime/bin/node --test scripts/patch-linux-window-ui.test.js

### DIFF
--- a/scripts/patches/webview-assets.js
+++ b/scripts/patches/webview-assets.js
@@ -265,10 +265,6 @@ function applyLinuxBrowserUseNonLocalNavigationPatch(currentSource) {
   const statsigNeedle = "3903563814";
   let changed = false;
 
-  if (currentSource.includes(`${messageNeedle}\`,{allowed:!0}`)) {
-    return currentSource;
-  }
-
   const dispatchPattern =
     /((?:[A-Za-z_$][\w$]*=)?[A-Za-z_$][\w$]*\(`3903563814`\)[\s\S]{0,1800}?dispatchMessage\(`browser-use-non-local-sites-allowed-changed`,\{allowed:)([A-Za-z_$][\w$]*)(\}\))/g;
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: scripts/patches/webview-assets.js [hunk 1]
```diff
@@ -212,6 +212,87 @@ function applyLinuxAppSunsetPatch(currentSource) {
...
+
+  return currentSource;
+}
+
+function applyLinuxBrowserUseNonLocalNavigationPatch(currentSource) {
+  const messageNeedle = "browser-use-non-local-sites-allowed-changed";
+  const statsigNeedle = "3903563814";
+  let changed = false;
+
+  if (currentSource.includes(`${messageNeedle}\`,{allowed:!0}`)) {
```

---

This file-level idempotency check can skip eligible occurrences: if a bundle contains one already-patched `browser-use-non-local-sites-allowed-changed` dispatch and another still using the rollout variable, this returns before the global replacement below scans the second one. The ASAR patch helpers in this repo need to process every eligible occurrence (or assert the needle is unique), because upstream bundles sometimes duplicate IPC paths. Since the regex below won't match `allowed:!0` anyway, letting it run and skip only non-matching occurrences would preserve idempotency without missing duplicates.

avifenesh (commented):

---

Found one ASAR patching invariant issue in the new Browser Use navigation helper.

ilysenko (approved):
Reviewed current head a92b522. The previous duplicate-dispatch concern is fixed, CI is green, local focused and full patcher tests pass, and the patch applies to the current local webview assets.

PR #416, opened 2026-06-05
[codex] Preserve enabled local features during wrapper updates

avifenesh:
## Summary
- stage enabled `linux-features/local/<id>` directories into fetched wrapper sources before wrapper rebuilds
- apply the same preservation path to packaged rebuilds and user-local helper managed checkouts
- keep upstream top-level feature ids authoritative, so a future built-in feature is not duplicated as local

## Root Cause
Wrapper updates rebuild from a freshly fetched wrapper source. If the current install had an enabled local Linux feature, the saved feature config still referenced that id, but the fetched source did not contain `linux-features/local/<id>`. `scripts/lib/linux-features.js` then rejected the enabled id and the wrapper update stayed pending.

## Validation
- `cargo fmt --all --check`
- `bash -n contrib/user-local-install/files/.local/lib/codex-desktop-linux/common.sh tests/scripts_smoke.sh`
- `git diff --check`
- `cargo test -p codex-update-manager wrapper_apply`

### DIFF
--- a/contrib/user-local-install/files/.local/lib/codex-desktop-linux/common.sh
+++ b/contrib/user-local-install/files/.local/lib/codex-desktop-linux/common.sh
@@ -373,6 +373,57 @@ apply_source_overlay() {
     done < <(source_repo_overlay_remove_paths "$base_ref")
 }
 
+copy_enabled_local_features() {
+    local config_path source_local_root target_local_root feature_id source_dir target_dir
+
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed current head 9a9c0bade0388082d528cd7df3b29980dc2cf29a. No blockers found. Verified GitHub checks are green and local focused updater/user-local tests pass, including the full scripts smoke suite.

PR #417, opened 2026-06-05
fix(read-aloud): repair current settings aliases

joshyorko:
## Summary
- resolve Read Aloud settings page aliases from current webview imports
- stop rendering the page through stale minified aliases after upstream bundle drift
- add regression coverage for the current alias shape and stale-page upgrade path

## Why
The upstream webview bundle changed the minified import aliases used by the Read Aloud settings page. The old patch could keep wiring the generated page through stale aliases, which breaks the dedicated Read Aloud settings view after bundle drift.

## User-visible behavior
The Read Aloud settings page should render through the current settings components again instead of showing a blank/error page.

## Source-of-truth files

## Validation

## Limitations, risks, and follow-up
- No known distro or package-format limitations; this only affects the optional Read Aloud feature patching path.
- No generated artifacts were edited.
- No follow-up work is intentionally left out of scope.


ilysenko (approved):
Reviewed current head 115e154e366e0163d90ce7974773f1599d6be5ba. No blockers found; focused read-aloud tests pass locally.

PR #418, opened 2026-06-05
[codex] Keep Linux Chrome plugin available

colonelpanic8:
## Summary

- Keep the bundled Chrome plugin available on Linux even when upstream `externalBrowserUseAllowed` is false.
- Preserve upstream Chrome availability gating on non-Linux platforms.
- Add regression coverage for the auto-installed Chrome gate and idempotent Linux-enabled gates.

## Root Cause

The Linux launcher staged the Chrome plugin resources and native host, but the app's bundled-plugin marketplace resolver still filtered Chrome out when upstream `externalBrowserUseAllowed` was false. That left only `computer-use` in the runtime marketplace, causing Chrome to be uninstalled and its native host manifest to be removed.

## Validation

- `bash -n install.sh scripts/lib/*.sh launcher/start.sh.template scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh scripts/build-appimage.sh`
- Locally verified the patched gate against the installed ASAR shape before rebasing.
- Previously built `nix build .#codex-desktop-computer-use-ui-remote-mobile-control` successfully with the patch.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -2820,7 +2820,7 @@ test("auto-installs the current Chrome plugin gate shape", () => {
 
   assert.match(
     patched,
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed current head c8aa87697e2cc74c4b8ed6ab9c9295d9a940355d; checks pass.

PR #419, opened 2026-06-05
[codex] Sync embedded computer-use-linux v0.2.5

avifenesh:
## Summary
- sync embedded `codex-computer-use-linux` to the standalone `computer-use-linux` v0.2.5 screenshot payload behavior
- add bounded PNG/JPEG screenshot payload controls and coordinate metadata to `get_app_state` / `screenshot`
- opt JavaScript GitHub Actions into Node 24 runtime while keeping existing job Node 24 setup

Tracks agent-sh/computer-use-linux#19.

## Validation
- `cargo fmt --check -p codex-computer-use-linux`
- `cargo check -p codex-computer-use-linux --all-targets`
- `node --test linux-features/agent-workspace/test.js`
- `cargo build -p codex-computer-use-linux`
- `node /home/avifenesh/projects/computer-use-linux/scripts/zod-check/check.mjs --command target/debug/codex-computer-use-linux`
- `git diff --check`

### DIFF
--- a/computer-use-linux/src/server.rs
+++ b/computer-use-linux/src/server.rs
@@ -463,24 +463,54 @@ impl ComputerUseLinux {
         )
     )]
     async fn click(&self, Parameters(mut params): Parameters<ClickParams>) -> Json<ActionOutput> {
-        let received = Some(serde_json::json!(params));
+        let received = Some(serde_json::json!(params.clone()));
         // Raise the target window first (if specified) so the click lands on the
         // intended app rather than whatever is stacked on top at that pixel.
-        if let Some(target) = params.window_target() {
-            let _ = focus_window_target(&target).await;
+        let window_target = params.window_target();
+        if params.relative == Some(true) && window_target.is_none() {
+            return Json(ActionOutput {
+                ok: false,
+                implemented: true,
+                action: "click".to_string(),
+                message: "Relative coordinate clicks require a window target.".to_string(),
+                received,
+            });
+        }
+        if let Some(target) = window_target {
+            let focus = match self.focus_target_for_input(&target).await {
+                Ok(focus) => focus,
+                Err(message) => {
+                    return Json(ActionOutput {
+                        ok: false,
+                        implemented: true,
+                        action: "click".to_string(),
+                        message,
+                        received,
+                    });
+                }
+            };
             tokio::time::sleep(std::time::Duration::from_millis(120)).await;
             // Window-relative coordinates: translate by the window's top-left so
             // the agent can click the pixel it saw in a window-cropped screenshot.
             if params.relative == Some(true) {
-                if let (Some(rx), Some(ry)) = (params.x, params.y) {
-                    if let Ok(windows) = list_windows().await {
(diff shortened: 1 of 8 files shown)


avifenesh:
@ilysenko @Leay15 I'm considering to extract comp use out and adding it directly to the mcp, instead of correlating between the two on any change. Thoughts?

ilysenko:
One suggestion before merging: can we make the new window-targeted click path fail closed where we already can verify the target?

`type_text` / `press_key` already use `focus_target_for_input(...)` and avoid sending input when focus verification fails. For `click`, the new target focus result is ignored, and with `relative=true` list/resolve/bounds failures can leave the original `x/y` to be sent as absolute desktop coordinates.

I think `click` should reuse the same target verification path when a window target is provided, and `relative=true` should require resolved window bounds before sending input.

The screenshot payload changes look fine to me; this is mostly about input safety.

ilysenko:
@avifenesh 
Decision is yours, of course. The way I see it: extracting it is possible, but it may create more maintenance than the current embedded sync.

With the current layout, the backend is in the workspace, CI builds it, Nix sees it, native packages include it, and updater rebuilds use the same source.

If it moves out, we need pinned versions/checksums, reproducible Nix builds, CI coverage for the exact shipped backend, native package integration, updater rebuild consistency, and no accidental dependency on a moving external release.

So the main requirement would be: keep it fully pinned and reproducible. Otherwise the embedded workspace copy is fine as-is.

computer-use-linux/src/server.rs
```diff
@@ -428,16 +454,36 @@ impl ComputerUseLinux {
...
             open_world_hint = true
         )
     )]
-    async fn click(&self, Parameters(params): Parameters<ClickParams>) -> Json<ActionOutput> {
+    async fn click(&self, Parameters(mut params): Parameters<ClickParams>) -> Json<ActionOutput> {
         let received = Some(serde_json::json!(params));
+        // Raise the target window first (if specified) so the click lands on the
+        // intended app rather than whatever is stacked on top at that pixel.
+        if let Some(target) = params.window_target() {
+            let _ = focus_window_target(&target).await;
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

Targeted clicks should fail closed here. `focus_window_target` can return an error or a focus result that does not satisfy the requested target (the keyboard paths handle this via `focus_target_for_input`/`focus_satisfies_target`), but this call discards that result and still proceeds to send the click. In the same block, `relative=true` bounds resolution is best-effort, so if listing/resolution/bounds are unavailable the original `x`/`y` are sent as absolute desktop coordinates. For this destructive tool that can click the currently focused or occluding window instead of the requested target; please require verified focus, and require resolved bounds before translating relative clicks.

avifenesh (commented):

---

I found one input-safety issue in the new targeted click path.

ilysenko (approved):
Reviewed current head 8676f8d128a36e998320469ef5b7a26a30731bd8; targeted click now fails closed and checks pass.

PR #421, opened 2026-06-06
Fix Linux titlebar tooltip overlap

lw33:
## Summary
- Reserve Linux window-control space in the webview safe-area map so header chrome does not render under the native minimize/maximize/close controls.
- Increase Linux tooltip collision padding near the titlebar so floating tooltips avoid the native controls.
- Remove the native `title` attribute from the thread side-panel toggle while keeping `aria-label` and the Codex tooltip, preventing duplicate tooltips.

## Why
On Linux, the top-right thread side-panel toggle could show both the Codex tooltip and the browser/Electron native title tooltip. Tooltips near the native titlebar controls could also overlap the minimize/close buttons because Linux titlebar geometry was not reserved consistently.

## Source Of Truth Files Edited
- `scripts/patches/webview-assets.js`
- `scripts/patches/core/all-linux/webview/theme-and-sunset/patch.js`

## Validation
- `node --test scripts/patch-linux-window-ui.test.js` passes: 187/187.
- Dry-ran the new thread side-panel patch against the generated `thread-app-shell-chrome` chunk and confirmed it removes the native `title` while preserving `aria-label` and `tooltipContent`.

## Environment / Limitations
- Validated against the currently generated local webview assets in this checkout.
- Did not rebuild native packages; this change is limited to ASAR/webview patch descriptors and patcher tests.

## Risks / Follow-up
- These patches target minified upstream webview chunks and remain fail-soft. If upstream chunk structure drifts, patch reports will warn and the regression tests should be updated with the new bundle shape.

### DIFF
--- a/scripts/patches/webview-assets.js
+++ b/scripts/patches/webview-assets.js
@@ -7,6 +7,8 @@ const path = require("node:path");
 // They stay fail-soft because upstream chunk names and minified symbols drift.
 const LINUX_SAFE_MONOSPACE_FONT_STACK =
   "\"Noto Sans Mono\", \"DejaVu Sans Mono\", \"Liberation Mono\", \"Ubuntu Mono\", ui-monospace, \"SFMono-Regular\", \"SF Mono\", Menlo, Consolas, monospace";
+const LINUX_TOOLTIP_COLLISION_PADDING_TOP = 44;
+const LINUX_WINDOW_CONTROLS_SAFE_AREA_RIGHT = 138;
 
 function applyLinuxSafeMonospaceFontStackPatch(currentSource) {
   const safeLinuxMonoFontPattern =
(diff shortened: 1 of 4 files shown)


scripts/patches/webview-assets.js
```diff
@@ -192,6 +194,74 @@ function applyLinuxOpaqueWindowsDefaultPatch(currentSource) {
...
 
+function applyLinuxWindowControlsSafeAreaPatch(currentSource) {
+  const currentInset = `applicationMenu:Object.freeze({left:0,right:${LINUX_WINDOW_CONTROLS_SAFE_AREA_RIGHT}})`;
+  if (currentSource.includes(currentInset)) {
+    return currentSource;
+  }
+
+  const defaultInset = "applicationMenu:Object.freeze({left:0,right:0})";
+  if (currentSource.includes(defaultInset)) {
+    return currentSource.replace(defaultInset, currentInset);
```

---

This new webview/ASAR patch uses `String.prototype.replace`, so it only updates the first matching safe-area literal. The repo's ASAR patching invariant (P1) requires patch helpers to process every eligible occurrence, or to explicitly assert that the needle is unique, because upstream minified bundles can contain duplicate literals/call sites. If `applicationMenu:Object.freeze({left:0,right:0})` appears more than once in this chunk after upstream drift, the remaining maps will keep `right:0` and titlebar chrome can still render under the Linux window controls. Please switch this to match-index/all-occurrence replacement or add a uniqueness assertion for this needle.

avifenesh (commented):

---

Found one patching invariant issue in the new webview asset helper.

ilysenko (approved):
Reviewed current head. The earlier patching invariant issue is fixed by patching all matching occurrences and covering mixed already-patched/default cases in tests.

PR #423, opened 2026-06-06
fix(chrome): enable external browser use gate on Linux

losese:
## Summary

- Add a Linux webview patch for the `browser_use_external` availability gate.
- Allow the Google Chrome Computer Use row to be available on Linux without the upstream external Browser Use rollout flag.
- Keep existing `chrome-extension` window-type behavior unchanged.
- Add patcher tests for the current minified `use-is-plugins-enabled` shape and idempotency.

## Testing

- `codex-app/resources/node-runtime/bin/node --test scripts/patch-linux-window-ui.test.js`
- `bash -n install.sh scripts/lib/*.sh launcher/start.sh.template scripts/build-*.sh`
- `git diff --check`


ilysenko (approved):
Reviewed current head. The patch is scoped to the external Browser Use availability gate, keeps chrome-extension behavior intact, and patcher tests plus upstream/package CI are green.

PR #424, opened 2026-06-06
[codex] Sync embedded computer-use-linux v0.2.6

avifenesh:
## Summary
- Sync the embedded `codex-computer-use-linux` to the standalone `computer-use-linux` **v0.2.6** screenshot fix.
- Adds `gnome-screenshot` as a third capture fallback (chain: GNOME Shell → portal → gnome-screenshot) for background / systemd contexts where the DBus paths are denied.
- Bumps the embedded crate to `0.2.6-linux-alpha1`.

## Why
The `screenshot` tool failed from background processes (systemd user services, non-interactive parent shells) on GNOME Wayland. Both existing backends are structurally unavailable there:
- GNOME Shell's `org.gnome.Shell.Screenshot` rejects callers that don't own an allowlisted bus name (`DBusSenderChecker`) → `ACCESS_DENIED`.
- The XDG portal cancels the non-interactive request (response code 2) when there's no foreground window to parent it to.

`gnome-screenshot` claims an allowlisted bus name and works regardless of session context, so it's the final fallback. Spawned via `tokio::process` and bounded by a 20s timeout, so a hung capture degrades to a clear error instead of blocking the tool.

## Changes
- `gnome-screenshot` fallback in `capture_screenshot_raw` (3-rung chain).
- `CODEX_COMPUTER_USE_SCREENSHOT_BACKEND` env var to force a single backend (`gnome-shell` | `portal` | `gnome-screenshot`), skipping the chain — keeps the embedded Codex naming convention.
- `doctor` probes `gnome-screenshot` and lists it under `capabilities.screenshot` when present.
- Enables the tokio `process` feature.

## Not included
The upstream mimalloc / release-profile change is **intentionally not ported** — switching the global allocator is a workspace-wide decision for this repo's maintainer, not something to fold in via a screenshot sync.

## Test
- `cargo test -p codex-computer-use-linux` → 111 pass (incl. backend-parse + env-override tests); clippy + fmt clean.
- Upstream fix verified end-to-end via an Xvfb A/B counterfactual: the pre-fix binary errors out with both DBus failures; the post-fix binary falls through and returns a real `source: gnome-screenshot` capture.

Upstream reference: agent-sh/computer-use-linux v0.2.6 (fixes the background-screenshot issue reported against GNOME Wayland).


computer-use-linux/src/screenshot.rs
```diff
@@ -607,6 +730,43 @@ mod tests {
...
+            ScreenshotBackend::parse("GNOME_SCREENSHOT"),
+            Some(ScreenshotBackend::GnomeScreenshot)
+        );
+        assert_eq!(ScreenshotBackend::parse("nonsense"), None);
+    }
+
+    #[test]
+    fn forced_backend_reads_env_override() {
+        // Only this test touches SCREENSHOT_BACKEND_ENV, so no cross-test race.
+        std::env::set_var(SCREENSHOT_BACKEND_ENV, "gnome-screenshot");
```
avifenesh, line 753: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This test mutates process-global environment state without a Drop guard. The repo’s Rust test invariant for `computer-use-linux` env-var mutation is to use an RAII guard that restores the previous value on all exits; otherwise a panic before the final `remove_var` leaves `CODEX_COMPUTER_USE_SCREENSHOT_BACKEND` set for the rest of the test process and can make later screenshot/backend tests order-dependent. Please wrap this in an env guard that restores the prior value in `Drop` instead of manually setting/removing it inline.

avifenesh (commented):

---

Found one test-isolation issue that matches the repo's env-mutation guard pattern.

computer-use-linux/src/screenshot.rs
avifenesh: Checked the claimed invariant — it doesn't hold for this repo. There's no env-guard / RAII-Drop pattern anywhere in `computer-use-linux` (no `EnvGuard`/`impl Drop` for env restoration); the only `env::set_var` calls are production hydration code in `diagnostics.rs`, not a test convention. And `CODEX_COMPUTER_USE_SCREENSHOT_BACKEND` is unique to this test — no other test reads it — so a panic leaving it set can't make later screenshot/backend tests order-dependent. The inline set/remove matches the same test as it landed upstream. Resolving as a false positive.

ilysenko (approved):
Reviewed current head. The screenshot fallback chain remains fail-soft, gnome-screenshot is only a final fallback or explicit override, diagnostics/tests are aligned, and local cargo test/fmt/clippy passed.

PR #430, opened 2026-06-08
Fix Linux About dialog icon handling

lw33:
## What changed

On Linux the About dialog showed no icon, or could crash with an uncaught promise rejection. This PR wires up a proper bundled icon path for Linux and adds null-safety guards on the icon return values so the dialog renders correctly regardless of which code path is taken.

## Why it changed

The upstream About-dialog helper (`bZ` in the minified main bundle) unconditionally calls `app.getFileIcon` and then calls `.isEmpty()` on the result without guarding against `null`. On Linux:

1. `app.getFileIcon` can reject on some desktop environments, leaving the icon promise unresolved or throwing.
2. The `windowIcon` field propagated downstream without a null check, so `d.windowIcon.isEmpty()` threw when the icon was `null`.
3. No bundled Linux asset was ever supplied to the dialog — the function only tried macOS thumbnail (`Fw`) or the system file-icon API, both of which are no-ops or failures on Linux.

The patch (`applyLinuxAboutDialogPatch`) addresses all three:
- On Linux it short-circuits the macOS-only `Fw` call to `null`.
- It replaces the `app.getFileIcon` call with `nativeImage.createFromPath(iconPathExpression)` pointing at the bundled `content/webview/assets/app-test.png` asset, wrapped in an IIFE that returns `null` on an empty image instead of propagating an empty `NativeImage`.
- It guards `i.isEmpty()` → `i==null||i.isEmpty()` and `d.windowIcon.isEmpty()` → `d.windowIcon==null||d.windowIcon.isEmpty()` and `windowIcon:i}` → `windowIcon:i??null}` throughout the downstream spread.
- `getFileIcon` on non-Linux paths gets `.catch(()=>null)` so a rejection no longer propagates.

The patch is `ciPolicy: "optional"` (fail-soft): if the minified bundle changes shape the dialog silently falls back to its previous behavior rather than breaking startup.

## Source-of-truth files changed

- `scripts/patches/main-process.js` — new `applyLinuxAboutDialogPatch` function
- `scripts/patches/core/all-linux/main-process/window-shell/patch.js` — registers the patch descriptor (`id: "linux-about-dialog"`, phase `main-bundle`, order 55)
- `scripts/patch-linux-window-ui.js` — imports and re-exports `applyLinuxAboutDialogPatch`
- `scripts/patch-linux-window-ui.test.js` — adds the patch to the uniqueness registry test and adds a focused unit test covering the icon-path substitution, null-safe guards, and idempotency

## Validation

```
node scripts/patch-linux-window-ui.test.js
```

All tests pass, including the new `"makes About dialog prefer the bundled Linux icon asset"` case. The test covers:
- `nativeImage.createFromPath(...)` substitution present in patched output
- `process.platform===\`linux\`?null:...` short-circuit for the macOS thumbnail call
- `windowIcon==null||d.windowIcon.isEmpty()?{}:{icon:d.windowIcon}` null guard
- `i==null||i.isEmpty()?null:i.resize(` null guard
- `windowIcon:i??null` coalescing
- Idempotency (patch applied twice produces identical output)
- Output is valid JavaScript (`new Function(patched)` does not throw)

## Limitations / follow-up

- The icon asset path (`content/webview/assets/app-test.png`) is resolved at patch time from `context.iconPathExpression`; if the upstream DMG renames the asset this patch will silently fall back. A follow-up could add a CI assertion that the asset path resolves.
- Patch is `optional` policy; if the minified signature drifts significantly only the null-safety guards (string-replace based) will still apply.

### DIFF
--- a/scripts/patch-linux-window-ui.js
+++ b/scripts/patch-linux-window-ui.js
@@ -64,6 +64,7 @@ const {
 } = require("./patches/launch-actions.js");
 const {
   applyBrowserUseNodeReplApprovalPatch,
+  applyLinuxAboutDialogPatch,
   applyLinuxBuildInfoTrayPatch,
   applyLinuxChromeExtensionStatusPatch,
   applyLinuxExplicitIpcQuitPatch,
@@ -171,6 +172,7 @@ module.exports = {
   applyLinuxDesktopSettingsIndexPatch,
   applyLinuxDesktopSettingsSectionsPatch,
   applyLinuxDesktopSettingsSharedPatch,
+  applyLinuxAboutDialogPatch,
   applyLinuxAppSunsetPatch,
   applyLinuxAppUpdaterBridgePatch,
   applyLinuxAppUpdaterMenuPatch,
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -27,6 +27,7 @@ const {
   applyLinuxChromePluginAutoInstallPatch,
   applyLinuxAppUpdaterBridgePatch,
   applyLinuxAppUpdaterMenuPatch,
+  applyLinuxAboutDialogPatch,
   applyLinuxBuildInfoTrayPatch,
   applyLinuxExplicitIpcQuitPatch,
   applyLinuxExplicitQuitPromptBypassPatch,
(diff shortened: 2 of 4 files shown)


ilysenko (approved):
Reviewed locally and against the current upstream DMG patch report. CI is green and no blockers were found.

PR #431, opened 2026-06-08
Fix Linux About dialog icon handling

lw33:
## Summary
This PR fixes the Linux `Help -> About` flow so the About dialog reliably opens and shows the app icon from the packaged Linux asset.

## User-visible change
On Linux, clicking `Help -> About` now opens the About dialog reliably and uses the bundled app icon instead of depending on desktop-specific file icon resolution.

## Root cause
The upstream About dialog path tried to build its icon state from `app.getFileIcon(process.execPath)`. On Linux, that call can fail or return an unusable result because `process.execPath` points at the launcher/Electron executable rather than a stable desktop-integrated app icon source. The About dialog code swallowed that failure, so users saw the menu item do nothing.

This fix makes the Linux wrapper patch prefer the packaged `webview/assets/app-*.png` icon asset for the About dialog, and keeps the remaining icon handling null-safe so the dialog still opens even if icon loading falls back.

## Source-of-truth files changed
- `scripts/patches/core/all-linux/main-process/window-shell/patch.js`
- `scripts/patch-linux-window-ui.js`

## Validation
- `node --test /home/lw/workspace/src/codex-desktop-linux/scripts/patch-linux-window-ui.test.js`

### DIFF
--- a/scripts/patch-linux-window-ui.js
+++ b/scripts/patch-linux-window-ui.js
@@ -64,6 +64,7 @@ const {
 } = require("./patches/launch-actions.js");
 const {
   applyBrowserUseNodeReplApprovalPatch,
+  applyLinuxAboutDialogPatch,
   applyLinuxBuildInfoTrayPatch,
   applyLinuxChromeExtensionStatusPatch,
   applyLinuxExplicitIpcQuitPatch,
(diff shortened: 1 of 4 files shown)


ilysenko:
Closing this as a duplicate of #430. The same Linux About dialog icon handling patch was already merged into main in #430, so there is nothing left to merge here.

PR #436, opened 2026-06-09
[codex] Show Linux build info in settings

colonelpanic8:
## Summary
- Render Linux build metadata inside the Linux Desktop settings page.
- Show the full source commit as selectable text and add Copy commit / Open commit actions.
- Reuse the same build-info helper for tray, Help menu, and renderer requests.

## Validation
- node --test scripts/patch-linux-window-ui.test.js

PR #438, opened 2026-06-09
fix(linux): avoid build-info helper variable shadowing

colonelpanic8:
## Summary
- Follow-up to #436. The injected Linux build-info main-process helpers used short local names (`e`, `t`, `n`, `a`, `s`, …) which, after minification, can collide with the bundle's own module-scope bindings (e.g. the `electron` / `node:path` aliases) and produce broken output like `let a=await a.dialog?.showMessageBox`.
- Rename every injected helper local to a `__codexBuildInfo*`-prefixed name so it can't shadow surrounding bundle bindings.
- Make request-handler insertion robust: insert directly before the existing `get-/set-global-state` handler key instead of at the enclosing `{`, and repair any previously mis-nested handler entry.

## Validation
- `node --test scripts/patch-linux-window-ui.test.js` — 201 pass, 0 fail (includes two new regression tests: no-shadow guarantee + correct handler-table insertion)

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -1738,6 +1738,30 @@ test("adds Linux build information request handlers for renderer settings", () =
   );
 });
 
+test("Linux build information helper locals do not shadow minified module bindings", () => {
+  const source =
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed against current main after #437. The change is scoped to the Linux build-info patcher, fixes helper variable shadowing/request-handler placement, and the merged result passes the patcher tests plus upstream build validation.

PR #439, opened 2026-06-09
Hide Linux titlebar menu chrome

moteea:
## Summary

- hide the Linux primary-window titlebar overlay by keeping Electron's `titleBarOverlay` object but making it transparent and 1px tall
- strengthen Linux menu suppression by calling both `setMenuBarVisibility(false)` and `removeMenu?.()` next to the existing Windows `removeMenu()` path
- map Linux webview window chrome back to `native` instead of `application-menu`, so the renderer no longer reserves application-menu/window-control chrome space

## Why

Recent Codex Desktop builds can show a Linux app/menu decoration row or a visible top-right Electron titlebar overlay. Removing the overlay object entirely, or switching the window to `frame: false`, can crash the Linux build. Keeping the overlay API present while making it visually transparent avoids that crash path while removing the unwanted chrome.

## Validation

- `tests/scripts_smoke.sh`

This behavior was also verified manually through a Nix/Home Manager override against Codex Desktop `26.602.71036` on Linux.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -1302,7 +1302,7 @@ test("patches remaining explicit quit handlers when another copy is already patc
   );
 });
 
-test("uses the frameless native Codex titlebar for primary Linux windows", () => {
+test("uses a hidden native titlebar overlay for primary Linux windows", () => {
   const source = [
     "function A2(e){return e===`avatarOverlay`}",
     "function I2({platform:e,appearance:t,opaqueWindowsEnabled:n,prefersDarkColors:r}){return n&&!A2(t)&&(e===`darwin`||e===`win32`)?{backgroundColor:r?a2:o2,backgroundMaterial:e===`win32`?`none`:null}:e===`linux`&&!A2(t)?{backgroundColor:r?a2:o2,backgroundMaterial:null}:{backgroundColor:i2,backgroundMaterial:null}}",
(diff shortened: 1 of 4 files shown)


ilysenko (commented):
Current upstream build applies correctly for me: the generated bundle has the transparent 1px Linux titleBarOverlay, Linux removeMenu?.(), and the webview chrome mapping changes Linux back to native.

One thing to tighten before this leaves draft: applyLinuxWindowControlsSafeAreaPatch now has currentInset equal to the default inset because the right safe area is 0. That makes the inset branch a no-op, and the final currentInset check can make future upstream drift look already-applied if upstream keeps applicationMenu:Object.freeze({left:0,right:0}) but changes/removes the chrome-mapping snippets. Since this PR's functional fix depends on those mapping snippets, please track whether the Linux chrome mapping or browser gate was actually found/already in the desired form, and warn if the safe-area bundle is present but neither mapping is recognizable. That keeps future upstream drift visible instead of silently shipping the old Linux application-menu layout.

ilysenko:
Thanks a lot for working on this and for the careful investigation around the Linux titlebar/menu chrome behavior. We ended up moving this into an opt-in Linux feature instead of applying it as a default core patch: the `frameless-titlebar` feature was merged in #472. That keeps the default build unchanged while still letting users who need compositor-managed/window-manager decorations enable it explicitly.\n\nSince #472 supersedes this draft PR, I'm going to close it. Appreciate the work and validation you put into this.

PR #442, opened 2026-06-10
Fix upstream drift for latest Codex Desktop release

ilysenko:
## Summary

Fixes latest upstream Codex Desktop drift for the Linux repackaging flow. The patch registry and Chrome plugin staging now handle the current upstream bundle shapes without required patch failures or stale `missing patch target` noise.

## Scope

- Updated ASAR patch helpers for current upstream minifier and bundle drift.
- Made optional/background-avatar-adjacent patch paths idempotent/fail-soft where upstream already has the Linux-compatible shape.
- Fixed current webview drift for automation schedules, local environment action drafts, settings sections, tooltips, updater bridge, subagent metadata, About dialog, native titlebar, and i18n gate detection.
- Updated Chrome plugin browser-client patching for the current alias family so Linux profile roots, profile metadata lookup, instance matching, and backend ranking are patched again.
- Added regression coverage for the new upstream shapes and focused Chrome plugin smoke coverage.

## Progress

- [x] Open tracking PR before starting the investigation.
- [x] Rebuild from a fresh upstream download.
- [x] Identify the failing patch/test surface.
- [x] Implement and validate the fix.
- [x] Push final drift fixes to this PR.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` — 218/218 passed.
- `node --test linux-features/codex-wrapper-updater/test.js` — 10/10 passed.
- Focused `test_chrome_browser_client_profile_root_variants` smoke check passed.
- `node --check scripts/lib/patch-chrome-plugin.js` passed.
- `bash -n tests/scripts_smoke.sh` passed.
- Fresh install from `/tmp/codex-upstream-ci/Codex.dmg` completed successfully into `/tmp/codex-driftfix-install-bIJjYS/codex-app`.
- Patch report validation passed with `bad-count=0` and `Required patch validation passed for profile upstream-build`.

## Notes

The only warnings left in the fresh install output are the known 7z DMG extraction symlink warnings; the app bundle is found and the build continues normally.

PR #443, opened 2026-06-10
Fix Linux opaque window patch drift

Igzela:
## Summary
- support the current `opaqueWindowSurfaceEnabled` BrowserWindow background helper shape so Linux windows stay opaque
- avoid minified Electron alias shadowing in the Linux native titlebar patch by using a stable `require(`electron`).nativeTheme` expression
- add regression coverage for the current surface-window helper shape and titlebar alias collision

## Testing


ilysenko:
Thanks for the fix. This PR is now superseded by the changes already merged into `main`:

- `opaqueWindowSurfaceEnabled` background helper support landed in #442.
- The native titlebar alias-shadowing startup crash was fixed in #444 by moving the Linux overlay calculation into a module-scoped helper instead of inlining `nativeTheme` inside the upstream titlebar function.

#444 passed the upstream DMG build and package/test checks, and the rebuilt local app was manually verified. Closing this as redundant with the current `main`.

PR #446, opened 2026-06-10
Fix Linux window and tray patch drift

Igzela:
## Summary
- support the current `opaqueWindowSurfaceEnabled` BrowserWindow background helper shape so Linux windows stay opaque
- avoid minified Electron alias shadowing in the Linux native titlebar patch by using a stable `require(`electron`).nativeTheme` expression
- avoid Linux tray icon fallback alias shadowing, and migrate already patched fallback blocks to safe local variable names
- add regression coverage for the current surface-window helper shape, titlebar alias collision, and tray icon alias collision

## Testing

Note: PR #443 was closed before the tray follow-up commit was added, so this supersedes it.


ilysenko:
Thanks for the follow-up. I reviewed this against current `main`, and this PR should not be merged as-is.

Why:
- The branch is now conflicted with `main` (`mergeStateStatus: DIRTY`) after #444.
- GitHub reports no checks for this head branch, so there is no green CI state to merge.
- The titlebar portion is superseded by #444, which fixed the alias-shadowing crash by using a module-scoped `codexLinuxTitleBarOverlay(...)` helper. This PR reintroduces an alternate inline `require(`electron`).nativeTheme` approach and also returns early once that primary titlebar shape is found, which would skip the rest of the titlebar sync patching path.
- The fresh-install tray alias-shadowing protection is already present on current `main`: the Linux tray fallback now uses collision-proof variables (`__codexLinuxTrayIcon`, `__codexLinuxAppIcon`, and `__codexLinuxUpstreamTrayIcon`) with regression coverage.

Closing this as superseded by the current `main`. If there is still a separate need to migrate old already-patched generated bundles, that should be proposed as a narrow follow-up PR without the superseded titlebar/background changes.

PR #447, opened 2026-06-10
Fix "Oops" crash on Settings → Codex mobile page (missing Sparkle autodownload stub)

jurgenmahn:
## Issue

Opening **Settings → Codex mobile** (and on app start) triggered the generic **"Oops, an error has occurred"** screen in the webview.

Root cause: the Linux update bridge swaps the upstream macOS Sparkle manager for `codexLinuxPackageUpdateBridge.manager`, but that object never implemented `setAutomaticBackgroundDownloadsEnabled`. The renderer still calls `s.setAutomaticBackgroundDownloadsEnabled(e)` via the `electron-sparkle-autodownload-changed` IPC when the updater preferences are read, so the main-process handler threw:

```
TypeError: s.setAutomaticBackgroundDownloadsEnabled is not a function
```

That rejected the `codex_desktop:message-from-view` invoke, which the renderer surfaced as the "Oops" error boundary.

## Fix

Add a no-op `setAutomaticBackgroundDownloadsEnabled` to the Linux manager in `scripts/lib/linux-update-bridge-patch.js`.

Background auto-download is a Sparkle-only concept that does not apply to the Linux package updater, so silently absorbing the toggle is the correct behavior and keeps the renderer IPC contract intact. The stub is added in the bridge **source** patch, so it survives auto-updater rebuilds.

## Validation

- Ubuntu 26.04, `.deb` path: rebuilt (`make build-app` + `make package`), reinstalled, and confirmed the `TypeError` no longer appears in `~/.cache/codex-desktop/launcher.log` and the Settings → Codex mobile view renders.
- Added a regression assertion in `scripts/patch-linux-window-ui.test.js`; full suite passes (218/218).


ilysenko (approved):
Reviewed the updated head f89d663. The Linux updater bridge now implements the Sparkle autodownload no-op in both bootstrap and class-style updater paths, and the focused patcher suite passes locally.

PR #449, opened 2026-06-11
fix(contrib): resolve the cached Codex.dmg where install.sh actually caches it

ryan-mt:
## What changed

The user-local helpers (`contrib/user-local-install`) looked for the cached DMG at `$OPT_ROOT/Codex.dmg`, but nothing ever writes the DMG there: `install.sh` caches it next to itself in the build repo checkout (`CACHED_DMG_PATH="$SCRIPT_DIR/Codex.dmg"` in `scripts/lib/install-helpers.sh`), and `CODEX_INSTALL_ROOT` only relocates the generated app directory.

Consequences in the update flow (`codex-desktop-update`):

- After every successful rebuild, `extract_icon` ran `7z` on a nonexistent file, so `set -euo pipefail` aborted the script **before** `record_metadata`.
- The weekly `codex-desktop-update.timer` unit therefore failed on every run.
- Because metadata never refreshed, `codex-desktop-check-update` kept reporting a pending update forever.
- The `Refreshing cached Codex.dmg...` step deleted the wrong path, so it never actually forced a refresh.

Fix:

- Resolve the cache via the effective build repo dir (`cached_dmg_file()` on top of the existing `effective_repo_dir()`), so icon extraction and DMG metadata work against the DMG that install.sh actually cached.
- Treat icon extraction as best-effort at the `codex-desktop-update` call sites (`extract_icon || log ...`), so a missing/unusable DMG can no longer abort the script before `record_metadata`. The extraction is still *attempted* every time, preserving the behavior the smoke suite pins.
- Record `DMG_SHA256` / `DMG_SIZE` as `unavailable` when no cached DMG exists (same convention as `repo_head`); `codex-desktop-version` already prints these fields defensively.
- Delete the real cache plus its `.metadata` sidecar when a DMG refresh is forced.

## Source-of-truth files

- `contrib/user-local-install/files/.local/lib/codex-desktop-linux/common.sh`
- `contrib/user-local-install/files/.local/bin/codex-desktop-update`

## Validation

- `bash -n` on all contrib scripts — clean
- `bash tests/scripts_smoke.sh` on this branch in Ubuntu (WSL): **All script smoke tests passed**, including the user-local overlay/refresh suites and `test_user_local_install_from_update_defers_record_only_metadata`
- Functional repro in a sandbox `$HOME`: sourcing the fixed `common.sh`, `cached_dmg_file` resolves into the build repo dir; the tolerated `extract_icon` failure no longer aborts under `set -e`; `record_metadata` completes and writes `DMG_SHA256/DMG_SIZE=unavailable` when no DMG is cached

## Notes / limitations

- Not exercised against a real `~/.local` install with the systemd timer; the failure mechanics and fix were verified by sourcing the helpers in a sandboxed `$HOME` plus the smoke suite.

### DIFF
--- a/contrib/user-local-install/files/.local/bin/codex-desktop-update
+++ b/contrib/user-local-install/files/.local/bin/codex-desktop-update
@@ -37,7 +37,9 @@ log() {
 }
 
 if [ "$record_only" -eq 1 ]; then
-    extract_icon
+    # Icon extraction is best-effort: without a usable cached Codex.dmg it
+    # must not abort metadata recording under set -e.
+    extract_icon || log "Icon extraction skipped: no usable cached Codex.dmg."
     record_metadata
     log "Metadata refreshed."
     exit 0
@@ -109,7 +111,8 @@ fi
 
 if [ "$dmg_changed" -eq 1 ]; then
     log "Refreshing cached Codex.dmg..."
-    rm -f "$DMG_FILE"
+    dmg_file="$(cached_dmg_file)"
+    rm -f "$dmg_file" "$dmg_file.metadata"
 fi
 
 log "Rebuilding wrapper..."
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed. No blocking issues found; CI is green and the contrib user-local cache path fix is narrowly scoped.

PR #450, opened 2026-06-11
fix(updater): rollback block identity, zero-interval panic, atomic DMG cache

ryan-mt:
## What changed

Three correctness fixes in the Rust update manager, plus a patch version bump to 0.8.3.

1. **The rollback block could never match, so a rolled-back bad update was auto-reinstalled on the failed-check retry path.** Generated candidate versions are `{download-timestamp}+{dmg-sha8}` (`derive_candidate_version`). `installed_version_matches_candidate` strips everything after `+`/`-` and compares only the four timestamp segments — which are *download times* and always differ between the originally installed download and a later re-download of the same DMG. `Ordering::Equal` was structurally unreachable, making the `rollback_blocked_candidate_version` guard dead code. On the `Failed`-status retry path (`retrying_failed_update`, which deliberately bypasses both the headers-fingerprint guard and the `dmg_sha256` guard), any transient check error after a rollback caused the same bad DMG to be rebuilt under a fresh version and auto-reinstalled — exactly what the guard, its log line, and its error message exist to prevent. The new `candidate_matches_rolled_back_version` compares the `+sha8` DMG identity, falling back to the old comparison for versions without a hash suffix.

2. **`check_interval_hours = 0` in `config.toml` panicked the daemon.** `tokio::time::interval` asserts a non-zero period (``"`period` must be non-zero"``), and the user-editable config value was multiplied straight into it, killing the reconcile loop right after the initial check. The new `check_interval_period` clamps to a 1-hour minimum (and saturates the multiply).

3. **A failed re-download corrupted the cached DMG in place.** `download_dmg` truncated `downloads/Codex.dmg` via `File::create` *before* sending the GET, so a mid-stream failure left a partial file at the exact path that `state.artifact_paths.dmg_path` still references — and `wrapper_apply::cached_or_downloaded_dmg` reuses it via a bare `exists()` check, feeding a truncated DMG into wrapper rebuilds. Downloads now stream into `Codex.dmg.part` and rename into place on success; the sidecar is removed on failure.

## Source-of-truth files

- `updater/src/app.rs`, `updater/src/upstream.rs`
- `updater/Cargo.toml` (0.8.2 → 0.8.3, patch bump per the repo versioning rules), `Cargo.lock`

## Validation (Ubuntu/WSL)

- `cargo fmt -p codex-update-manager -- --check` — clean
- `cargo clippy -p codex-update-manager --all-targets -- -D warnings` — clean
- `cargo test -p codex-update-manager` — 169 passed; the 3 failures (`feature_picker::tests::{dont_ask_sentinel_writes_setting, selection_preserves_unknown_existing_feature_ids, selection_writes_feature_config_outside_wrapper_src}`) fail identically on `main` in my environment (env-specific, unrelated to this change)
- New tests: `rollback_block_matches_same_dmg_across_redownload_timestamps`, `check_interval_period_clamps_zero_hours`, `failed_download_preserves_existing_cached_dmg` (asserts a 500 response leaves the previously cached DMG intact and no `.part` behind); existing `downloads_dmg_and_hashes_contents` still passes

## Notes / limitations

- Persisted state is unchanged (`rollback_blocked_candidate_version` keeps its existing format); only the comparison is fixed, so no state migration is needed.

### DIFF
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -318,8 +318,7 @@ async fn run_daemon(
         error!(?error, "initial reconciliation failed");
     }
 
-    let mut check_interval =
-        time::interval(Duration::from_secs(config.check_interval_hours * 3600));
+    let mut check_interval = time::interval(check_interval_period(config.check_interval_hours));
     let mut reconcile_interval = time::interval(Duration::from_secs(RECONCILE_INTERVAL_SECONDS));
     check_interval.tick().await;
     reconcile_interval.tick().await;
@@ -835,7 +834,7 @@ async fn run_check_cycle(
             .rollback_blocked_candidate_version
             .as_deref()
             .is_some_and(|blocked| {
-                installed_version_matches_candidate(blocked, &downloaded.candidate_version)
+                candidate_matches_rolled_back_version(blocked, &downloaded.candidate_version)
             })
         {
             state.status = UpdateStatus::Idle;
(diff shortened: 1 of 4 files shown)


updater/src/app.rs
```diff
@@ -1219,6 +1218,33 @@ fn installed_version_matches_candidate(installed: &str, candidate: &str) -> bool
...
 
+/// `tokio::time::interval` panics on a zero period; clamp a hand-edited
+/// `check_interval_hours = 0` to the smallest supported cadence.
+fn check_interval_period(check_interval_hours: u64) -> Duration {
+    Duration::from_secs(check_interval_hours.max(1).saturating_mul(3600))
+}
+
+fn version_dmg_short_hash(version: &str) -> Option<&str> {
+    version
+        .split_once('+')
```
avifenesh, line 1229: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This only extracts hashes from raw `+<sha8>` candidate strings, but the rollback blocker is usually populated from `state.installed_version` (`rollback::trigger_rollback` uses that whenever it is known). For RPM, `installed_package_version()` returns `%{VERSION}-%{RELEASE}` and `build-rpm.sh` puts the hash in `Release`, so the blocked value looks like `2026.05.04.131500-abcdef12.fc43` and falls back to timestamp comparison. For pacman, `pacman -Q` returns `2026.05.04.131500+abcdef12-1`, so this helper compares `abcdef12-1` to the candidate's `abcdef12`. In both cases a re-download of the same DMG with a fresh timestamp still bypasses the rollback block. Please normalize the installed package-version forms (or store/compare the original candidate hash) before comparing.

avifenesh (commented):

---

Found one correctness gap in the rollback identity comparison for non-Debian package formats.

ilysenko:
Blocking issue: the rollback identity comparison still misses RPM and pacman installed-version forms. The new helper only extracts a clean +sha8 suffix, but rollback stores installed versions such as RPM VERSION-RELEASE (for example 2026.05.04.131500-abcdef12.fc43) and pacman pkgver-pkgrel (for example 2026.05.04.131500+abcdef12-1). In those cases the same rolled-back DMG can be rebuilt under a new candidate timestamp and auto-installed again. Please normalize deb/rpm/pacman installed-version forms or store the original candidate DMG hash separately, and add tests for RPM and pacman forms.

ilysenko:
Closing this for now. If you'd like to come back to it, please feel free to reopen.

PR #451, opened 2026-06-11
fix(patches): align Computer Use settings resolution and quit-guard insertion

ryan-mt:
## What changed

Two core ASAR patch fixes.

1. **`computer-use.js` resolved the settings identity inversely to the runtime that writes the flag.** `computerUseUiSettingsAppId` used `CODEX_APP_ID || CODEX_LINUX_APP_ID`, while every other resolver in the repo — including the injected `codexLinuxSettingsAppId()` runtime that *persists* `codex-linux-computer-use-ui-enabled`, plus `codexLinuxWrapAppId`, `codexLinuxReadAloudSettingsAppId`, `bootstrap-wizard.sh`, `apply-pending.sh`, and the updater's documented resolution order (`updater/src/config.rs`) — uses `CODEX_LINUX_APP_ID || CODEX_APP_ID`. With side-by-side app identities (e.g. `make build-dev-app`, where the repo's own tests set the two vars to different values) the build-time reader checked a different `settings.json` than the one the running app writes, silently ignoring the persisted Computer Use UI opt-in. The reader now also honours `CODEX_LINUX_SETTINGS_FILE`, like the runtime `codexLinuxSettingsPath()` it mirrors.

2. **`applyLinuxQuitGuardPatch`'s last-resort fallback de-stricted the main bundle, and its drift warning was dead code.** The fallback prepended the quit-guard helper at byte 0; the `.vite/build` bundles start with `"use strict";`, and a statement before the directive demotes it to a plain expression, switching the whole bundle to sloppy mode. The helper is now inserted after a leading directive (the same technique the other helper-injection patches use). The "could not find insertion point" warning was guarded by a condition that the preceding `return` made unreachable; it now actually fires when no anchor exists.

## Source-of-truth files

- `scripts/patches/computer-use.js`
- `scripts/patch-linux-window-ui.test.js` (new coverage)

## Validation (Ubuntu/WSL, node 22)

- `node --test scripts/patch-linux-window-ui.test.js`: **222/222 pass** (218 baseline + 4 new: LINUX-first app-id precedence, `CODEX_LINUX_SETTINGS_FILE` override, use-strict-preserving fallback insertion, warning on missing anchor)
- `node --test linux-features/*/test.js`: all pass
- Baseline on `main`: 218/218 pass, so no regressions

## Notes / limitations

- The existing fallback test ("adds the Linux quit guard when only the Electron require is recognizable") still passes unchanged — insertion at byte 0 is preserved for sources without a directive.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -1144,6 +1144,25 @@ test("adds the Linux quit guard when only the Electron require is recognizable",
   assert.equal((patched.match(/codexLinuxQuitInProgress=!1/g) ?? []).length, 1);
 });
 
+test("keeps a leading use strict directive ahead of the quit guard fallback", () => {
+  const source =
+    '"use strict";const e=require(`./app-session.js`);let t=require(`electron`);class WindowManager{}';
+
+  const patched = applyPatchTwice(applyLinuxQuitGuardPatch, source);
+
+  assert.match(patched, /^"use strict";let codexLinuxQuitInProgress=!1/);
+  assert.equal((patched.match(/codexLinuxQuitInProgress=!1/g) ?? []).length, 1);
+});
+
+test("warns and skips the quit guard when no Electron require exists", () => {
+  const source = "const e=1;";
+
+  const { value, warnings } = captureWarns(() => applyLinuxQuitGuardPatch(source));
+
+  assert.equal(value, source);
+  assert.match(warnings.join("\n"), /quit guard insertion point/);
+});
+
 test("upgrades the legacy Linux quit guard helper when re-patching older bundles", () => {
   const source =
     "let n=require(`electron`),i=require(`node:path`),o=require(`node:fs`);let codexLinuxQuitInProgress=!1,codexLinuxMarkQuitInProgress=()=>{codexLinuxQuitInProgress=!0},codexLinuxIsQuitInProgress=()=>codexLinuxQuitInProgress===!0;var x=1;";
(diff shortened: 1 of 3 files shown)


scripts/patch-linux-window-ui.test.js
```diff
@@ -4046,11 +4065,13 @@ function withIsolatedHome(body) {
   const previousXdg = process.env.XDG_CONFIG_HOME;
   const previousAppId = process.env.CODEX_APP_ID;
   const previousLinuxAppId = process.env.CODEX_LINUX_APP_ID;
+  const previousSettingsFile = process.env.CODEX_LINUX_SETTINGS_FILE;
```
avifenesh, line 4068: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This saves the prior `CODEX_LINUX_SETTINGS_FILE`, but the `finally` block never restores it. The new test below sets this env var to a file under `tempHome`, and line 4104 then removes that directory, leaving `process.env.CODEX_LINUX_SETTINGS_FILE` pointing at a deleted file until another helper call clears it; if the caller had the variable set before the suite, the first `withIsolatedHome` call also permanently drops it. Please restore/delete it in the same way as `HOME`, `XDG_CONFIG_HOME`, and the app-id vars.

avifenesh (commented):

---

Found one test-isolation issue in the new settings-file coverage.

ilysenko:
Blocking issue: withIsolatedHome() saves and deletes CODEX_LINUX_SETTINGS_FILE but does not restore it in the finally block. The new test can leave process.env.CODEX_LINUX_SETTINGS_FILE pointing at a removed temp file, or lose a value that existed before the test suite. Please restore/delete CODEX_LINUX_SETTINGS_FILE the same way HOME, XDG_CONFIG_HOME, CODEX_APP_ID, CODEX_LINUX_APP_ID, and the Computer Use flag are restored.

ilysenko:
Closing this for now. If you'd like to come back to it, please feel free to reopen.

PR #452, opened 2026-06-11
fix(linux-features): espeak voice precedence and use-strict-safe helper insertion

ryan-mt:
## What changed

Two bug fixes in optional Linux feature patches.

1. **read-aloud: a custom voice forced the Hebrew espeak voice.** The espeak-ng fallback computed `let espeakVoice=voice||hasHebrew?`he`:`en-us``, which parses as `(voice||hasHebrew) ? "he" : "en-us"` — so whenever `CODEX_LINUX_READ_ALOUD_VOICE` was set, the configured voice was discarded and espeak-ng ran with `-v he` regardless of the text language. The intent is unambiguous from the parallel `spd-say` path on the same line, which passes the custom voice (`-y voice`) and the language (`-l hasHebrew?he:en`) separately. Parenthesized so a custom voice wins and the language fallback only applies without one.

2. **remote-mobile-control: helper insertion de-stricted the app-server bundle.** `applyLinuxRemoteMobileAppServerRemoteControlPatch` prepended its helper function at byte 0 of the `.vite/build` bundle. Those bundles start with `"use strict";`, and a statement before the directive demotes it to a plain expression, silently switching the whole bundle to sloppy mode. The helper is now inserted after a leading directive, the same technique the read-aloud main-bundle patch already uses.

## Source-of-truth files

- `linux-features/read-aloud/patch.js`, `linux-features/read-aloud/test.js`
- `linux-features/remote-mobile-control/patch.js`, `linux-features/remote-mobile-control/test.js`

## Validation (Ubuntu/WSL, node 22)

- `node --test linux-features/read-aloud/test.js` — 39/39 pass, including a new assertion pinning `voice||(hasHebrew?...)`
- `node --test linux-features/remote-mobile-control/test.js` — 56/56 pass, including a new test asserting a leading `"use strict";` stays first after patching
- Full sweep: `node --test scripts/patch-linux-window-ui.test.js` and every `linux-features/*/test.js` pass on this branch; identical suites pass on `main` (baseline)

## Notes / limitations

- Both features are opt-in and disabled by default; no core build flow changes.


ilysenko (approved):
Reviewed. No blocking issues found; CI is green and the optional feature patch fixes are covered by focused tests.

PR #453, opened 2026-06-11
fix(build): portable awk header matching and correct stderr redirects

ryan-mt:
## What changed

Two small build/CI plumbing fixes.

1. **Upstream DMG metadata capture silently degraded to `unknown` under mawk.** `capture_upstream_metadata` in `scripts/ci/container-entrypoint.sh` and the equivalent step in `.github/workflows/upstream-build-app.yml` matched HTTP response headers with `awk 'BEGIN{IGNORECASE=1} /^last-modified:/ ...'`. `IGNORECASE` is a gawk extension; under mawk — the default `awk` in the `ubuntu:24.04` CI container and on ubuntu runners — it is an inert variable, and real responses send capitalized `Last-Modified` / `ETag` / `Content-Length`, so all three lookups returned empty and the recorded metadata fell back to `unknown` / `no-etag`. Now matches on `tolower($0)`, like the equivalent parser in `scripts/lib/dmg.sh`.

   Verified against real mawk:
   ```
   mawk old IGNORECASE etag: []
   mawk new tolower etag:    [abc]
   ```

2. **`2>&1 >&2` in `scripts/lib/native-modules.sh` did the opposite of intent.** Redirections apply left to right: `2>&1` points stderr at current stdout, then `>&2` points stdout at the (already-redirected) stderr — net effect, stderr merges *into stdout* and stdout is unchanged. The evident intent (keep npm/electron-rebuild chatter on stderr, the convention everywhere else in the install pipeline) is `>&2`.

## Source-of-truth files

- `scripts/ci/container-entrypoint.sh`
- `.github/workflows/upstream-build-app.yml`
- `scripts/lib/native-modules.sh`

## Validation

- `bash -n` across install/lib/launcher/build/ci scripts — clean
- `bash tests/scripts_smoke.sh` on this branch in Ubuntu (WSL): **All script smoke tests passed**
- mawk comparison above run against `mawk 1.3.4` on Ubuntu (WSL)

## Notes / limitations

- Impact of (1) is limited to CI metadata/reporting (cache keys still work since they hash whatever values were captured); (2) is stream classification only — no caller currently parses install.sh stdout.

### DIFF
--- a/.github/workflows/upstream-build-app.yml
+++ b/.github/workflows/upstream-build-app.yml
@@ -64,9 +64,11 @@ jobs:
 
           curl -fsSLI "$UPSTREAM_DMG_URL" > "$headers_file"
 
-          last_modified="$(awk 'BEGIN{IGNORECASE=1} /^last-modified:/ {sub(/\r$/,""); sub(/^[^:]+: /,""); print; exit}' "$headers_file")"
-          etag="$(awk 'BEGIN{IGNORECASE=1} /^etag:/ {sub(/\r$/,""); sub(/^[^:]+: /,""); gsub(/"/,""); print; exit}' "$headers_file")"
-          content_length="$(awk 'BEGIN{IGNORECASE=1} /^content-length:/ {sub(/\r$/,""); sub(/^[^:]+: /,""); print; exit}' "$headers_file")"
+          # Match header names case-insensitively without gawk's IGNORECASE,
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed. No blocking issues found; CI is green and the build/CI portability fixes are scoped and safe.

PR #455, opened 2026-06-11
fix(linux): add resize repaint hook

mayrainLN:
## Problem

On some Linux desktop/compositor setups, resizing the Codex Desktop window can leave stale compositor frames behind. The visible symptom is a "ghosting" or repeated-frame trail after dragging the window border: portions of the old window contents remain painted until the window is forced through a full refresh, for example by toggling fullscreen and restoring the window.

Resize ghosting after dragging the window border

The issue appears to be in Electron/Chromium's GPU compositor path on the affected Linux setup. In local testing, a normal launch still showed stale frame trails after resize, while launching with `CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=1` removed the ghosting.

## Why this change

This PR adds a small Linux-only repaint hook around `BrowserWindow` resize events. When a window emits `resize` or `resized`, the hook schedules a bounded `webContents.invalidate()` call on the next tick. That gives affected compositor paths an explicit repaint nudge after border drags without changing non-Linux behavior.

The hook is intentionally:

- Linux-only.
- Idempotent per `BrowserWindow` instance.
- Guarded against destroyed windows / webContents.
- Throttled through a short scheduled callback instead of invalidating repeatedly in the same resize burst.

This does not force software rendering for everyone. Instead, the code path provides a low-risk repaint nudge, while the docs now call out `--disable-gpu-compositing` as the stronger workaround for systems where Electron's GPU compositor still produces stale frames.

## User workaround documented

For users who still see resize ghosting or stale frame trails, the documented workaround is either:

```bash
CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=1 ./codex-app/start.sh
```

or adding this persistent Electron flag:

```text
--disable-gpu-compositing
```

## Additional test fix

While validating this from a side-by-side Codex app identity, `tests/scripts_smoke.sh` inherited `CODEX_LINUX_APP_ID` / `CODEX_LINUX_SETTINGS_FILE` from the running app. That made the Computer Use UI opt-in smoke test read the caller app's settings file instead of its isolated fake `$HOME`.

The smoke test now unsets those app-identity variables for the relevant patcher invocations, keeping the fixture hermetic.

## Validation

- `bash -n install.sh scripts/lib/*.sh launcher/start.sh.template scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh scripts/build-appimage.sh`


ilysenko (approved):
Reviewed. No blocking issues found; CI is green and the resize repaint hook is optional, descriptor-based, and covered by tests.

Issue #454, opened 2026-06-11
App-server disconnect loop: IAB_LIFECYCLE "route window is not live" → stop_process


1WYUSE:
## Environment
- OS: Ubuntu 24.04.4 LTS (Noble Numbat), X11
- Electron: 42.1.0
- Codex Desktop: 2026.06.11.045753 (commit 4203550)
- Desktop: GNOME/X11

## Problem
Codex Desktop enters a continuous disconnect loop (stop_process → connected → stop_process) every few seconds, displaying "Reconnecting..." constantly.

## Root Cause
Every disconnect is triggered by the same cascade:

```
[browser-sidebar-manager] IAB_LIFECYCLE route window is not live hasWindowState=false ownerDestroyed=null windowDestroyed=null windowId=1
[browser-sidebar-manager] IAB_LIFECYCLE browser use route window missing for active state
[browser-session-registry] IAB_LIFECYCLE removed browser session window windowId=1
[AppServerConnection] Stopping app-server transport connectionId=1
[AppServerConnection] app_server_connection.state_changed cause=stop_process
```

browser-sidebar-manager reports hasWindowState=false even though the window is NOT destroyed. This false negative tears down the session, cascading into app-server disconnect.

## Frequency
110+ times across multiple sessions. Every conversation triggers it within seconds.

## Possible fix
The window liveness check in browser-sidebar-manager/browser-session-registry likely relies on macOS-specific Electron window state APIs. On Linux/X11 they don't behave identically. A patch that relaxes the liveness check on Linux should fix this.

ilysenko:
Thanks for the detailed report. We merged PR #456 with a Linux-only Browser Use route liveness fallback that should prevent the false `route window is not live` result from tearing down the app-server session when the app window is still alive.

Could you please try a fresh build from current `main` and let us know whether the reconnect loop is gone on Ubuntu/GNOME/X11?

PR #456, opened 2026-06-11
[codex] Fix Linux Browser Use route liveness fallback

ilysenko:
## Summary
- add a Linux-only fallback for Browser Use route window liveness when upstream cannot resolve the recorded window id
- keep the fallback conservative: use one live BrowserWindow/window state only, and preserve upstream behavior when multiple windows make the route ambiguous
- register the patch as an optional core main-bundle descriptor and cover it with regression tests

## Root cause
Issue #454 reports a reconnect loop where Browser Use route liveness returns false with hasWindowState=false even though the app window is still alive. In the fresh upstream 26.608.12217 bundle, the main-process route helper logs this exact IAB_LIFECYCLE route window is not live warning before downstream Browser Use/session teardown.

## Validation
- verified applyLinuxBrowserUseRouteLivenessPatch changes the fresh main-CIL4OHS5.js bundle from upstream Codex.dmg

Addresses #454

### DIFF
--- a/scripts/patches/main-process.js
+++ b/scripts/patches/main-process.js
@@ -1293,6 +1293,45 @@ function applyBrowserUseNodeReplApprovalPatch(currentSource) {
   return patchedSource;
 }
 
+function applyLinuxBrowserUseRouteLivenessPatch(currentSource) {
+  if (currentSource.includes("codexLinuxResolveLiveBrowserUseRouteWindow")) {
+    return currentSource;
+  }
+
+  const routeWindowPattern =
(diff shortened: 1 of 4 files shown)


scripts/patches/main-process.js
```diff
@@ -1293,6 +1293,45 @@ function applyBrowserUseNodeReplApprovalPatch(currentSource) {
...
+  const replacement = `${helper}function ${functionName}({ensureWindowState:${ensureWindowStateVar},windowId:${windowIdVar},windows:${windowsVar}}){let ${stateVar}=${windowsVar}.get(${windowIdVar})??null;if(${stateVar}==null){let ${browserWindowVar}=${electronVar}.BrowserWindow.fromId(${windowIdVar});${browserWindowVar}!=null&&!${browserWindowVar}.isDestroyed()&&!${browserWindowVar}.webContents.isDestroyed()&&(${stateVar}=${ensureWindowStateVar}(${browserWindowVar},${browserWindowVar}.webContents))}${stateVar}==null&&(${stateVar}=codexLinuxResolveLiveBrowserUseRouteWindow(${ensureWindowStateVar},${windowIdVar},${windowsVar},${electronVar}));return ${stateVar}==null||${stateVar}.window.isDestroyed()||${stateVar}.owner.isDestroyed()?(${loggerVar}().warning(\`IAB_LIFECYCLE route window is not live\`,{safe:{hasWindowState:${stateVar}!=null,ownerDestroyed:${stateVar}?.owner.isDestroyed()??null,windowDestroyed:${stateVar}?.window.isDestroyed()??null,windowId:${windowIdVar}},sensitive:{}}),null):${stateVar}}`;
+
+  return currentSource.replace(original, replacement);
```
avifenesh, line 1332: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This patch helper uses `currentSource.match(routeWindowPattern)` and then a single `currentSource.replace(original, replacement)`, so it only patches the first eligible route-liveness helper. Per the ASAR patch invariant for `scripts/patches/*`, helpers should replace all eligible occurrences (or explicitly assert uniqueness) because upstream bundles can contain duplicate minified call sites; otherwise the descriptor can report success while a later `IAB_LIFECYCLE route window is not live` helper still lacks the Linux fallback. Please either iterate/count matches and patch all of them, or warn/fail unless this pattern is unique.

avifenesh (commented):

---

Found one ASAR patching robustness issue.

PR #461, opened 2026-06-11
chrome-plugin: prefer active Linux Chrome profiles

moxw:
## Summary

- Filter Linux Chrome extension backends that have zero user tabs when another extension backend has active tabs.
- Update the Chrome skill bootstrap to enumerate extension backends and bind the single active profile before creating tabs.
- Add smoke coverage for the installed Chrome plugin bundle shape and the staged skill bootstrap.

## Why

Closed or idle Chrome profiles can leave extension native messaging pipes connected. Picking the generic `extension` backend could then bind an idle profile instead of the visible active profile, which can start the wrong Chrome profile.

## Validation

- `node --check scripts/lib/patch-chrome-plugin.js`
- Temporary patch against the installed Chrome plugin cache plus `node --check` on the patched `browser-client.mjs`


scripts/lib/patch-chrome-plugin.js
```diff
@@ -474,10 +486,148 @@ patchFileFirstMatch(path.join(scriptsDir, "browser-client.mjs"), {
...
+    newText: String.raw`function lk({browserId:e,clientInfo:t,requestedBrowserId:r}){return cd(r)?eg(t.type)===r:e===r}var B$`,
+  },
+  {
+    label: "remove Linux ambiguous Chrome extension alias check",
+    oldText: String.raw`if(cd(p.browser_id)){let h=dd(p.browser_id);tg(h)||hk({diagnostics:n,reason:"backend-disabled",requestedBrowserId:p.browser_id}),ck(h),codexLinuxRejectAmbiguousBrowserAlias(p.browser_id,i)}let f=i.find`,
+    newText: String.raw`if(cd(p.browser_id)){let h=dd(p.browser_id);tg(h)||hk({diagnostics:n,reason:"backend-disabled",requestedBrowserId:p.browser_id}),ck(h)}let f=i.find`,
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This removes the Linux-only ambiguity guard for `browser_id: "extension"`, but the new filter above only drops extension backends with `userTabCount === 0`. If two Chrome/Brave profiles both have user tabs, both survive and the later `i.find(...)` path will silently pick whichever ranked backend comes first instead of preserving the previous invariant that the generic `"extension"` alias is rejected when multiple extension instances remain. That can still bind the wrong visible profile; the guard should continue to reject the alias after filtering whenever more than one active extension backend is left.

scripts/lib/patch-chrome-plugin.js
```diff
@@ -474,10 +486,148 @@ patchFileFirstMatch(path.join(scriptsDir, "browser-client.mjs"), {
...
+  const summaries = [];
+  for (const info of extensionInfos) {
+    const candidate = await agent.browsers.get(info.id);
+    const tabs = await candidate.user.openTabs().catch((error) => [
+      { error: String(error) },
+    ]);
+    summaries.push({ id: info.id, metadata: info.metadata, tabs });
+  }
+  const activeSummaries = summaries.filter(
+    ({ tabs }) => Array.isArray(tabs) && tabs.length > 0,
```
avifenesh, line 595: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

The catch path returns an array containing an `{ error: ... }` object, and the `activeSummaries` predicate below treats any non-empty array as active. That means a stale/failing extension backend can be counted as an active profile, causing the bootstrap to report multiple active profiles (or select the erroring one) even though it did not actually return any user tabs. Please keep errors in the summary separately, or filter only arrays of real tab entries. The same snippet is repeated in the `prefer active Chrome profile bootstrap` template below.

avifenesh (commented):

---

I found a couple of correctness issues in the Chrome multi-profile handling that look likely to preserve (or reintroduce) wrong-profile selection in edge cases.

scripts/lib/patch-chrome-plugin.js
moxw: Fixed in `ac0c926`: `openTabs()` failures are now recorded separately as `error`, while `tabs` stays an array of real returned tabs or `[]`. The same handling is applied to both bootstrap templates, and smoke coverage now asserts we do not emit the fake `{ error: String(error) }` tab entry.

ilysenko (changes requested):
Blocking this for now. The previous review comments look addressed, but I found two runtime risks in the current head. First, `codexLinuxFilterBrowserBackends()` filters idle extension backends out of the returned browser list without closing their `browser.api`. Those discarded extension connections will not be covered by the later runtime disposer, so stale Chrome/Brave native messaging pipes can remain open across bootstrap/re-init. Please close discarded extension backends before returning the kept list. Second, the `control-chrome` skill bootstrap probes `candidate.user.openTabs()` without the 750ms timeout used by `codexLinuxExtensionUserTabCount()`. A stale or blocked backend can hang the bootstrap before it prints summaries or reaches the intended profile-selection error. Please use a bounded tab probe there too and record timeout/error separately.

moxw:
Addressed in `d0eb6eb`.

- `codexLinuxFilterBrowserBackends()` now awaits closing of discarded idle extension backends before returning the kept backend list.
- The `control-chrome` skill bootstrap wraps `candidate.user.openTabs()` in a 750ms timeout and records timeout/error separately from real returned tabs.
- Added current `26.609` bundle-shape patch variants so the fix applies to the current installed Chrome plugin cache too.

Validated with:

- `node --check scripts/lib/patch-chrome-plugin.js`
- temp patch against installed Chrome plugin cache `26.609.30741` plus `node --check` on patched `browser-client.mjs`

ilysenko (approved):
Reviewed: fixes 4 active patch-target misses against the current upstream DMG plugin, adds idle-profile filtering; verified locally — patched output valid and idempotent, merged-tree smoke green.

PR #465, opened 2026-06-12
[codex] Keep rate limit footer visible in current composer

nofihq:
## Summary

- keeps the Linux rate-limit footer visible in the current composer bundles, including restored chats and new chats
- patches current external footer chunks instead of skipping them
- adds a visible `Usage limits` fallback when upstream limit data/model lookup is unavailable instead of returning `null`
- avoids caching the injected footer inside stale composer footer memo entries where the active chat can change

## Root Cause

The previous patch still missed current composer footer paths after upstream bundle drift. In restored/follow-up chats, the footer helper could throw or have no matching limit rows and return `null`, so the injected footer disappeared even though the patch had applied. External composer footer chunks were also excluded from the descriptor scan.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js --test-name-pattern='persistent rate limit footer'` passed: 226/226
- patched the installed `/opt/codex-desktop` app locally and restarted with fresh Electron caches
- visually verified the real desktop app with screenshots in all four states:
  - new chat inside a project
  - restored/existing project chat
  - restored/existing projectless chat
  - new projectless chat
- DOM checks for each screenshot reported `Usage limits` visible in the live app

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -4788,7 +4788,7 @@ test("persistent rate limit footer uses existing account signal without dropdown
   assert.equal((patched.match(/Rate limits remaining/g) || []).length, 1);
 });
 
-test("persistent rate limit footer descriptor ignores external footer chunks", () => {
+test("persistent rate limit footer descriptor includes external footer chunks", () => {
   const descriptor = corePatchDescriptors().find(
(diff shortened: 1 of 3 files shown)


nofihq:
Closing this draft because the visual result was too noisy: it showed multiple placeholder Usage limits badges instead of one numeric footer status. I’ll rework this locally before opening anything new.

PR #466, opened 2026-06-12
nix: wrap launcher to bake CODEX_CLI_PATH via cliPackage option

colonelpanic8:
## Problem

Codex Desktop resolves the local Codex CLI from `CODEX_CLI_PATH` first, then a bundled `resources/bin/codex`. The Nix package ships **neither**, so it depends on the launcher's `find_codex_cli` (cold start only) finding `codex` on `PATH`.

Graphical autostart entries, application launchers, and **warm-start handoffs** to an already-running instance don't carry the Nix profile on `PATH`, and the warm-start path explicitly logs `Using CODEX_CLI_PATH=warm-start-skip`. A long-running instance started without a resolvable CLI then fails on every new app-server connection with:

```
Unable to locate the Codex CLI binary. Set CODEX_CLI_PATH or ensure the Electron resources include bin/codex.
```

Today this is documentation-only (`docs/nix.md` tells users to set `CODEX_CLI_PATH` by hand). Related: #395.

## Change

Add a `programs.codexDesktopLinux.cliPackage` option to the Home Manager and NixOS modules. When set, the module wraps the installed launcher with a thin `makeWrapper` layer that does `--set-default CODEX_CLI_PATH <pkg>/bin/codex`, and repoints the package's `.desktop` entry (which launches `<pkg>/bin/codex-desktop` by absolute path) at the wrapper.

The CLI path is **baked into the launcher**, not exported as a session variable, so it:
- works for graphical, terminal, and warm-start launches alike;
- takes effect on the next app launch — **no re-login required**;
- still lets an explicit `CODEX_CLI_PATH` in the environment win (`--set-default`).

The wrapper is a cheap `symlinkJoin` layer; it does not rebuild the Electron payload.

- Default `null` preserves current PATH-based behavior (launcher left unwrapped).
- When unset but `remoteControl.enable = true`, it reuses `remoteControl.package` so the desktop and the remote-control service share one CLI.
- Docs + CHANGELOG updated.

> Re: the earlier auto-review note about also setting `systemd.user.sessionVariables` in the NixOS module — that no longer applies. The first revision wired `CODEX_CLI_PATH` through `environment.sessionVariables`, which (a) needs a re-login to apply and (b) splits the GUI/login-shell vs systemd-user surfaces. This revision drops the session-variable approach entirely in favor of the launcher wrapper, so there is no env-propagation surface to mirror.

## Validation

- `nix-instantiate --parse` on both module files.
- Built the `symlinkJoin` wrapper over a real package store path and confirmed `bin/codex-desktop` becomes `export CODEX_CLI_PATH=${CODEX_CLI_PATH-'.../codex'}; exec <base>/bin/codex-desktop "$@"` and that both `.desktop` `Exec=` lines (main + "New Window") now point at the wrapper.

### DIFF
--- a/nix/nixos-module.nix
+++ b/nix/nixos-module.nix
@@ -19,7 +19,47 @@ let
       "codex-desktop-computer-use-ui"
     else
       "codex-desktop";
-  desktopPackage = if cfg.package != null then cfg.package else flakePackages.${packageName};
+  basePackage = if cfg.package != null then cfg.package else flakePackages.${packageName};
+  codexCliPackage =
+    if cfg.cliPackage != null then
+      cfg.cliPackage
+    else if remoteCfg.enable then
+      remoteCfg.package
+    else
+      null;
+  codexCliPath = if codexCliPackage != null then lib.getExe' codexCliPackage "codex" else null;
+  # Thin wrapper that bakes CODEX_CLI_PATH into the launcher. The `.desktop`
+  # entry shipped by the package launches `<pkg>/bin/codex-desktop` by absolute
+  # path, so wrapping that binary (and repointing the desktop entry at the
+  # wrapper) makes Codex Desktop locate the CLI no matter how it is started --
+  # graphical autostart, application launcher, terminal, or a warm-start handoff
+  # to an already-running instance -- without depending on the session/login
+  # `PATH` and without requiring a re-login for a config change to take effect.
+  # `--set-default` leaves an explicit `CODEX_CLI_PATH` in the environment in
+  # control, so users can still override the launched CLI.
+  withCodexCliPath =
+    base:
+    pkgs.symlinkJoin {
+      name = "${base.name}-codex-cli-path";
+      paths = [ base ];
+      nativeBuildInputs = [ pkgs.makeWrapper ];
+      postBuild = ''
+        if [ -e "$out/bin/codex-desktop" ]; then
+          rm -f "$out/bin/codex-desktop"
+          makeWrapper "${base}/bin/codex-desktop" "$out/bin/codex-desktop" \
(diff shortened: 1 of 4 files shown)


nix/nixos-module.nix
```diff
@@ -163,9 +200,7 @@ in
       desktopPackage
     ];
 
-    environment.sessionVariables = lib.mkIf (remoteCfg.enable && remoteCfg.disableLauncherAutostart) {
-      CODEX_REMOTE_CONTROL_DAEMON_AUTOSTART_DISABLED = "1";
-    };
+    environment.sessionVariables = sessionVariables;
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This wires the new `CODEX_CLI_PATH` only through `environment.sessionVariables`. For Nix module environment variables that affect launcher/desktop startup, the repo invariant is to publish them on the systemd-user surface as well, because GUI/login-shell and systemd-launched processes read different environments. As written, the NixOS `cliPackage` / `remoteControl.package` fallback can still be missing for user units or systemd-launched sessions, so this should also set `systemd.user.sessionVariables = sessionVariables;` (mirroring the Home Manager module's `home.sessionVariables` + `systemd.user.sessionVariables`).

avifenesh (commented):

---

Found one environment propagation issue in the NixOS module.

ilysenko (approved):
Reviewed current head 1187f3601c7bb37f6ad78d7e93023bd3f99f7056. The previous environment propagation concern is addressed by baking CODEX_CLI_PATH into the wrapped launcher instead of relying on session variables. CI is green and I do not see a blocker.

PR #468, opened 2026-06-12
[codex] Keep rate limit footer visible in current composer

nofihq:
## Summary
- keep the persistent rate-limit footer patch on the current composer bundle instead of the external/shared footer path
- render exactly one numeric footer badge after the existing Full access control
- preserve the previous transparent oval chip styling while removing the fake `Usage limits` fallback

## Root cause
The earlier patch targeted broad/shared footer surfaces, which could produce duplicate badges and miss the current composer path after upstream bundle drift. The current bundle also exposes rate-limit data through the latest rate-limit status query shape, so the helper now detects that query and fails soft when data is unavailable.

## Validation
- `node --test scripts/patch-linux-window-ui.test.js --test-name-pattern='persistent rate limit footer'`
- Installed the patch into local `/opt/codex-desktop/content/webview/assets` and restarted Codex Desktop.
- Verified in the running app via DevTools DOM inspection:
  - fresh composer: exactly one `.composer-footer__label--sm`, text `5h 49% / 7d 86%`, `borderRadius: 9999px`, `backgroundColor: rgba(0, 0, 0, 0)`, `boxShadow: none`, positioned immediately after `Full access`
  - existing chat: exactly one `.composer-footer__label--sm`, same transparent oval styling, positioned immediately after `Full access`
- Captured local renderer screenshots for both fresh composer and existing chat before pushing.


ilysenko (changes requested):
Blocking this for now. On the current real `composer-D9quMRYW.js` bundle, `applyPersistentRateLimitFooterPatch` is not idempotent: applying the PR patch once changes the helper, and applying it again changes the bundle from 436950 bytes to 431747 bytes and removes an intervening `function lh`. The risky path is the existing-helper replacement that uses `latestFooterControls.insertionNeedle` as the end boundary, so it can replace everything between `function codexLinuxRateLimitFooter` and the later footer controls function, including unrelated functions. Please bound the replacement to the helper function itself or skip it when the existing helper is already valid, and add a regression fixture for an existing helper separated from the latest composer controls by intervening functions.

ilysenko (approved):
Re-reviewed current head after resolving the idempotency blocker. Local patcher tests pass and CI is green.

PR #469, opened 2026-06-12
fix(linux): keep titlebar overlay refreshes native

jojomini1231-cloud:
## Summary
- route Linux titlebar overlay refreshes through the Linux-specific helper instead of reusing the Windows overlay helper
- keep nativeTheme/titlebar sync idempotent for Linux primary windows
- add regression coverage for zoom and renamed titlebar overlay sync paths

## Test Plan

### DIFF
--- a/scripts/patches/main-process.js
+++ b/scripts/patches/main-process.js
@@ -190,56 +190,73 @@ function applyLinuxNativeTitlebarPatch(currentSource) {
     return currentSource;
   }
 
+  // Zoom-change overlay call site: newer upstream includes Linux in the
+  // win32 branch and reuses the transparent win32 overlay helper there.
+  const zoomOverlayRegex =
+    /\(process\.platform===`win32`\|\|process\.platform===`linux`\)&&\(this\.windowZooms\.set\(([A-Za-z_$][\w$]*)\.id,([A-Za-z_$][\w$]*)\),\1\.setTitleBarOverlay\(([A-Za-z_$][\w$]*)\(\2\)\)\)/;
+  const zoomOverlayMatch = patchedSource.match(zoomOverlayRegex);
+  if (zoomOverlayMatch != null && zoomOverlayMatch[3] !== LINUX_TITLEBAR_OVERLAY_HELPER) {
+    const [zoomMatchText, zoomWindowAlias, zoomValueAlias, zoomHelperAlias] = zoomOverlayMatch;
+    patchedSource = patchedSource.replace(
+      zoomMatchText,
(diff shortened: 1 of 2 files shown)


scripts/patches/main-process.js
```diff
@@ -190,56 +190,73 @@ function applyLinuxNativeTitlebarPatch(currentSource) {
...
   }
 
+  // Zoom-change overlay call site: newer upstream includes Linux in the
+  // win32 branch and reuses the transparent win32 overlay helper there.
+  const zoomOverlayRegex =
+    /\(process\.platform===`win32`\|\|process\.platform===`linux`\)&&\(this\.windowZooms\.set\(([A-Za-z_$][\w$]*)\.id,([A-Za-z_$][\w$]*)\),\1\.setTitleBarOverlay\(([A-Za-z_$][\w$]*)\(\2\)\)\)/;
+  const zoomOverlayMatch = patchedSource.match(zoomOverlayRegex);
+  if (zoomOverlayMatch != null && zoomOverlayMatch[3] !== LINUX_TITLEBAR_OVERLAY_HELPER) {
+    const [zoomMatchText, zoomWindowAlias, zoomValueAlias, zoomHelperAlias] = zoomOverlayMatch;
+    patchedSource = patchedSource.replace(
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This new zoom-overlay rewrite only patches the first matching call site: `match()` returns one match and `replace(zoomMatchText, ...)` replaces one string occurrence. The repo's ASAR patch invariant is that patch helpers must replace all eligible bundle occurrences (or assert the needle is unique); otherwise a duplicate zoom handler in an upstream bundle would continue using the transparent win32 helper on Linux. Please iterate over all `zoomOverlayRegex` matches (or make/verify a uniqueness assertion) instead of stopping after the first match.

avifenesh (commented):

---

Found one correctness issue in the ASAR patch update: the new zoom overlay call-site patch does not process all eligible occurrences.

scripts/patches/main-process.js
jojomini1231-cloud: Thanks, fixed in f3c5d9848da1f520495018318335471274b7e4e1. The zoom overlay rewrite now uses a global regex replacement so every eligible call site is patched, while already-patched Linux helper calls are left unchanged for idempotency. Added a regression fixture with two zoom refresh call sites and verified locally with `node --test scripts/patch-linux-window-ui.test.js`.

ilysenko (approved):
Reviewed current head f3c5d98. No blockers found; CI and focused patch tests are green.

PR #470, opened 2026-06-12
Fix Linux plugin marketplace feature sync

joshyorko:
## Summary
- Preserve the upstream `remote_plugin` feature when Linux webview feature-sync patches add Linux-only feature flags.
- Keep the remote mobile feature patch from narrowing the supported feature list to `remote_control` only.
- Add regression tests for both the core webview patch and the optional remote mobile feature patch.

## Root Cause
After the upstream DMG started relying on the dynamic `remote_plugin` feature, the Linux feature-sync patches replaced that dynamic feature list instead of extending it. That disabled the remote plugin catalog path in the packaged Linux app, so the OpenAI Curated marketplace loaded only locally installed plugins.

## Validation

### DIFF
--- a/scripts/patches/webview-assets.js
+++ b/scripts/patches/webview-assets.js
@@ -4,6 +4,7 @@ const fs = require("node:fs");
 const path = require("node:path");
 
 const {
+  escapeRegExp,
   findMatchingBrace,
 } = require("./shared.js");
 
@@ -489,7 +490,7 @@ function applyLinuxAppServerFeatureEnablementPatch(currentSource) {
     if (dynamicBuilderExtraMatch != null) {
       const [, loopBlock, , , enablementVar, featureKeyVar] = dynamicBuilderExtraMatch;
       const featureKeyDeclaration = new RegExp(
-        `${featureKeyVar}=\`remote_plugin\``,
+        `${escapeRegExp(featureKeyVar)}=\`remote_plugin\``,
         "u",
       );
       if (featureKeyDeclaration.test(currentSource)) {
--- a/scripts/patches/webview-assets.js
+++ b/scripts/patches/webview-assets.js
@@ -462,6 +462,7 @@ function applyLinuxAppServerFeatureEnablementPatch(currentSource) {
     "mentions_v2",
     "plugins",
     "remote_control",
+    "remote_plugin",
     "tool_call_mcp_elicitation",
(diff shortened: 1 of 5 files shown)


scripts/patches/webview-assets.js
```diff
@@ -486,7 +487,14 @@ function applyLinuxAppServerFeatureEnablementPatch(currentSource) {
       /(for\(let ([A-Za-z_$][\w$]*) of [A-Za-z_$][\w$]*\)\{let ([A-Za-z_$][\w$]*)=[A-Za-z_$][\w$]*\[\2\];\3!=null&&\(([A-Za-z_$][\w$]*)\[\2\]=\3\)\})return \4\[([A-Za-z_$][\w$]*)\]=([A-Za-z_$][\w$]*),\4\}/u;
     const dynamicBuilderExtraMatch = currentSource.match(dynamicBuilderExtraRegex);
     if (dynamicBuilderExtraMatch != null) {
-      const [, loopBlock, , , enablementVar] = dynamicBuilderExtraMatch;
+      const [, loopBlock, , , enablementVar, featureKeyVar] = dynamicBuilderExtraMatch;
+      const featureKeyDeclaration = new RegExp(
+        `${featureKeyVar}=\`remote_plugin\``,
+        "u",
```
avifenesh, line 494: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`featureKeyVar` comes from `([A-Za-z_$][\w$]*)` above, so `$` is a valid captured minified identifier character. Interpolating it directly into a `RegExp` makes `$` act as an end-anchor; for an upstream shape like `v$I=\`remote_plugin\`` this test returns false and the fallback below still rewrites `return n[v$I]=t,n` to `return n}`, dropping `remote_plugin` despite this PR's fix. Please escape the captured identifier (or avoid regex construction and match the declaration structurally) before testing it.

avifenesh (commented):

---

I found one edge case where the new preservation check can still strip `remote_plugin` for a valid minified identifier.

ilysenko (approved):
Reviewed current head 3a60c32 after pushing the follow-up fix. The remote_plugin sync no longer treats minified feature-key variables as regex syntax, the remote-mobile-control feature remains opt-in/disabled by default with optional patch descriptors, and local patcher plus feature tests pass.

PR #472, opened 2026-06-12
Add frameless-titlebar Linux feature

jwiseuk:
## What

Adds an opt-in `frameless-titlebar` Linux feature that hides the Electron titlebar overlay controls and native menu chrome on the main Codex window, for Wayland compositors / window managers where compositor-managed decorations already provide window controls (developed and tested on Hyprland).

Disabled by default per the feature contract; enabled via `linux-features/features.json`.

## Why

The native Linux titlebar overlay (#413, refined in #460) renders Electron-drawn window buttons that visually conflict with compositor-managed decorations on Hyprland — the overlay strip and buttons are redundant there and don't match the compositor theme.

## How

Two patch descriptors (orders 20720/20730, after the core `linux-native-titlebar` patch at order 85):

- **main-bundle**: collapses the Linux primary-window branch to `titleBarStyle: hidden` with no `titleBarOverlay`, removes all Linux `setTitleBarOverlay` call paths (zoom sync and nativeTheme sync — these throw on a window created without an overlay), and hides/removes the window menu. Handles the raw upstream shape, the current upstream shape, and the core-patched shape, and is idempotent.
- **webview-asset** (`use-window-controls-safe-area-*.js`): zeroes the window-controls safe-area inset and maps Linux chrome from `application-menu` to `native` so the UI doesn't reserve space for buttons that no longer exist.

## Testing

- `node --test linux-features/frameless-titlebar/test.js` (3 tests: loader gating, main-bundle transform + idempotency, webview transform + idempotency)
- Verified end-to-end with `make install-native` on Arch/Hyprland: no overlay buttons, no menu bar, window management via compositor bindings, dark/light theme flips clean.
- Dry-run against the real upstream bundle confirms only the intended call sites change and the patch is a no-op when re-applied.

## Note for the maintainer

While building this I noticed upstream renamed `installWindowsTitleBarOverlaySync` to `installApplicationMenuTitleBarOverlaySync`. The core `linux-native-titlebar` patch in `scripts/patches/main-process.js` only matches the old name, so its nativeTheme-sync portion silently no-ops on current upstream bundles (the warning is gated on the old name being present). That's independent of this PR but probably worth a core fix.

### DIFF
--- a/linux-features/frameless-titlebar/README.md
+++ b/linux-features/frameless-titlebar/README.md
@@ -1,0 +1,49 @@
+# Frameless Titlebar
+
+This optional feature hides the Linux Electron titlebar overlay controls and
+removes the native menu chrome from the main Codex window. It is intended for
+Wayland compositors or window managers where compositor-managed decorations
+already provide the expected window controls, such as Hyprland setups.
+
+The default build leaves the existing Linux titlebar overlay behavior in place.
+Enable this only when the built-in Codex titlebar/buttons visually conflict
+with your desktop environment.
+
+Enable it by copying `linux-features/features.example.json` to
+`linux-features/features.json` and listing the feature id:
+
+```json
+{
+  "enabled": [
+    "frameless-titlebar"
+  ]
+}
+```
+
+Then rerun `./install.sh` or the native package build flow so the ASAR patches
+are regenerated with this feature enabled.
+
+## Testing
+
+Run the feature's unit tests from the repository root:
+
+```bash
+node --test linux-features/frameless-titlebar/test.js
+```
+
+For a manual check, enable the feature as above, rebuild, and launch the app:
+
+- The primary window should show no Electron-drawn titlebar overlay buttons
+  (minimize/maximize/close in the top-right corner) and no menu bar.
+- Window move, resize, and close/minimize/maximize should work through your
+  compositor's bindings (for example Hyprland's `bindm` mouse binds and
+  `killactive`/`fullscreen` dispatchers).
+- Changing the system dark/light theme must not crash the app or repaint a
+  titlebar strip; the patch removes all Linux `setTitleBarOverlay` calls,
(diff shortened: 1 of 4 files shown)


ilysenko (changes requested):
Thanks for keeping this as an opt-in Linux feature. I found one blocker on the current main bundle shape: after the core Linux titlebar patch, `setWindowZoom` calls `setTitleBarOverlay(process.platform===`linux` ? codexLinuxTitleBarOverlay(...) : ...)`. This feature removes `titleBarOverlay` from the Linux primary BrowserWindow options, but the zoom path is not rewritten by `applyFramelessTitlebarOverlaySyncPatch`, so changing window zoom can still call `setTitleBarOverlay` on Linux and hit the crash path this feature is meant to avoid. Please handle that current ternary zoom shape and add a regression fixture for it.

jwiseuk:
Fixed in d6fa1ce. `applyFramelessTitlebarOverlaySyncPatch` now also collapses the core-patched zoom shape — `` (win32||linux)&&(this.windowZooms.set(...),setTitleBarOverlay(process.platform===`linux`?codexLinuxTitleBarOverlay(...):helper(...))) `` — to the win32-only form, same as the plain-helper shape. The linux arm matches `codexLinuxTitleBarOverlay` by name with any arguments, so it tolerates argument drift in the core patch.

Added a regression fixture with that exact ternary body asserting the collapse, that no Linux `setTitleBarOverlay` path survives, and idempotency.

One note: the bundle I have locally still has the plain-helper zoom body (the new regex is exercised by the fixture, not a live bundle here), so if you can confirm against your current bundle shape that would close this out.

ilysenko (approved):
Reviewed current head d6fa1ce. The previous Linux setTitleBarOverlay blocker is fixed, the feature stays opt-in/disabled by default, and the descriptors are optional/fail-soft on drift. Local feature tests, all linux-feature tests, syntax/diff checks, and a dry-run against the generated app.asar passed.

PR #474, opened 2026-06-12
Keep Linux menu patch idempotent over frameless-titlebar output

jwiseuk:
## What

Follow-up to #472 — enabling `frameless-titlebar` surfaces two interactions this PR cleans up:

1. **Menu patch idempotency.** #472's menu patch upgrades the inserted snippet to also call `removeMenu?.()`. `applyLinuxMenuPatch` predates that form, so re-running the pipeline over feature-patched output added a duplicate snippet. This teaches the already-applied guard the upgraded form, with a regression test.

2. **Test isolation against feature opt-ins.** The core suite resolves the developer-local gitignored `linux-features/features.json`, so locally enabled features change which descriptors `patchMainBundleSource` applies during core tests — with `frameless-titlebar` enabled, three icon-path tests fail (two on `applyPatchTwice` idempotency via the duplication above, one asserting the pre-#472 menu form). The suite now pins `CODEX_LINUX_FEATURES_CONFIG` to `features.example.json` so core tests always run the same zero-feature configuration as CI. The build-info test, which reads `features.json` from its own temp `featuresRoot`, clears and restores the pin around its body.

## Testing

- `node --test scripts/patch-linux-window-ui.test.js`: 226/226 with no `features.json`, and 226/226 with a local `features.json` enabling `frameless-titlebar` (previously 3 failures).
- `node --test linux-features/frameless-titlebar/test.js`: 4/4.


ilysenko (approved):
Re-reviewed after merging current main. The idempotency fix now lives in main-process/window.js, local patcher tests pass, and CI is green.

PR #475, opened 2026-06-12
fix(updater): ignore already installed DMG updates

moxw:
## Summary
- Suppress DMG update candidates when installed build metadata already reports the same upstream DMG SHA.
- Clear stale in-progress, failed, or ready updater states for same-DMG candidates.
- Add updater tests for same-hash and different-hash in-progress candidates.

## User-visible behavior
Codex Desktop no longer shows or prepares a Desktop update when the downloaded upstream DMG already matches the installed app, even if the locally generated package timestamp is newer.

## Source of truth
- updater/src/app.rs

## Validation
- cargo fmt -p codex-update-manager

### DIFF
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -13,6 +13,7 @@ use anyhow::{Context, Result};
 use chrono::{Duration as ChronoDuration, Utc};
 use fs4::fs_std::FileExt;
 use reqwest::Client;
+use serde::Deserialize;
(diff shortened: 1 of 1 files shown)


ilysenko (approved):
Reviewed current head 2152ca2 after same-DMG recovery guard and regression tests. Local cargo check/test passed; waiting for CI before merge.

PR #476, opened 2026-06-12
[codex] fix Chrome Browser Use CDP access on Linux

moxw:
## Summary

- Stage a Browser Use `nodeRepl.config` compatibility shim for bundled Browser and Chrome plugin clients on Linux.
- Recognize the current upstream `cua_node/bin/node_repl` layout before falling back to the Linux primary runtime archive.
- Keep the Browser Use `node_repl` approval patch working across current and Electron 42 runtime-builder shapes.
- Preserve raw CDP call/event/response routing through the Linux Chrome native host and cover it with tests.

## User-visible behavior

Linux Browser Use / `@chrome` sessions can continue discovering Chrome extension backends when the bundled `node_repl` host is sealed and does not expose `nodeRepl.config`. Full-CDP tab capabilities can still flow through the native host, and the staged Browser Use client gets the config methods expected by newer macOS clients.

## Root Cause

Newer Browser Use clients query `nodeRepl.config` while building backend metadata for the full-CDP access gate. The current macOS bundle exposes that through its trusted `node_repl` runtime, while the Linux fallback runtime can be non-extensible and omit that API. That caused otherwise live extension sockets to be filtered out before tab capabilities could be inspected.

## Validation

- [x] `bash -n scripts/lib/bundled-plugins.sh tests/scripts_smoke.sh`
- [x] `node --test scripts/patch-linux-window-ui.test.js`
- [x] `cargo fmt --check`
- [x] `cargo test -p codex-computer-use-linux --bin codex-chrome-extension-host`
- [x] `bash tests/scripts_smoke.sh`
- [x] `git diff --check`
- [x] Live `@chrome` check: selected non-work profile `1Profile 1`, opened a fresh `https://example.com/` tab, and verified `Runtime.evaluate` over CDP returned the expected URL/title.


ilysenko (approved):
Reviewed current head 954ee61 after rebase onto split patch modules and safe read-only nodeRepl.config shim. Local patch tests, Chrome host tests, smoke suite, syntax, fmt, and diff checks passed; waiting for CI before merge.

PR #477, opened 2026-06-12
[codex] Hydrate chat search before navigation

moxw:
## Summary
- add a core webview patch that preserves host metadata for chat search results
- hydrate selected local search results before navigating to their conversation route
- add regression coverage for the minified app-main search/navigation shapes

## Why
Selecting an older chat from command-menu search could navigate with only the local thread key before that conversation was loaded into renderer state. The local conversation page could then remain on its loading/blank state.

## Source of truth
- scripts/patches/webview-assets.js
- scripts/patches/core/all-linux/webview/chat-search-hydration/patch.js

## Testing
- git diff --check
- built a side-by-side dev app with DEV_APP_ID=codex-search-test and verified the generated app-main bundle contains the hydration helper
- manual verification: search-loading an old chat in the side-by-side app now opens correctly


ilysenko:
Thanks for the patch, this looks like the right direction and the feature should stay fail-soft/optional.

I found one blocker before merge: the patch changes the navigation callback from passing `result.threadKey` to passing the full `result`, so the callback now depends on result metadata such as `hostId`. However the React compiler cache guard still keys the callback only on `s.threadKey` (`t[23] !== s.threadKey` / `t[23] = s.threadKey`). If two search results share the same threadKey but differ by hostId, the cached callback can keep the stale result object and hydrate/navigate the wrong host.

Could you update the patch to also rewrite that cache dependency to include the full result object, or at least `kind/threadKey/hostId`, and add a regression assertion for the cache guard? The current test asserts `FS(s,o,d)` but does not verify that the callback cache invalidates when hostId changes.

moxw:
Addressed in 382dc98: the chat search patch now rewrites the React compiler cache guard/store to key on the full result object instead of result.threadKey, so hostId changes invalidate the callback. I also added regression assertions that the minified guard uses t[23] !== s / t[23] = s and no longer leaves the stale s.threadKey dependency behind.\n\nValidation:\n- node --test scripts/patch-linux-window-ui.test.js\n- git diff --check

ilysenko (approved):
Reviewed current head and the combined state after #488; no blockers found.

PR #481, opened 2026-06-12
fix(linux): prevent duplicate app instances (launcher detection + enforced bootstrap single-instance lock)

Sesma1510:
Two fixes for the duplicate-instance restart loop observed in production on SteamOS:

**1. Launcher: running-app detection survives lost markers** (`launcher/start.sh.template`)
- A hidden-to-tray instance could lose both its `app.pid` (clobbered by a failed duplicate launch's cleanup) and its launch-action socket; the `/proc` scan was gated on the socket existing, so the instance became undetectable and every launch cold-started a duplicate sharing the same user-data dir.
- The scan now always runs as fallback, filtered by instance identity from `/proc/<pid>/environ` so default and side-by-side multi-launch instances never adopt each other.
- The detection → spawn → `app.pid` write critical section is serialized with a bounded `flock`.

**2. Bootstrap: enforce the single-instance lock on Linux** (`scripts/patches/bootstrap.js`)
- Upstream gates `requestSingleInstanceLock()` behind a flag computed from `{isMacOS, isPackaged}` — always false on Linux — so packaged Linux builds ran with **no** duplicate-instance protection (verified: no `SingletonLock` in the profile dir while the app runs). The previous patch only added the multi-launch bypass and preserved the `!flag ||` short-circuit.
- The gate is rewritten to a platform ternary: Linux always takes the lock unless `CODEX_LINUX_MULTI_LAUNCH=1`; other platforms keep upstream semantics. Legacy-guarded bundles are upgraded in place.
- Patch promoted to `required-upstream` since main-process defers single-instance handling to bootstrap on current bundles.

Verified end-to-end on SteamOS against DMG 26.608.12217 and 26.609.30741: `SingletonLock` now appears in the profile dir, and a second launch with all launcher markers wiped rediscovers the live instance and hands off instead of duplicating. `node --test scripts/patch-linux-window-ui.test.js` (204 pass) and `tests/scripts_smoke.sh` launcher assertions pass.

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -1416,6 +1416,25 @@ find_running_app_pid() {
     return 1
 }
 
+pid_in_same_launch_instance() {
+    # The /proc scan matches by electron binary path, which side-by-side
+    # multi-launch instances share with the default app. Match instance
+    # identity via the process environment so the default app never adopts
+    # a side-by-side instance and vice versa. Substring checks on the
+    # newline-joined environ avoid grep -q pipelines, which can fail under
+    # pipefail when grep exits before the writer finishes.
+    local pid="$1"
+    local environ="/proc/$pid/environ"
+    local env_lines
+
+    env_lines=$'\n'"$(tr '\0' '\n' < "$environ" 2>/dev/null || true)"$'\n'
+    if [ "$MULTI_LAUNCH_ACTIVE" -eq 1 ]; then
+        [[ "$env_lines" == *$'\n'"CODEX_LINUX_INSTANCE_ID=$CODEX_LINUX_INSTANCE_ID"$'\n'* ]]
+    else
+        [[ "$env_lines" != *$'\n'"CODEX_LINUX_MULTI_LAUNCH=1"$'\n'* ]]
+    fi
+}
+
 discover_running_app_pid() {
     local pid
     local proc_exe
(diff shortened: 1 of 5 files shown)


IvanTheGeek:
**Reproduced the second-instance fallback on Debian 13 — plus a small defensive complement in the launcher**

Confirmed on Debian 13 trixie (X11): when an instance is already running and warm-start IPC is unavailable, the launcher logs `Warm-start IPC unavailable; falling back to Electron second-instance handoff`, then `Using CODEX_CLI_PATH=warm-start-skip`, and if Electron's single-instance handoff doesn't take effect the second process boots a full app-server and dies with `Unable to locate the Codex CLI binary` → `bootstrap-import-main`. PR #481 fixes the handoff itself (the right fix).

As belt-and-suspenders, the CLI-path *lookup* in `launcher/start.sh.template` is gated by `needs_cold_start`, which is **false** on this fallback path: `detect_warm_start` sets `WARM_START=1`, IPC failure leaves it `1`, so `needs_cold_start = [ "$WARM_START" -eq 0 ] && …` is false and the cheap, read-only `find_codex_cli` lookup is skipped. Ungating just the lookup (the heavier preflight install/update stays gated) lets a process that *does* become primary still start:

```diff
-if needs_cold_start && [ -z "${CODEX_CLI_PATH:-}" ]; then
+if [ -z "${CODEX_CLI_PATH:-}" ]; then
     CODEX_CLI_PATH="$(find_codex_cli || true)"
     export CODEX_CLI_PATH
     log_phase "cli_lookup"
 fi
```

With this applied locally, the same fallback relaunch logs `Using CODEX_CLI_PATH=/…/codex` and produces zero CLI/bootstrap errors (the second instance handed off to the primary cleanly, single window).

(Note: gating on `[ "$WARM_START" -eq 0 ]` instead wouldn't help — `WARM_START` is still `1` on the IPC-failure fallback.)

ilysenko (approved):
Reviewed current head 8c9947ae7608e593c223ae911f92a1819d98c382. The Debian 13 second-instance fallback comment is addressed, launcher locking is scoped to the final cold-launch critical section, local focused checks passed, and GitHub CI is green.

PR #482, opened 2026-06-12
feat(linux): opt-in node-repl-reaper feature for leaked Browser Use helpers

Sesma1510:
The upstream `codex app-server` spawns `node_repl` helpers for Browser Use and does not always reap them — helpers accumulate over long sessions and survive their owner (observed on SteamOS: six leaked in 15 minutes, persisting for over a day under a hidden-to-tray instance), each holding memory and fds indefinitely.

This adds an **opt-in** linux-feature that reaps only *leaked* helpers — those whose parent is no longer a live `codex app-server` — so active Browser Use sessions are never touched. Matching is scoped to the install's `resources/node_repl` path, so side-by-side installs reap independently.

- Cold-start hook: one watchdog per install (pid file in the app state dir), reaps every 5 min (`CODEX_NODE_REPL_REAPER_INTERVAL`), self-terminates with a final pass when no electron from the install remains.
- After-exit hook: immediate pass on quit.
- SIGTERM → SIGKILL after a grace period (`CODEX_NODE_REPL_REAPER_KILL_GRACE`), with an identity re-check before escalation to guard against pid reuse.
- argv matching accepts both NUL-separated and space-joined `/proc/<pid>/cmdline` (Chromium rewrites its argv area — same quirk as #480).

Tests: `node --test linux-features/node-repl-reaper/test.js` exercises the reap and protect paths against live fake processes. Verified on a live install: a helper owned by a running app-server is left alone; the same helper is reaped once its owner dies.

Note: this branch includes the two commits from #481 (it builds on them).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -81,10 +81,10 @@
 
         codexDmg = pkgs.fetchurl {
           url = "https://persistent.oaistatic.com/codex-app-prod/Codex.dmg";
-          hash = "sha256-Ue7rpYOUxHR8vJ2f7nqmE1ACU/7detWxFPSN/LiabLs=";
+          hash = "sha256-/MTamgmXP9NcA5kBhDSqgOTjoEES0OqYyw7utw+52eg=";
         };
 
-        codexVersion = "26.609.30741";
+        codexVersion = "26.609.41114";
         electronVersion = "42.1.0";
         electronPlatform =
           {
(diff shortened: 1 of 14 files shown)


ilysenko (approved):
Reviewed current head 00a9df327e5b24ab42bf96808b33ce78f812f32a. Rebased/merged onto the already-merged #481 behavior, kept the PR diff feature-only, fixed the watchdog startup race, and verified GitHub CI plus local Linux feature tests.

PR #484, opened 2026-06-13
fix(packaging): order libasound2t64 first to match other t64 deps

IvanTheGeek:
The `Depends` line lists `libasound2 | libasound2t64` (non-t64 first), which is inconsistent with the other three t64 libraries that are already ordered t64-first (`libcups2t64 | libcups2`, `libglib2.0-0t64 | libglib2.0-0`, `libgtk-3-0t64 | libgtk-3-0`).

On Debian 13 (trixie) the real package is `libasound2t64`; `libasound2` is not provided, so the alternative still resolves — but listing the t64 variant first is the correct, consistent form and avoids the resolver preferring a non-existent/transitional name on mixed systems.

Verified: the package builds, installs, and runs on Debian 13 trixie with `libasound2t64 1.2.14-1` (upstream app `26.609.41114`, Electron `42.1.0`).

---

### DIFF
--- a/packaging/linux/control
+++ b/packaging/linux/control
@@ -4,7 +4,7 @@ Section: utils
 Priority: optional
 Architecture: __ARCH__
 Maintainer: Codex Desktop Linux Maintainers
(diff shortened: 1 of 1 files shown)


ilysenko (approved):
Reviewed current head 3034bb1d51736892f85417a1cd35e8ae2bb933f1. CI is green, the local diff is clean, and the change only reorders the Debian libasound alternative to match the existing t64-first dependency policy.

PR #486, opened 2026-06-13
fix(linux): prevent IAB_LIFECYCLE sidebar rebound loop (windowId-based lookup)

### LINKED ISSUE #485: fix(linux): IAB_LIFECYCLE sidebar rebound loop causes text overlap/ghosting
## Bug Description
Codex Desktop Linux shows severe UI glitch in the left sidebar: text overlapping, ghosting, duplicated rendering. Looks like two copies of the sidebar are stacked on top of each other.

## Root Cause
`scripts/patches/main-process/browser.js` → `applyLinuxBrowserUseRouteLivenessPatch` → injected `codexLinuxResolveLiveBrowserUseRouteWindow()` helper.

**The buggy logic (current):**
```js
// 1. Iterates all live windows in the map, if exactly 1 → returns it (regardless of windowId!)
// 2. Calls BrowserWindow.getAllWindows(), if exactly 1 → forces bind to it (regardless of windowId!)
// 3. Otherwise → null (and falls through to "route window is not live" warning)
```

**The problem:**
- During Electron startup, the main window is the ONLY window → step 2 matches and eagerly returns a window state bound to that single window
- Moments later, hidden/utility windows appear (crashpad, gpu, clipboard, etc.) → the bound window is no longer the only one
- IAB_LIFECYCLE tries to resolve the route window, gets the now-stale binding, triggers `rebound` to the same `ownerWebContentsId`
- This creates a cycle of destroy+recreate for the sidebar webview → visual overlap/ghosting

**Evidence from launcher log:**
```
[browser-session-registry] IAB_LIFECYCLE registered browser sidebar window ownerWebContentsId=1 previousOwnerWebContentsId=undefined windowId=1
[browser-session-registry] IAB_LIFECYCLE rebound browser sidebar window ownerWebContentsId=1 previousOwnerWebContentsId=1 windowId=1
[browser-session-registry] IAB_LIFECYCLE rebound browser sidebar window ownerWebContentsId=1 previousOwnerWebContentsId=1 windowId=1
... (infinite loop)
```

## Fix
Replace the heuristic "find any live window" logic with a direct `windowId`-based lookup:

```js
function codexLinuxResolveLiveBrowserUseRouteWindow(e,t,n,r){
  if(process.platform!==`linux`)return null;
  // Direct lookup by windowId
  let o=r.BrowserWindow.fromId(t);
  if(o!=null&&!o.isDestroyed()&&!o.webContents.isDestroyed())
    return e(o,o.webContents);
  // Fallback: check if the requested windowId exists in the state map
  let s=n.get(t)??null;
  return s!=null&&!s.window.isDestroyed()&&!s.owner.isDestroyed()?s:null
}
```

The fix preserves the route liveness fallback for Browser Use but eliminates the incorrect heuristic that caused sidebar lifecycle chaos.

## Affected Versions
- Introduced in commit `3bb0548` / `3a5a7cf` (June 8-12, 2026)
- Affects all builds with ASAR version `26.608.12217` and later

## Environment
- Ubuntu 24.04 LTS, X11
- Electron 42.1.0
- AMD Phoenix3 GPU (RDNA3)
- Mesa 25.2.8

## Tested Fix
Verified on local build — sidebar renders correctly, no IAB_LIFECYCLE rebound loops, no text overlap.

No1Hoo:
## Problem
`codexLinuxResolveLiveBrowserUseRouteWindow()` uses a "first live window" heuristic instead of matching the requested `windowId`. During Electron startup when the main window is the only one, this eagerly binds to it, then hidden windows appear and IAB_LIFECYCLE enters a destroy+recreate rebound loop → sidebar text overlap/ghosting.

Fixes #485

## Change
Replace the heuristic with a direct `BrowserWindow.fromId(windowId)` lookup — the function now correctly uses the `windowId` parameter it was already receiving:

- **Old**: iterate `getAllWindows()` → if exactly 1 → force bind (regardless of windowId)
- **New**: `fromId(windowId)` → if live → create state from it; otherwise check state map by id

## Tested
- Ubuntu 24.04 LTS, X11, Electron 42.1.0, AMD Phoenix3 GPU
- Before: sidebar text overlap/ghosting, IAB_LIFECYCLE rebound loop in logs
- After: sidebar renders correctly, clean IAB_LIFECYCLE init (registered once, no rebounds)
- Browser Use functionality preserved (route window lookup still works)

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -3956,7 +3956,7 @@ test("patches later Browser Use navigation dispatches when an earlier one is alr
   assert.doesNotMatch(patched, /browser-use-non-local-sites-allowed-changed`,\{allowed:p\}/);
 });
 
-test("falls back to the single live Linux Browser Use route window", () => {
+test("resolves the requested live Linux Browser Use route window by id", () => {
   const source =
     "var kK=t.Ur(`browser-sidebar-manager`);" +
     "function AK({ensureWindowState:e,windowId:t,windows:n}){let i=n.get(t)??null;if(i==null){let n=r.BrowserWindow.fromId(t);n!=null&&!n.isDestroyed()&&!n.webContents.isDestroyed()&&(i=e(n,n.webContents))}return i==null||i.window.isDestroyed()||i.owner.isDestroyed()?(kK().warning(`IAB_LIFECYCLE route window is not live`,{safe:{hasWindowState:i!=null,ownerDestroyed:i?.owner.isDestroyed()??null,windowDestroyed:i?.window.isDestroyed()??null,windowId:t},sensitive:{}}),null):i}";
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed current head bc1c6ea. The runtime change now has targeted route-window lookup coverage: exact windowId resolves, unrelated single live windows are not used. Local focused patcher suite passed (244/244), and GitHub checks are green.

PR #489, opened 2026-06-13
[codex] Fix Linux avatar overlay reply focus

joshyorko:
## Summary

- make the avatar overlay focusable on Linux so inline reply inputs accept keyboard input
- keep the focusable Linux overlay above the main Codex window with Linux-only overlay window options
- add regression coverage for the reply-input and overlay stacking behavior

## Root cause

The Linux avatar overlay was still created as non-focusable, which lets it receive mouse interactions but prevents the inline reply input from reliably accepting keyboard input. Making it focusable fixes typing, but on Linux it also removes the implicit window-manager behavior that kept the old non-focusable overlay above the app, so the patch adds explicit Linux-only overlay options for that case.

## Validation

- `node --test --test-name-pattern "avatar overlay|keeps Linux avatar overlay above" scripts/patch-linux-window-ui.test.js`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- The avatar overlay is focusable on Linux so inline pet reply inputs can accept
+  keyboard input after being clicked.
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed current head; no critical blockers found.

PR #490, opened 2026-06-13
[codex] Fix AppShots availability patch

moxw:
## Summary

- update the AppShots availability patch to match the current appshot-availability bundle
- remove stale legacy AppShots compatibility paths from the optional feature patcher
- update focused AppShots feature tests for the current upstream bundle shape

## User-visible behavior

- AppShots availability can be enabled on Linux again when the appshots feature is enabled.

## Root cause

Upstream moved AppShots availability from the old use-is-appshot-available bundle shape to the current appshot-availability atom bundle, so the Linux feature descriptor no longer matched the asset that gates AppShots settings/UI availability.

## Validation

- node --test linux-features/appshots/test.js
- git diff --check
- ./install.sh ./Codex.dmg

### DIFF
--- a/linux-features/appshots/patch.js
+++ b/linux-features/appshots/patch.js
@@ -12,16 +12,16 @@ function warn(message, patchName) {
 }
 
 function applyLinuxAppshotAvailabilityPatch(currentSource) {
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed current head with AppShots feature build; no critical blockers found.

PR #491, opened 2026-06-13
fix: allow address reuse and daemonize threads in webview server

MatrixNeoKozak:
This change ensures the webview backend server can be restarted immediately by enabling SO_REUSEADDR, which prevents 'Address already in use' errors caused by sockets remaining in a TIME_WAIT state. Additionally, daemon_threads is enabled to ensure request threads do not block process termination, improving overall reliability and handling of rapid application restarts.

### DIFF
--- a/launcher/webview-server.py
+++ b/launcher/webview-server.py
@@ -52,6 +52,11 @@ class CodexWebviewHandler(http.server.SimpleHTTPRequestHandler):
         super().end_headers()
 
 
+class CodexHTTPServer(http.server.ThreadingHTTPServer):
+    allow_reuse_address = True
+    daemon_threads = True
+
+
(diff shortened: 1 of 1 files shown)


ilysenko:
Thanks for the PR. I checked the current Python behavior, and `http.server.ThreadingHTTPServer` already has `allow_reuse_address = 1` and `daemon_threads = True`, so this change appears to be a no-op for our current runtime.

If the goal is to fix "Address already in use" during rapid restarts, this likely will not address cases where another live or stale webview server process is still bound to the port. That probably needs to be handled in the launcher lifecycle/cleanup/adoption logic instead.

Could you share a reproduction where the current server fails specifically because address reuse is not enabled? That would help us understand whether there is a real runtime difference we are missing.

PR #497, opened 2026-06-15
[codex] fix Computer Use settings path

ryan-mt:
## Summary

Fix the build-time Computer Use UI settings reader so it resolves the same settings file that the Linux runtime writes.

## Root cause

The patched runtime persists Linux settings through `CODEX_LINUX_SETTINGS_FILE`, or through an app-specific config directory resolved with `CODEX_LINUX_APP_ID` before `CODEX_APP_ID`. The Computer Use UI opt-in reader used during patching only looked under the config directory and preferred `CODEX_APP_ID`, so side-by-side installs and explicit settings-file launches could write the opt-in flag successfully while the patcher still treated Computer Use UI as disabled.

## Changes

- Honor `CODEX_LINUX_SETTINGS_FILE` in `computerUseUiSettingsPath`.
- Prefer `CODEX_LINUX_APP_ID` before `CODEX_APP_ID`, matching launcher/runtime settings resolution.
- Add regression coverage for both settings-path cases and restore the env var in the isolated test helper.

## Verification

- `node --test --test-name-pattern "isComputerUseUiEnabled" scripts/patch-linux-window-ui.test.js`
- `node -c scripts/patches/computer-use.js`
- `node -c scripts/patch-linux-window-ui.test.js`
- `git diff --check --cached`

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -4808,11 +4808,13 @@ function withIsolatedHome(body) {
   const previousXdg = process.env.XDG_CONFIG_HOME;
   const previousAppId = process.env.CODEX_APP_ID;
   const previousLinuxAppId = process.env.CODEX_LINUX_APP_ID;
+  const previousSettingsFile = process.env.CODEX_LINUX_SETTINGS_FILE;
   const previousFlag = process.env[COMPUTER_USE_UI_ENV_VAR];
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed current head eae81357707f05fabcd7f2d1902d2e3bb14c3de2; no blockers found. CI is green and local focused/full patcher validation passed.

PR #498, opened 2026-06-15
[codex] fix Linux XDG documents directory

RockingSisyphus:
## Summary

- add a Linux XDG-aware documents directory resolver for main-process document path lookups
- patch the projectless workspace bundle so generated Codex folders use `$XDG_DOCUMENTS_DIR/Codex` instead of hardcoded `$HOME/Documents/Codex`
- add regression coverage for XDG parsing, projectless bundle patching, and escaped `user-dirs.dirs` values

## Root Cause

Projectless threads ultimately create their generated workspace through a shared `.vite/build/src-*.js` bundle. That bundle hardcoded `path.join(homeDirectory, `Documents`, `Codex`)`, so localized Linux desktops created an extra English `~/Documents` directory even when `~/.config/user-dirs.dirs` points documents to `~/文档`, `~/Dokumente`, etc.

## Validation

- `node scripts/patch-linux-window-ui.test.js`
- rebuilt and installed a local Debian package with `make install-native`
- verified the installed patch report contains `linux-xdg-documents-dir` and `linux-projectless-xdg-documents-dir`

### DIFF
--- a/scripts/patches/core/all-linux/extracted-app/projectless-xdg-documents/patch.js
+++ b/scripts/patches/core/all-linux/extracted-app/projectless-xdg-documents/patch.js
@@ -1,0 +1,19 @@
+"use strict";
+
+const { patchProjectlessDocumentsAssets } = require("../../../../projectless-documents.js");
+
+module.exports = {
+  id: "linux-projectless-xdg-documents-dir",
+  phase: "extracted-app",
+  order: 245,
+  ciPolicy: "optional",
+  apply: patchProjectlessDocumentsAssets,
+  status: (result, warnings) => ({
+    status: result?.changed
+      ? "applied"
+      : result?.matched
+        ? "already-applied"
+        : "skipped-optional",
+    reason: result?.reason ?? warnings[0] ?? null,
+  }),
+};
--- a/scripts/patches/core/all-linux/extracted-app/projectless-xdg-documents/patch.js
+++ b/scripts/patches/core/all-linux/extracted-app/projectless-xdg-documents/patch.js
@@ -11,9 +11,11 @@ module.exports = {
   status: (result, warnings) => ({
     status: result?.changed
       ? "applied"
-      : result?.matched
-        ? "already-applied"
-        : "skipped-optional",
+      : warnings.length > 0
+        ? "skipped-optional"
+        : result?.matched
+          ? "already-applied"
+          : "skipped-optional",
     reason: result?.reason ?? warnings[0] ?? null,
   }),
 };
--- a/scripts/patches/projectless-documents.js
+++ b/scripts/patches/projectless-documents.js
@@ -1,0 +1,96 @@
+"use strict";
+
+const fs = require("node:fs");
+const path = require("node:path");
+
+const {
+  readDirectoryNames,
+} = require("./shared.js");
+
+const PROJECTLESS_DOCUMENTS_MARKER = "function codexLinuxProjectlessDocumentsDir(";
+
+function xdgDocumentsHelperSource() {
+  return [
+    "function codexLinuxProjectlessDocumentsDir({homeDirectory:e,path:t}){try{",
+    "if(process.platform!==`linux`)return t.join(e,`Documents`,`Codex`);",
+    "let n=process.env.XDG_CONFIG_HOME?.trim(),r=n&&t.isAbsolute(n)?t.join(n,`user-dirs.dirs`):t.join(e,`.config`,`user-dirs.dirs`);",
+    "if(!require(`node:fs`).existsSync(r))return t.join(e,`Documents`,`Codex`);",
+    "let i=require(`node:fs`).readFileSync(r,`utf8`).match(/^XDG_DOCUMENTS_DIR=([\"'])(.*)\\1/m);",
+    "if(i==null)return t.join(e,`Documents`,`Codex`);",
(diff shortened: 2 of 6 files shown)


ilysenko:
Thinking out loud here: the XDG Documents change looks useful, and the patch applies cleanly against the current bundle.

The one thing I would like to avoid is hiding future drift in the patch report. Right now `linux-projectless-xdg-documents-dir` can report `already-applied` when it matched a candidate bundle and made no changes, even if the helper emitted a warning because the resolver shape drifted. That makes it harder for us to notice when the optional patch stopped applying correctly.

Could you adjust the status logic so warnings are surfaced as `skipped-optional` or otherwise distinguish "already patched" from "matched but not patched"? After that, please mark the PR ready when you want it reviewed for merge.

RockingSisyphus:
Thanks, that makes sense. I updated `linux-projectless-xdg-documents-dir` so warnings take precedence over the matched/no-change path: if the candidate bundle is matched but the resolver shape drifts and emits a warning, the report now records `skipped-optional` instead of `already-applied`.

I also added a regression test with a deliberately drifted projectless resolver shape to assert that the patch report surfaces the warning as `skipped-optional`.

Validation: `node scripts/patch-linux-window-ui.test.js` passes with 268 tests.

scripts/patches/core/all-linux/extracted-app/projectless-xdg-documents/patch.js
```diff
@@ -0,0 +1,21 @@
...
+const { patchProjectlessDocumentsAssets } = require("../../../../projectless-documents.js");
+
+module.exports = {
+  id: "linux-projectless-xdg-documents-dir",
+  phase: "extracted-app",
+  order: 245,
+  ciPolicy: "optional",
+  apply: patchProjectlessDocumentsAssets,
+  status: (result, warnings) => ({
+    status: result?.changed
```
avifenesh, line 12: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This `status` resolver has a hole when the bundle is found but the resolver shape has drifted.

In that case `patchProjectlessDocumentsAssets` returns `{matched: 1, changed: 0}` and `applyLinuxProjectlessXdgDocumentsDirPatch` emits the drift warning ("Could not find projectless documents directory resolver"). But here the branch order is:

```
result?.changed ? "applied"
: warnings.length > 0 ? "skipped-optional"
: result?.matched ? "already-applied"
: "skipped-optional"
```

That ordering is correct for the drift case (warnings take precedence over `matched` → `skipped-optional`), which matches the test `projectless documents descriptor surfaces resolver drift as skipped optional`.

However, note the inverse: when the patch is genuinely already applied on a re-run, `applyLinuxProjectlessXdgDocumentsDirPatch` returns early via the `PROJECTLESS_DOCUMENTS_MARKER` check with NO warning, so `matched:1, changed:0, warnings:[]` → `already-applied`. Good.

The real gap: `findProjectlessBundles` treats a bundle as a candidate if it contains the marker OR the (`Documents`,`Codex` + `Projectless thread directory`) heuristic. If the upstream resolver drifts such that it still contains both heuristic substrings but no longer matches `resolverRegex`, every candidate yields `changed:0` and emits a warning per file. `matched` is the candidate count, so the resolver returns `skipped-optional` — correct. But because `ciPolicy: "optional"`, this silent demotion of a real bundle-drift to optional means upstream changes to the projectless workspace path will never fail CI. Confirm this is intended: the analogous main-process resolver is also `optional`, so a drift that breaks localized-documents behavior on both paths is invisible in required-CI. Consider `required-upstream` if this is meant to be load-bearing.

scripts/patches/projectless-documents.js
```diff
@@ -0,0 +1,96 @@
...
+
+const PROJECTLESS_DOCUMENTS_MARKER = "function codexLinuxProjectlessDocumentsDir(";
+
+function xdgDocumentsHelperSource() {
+  return [
+    "function codexLinuxProjectlessDocumentsDir({homeDirectory:e,path:t}){try{",
+    "if(process.platform!==`linux`)return t.join(e,`Documents`,`Codex`);",
+    "let n=process.env.XDG_CONFIG_HOME?.trim(),r=n&&t.isAbsolute(n)?t.join(n,`user-dirs.dirs`):t.join(e,`.config`,`user-dirs.dirs`);",
+    "if(!require(`node:fs`).existsSync(r))return t.join(e,`Documents`,`Codex`);",
+    "let i=require(`node:fs`).readFileSync(r,`utf8`).match(/^XDG_DOCUMENTS_DIR=([\"'])(.*)\\1/m);",
```
avifenesh, line 18: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

The `XDG_DOCUMENTS_DIR` value parsing uses `i[2].replace(/\\(.)/g, "$1")` to unescape, and the regex `/^XDG_DOCUMENTS_DIR=(["'])(.*)\1/m` greedily captures to the last matching quote on the line. For a well-formed `user-dirs.dirs` this is fine, but `.*` is greedy and will over-capture if the line contains a trailing quote later (e.g. an inline comment with a quote). Low risk given the file format, but worth confirming `user-dirs.dirs` never has trailing content after the closing quote on the documents line.

avifenesh (commented):

---

Review of the XDG documents directory patch. One correctness concern on the projectless descriptor's `status` resolver where `matched`-but-`changed:0` with a drift warning is mishandled, plus a verification note on the main-process resolver regex. Detail inline.

ilysenko (approved):
Reviewed current head 1ef2a97741b9ef59f9aa6fe45411e1da0c44fef8. No blockers found; CI and local focused patcher tests passed.

RockingSisyphus:
I tightened the `XDG_DOCUMENTS_DIR` parsing so the quoted path stops at the closing quote and only allows trailing whitespace or a `#` comment. This avoids over-capturing when a trailing comment contains another quote, and I applied the same parser to both documents helpers.

PR #504, opened 2026-06-17
[codex] Fix avatar overlay patch drift

ramazangur:
## Summary

Fixes required Linux avatar overlay ASAR patch drift caused by the current upstream DMG bundle shape.

## Root cause

The avatar overlay patch still depended on exact minified needles for `startDrag`, `endDrag`, `applyLayout`, and `showWindow`. The current upstream bundle changed those method bodies: drag state moved to a class-style object, drag/end handlers accept pointer event data, `applyLayout` has additional parameters, and `showWindow` now groups presentation completion with open-state broadcasting. Those valid upstream changes made the required `linux-avatar-overlay-mouse-passthrough` descriptor report warnings, which then made `main-process-ui` fail under `--enforce-critical` and blocked `bootstrap-native`.

## What changed

- Switched the avatar overlay patcher to method-scoped matching for the drifted methods instead of brittle exact cross-method needles.
- Scoped avatar method lookups to the `/avatar-overlay` region so unrelated earlier methods in the main bundle cannot be patched silently.
- Kept the Linux pointer interactivity, layout sync, compositor hint, and drag cleanup injections idempotent.
- Updated the current avatar overlay fixture to cover the new upstream method shapes and assert zero avatar drift warnings.
- Added a regression test for an unrelated earlier class with matching method names.

## Source of truth

- `scripts/patches/avatar-overlay.js`

## Validation

- `git diff --check`
- `node --test scripts/patch-linux-window-ui.test.js` (`269/269` passing)
- Fresh cached `Codex.dmg` ASAR extract with `node scripts/patch-linux-window-ui.js --enforce-critical --report-json ...` reports no critical failures; `linux-avatar-overlay-mouse-passthrough` and `main-process-ui` both apply.

## Notes

The enforced DMG run still reports pre-existing optional drift warnings for Browser Use `node_repl`, app-server backfill wait, and external Browser Use availability. They are not critical blockers for this fix.

### DIFF
--- a/scripts/patches/avatar-overlay.js
+++ b/scripts/patches/avatar-overlay.js
@@ -24,15 +24,15 @@ function findAvatarMethod(source, signatureRegex) {
   };
 }
 
-function findAvatarMethodAfter(source, signatureRegex, startIndex) {
-  const match = source.slice(startIndex).match(signatureRegex);
+function findAvatarMethodAfter(source, signatureRegex, startIndex, endIndex = source.length) {
+  const match = source.slice(startIndex, endIndex).match(signatureRegex);
   if (match == null) {
     return null;
   }
   const absoluteIndex = startIndex + match.index;
   const openIndex = absoluteIndex + match[0].length - 1;
   const closeIndex = findMatchingBrace(source, openIndex);
-  if (closeIndex === -1) {
+  if (closeIndex === -1 || closeIndex + 1 > endIndex) {
     return null;
   }
   return {
@@ -52,8 +52,38 @@ function avatarOverlayRegionStart(source) {
   return stateMessageIndex === -1 ? 0 : stateMessageIndex;
 }
 
+function findAvatarOverlayClass(source) {
+  const classRegex = /class(?:\s+[A-Za-z_$][\w$]*)?(?:\s+extends\s+[A-Za-z_$][\w$.]*)?\{/g;
+  classRegex.lastIndex = avatarOverlayRegionStart(source);
+  let match;
+  while ((match = classRegex.exec(source)) != null) {
+    const openIndex = match.index + match[0].length - 1;
+    const closeIndex = findMatchingBrace(source, openIndex);
+    if (closeIndex === -1) {
+      return null;
+    }
+    const text = source.slice(match.index, closeIndex + 1);
+    if (
+      text.includes("appearance:`avatarOverlay`") ||
+      text.includes("avatar-overlay-open-state-changed")
+    ) {
+      return {
+        start: match.index,
+        end: closeIndex + 1,
+        text,
+      };
+    }
+    classRegex.lastIndex = closeIndex + 1;
+  }
+  return null;
+}
+
 function findAvatarOverlayMethod(source, signatureRegex) {
-  return findAvatarMethodAfter(source, signatureRegex, avatarOverlayRegionStart(source));
+  const overlayClass = findAvatarOverlayClass(source);
+  if (overlayClass == null) {
+    return null;
+  }
+  return findAvatarMethodAfter(source, signatureRegex, overlayClass.start, overlayClass.end);
 }
 
 function replaceAvatarMethod(source, signatureRegex, replacement) {
--- a/scripts/patches/avatar-overlay.js
+++ b/scripts/patches/avatar-overlay.js
@@ -24,8 +24,47 @@ function findAvatarMethod(source, signatureRegex) {
   };
 }
 
+function findAvatarMethodAfter(source, signatureRegex, startIndex) {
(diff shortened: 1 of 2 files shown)


avifenesh (commented):

---

Reviewed the avatar-overlay patch drift fix. The switch from cross-method exact needles to method-scoped matching (`findAvatarMethod` + `replaceAvatarMethodText`) is sound: each method (startDrag, endDrag, applyLayout, showWindow) is located by signature, the matching brace is found via `findMatchingBrace`, and the injection is idempotent (each branch first checks for the already-applied marker). The `replaceAvatarMethodText` helper mirrors the existing `replaceAvatarMethod` and operates on a single bounded method slice, so there is no risk of a `string.replace()` hitting the wrong occurrence (P1) — the `.replace()` calls run on `method.text`, which is scoped to one method body.

One item worth confirming (not blocking):

- `findAvatarMethod` uses `source.match(signatureRegex)` which returns the *first* match in the whole bundle. For `startDrag\([^)]*\)\{` and `showWindow\(([A-Za-z_$][\w$]*)\)\{`, verify these signatures are unique across the bundle so the first match is always the avatar overlay method. The `applyLayout` regex already guards against member-access via `(?<![\w$.])`, but `startDrag`/`endDrag`/`showWindow` have no such guard. If any other class defines a same-named method earlier in the bundle, the wrong method would be patched silently. The fixture covers the avatar class in isolation, so the test passing does not prove uniqueness in the real DMG.

No correctness defect with citable evidence beyond that; the descriptor ciPolicy handling and warning behavior are unchanged by this diff.

avifenesh (commented):

---

Reviewed the avatar overlay patch drift fix. The migration from cross-method exact needles to method-scoped matching (`findAvatarOverlayMethod`) is sound, and scoping lookups to the `/avatar-overlay` region addresses the P3/P1-adjacent risk of patching unrelated earlier methods. One concern below worth confirming about the region-scoping anchor.

scripts/patches/avatar-overlay.js
```diff
@@ -148,46 +187,57 @@ function applyLinuxAvatarOverlayMousePassthroughPatch(currentSource) {
...
-  const startDragPatch =
-    `displayBounds:${electronVar}.screen.getDisplayNearestPoint(${electronVar}.screen.getCursorScreenPoint()).bounds},process.platform===\`linux\`&&(this.pointerInteractive=!0,this.applyPointerInteractivityPolicy())}moveDrag(e){`;
-  const startDragRegex =
-    /displayBounds:([A-Za-z_$][\w$]*)\.screen\.getDisplayNearestPoint\(\1\.screen\.getCursorScreenPoint\(\)\)\.bounds\}\}moveDrag\(e\)\{/;
-  const currentStartDragPatchRegex =
-    /displayBounds:([A-Za-z_$][\w$]*)\.screen\.getDisplayNearestPoint\(\1\.screen\.getCursorScreenPoint\(\)\)\.bounds\},process\.platform===`linux`&&\(this\.pointerInteractive=!0,this\.applyPointerInteractivityPolicy\(\)\)\}moveDrag\(e\)\{/;
-  if (currentStartDragPatchRegex.test(patchedSource)) {
+  const startDragMethod = findAvatarOverlayMethod(
+    patchedSource,
+    /startDrag\([^)]*\)\{/,
```
avifenesh, line 222: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`findAvatarOverlayMethod` scopes matching to the first occurrence of `` `/avatar-overlay` `` (the `rV` route literal), which in the current bundle sits before the `fV` overlay class. But the new `h2` drag-state helper class is also defined after that anchor and contains a `recordMovementIntent`/`getCursorPointForSource` body — not `startDrag`/`endDrag`. That's fine here. The real fragility: the region anchor only marks the *start*; there is no end bound. If a future upstream bundle places an unrelated class with a `startDrag`/`endDrag`/`showWindow`/`applyLayout` method *after* the `` `/avatar-overlay` `` literal but *before* the `fV` overlay class (e.g. the `h2`-style helpers gaining such a method, or reordering), `findAvatarMethodAfter` will match that earlier method instead of the overlay's. The added regression test only covers an unrelated class placed *before* the anchor. Consider anchoring to the overlay class body start (e.g. the `fV=class{...compositionHost` signature) rather than the route literal, or asserting the matched method falls within the overlay class braces, to avoid silent mis-patching on the next drift.

ilysenko (approved):
Reviewed the current head after the matcher tightening. The avatar overlay method lookup is now bounded to the overlay class body, and the added regression covers an unrelated class after the /avatar-overlay anchor. Local patcher tests and all GitHub checks are green.

PR #511, opened 2026-06-17
fix(linux): backfill helpers at top-level scope (not inside fi())

No1Hoo:
## Summary

- `applyLinuxAppServerBackfillWaitPatch()` helper functions (`codexLinuxAppServerBackfillTimeoutMs`, etc.) were inserted inside `function fi()` (Sentry error handler) when neither legacy anchors nor `"use strict"` matched
- `createRequest()` calls `codexLinuxAppServerBackfillTimeoutMs()` from a class method at depth 2 — invisible inside `fi()` at depth 1
- Caused `ReferenceError: codexLinuxAppServerBackfillTimeoutMs is not defined` at login time

## Root Cause

The helper insertion logic used `"function za(e){...}"` as the anchor. When the DMG's webview bundle didn't contain that anchor (new bundle layout), the fallback matched `function fi()` — a **different top-level function**. Helpers ended up at depth 1 inside `fi()`, invisible to `createRequest()` at depth 2 in a class method.

## Fix

Add `"function fi("` as the **primary anchor** for helper insertion. Since `fi()` is a top-level function in the webview bundle, inserting the helpers **before** it places them at depth 0, visible to all scopes. Legacy anchors preserved as fallback for older bundle layouts.

**Verification:**
```
Helper def at depth: 0 ✅ (top-level)
fi() follows helper: ✅
Call site depth: 2 (visible from depth 0)
```

## Files Changed

- `scripts/patches/webview-assets.js` — anchor priority chain for backfill helper insertion

### DIFF
--- a/scripts/patches/webview-assets.js
+++ b/scripts/patches/webview-assets.js
@@ -701,15 +701,29 @@ function applyLinuxAppServerBackfillWaitPatch(currentSource) {
   let changed = false;
 
   if (!patchedSource.includes("function codexLinuxIsStateDbBackfillMessage(")) {
-    const helperAnchors = [
+    // Insert helpers at module top-level so they're visible to ALL scopes.
+    // The helpers must not land inside function fi() (Sentry handler) —
+    // createRequest() calls them from a different scope.
+    const fiAnchor = "function fi(";
+    const legacyAnchors = [
       "function za(e){let t=La.safeParse(e);return t.success?new Ba(t.data):e}",
       "function za(e){",
     ];
-    const helperAnchor = helperAnchors.find((anchor) => patchedSource.includes(anchor));
-    if (helperAnchor != null) {
-      patchedSource = patchedSource.replace(helperAnchor, `${helperSource}${helperAnchor}`);
+    let inserted = false;
+    if (patchedSource.includes(fiAnchor)) {
+      patchedSource = patchedSource.replace(fiAnchor, `${helperSource}${fiAnchor}`);
       changed = true;
-    } else if (shouldPatchTimeout) {
+      inserted = true;
+    }
+    if (!inserted) {
+      const legacyAnchor = legacyAnchors.find((anchor) => patchedSource.includes(anchor));
+      if (legacyAnchor != null) {
+        patchedSource = patchedSource.replace(legacyAnchor, `${helperSource}${legacyAnchor}`);
+        changed = true;
+        inserted = true;
+      }
+    }
+    if (!inserted && shouldPatchTimeout) {
       const timeoutMatch = patchedSource.match(timeoutNeedle);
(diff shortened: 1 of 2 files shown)


scripts/patches/webview-assets.js
```diff
@@ -701,15 +701,29 @@ function applyLinuxAppServerBackfillWaitPatch(currentSource) {
...
+    const legacyAnchors = [
       "function za(e){let t=La.safeParse(e);return t.success?new Ba(t.data):e}",
       "function za(e){",
     ];
-    const helperAnchor = helperAnchors.find((anchor) => patchedSource.includes(anchor));
-    if (helperAnchor != null) {
-      patchedSource = patchedSource.replace(helperAnchor, `${helperSource}${helperAnchor}`);
+    let inserted = false;
+    if (patchedSource.includes(fiAnchor)) {
+      patchedSource = patchedSource.replace(fiAnchor, `${helperSource}${fiAnchor}`);
```

---

The new primary anchor `"function fi("` is matched via `patchedSource.includes(...)` and then patched with a single `String.replace()`, which only rewrites the **first** occurrence. Two related risks per the ASAR-patch correctness pattern (replace ALL eligible occurrences / assert the needle is unique):

1. `"function fi("` is an extremely short, generic minified-identifier anchor. In a large minified webview bundle, a 2-char top-level function name like `fi` is not guaranteed unique — other functions/identifiers could collide, and even an unrelated `function fi(` declaration would be the insertion point. Unlike the legacy anchors (which carry a distinctive body), this anchor has no body specificity, so a collision silently inserts the helpers before the wrong function.

2. The helper block is inserted before the **first** `function fi(` match. If the bundle minifier ever assigns `fi` to a function that is itself nested (not top-level), the depth-0 guarantee the PR relies on breaks — and there is no assertion verifying `fi` is top-level at the matched site.

Consider anchoring on a more specific signature (e.g. `function fi(` plus a stable prefix of its body, mirroring how the legacy `za` anchor includes its body), or assert the match is unique/top-level before inserting. Note the added tests (lines 3531, 3547) do not exercise the new `fiAnchor` path at all — both existing fixtures hit the legacy `za(` / class-scan branches, so the new primary branch is currently untested.

avifenesh (commented):

---

Review of the backfill helper insertion anchor change in `scripts/patches/webview-assets.js`.

One concern about the new `"function fi("` primary anchor — flagged inline. The fix logic itself (top-level insertion to make helpers visible from `createRequest()` at depth 2) is sound; the worry is about anchor uniqueness/specificity given this repo's match-correctness pattern (P1).

ilysenko (approved):
Reviewed the updated head after tightening the helper insertion anchor and adding regression coverage. Local build and GitHub CI are green; no blockers found.

PR #513, opened 2026-06-17
[codex] x11 computer use improvements

Trevongit:
Adds the current computer-use work to the Linux wrapper and backend: shared-pid accessibility selection, Cinnamon/X11 compositor fallback, X11 package dependency wiring, and the opt-in X11/EWMH feature staging/tests.

Also includes the two local recovery notes written during the package-repair step.

Checks:
- node --test linux-features/x11-ewmh-computer-use/test.js
- cargo test -p codex-computer-use-linux select_accessibility_object_ref -- --nocapture
- git diff --cached --check


ilysenko:
Thanks for the work here. I can see useful ideas in the X11/computer-use part, but this PR has grown far beyond that scope.

The main issue is that it mixes a focused feature change with a lot of unrelated personal/local workflow infrastructure. We need changes in this repo to stay narrowly scoped and reviewable, especially around fresh install behavior, packaging, launcher logic, and updater rebuilds. Personal workflow helpers, local root certificates, fork-sync scripts, recovery notes, and machine-specific runbooks should not be bundled into core wrapper changes.

Concrete issues I found:

- `build-info.js` now depends on `scripts/organ-registry.js`, but the packaged update-builder does not stage that file, so updater rebuilds would break.
- Build metadata would include local machine paths and personal fork identity, for example `/home/trev/...` and `Trevongit`, and expose that in the app UI.
- The new `easy-update` flow assumes a remote named `upstream`, which is not true for normal clones.
- `publish-fork` blindly pushes to `origin`, which may be the main repository depending on the checkout.
- The launcher silently edits the user’s existing `~/.codex/state_5.sqlite` on cold start. Our supported path is fresh install behavior, not mutating old installed state.
- User-local install/update now always recreates desktop shortcuts and even creates `~/Desktop`, which ignores user preference and breaks idempotency.
- `make deploy` runs a report that fetches external OpenAI pages without a timeout, so a local install/deploy flow can hang on network issues.
- The X11 feature dependency handling is inconsistent: `xdotool` becomes a hard dependency for all native packages, even though the feature is optional, while the actual X11 plugin also needs `wmctrl` and `xprop`.
- The source overlay staging is not idempotent if the source tree is already patched.
- Several added docs/scripts are clearly local-machine workflow material rather than shared project behavior.

Please rework this as a much smaller, focused change. The next PR should contain only the X11/computer-use fresh-install pieces, with no personal workflow docs, no root-certificate metadata, no fork-sync helpers, no old-state database migration, and no unrelated desktop/update helper changes.

In general, please send changes in smaller steps: one purpose per PR, only the files needed for that purpose, and behavior that works from a normal fresh clone/fresh install.

Trevongit:
Thanks Gary — that makes sense.

I see this PR is now closed, which is fine. This was not intended to become
a broad upstream contribution in its current form. I am developing my fork
for my own local workflow and accidentally let personal/local workflow
infrastructure get bundled into a PR against upstream.

I agree that the X11/computer-use part is the only piece that might make
sense upstream, and that it would need to be separated into a much smaller
focused PR with no local workflow docs, root-certificate metadata,
fork-sync scripts, recovery notes, or machine-specific changes.

I’ll continue this work in my personal fork. If I later prepare anything
for upstream, I’ll keep it narrow, fresh-install focused, and limited to
the specific feature area.

Thanks for the review — it helped clarify the boundary.

PR #517, opened 2026-06-18
[codex] Fix Linux desktop settings row import

colonelpanic8:
## Summary

- Fix Linux Desktop settings rows after upstream `settings-row` export drift.
- Infer the correct `SettingsRow` export from the actual settings row asset instead of assuming `n`.
- Render the Linux source commit value as a direct clickable link when build metadata includes a commit URL.
- Update the native Keyboard Shortcuts fixture to model the current upstream row export map.

## Root Cause

Current upstream exports the `{ label, description, control }` settings row component as `r`, while `n` is now a grid row that accepts `{ label, children }`. The generated Linux Desktop settings page imported `n`, so labels rendered but descriptions and controls were empty.

## Validation

- `git diff --check`
- `node --test --test-name-pattern 'Linux desktop|Keybinds settings' scripts/patch-linux-window-ui.test.js`


ilysenko (approved):
Reviewed the current head and found no blockers. CI is green and the local patcher test suite passes.

PR #518, opened 2026-06-19
Fix Electron 42 required patch drift

moxw:
## Summary

- update the Linux tray patch for the Electron 42 main-process bundle shapes
- retarget the subagent nickname metadata descriptor to the Electron 42 webview bundle name
- keep both fixes in one PR because upstream-build tests all required patches against the same upstream DMG

## Validation

- MAX_BUILD_THREADS=0 ./install.sh --inspect --report-dir /tmp/codex-inspect-42-fixed ./Codex.dmg
- make build-app-fresh
- make package


scripts/patches/core/all-linux/extracted-app/owl-feature-binding/patch.js
```diff
@@ -0,0 +1,11 @@
...
+
+const { patchLinuxOwlFeatureBindingFallbackAssets } = require("../../../../main-process.js");
+
+module.exports = {
+  id: "linux-owl-feature-binding-fallback",
+  phase: "extracted-app",
+  order: 190,
+  ciPolicy: "required-upstream",
+  apply: patchLinuxOwlFeatureBindingFallbackAssets,
+};
```
avifenesh, line 20: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This descriptor is `phase: "extracted-app"`, `ciPolicy: "required-upstream"`, and `apply` returns `patchLinuxOwlFeatureBindingFallbackAssets`, which returns `{ matched: 0, changed: 0 }` whenever no bundle in `.vite/build` contains `electron_common_owl_features` (e.g., the upstream binding loader is renamed/relocated so the string no longer appears, or the dir is missing).

Unlike the webview-asset path (which routes `matched === 0` through `recordAssetDescriptorPatch` → `descriptorFailureStatus` → `failed-required`), the extracted-app path in `engine.js` `applyExtractedAppPatchDescriptors` computes status purely from `result.changed`:

```js
result?.changed != null
  ? patchStatusFromChange(Boolean(result.changed), warnings, descriptor.ciPolicy)
  : "applied"
```

With `changed: 0` and no warnings, `patchStatusFromChange` returns `"already-applied"` — a SUCCESS_STATUS — so genuine upstream drift (binding string gone) is silently treated as success and never surfaces as `failed-required`. This is exactly the P2 failure mode (required-upstream patch demoted to skipped/passing). `patchLinuxOwlFeatureBindingFallbackPatch` also only `console.warn`s when the string is present but the regex shape drifts; if the whole bundle is gone there is no warning at all.

Sibling extracted-app descriptors that need matched-vs-changed disambiguation (e.g. `package-metadata`, `projectless-xdg-documents`) supply a custom `status` function. Consider adding a `status` that maps `matched === 0` to `failed-required` here so binding drift is not invisible.

avifenesh (commented):

---

Reviewed the Electron 42 patch-drift fixes. The tray-bundle retargeting (`appBrand`/`small` size handling), subagent-metadata descriptor widening, and the new Owl-feature-binding fallback patch are well-covered by tests. One concern about the new required-upstream extracted-app descriptor's drift visibility.

scripts/patches/core/all-linux/extracted-app/owl-feature-binding/patch.js
moxw: Addressed in 232c1c8: the descriptor now has a custom status that maps matched === 0 to failed-required for required-upstream drift, while preserving patchStatusFromChange for matched bundles. Added a regression covering the missing Owl binding bundle path in upstream-build validation.

scripts/patches/core/all-linux/webview/subagent-metadata/patch.js
```diff
@@ -10,8 +10,8 @@ module.exports = [
     phase: "webview-asset",
     order: 1050,
     ciPolicy: "required-upstream",
-    pattern: /^app-server-manager-signals-.*\.js$/,
-    missingDescription: "app-server manager webview bundle",
+    pattern: /^(?:app-server-manager-signals|use-host-config)-.*\.js$/,
```
avifenesh, line 13: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This pattern now matches two distinct bundle families (`app-server-manager-signals-*` and `use-host-config-*`). Because `apply: applySubagentNicknameMetadataPatch` runs per matched file and the descriptor is `ciPolicy: "required-upstream"`, the engine's `recordAssetDescriptorPatch` reports success when `matched > 0` and `changed > 0` for the set as a whole — it does not require that *every* matching bundle was actually changed.

If both `app-server-manager-signals-*.js` and `use-host-config-*.js` are present but the metadata shape only lives in one of them, the patch is fine. But if the shape moved to `use-host-config` AND `app-server-manager-signals` still exists without the shape, `applySubagentNicknameMetadataPatch` should be a no-op on the non-matching file (not a warning), otherwise `patchAssetFiles` could emit a warning and flip a required descriptor to `failed-required` on a benign file. Please confirm `applySubagentNicknameMetadataPatch` returns the source unchanged without warning when the metadata literal is absent, so the widened pattern doesn't produce false required failures on the sibling bundle.

avifenesh (commented):

---

Reviewed the Electron 42 patch-drift fixes (tray bundle retargeting, subagent metadata pattern widening, and the new Owl feature binding fallback patch).

Key observations:

- The Owl feature binding fallback is correctly wired as `phase: "extracted-app"`, `ciPolicy: "required-upstream"`, with a custom `status` mapping `matched === 0 → failed-required`, so upstream bundle drift surfaces as a required failure rather than being silently demoted (matches the prior review feedback). Verified against `applyExtractedAppPatchDescriptors`/`patchStatusFromChange`.
- `patchLinuxOwlFeatureBindingFallbackAssets` scans all `.vite/build/*.js` candidates and processes every matching occurrence (no single-`replace()` blind spot), with deterministic `.sort()` and a stable pending-write phase.
- The tray patch's `legacyGetFileIconSize` capture group and `sizeExpression` callback preserve the original `getFileIcon` size argument across both the new `small`-only and legacy `win32?small:normal` bundle shapes.
- The subagent-metadata pattern widening to `(?:app-server-manager-signals|use-host-config)-.*\.js` retains `required-upstream` and is covered by the new descriptor test asserting both bundle names match and an unrelated bundle does not.

One point worth confirming inline below regarding the widened webview pattern.

scripts/patches/core/all-linux/webview/subagent-metadata/patch.js
moxw: Confirmed and covered in a1407e2: applySubagentNicknameMetadataPatch stays quiet for matched sibling bundles with no metadata markers. The regression keeps both app-server-manager-signals-* and use-host-config-* present, then asserts matched=2, changed=1, no warnings, the sibling file remains unchanged, and only the metadata bundle is patched.

ilysenko (approved):
Reviewed the current head for the fresh-install/current-upstream flow. No blockers found; CI and local patcher tests are green.

PR #519, opened 2026-06-19
[codex] Retarget subagent metadata patch for Electron 42

moxw:
## Summary

- retarget the required subagent nickname metadata patch to the Electron 42 `use-host-config-*` webview bundle
- keep the legacy `app-server-manager-signals-*` match for older upstream bundles
- add descriptor coverage for both bundle names

## Root cause

The upstream Electron 42.1.0 DMG moved the subagent metadata shape out of `app-server-manager-signals-*` and into `use-host-config-*`. The patch body still applies, but the descriptor no longer found the asset, causing critical patch enforcement to fail.

## Validation

- Aggregate branch validation with the tray fix from #518:
  - `MAX_BUILD_THREADS=0 ./install.sh --inspect --report-dir /tmp/codex-inspect-42-fixed ./Codex.dmg`
  - `make build-app-fresh`
  - `make package` after rerunning unsandboxed because sandboxed `fakeroot` could not preserve ownership

The aggregate build generated `codex-app/.codex-linux/patch-report.json` with required core counts `applied=13, already-applied=3` and `failed-required=0`.

Note: this PR is intentionally split from the tray fix. Full Electron 42 rebuild validation requires both required-patch fixes together.


moxw:
Superseded by #518, which now contains both required Electron 42 patch drift fixes so upstream-build can test them together.

PR #522, opened 2026-06-19
fix(linux): stub Owl feature binding to fix startup crash

### LINKED ISSUE #520: App crashes on startup: "No such binding was linked: electron_common_owl_features" (plus broken patches in latest version)
Image
### Describe the bug
When building the latest version of the application, the build process fails at the patching stage. If bypassed using `CODEX_ENFORCE_CRITICAL_PATCHES=0`, the build completes but the resulting app crashes immediately on startup with the following error dialog:

`Codex failed to start. No such binding was linked: electron_common_owl_features`

### To Reproduce
1. Run the standard build pipeline (which fetches the latest upstream macOS `.dmg`).
2. The build will initially fail with the following critical patch failures:
   - `linux-tray` (Could not find tray icon fallback)
   - `main-process-ui` (Could not find tray icon fallback)
   - `subagent-nickname-metadata-shape` (Could not find app-server manager webview bundle)
3. Set `CODEX_ENFORCE_CRITICAL_PATCHES=0` to bypass the build failures.
4. Install the resulting `.deb` or `AppImage` on Linux.
5. Launch the application. It will immediately show the `electron_common_owl_features` crash dialog.

### Expected behavior
The application should successfully launch without trying to load upstream custom bindings that aren't present in standard Linux Electron builds. The broken patches also need to be updated for the new upstream code structure.

### Root Cause & Potential Fix
It seems the newest upstream app version introduces a custom Electron binding (`electron_common_owl_features`). Because standard Linux Electron doesn't have this native module, calling `process.binding('electron_common_owl_features')` causes a fatal crash.

As a temporary workaround, I was able to get the app to launch by injecting a new patch into `scripts/patches/core/all-linux/main-process/owl-stub/patch.js` to stub it out with a Proxy:

```javascript
module.exports = [{
  id: "linux-owl-features-stub",
  phase: "main-bundle",
  order: 10,
  ciPolicy: "optional",
  apply: (source) => source.replace(/process\.binding$['"\`]?electron_common_owl_features['"\`]?$/g, "(new Proxy({}, { get: () => () => {} }))")
}];

### PR DESCRIPTION
**Pull request by kj49**
## Summary

Adds a patch that stubs OpenAI's internal "Owl" feature-flag binding so the rebuilt Linux app stops crashing on startup. Fixes #520.

Codex `26.616.x+` reads Owl flags through a native binding `electron_common_owl_features` that exists only in OpenAI's private Electron build. On the stock Electron this wrapper rebuilds against, the binding is absent, so the accessor throws **`No such binding was linked: electron_common_owl_features`** at `bootstrap-import-main` and the app dies with a "Codex failed to start" dialog.

The accessor is the binding's sole consumer and its only used member is `isOwlFeatureEnabled(name)`. This rewrites the function body to a stub returning `{ isOwlFeatureEnabled: () => false }`, dropping the binding call entirely. Owl-gated features default off; everything else is untouched. The file-based owl-feature-bootstrap-cache path is independent and keeps working.

## Why an extracted-app patch (not main-bundle)

The workaround posted in #520 targets `process.binding('electron_common_owl_features')` with a `main-bundle` patch. On current builds that misses, because:

- the real call is `process._linkedBinding(...)`, and
- the accessor lives in a **content-hashed chunk** (`workspace-root-drop-handler-*.js`), not the main bundle.

So this patch runs in the `extracted-app` phase and scans every `.vite/build/*.js`, patching whichever chunk contains the binding literal.

## Implementation notes

- Anchored on the stable string literals + structural shape; minified identifiers change every build, so the captured function name is preserved so existing callers still resolve.
- `ciPolicy: "optional"` so a future shape drift emits a loud warning (`owl stub NOT applied; app will crash...`) instead of hard-failing the whole build — the post-install launch is the real safety net.
- No-op when the literal is absent (i.e. once upstream drops the binding).

## Validation

- `make build-app DMG=./Codex.dmg` then `make rpm` + install: patch-report shows `linux-owl-feature-binding-stub` = `applied`; repacked `app.asar` has the stub and zero `electron_common_owl_features` references.
- Installed build launches cleanly on Fedora 44 / stock Electron 42.1.0 — no bootstrap/owl error, all windows load.

> Orthogonal to #518, which fixes the `linux-tray` / `subagent-nickname-metadata-shape` patch drift the same issue reports. This PR only addresses the Owl binding crash.

### DIFF
--- a/scripts/patches/core/all-linux/extracted-app/owl-feature-binding/patch.js
+++ b/scripts/patches/core/all-linux/extracted-app/owl-feature-binding/patch.js
@@ -1,0 +1,82 @@
+"use strict";
+
+const fs = require("node:fs");
+const path = require("node:path");
+
+// Codex 26.616.x+ reads an internal "Owl" feature-flag set through a native
+// binding that only exists in OpenAI's private Electron build:
+//   function Q(){let e=process._linkedBinding;...;return G.parse(e.call(process,`electron_common_owl_features`))}
+// On stock upstream Electron (what this wrapper rebuilds against) that binding
+// is absent, so the call throws "No such binding was linked:
+// electron_common_owl_features" the moment the main bundle imports — the app
+// dies at bootstrap-import-main with a "Codex failed to start" dialog.
+//
+// The accessor is the sole consumer of the binding; its only used member is
+// isOwlFeatureEnabled(name). Rewrite the function body to return a stub that
+// reports every Owl feature as disabled, dropping the binding call entirely.
+// Owl-gated features default off; everything else is untouched. The file-based
+// owl-feature-bootstrap-cache path is independent and keeps working.
+//
+// Minified identifiers (the function name and the schema-parser name) change
+// every build, so anchor on the stable string literals and structural shape,
+// and preserve the captured function name so existing callers still resolve.
+const OWL_BINDING_REGEX =
+  /function (\w+)\(\)\{let \w+=process\._linkedBinding;if\(typeof \w+!=`function`\)throw [\w.]*Error\(`Owl feature binding is unavailable`\);return \w+\.parse\(\w+\.call\(process,`electron_common_owl_features`\)\)\}/;
+
+const OWL_BINDING_LITERAL = "electron_common_owl_features";
+
+function stubOwlAccessor(source) {
+  return source.replace(
+    OWL_BINDING_REGEX,
+    (_match, fnName) => `function ${fnName}(){return{isOwlFeatureEnabled:()=>!1}}`,
+  );
+}
+
+function patchLinuxOwlFeatureBinding(extractedDir) {
+  const buildDir = path.join(extractedDir, ".vite", "build");
+  if (!fs.existsSync(buildDir)) {
+    return { changed: false, reason: ".vite/build not found" };
+  }
+
+  let changedFiles = 0;
+  let sawLiteral = false;
+  for (const name of fs.readdirSync(buildDir).sort()) {
+    if (!name.endsWith(".js")) {
+      continue;
+    }
+    const target = path.join(buildDir, name);
+    const source = fs.readFileSync(target, "utf8");
+    if (!source.includes(OWL_BINDING_LITERAL)) {
+      continue;
+    }
+    sawLiteral = true;
+    const patched = stubOwlAccessor(source);
+    if (patched !== source) {
+      fs.writeFileSync(target, patched, "utf8");
+      changedFiles += 1;
+    }
+  }
+
+  if (changedFiles === 0) {
+    if (sawLiteral) {
+      console.warn(
+        "WARN: Owl feature binding accessor shape changed — owl stub NOT applied; " +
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scripts/patches/core/all-linux/extracted-app/owl-feature-binding/patch.js [hunk 1]
```diff
@@ -0,0 +1,82 @@
...
+    if (patched !== source) {
+      fs.writeFileSync(target, patched, "utf8");
+      changedFiles += 1;
+    }
+  }
+
+  if (changedFiles === 0) {
+    if (sawLiteral) {
+      console.warn(
+        "WARN: Owl feature binding accessor shape changed — owl stub NOT applied; " +
```
avifenesh, line 63: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`stubOwlAccessor` only rewrites the accessor function body, but the `electron_common_owl_features` literal also appears in the helper comment / regex matching context — more importantly, after a successful rewrite the resulting stub no longer contains the literal, but the `sawLiteral`/warning logic keys off the literal being present in the *pre-patch* source. That part is fine.

The real risk: if the binding accessor's minified shape drifts so `OWL_BINDING_REGEX` no longer matches, but the literal `electron_common_owl_features` still appears elsewhere in the chunk (e.g. inside the file-based bootstrap-cache path the PR notes is independent), then `sawLiteral` is true, `changedFiles` stays 0, and the code emits the 'shape changed — owl stub NOT applied' warning even though the crashing accessor may genuinely be gone. Conversely, if the literal only exists in that independent cache path and the crashing accessor was renamed/removed, the warning is a false alarm. Consider anchoring `sawLiteral` on a match of the accessor structural shape (process._linkedBinding + the literal together), not the bare literal, so the drift warning tracks the actual accessor rather than any occurrence of the string. Worth confirming the literal appears only in the accessor on current bundles.

avifenesh (commented):

---

Review of the new `linux-owl-feature-binding-stub` extracted-app patch.

One concern worth confirming before merge regarding the warning path when the literal is present but the regex misses (see inline). Otherwise the patch follows repo conventions: it scans every `.vite/build/*.js` (P1 — replaces in every chunk that contains the literal rather than a single bundle), preserves the captured function name so callers still resolve, and uses `ciPolicy: "optional"` with a loud warn-on-drift consistent with sibling optional extracted-app patches.

ilysenko:
Thanks for the quick Owl binding fix. I rechecked this against the fresh-install/current-upstream flow after #518 landed.

This PR is now superseded by #518. #518 already includes the Owl binding fallback, plus the Electron 42 tray and subagent metadata drift fixes needed for the current upstream DMG, and its upstream-build check is green.

This branch is now conflicting with main, and its upstream-DMG check fails on the tray/subagent required patches that #518 fixed. Reapplying it on top of main would also replace the #518 Owl implementation with a weaker optional stub and drop the required-upstream drift guard/tests.

For the current fresh-install/current-upstream scope, I’m closing this as superseded by #518.

ilysenko:
Closing as superseded by #518 for the current fresh-install/current-upstream flow.

PR #530, opened 2026-06-20
Fix Linux remote control toggle params

joshyorko:
## Summary
- Patch the current `set-remote-control-enabled-for-host` webview handler so Linux sends omitted params to `remoteControl/enable` and `remoteControl/disable` instead of `null`.
- Add a current-DMG-style remote-control fixture covering the handler shape introduced around 26.616.41845.
- Preserve the existing older enablement bridge behavior and idempotence.

## Validation
- `devcontainer exec --workspace-folder . sh -lc 'cd /workspaces/codex-desktop-linux-remote-pr && node --test linux-features/remote-mobile-control/test.js'`
- `devcontainer exec --workspace-folder . sh -lc 'cd /workspaces/codex-desktop-linux-remote-pr && node --test scripts/patch-linux-window-ui.test.js'`
- `git diff --check`

### DIFF
--- a/linux-features/remote-mobile-control/patch.js
+++ b/linux-features/remote-mobile-control/patch.js
@@ -1175,10 +1175,17 @@ function applyLinuxRemoteControlEnablementBridgePatch(source) {
   if (!patched.includes(REMOTE_CONTROL_ENABLE_FOR_HOST_PARAMS_MARKER)) {
     const enabledForHostNullParamsPattern =
       /("set-remote-control-enabled-for-host":[A-Za-z_$][\w$]*\(\([A-Za-z_$][\w$]*,\{enabled:[A-Za-z_$][\w$]*\}\)=>[A-Za-z_$][\w$]*\.sendRequest\([A-Za-z_$][\w$]*\?`remoteControl\/enable`:`remoteControl\/disable`,)null(\)\))/u;
+    const beforeEnableForHostParamsPatch = patched;
     patched = patched.replace(
       enabledForHostNullParamsPattern,
       `$1void 0/*${REMOTE_CONTROL_ENABLE_FOR_HOST_PARAMS_MARKER}*/$2`,
     );
+    if (
+      patched === beforeEnableForHostParamsPatch &&
+      patched.includes("set-remote-control-enabled-for-host")
+    ) {
+      console.warn("WARN: Could not find remote-control enable-for-host params needle - skipping Linux remote-control host params patch");
+    }
   }
 
   const markerIndex = patched.indexOf("[remote-connections/slingshot-gate-bridge]");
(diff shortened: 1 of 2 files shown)


linux-features/remote-mobile-control/patch.js
```diff
@@ -1169,12 +1170,22 @@ function applyLinuxRemoteControlStatusReadGuardPatch(source) {
...
+  let patched = source;
+
+  if (!patched.includes(REMOTE_CONTROL_ENABLE_FOR_HOST_PARAMS_MARKER)) {
+    const enabledForHostNullParamsPattern =
+      /("set-remote-control-enabled-for-host":[A-Za-z_$][\w$]*\(\([A-Za-z_$][\w$]*,\{enabled:[A-Za-z_$][\w$]*\}\)=>[A-Za-z_$][\w$]*\.sendRequest\([A-Za-z_$][\w$]*\?`remoteControl\/enable`:`remoteControl\/disable`,)null(\)\))/u;
+    patched = patched.replace(
+      enabledForHostNullParamsPattern,
+      `$1void 0/*${REMOTE_CONTROL_ENABLE_FOR_HOST_PARAMS_MARKER}*/$2`,
+    );
+  }
```
avifenesh, line 1189: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This new sub-patch has no diagnostic when the needle fails to match. If the upstream `set-remote-control-enabled-for-host` handler shape drifts, `patched.replace(...)` silently no-ops and the function proceeds — Linux would keep sending `null` params with zero signal, and since this descriptor is `ciPolicy: "optional"`, the drift is invisible in CI.

Every other needle in this function warns on a miss (e.g. lines 1201 and 1234: `console.warn("WARN: Could not find ...")`). Consider mirroring that here: when the handler literal is present but the null-params pattern fails to match, emit a `console.warn` so bundle drift surfaces instead of silently reverting to `null` params.

avifenesh (commented):

---

Reviewed the `set-remote-control-enabled-for-host` null→`void 0` param fix and the new fixture/test in `linux-features/remote-mobile-control/`.

One concern about the new params-replacement code path (P2 — required/optional-upstream patch silently demoted). The new regex-based replacement at lines 1176-1181 has no warning/failure signal when the needle is not found: if the upstream handler shape drifts (e.g. minifier renames or whitespace changes), `patched.replace()` is a no-op and the function silently continues. Because the bridge patch is `ciPolicy: "optional"`, a real drift in this sub-patch is invisible — Linux would resume sending `null` params with no diagnostic. Unlike the other sub-patches in this function (bridge needle, self-auto-connect needle) which both emit `console.warn("WARN: Could not find ...")` on failure, this new block has no such guard.

This is a robustness gap rather than a hard correctness break on the current bundle (the test confirms the happy path), so flagging at MEDIUM for consideration — consider emitting a `console.warn` when the `set-remote-control-enabled-for-host` handler is present (`patched.includes("set-remote-control-enabled-for-host")`) but the null-params pattern fails to match, mirroring the existing needle-miss warnings in this function.

PR #532, opened 2026-06-20
Updater v0.8.3: Polkit detection enhance

Leay15:
This pull request updates dependencies and refines how the application detects graphical and user session environments, especially for polkit (policy kit) support. The main focus is on improving session detection logic, clarifying function responsibilities, and adding targeted tests.

Dependency updates:

* Updated versions for several dependencies in `Cargo.toml`, including `chrono`, `clap`, `notify-rust`, `reqwest`, `serde_json`, `semver`, and `tokio` for bug fixes and compatibility.

Session detection improvements:

* Renamed `has_graphical_session` to `has_interactive_graphical_session` to clarify its purpose, and updated all call sites to use the new name.
* Introduced a new function `has_user_session_bus_for_polkit`, which checks for a user session bus environment and is now used in polkit agent detection. 

Testing:

* Added a unit test to verify that `has_user_session_bus_for_polkit` returns true when the appropriate environment variables are set, even without a display.

### DIFF
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -656,7 +656,7 @@ fn prompt_install_cli(
         return Ok(PromptInstallCliOutcome::Cancelled);
     }
 
-    if !has_graphical_session() {
+    if !has_interactive_graphical_session() {
         return Ok(PromptInstallCliOutcome::NoBackend);
     }
 
@@ -687,12 +687,17 @@ fn recently_dismissed_cli_prompt(state: &PersistedState) -> bool {
     })
 }
 
-fn has_graphical_session() -> bool {
+fn has_interactive_graphical_session() -> bool {
     let has_display =
         std::env::var_os("DISPLAY").is_some() || std::env::var_os("WAYLAND_DISPLAY").is_some();
     let has_dbus = std::env::var_os("DBUS_SESSION_BUS_ADDRESS").is_some()
         || std::env::var_os("XDG_RUNTIME_DIR").is_some();
     has_display && has_dbus
+}
+
+fn has_user_session_bus_for_polkit() -> bool {
+    std::env::var_os("DBUS_SESSION_BUS_ADDRESS").is_some()
+        || std::env::var_os("XDG_RUNTIME_DIR").is_some()
 }
 
 fn prefers_kdialog() -> bool {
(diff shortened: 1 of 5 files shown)


avifenesh (commented):

---

One finding on the new test's env-var handling. The dependency bumps (Cargo.toml/Cargo.lock are consistent), the `has_graphical_session` → `has_interactive_graphical_session` rename (no stragglers), and the relaxed polkit gate (`has_user_session_bus_for_polkit`) all look correct.

updater/src/app.rs
```diff
@@ -2786,6 +2791,43 @@ mod tests {
...
 
+    #[test]
+    fn user_session_bus_for_polkit_allows_user_service_env_without_display() {
+        let _env_guard = crate::test_util::env_lock();
+        let original_display = std::env::var_os("DISPLAY");
+        let original_wayland_display = std::env::var_os("WAYLAND_DISPLAY");
+        let original_dbus_session_bus_address = std::env::var_os("DBUS_SESSION_BUS_ADDRESS");
+        let original_xdg_runtime_dir = std::env::var_os("XDG_RUNTIME_DIR");
+
+        std::env::remove_var("DISPLAY");
```
avifenesh, line 2804: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This test mutates process-wide env vars (`DISPLAY`, `WAYLAND_DISPLAY`, `DBUS_SESSION_BUS_ADDRESS`, `XDG_RUNTIME_DIR`) and restores them with manual save/restore code at the tail (lines 2809-2828). If the `assert!(has_user_session_bus_for_polkit())` on line 2807 panics, the restore block never runs, leaking the removed `DISPLAY`/`WAYLAND_DISPLAY` and the faked `DBUS_SESSION_BUS_ADDRESS`/`XDG_RUNTIME_DIR` into other env-dependent tests in the crate (e.g. the `feature_picker` tests that read `DISPLAY`/`WAYLAND_DISPLAY`, and `has_interactive_graphical_session` callers). `env_lock()` only serializes access — it does not roll back leaked state — so a single failure here can cascade into misleading failures elsewhere.

Use a Drop-based RAII guard so restoration happens on every exit path including panic. The crate already has this exact pattern in `codex_cli.rs` (`EnvRestoreGuard::capture(&[...])`); reusing/lifting that (e.g. into `test_util`) would keep restoration panic-safe and drop the ~20 lines of manual restore here.

ilysenko:
Thanks, the runtime change looks good and the updater tests pass locally.

One thing to fix before merge: the new test mutates process-wide env vars and restores them manually after the assertion. If the assertion panics, the restore block will not run and the modified env can leak into later tests. `env_lock()` only serializes access; it does not restore state.

Please use a `Drop`/RAII env restore guard here, or move the existing `EnvRestoreGuard` pattern from `codex_cli.rs` into `test_util` and reuse it.

ilysenko (approved):
Reviewed current head 6e43d6a. The env restoration issue is fixed with a shared RAII guard, GitHub checks are green, and local codex-update-manager tests pass.

PR #543, opened 2026-06-22
Sync embedded computer use to v0.2.9

avifenesh:
## Summary
- sync the embedded `codex-computer-use-linux` copy from the current upstream baseline (`0.2.6-linux-alpha1`) through the standalone `computer-use-linux` v0.2.9 release
- preserve Codex-specific binary/tool/env identity while taking the standalone v0.2.7/v0.2.8/v0.2.9 behavior deltas
- port the v0.2.9 GTK4 accessibility fix so Nautilus/Text Editor/baobab-style apps read via destination-pinned AT-SPI proxies instead of the P2P no-destination fallback

Refs agent-sh/computer-use-linux#33
Refs agent-sh/computer-use-linux#34

## Release/User-visible delta
- Embedded version parity moves to `0.2.9-linux-alpha1` in `Cargo.toml`, `Cargo.lock`, and the MCP tool handler literal.
- GTK4 app accessibility trees now return real application/window nodes instead of the old empty `role: unknown`, `child_count: 0` result when the app does not expose `GetApplicationBusAddress`.
- Prior unpublished embedded deltas from v0.2.7/v0.2.8 remain included: target/input/backend fixes, COSMIC helper packaging, and the registry `cosmic-protocols` alignment used by the embedded build.

## Validation
- `cargo fmt --check -p codex-computer-use-linux`
- `git diff --check`
- `cargo check -p codex-computer-use-linux --all-targets`
- `cargo build -p codex-computer-use-linux`
- `node /home/avifenesh/projects/computer-use-linux/scripts/zod-check/check.mjs --command target/debug/codex-computer-use-linux`
- live GTK4 probe: launched Nautilus and verified `target/debug/codex-computer-use-linux state nautilus` returned `org.gnome.Nautilus` as an `application` root with `child_count: 1` and a `window` node named `tmp` with action metadata


computer-use-linux/src/server.rs
```diff
@@ -3702,6 +4001,32 @@ mod tests {
...
+            requested_or_primary_action(Some(" show-menu ")),
+            "show-menu"
+        );
+    }
+
+    #[test]
+    fn explicit_ydotool_socket_is_used_without_connectability_probe() {
+        let key = "YDOTOOL_SOCKET";
+        let original = std::env::var_os(key);
+        std::env::set_var(key, " /does/not/exist.sock ");
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This test mutates the process-global `YDOTOOL_SOCKET` env var with `set_var`/`remove_var` and restores it manually. Per repo convention (P12), env-var mutation in computer-use tests must use a Drop-based RAII guard, not manual save/restore: if `explicit_ydotool_socket()` or the assertion panics, the restore lines never run and the var leaks into other tests. Concretely, `ydotool_socket_selection_skips_unconnectable_candidates`/`accepts_datagram_socket` and the backend-preference helpers (`should_prefer_*_text_backend`) all read `YDOTOOL_SOCKET` via `ydotool_socket()`, and Rust runs tests in parallel by default — a leaked `/does/not/exist.sock` value can flip their behavior nondeterministically. Wrap the save/set/restore in an RAII guard so the original value is restored on every exit path.

avifenesh (commented):

---

Reviewed the v0.2.9 embedded computer-use sync. The async migration (ydotool/zbus/spawn_blocking), the GTK4 AT-SPI `as_accessible_proxy` destination-pinning fix, the cosmic-protocols move to crates.io (with matching `flake.nix` `outputHashes` removal — no `git+` sources remain in `Cargo.lock`), the new `codex-computer-use-cosmic` bin (present on disk and wired into `flake.nix` install + Cargo.toml), and the relative-click bounds clamping all check out. The KDE clipboard path keeps a ydotool fallback (P14) and treats restore-after-paste failure as a warning, not an error (P15). One concern below.

ilysenko (approved):
Reviewed current head 93b714f after the env-var guard cleanup. Local validation: cargo fmt --check -p codex-computer-use-linux, cargo test -p codex-computer-use-linux, git diff --check. No blockers found.

PR #546, opened 2026-06-22
Fix Linux file manager target in worker registry

### LINKED ISSUE #545: Open in File Manager still fails on Codex.app 26.616.51431 because worker.js has no Linux fileManager target
Hi, I think the Linux file manager fix from #34 only covers one of the current open-target execution paths.

## Symptom

On Linux/KDE, all "Open in File Manager" actions are still shown in the UI, but clicking them does not open the folder.

Examples:

- The toolbar dropdown shows "Open in File Manager"
- The workspace/file context menu also shows "Open in File Manager"

The action fails with this error in the logs:

```text
Unknown open target "fileManager"
```

## What I found

I inspected the generated `app.asar`.

The main bundle appears to be patched correctly. In `.vite/build/main-*.js`, the `fileManager` target has a Linux implementation using `shell.openPath()`, so the #34 patch seems to apply to the main-process bundle.

However, the current app also has a separate open-target registry in:

```text
.vite/build/worker.js
```

That worker bundle has its own target map and lookup function. The lookup throws when a target id is missing:

```js
function W9(e) {
  let t = ale.get(e);
  if (t == null) throw Error(`Unknown open target "${e}"`);
  return t;
}
```

The worker-side `fileManager` target only defines `darwin` and `win32` platforms, but no `linux` platform. Because the worker target map is built from `e.platforms[process.platform]`, `fileManager` is not registered on Linux in that path.

```js
darwin: {...},
win32: {...}
```

This would explain why #34 fixed the main bundle path but the action still fails in the current build: the same Linux `fileManager` implementation also needs to be injected into the worker bundle, or the worker should reuse a registry that already includes the patched Linux target.

## Version details

From `codex-app/resources/codex-linux-build-info.json`:

- Upstream Codex.app version: `26.616.51431`
- Upstream DMG sha256: `e0d07909bd047d897db544e366103d4f14c06c1e038e8443aa97dc90405fcacf`
- Electron version: `42.1.0`
- codex-desktop-linux version: `0.8.4`
- codex-desktop-linux commit: `c868dcd19f3a176d833ac9f1d35873f6a21e8921`
- Target: `manjaro/pacman:kde`
- Session: Wayland
- Relevant enabled feature: `open-target-discovery`

The generated `app.asar` sha256 is:

```text
58c152953655980b2dc83717afbb35d12d37871fa35dae92f9382e757b7f113d  codex-app/resources/app.asar
```

ihipop:
Fixes #545.

## Summary

- add a core extracted-app patch for `.vite/build/worker.js`
- inject the missing Linux `fileManager` target into the worker open-target registry
- add regression coverage for the direct worker patch and the `patchExtractedApp` path

## Notes

The existing Linux file manager patch covers the main bundle, but current Codex.app builds also have a separate open-target registry in `worker.js`. That worker-side `fileManager` target only had `darwin` and `win32`, so Linux could still hit `Unknown open target "fileManager"` from the worker path.

## Validation

```bash
```

Result: 292 passed, 0 failed.

### DIFF
--- a/scripts/patches/main-process/misc.js
+++ b/scripts/patches/main-process/misc.js
@@ -56,6 +56,84 @@ function applyLinuxFileManagerPatch(currentSource) {
   return patchedSource;
 }
 
+function applyLinuxWorkerFileManagerPatch(currentSource) {
+  const block = findCallBlock(currentSource, "id:`fileManager`");
+  if (block == null) {
+    console.warn("Failed to apply Linux Worker File Manager Patch");
+    return currentSource;
+  }
+
+  if (block.text.includes("linux:{")) {
+    return currentSource;
+  }
+
+  const fsVar = requireName(currentSource, "node:fs");
+  const pathVar = requireName(currentSource, "node:path");
+  if (fsVar == null || pathVar == null) {
+    console.warn("Failed to apply Linux Worker File Manager Patch");
+    return currentSource;
+  }
+
+  const insertionPoint = block.text.lastIndexOf("}});");
+  if (insertionPoint === -1) {
+    console.warn("Failed to apply Linux Worker File Manager Patch");
+    return currentSource;
+  }
+
+  const linuxFileManager =
+    `,linux:{label:\`File Manager\`,icon:\`apps/file-explorer.png\`,detect:()=>\`linux-file-manager\`,args:e=>[e],open:async({path:e})=>{let t=e;for(;;){try{if(${fsVar}.existsSync(t))break}catch{}let e=${pathVar}.dirname(t);if(e===t)break;t=e}try{${fsVar}.existsSync(t)&&${fsVar}.statSync(t).isFile()&&(t=${pathVar}.dirname(t))}catch{}let i=await(await import(\`electron\`)).shell.openPath(t);if(i)throw Error(i)}}`;
+
+  const patchedBlock =
+    block.text.slice(0, insertionPoint + 1) +
+    linuxFileManager +
+    block.text.slice(insertionPoint + 1);
+  const patchedSource =
+    currentSource.slice(0, block.start) + patchedBlock + currentSource.slice(block.end);
+
+  const patchedBlockCheck = patchedSource.slice(block.start, block.start + patchedBlock.length);
+  if (
+    !patchedBlockCheck.includes("linux:{label:`File Manager`") ||
+    !patchedBlockCheck.includes("detect:()=>`linux-file-manager`") ||
+    !patchedBlockCheck.includes("import(`electron`)).shell.openPath(t)")
+  ) {
+    console.warn("Failed to apply Linux Worker File Manager Patch");
+    return currentSource;
+  }
+
+  return patchedSource;
+}
+
+function patchLinuxWorkerFileManagerTarget(extractedDir) {
+  const workerPath = path.join(extractedDir, ".vite", "build", "worker.js");
+  if (!fs.existsSync(workerPath)) {
+    console.warn(
+      `WARN: Could not find worker bundle at ${workerPath} — skipping Linux Worker File Manager Patch`,
+    );
+    return { matched: 0, changed: 0, reason: "worker bundle not found" };
+  }
+
+  const source = fs.readFileSync(workerPath, "utf8");
+  const patchedSource = applyLinuxWorkerFileManagerPatch(source);
+  if (patchedSource === source) {
+    const hasTarget = source.includes("id:`fileManager`");
+    const hasLinuxTarget = source.includes("linux:{label:`File Manager`");
+    const hasPatchableBlock = findCallBlock(source, "id:`fileManager`") != null;
+    return {
+      matched: hasPatchableBlock ? 1 : 0,
+      changed: 0,
+      reason: !hasTarget
+        ? "fileManager target not found"
+        : hasLinuxTarget
+          ? null
+          : hasPatchableBlock
+            ? "fileManager target found but Linux worker patch was not applied"
+            : "fileManager target found but patchable block not found",
+    };
+  }
+  fs.writeFileSync(workerPath, patchedSource, "utf8");
+  return { matched: 1, changed: 1 };
+}
+
 function applyLinuxGitOriginsSourceFallbackPatch(currentSource) {
   const fallbackSource = "linux_git_origins_missing_source_fallback";
   if (currentSource.includes(`source:\`${fallbackSource}\`,requestKind:`)) {
@@ -286,7 +364,9 @@ module.exports = {
   applyLinuxGitOriginsSourceFallbackPatch,
   applyLinuxLocalAppServerFeatureEnablementHandlerPatch,
   applyLinuxOwlFeatureBindingFallbackPatch,
+  applyLinuxWorkerFileManagerPatch,
   patchLinuxOwlFeatureBindingFallbackAssets,
+  patchLinuxWorkerFileManagerTarget,
   applyLinuxRemoteControlConfigPreservationPatch,
   applyLinuxXdgDocumentsDirPatch,
 };
--- a/scripts/patch-linux-window-ui.js
+++ b/scripts/patch-linux-window-ui.js
@@ -78,6 +78,7 @@ const {
   applyLinuxExplicitTrayQuitPatch,
   applyLinuxFileManagerPatch,
   applyLinuxGitOriginsSourceFallbackPatch,
+  applyLinuxWorkerFileManagerPatch,
   applyLinuxMenuPatch,
   applyLinuxNativeTitlebarPatch,
   applyLinuxLocalAppServerFeatureEnablementHandlerPatch,
@@ -94,6 +95,7 @@ const {
   applyLinuxWindowOptionsPatch,
   applyLinuxXdgDocumentsDirPatch,
   patchLinuxOwlFeatureBindingFallbackAssets,
+  patchLinuxWorkerFileManagerTarget,
 } = require("./patches/main-process.js");
 const {
   applyLinuxAvatarOverlayMousePassthroughPatch,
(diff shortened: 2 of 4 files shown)


scripts/patches/main-process/misc.js
```diff
@@ -56,6 +56,68 @@ function applyLinuxFileManagerPatch(currentSource) {
...
+  if (patchedSource === source) {
+    return {
+      matched: source.includes("id:`fileManager`") ? 1 : 0,
+      changed: 0,
+      reason: source.includes("id:`fileManager`") ? null : "fileManager target not found",
+    };
+  }
+  fs.writeFileSync(workerPath, patchedSource, "utf8");
+  return { matched: 1, changed: 1 };
+}
```
avifenesh, line 135: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

When `worker.js` exists but the `id:`fileManager`` target is absent (upstream bundle drift), `applyLinuxWorkerFileManagerPatch` returns the source unchanged and this returns `{matched: 0, changed: 0, reason: "fileManager target not found"}`. With `ciPolicy: "optional"`, the engine maps `changed:0` + reason/warning to `skipped-optional` (P2 territory), so a real upstream regression where the worker registry moves/renames would be silently demoted to optional drift rather than surfaced. That matches the existing main-bundle file-manager patch's policy, so it may be intentional — but note that the worker case has an additional silent path: when the *call block* is found but `block.text.includes("linux:{")` is false and `lastIndexOf("}});")` returns -1, `applyLinuxWorkerFileManagerPatch` logs a warn and returns unchanged, yet `patchLinuxWorkerFileManagerTarget` then reports `matched: 1, reason: null` (status `already-applied`) because `source.includes("id:`fileManager`")` is true. That misreports a genuine match-failure as success and drops the warning. Consider propagating the unchanged-but-found-and-not-applied case as drift rather than `already-applied`.

avifenesh (commented):

---

Reviewed the worker `fileManager` open-target patch. One issue worth confirming around the new descriptor's CI status semantics for the worker bundle; otherwise the patch logic mirrors the existing file-manager patch and is idempotent. Inline note below.

scripts/patches/main-process/misc.js
ihipop: Thanks, I updated the worker descriptor to report unchanged-but-not-applied cases as optional drift instead of already-applied, and added a regression test for worker fileManager patch drift. Validation now passes with `node --test scripts/patch-linux-window-ui.test.js` (293 passed, 0 failed).

MuQY1818:
It works on my Ubuntu 22.04

scripts/patches/main-process/misc.js
```diff
@@ -56,6 +56,74 @@ function applyLinuxFileManagerPatch(currentSource) {
...
+  if (!fs.existsSync(workerPath)) {
+    console.warn(
+      `WARN: Could not find worker bundle at ${workerPath} — skipping Linux Worker File Manager Patch`,
+    );
+    return { matched: 0, changed: 0, reason: "worker bundle not found" };
+  }
+
+  const source = fs.readFileSync(workerPath, "utf8");
+  const patchedSource = applyLinuxWorkerFileManagerPatch(source);
+  if (patchedSource === source) {
```
avifenesh, line 117: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

Minor robustness note (not blocking): when `applyLinuxWorkerFileManagerPatch` fails to *match* the call block, it logs `console.warn("Failed to apply Linux Worker File Manager Patch")` and returns the source unchanged. In that path `patchLinuxWorkerFileManagerTarget` then hits this `patchedSource === source` branch and, because the `id:`fileManager`` literal is present but `findCallBlock` returned null (e.g. no preceding `var`/`let`/`const` or no `});`), reports `reason: "fileManager target found but Linux worker patch was not applied"`. The engine's `status` callback already routes that to `skipped-optional` via `result?.reason != null`, plus the warning is captured — so drift is still visible. Just flagging that the `matched: hasTarget ? 1 : 0` count reports `1` (matched) even though `findCallBlock` actually failed to locate a usable block; the `reason` carries the real signal, so this is cosmetic, not a correctness break.

avifenesh (commented):

---

Reviewed the worker `fileManager` open-target patch. The core logic is sound: `applyLinuxWorkerFileManagerPatch` reuses the established `findCallBlock` + `}});` insertion approach, is idempotent (guarded by `block.text.includes("linux:{")`), and `patchLinuxWorkerFileManagerTarget` distinguishes the three cases (target missing, target present + already patched, target present + patch not applied) so upstream drift surfaces as `skipped-optional` rather than being silently swallowed. The descriptor's `status` callback and the regression tests align with the engine semantics. One non-blocking observation below.

scripts/patches/main-process/misc.js
```diff
@@ -56,6 +56,77 @@ function applyLinuxFileManagerPatch(currentSource) {
...
+  }
+
+  const insertionPoint = block.text.lastIndexOf("}});");
+  if (insertionPoint === -1) {
+    console.warn("Failed to apply Linux Worker File Manager Patch");
+    return currentSource;
+  }
+
+  const linuxFileManager =
+    `,linux:{label:\`File Manager\`,icon:\`apps/file-explorer.png\`,detect:()=>\`linux-file-manager\`,args:e=>[e],open:async({path:e})=>{let n=require(\`node:fs\`),r=require(\`node:path\`),t=e;for(;;){try{if(n.existsSync(t))break}catch{}let e=r.dirname(t);if(e===t)break;t=e}try{n.existsSync(t)&&n.statSync(t).isFile()&&(t=r.dirname(t))}catch{}let i=await(await import(\`electron\`)).shell.openPath(t);if(i)throw Error(i)}}`;
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

The injected worker `open()` body uses bare `require(\`node:fs\`)`/`require(\`node:path\`)` plus a dynamic `await import(\`electron\`)`, whereas the main-bundle `applyLinuxFileManagerPatch` (just above) resolves the existing module aliases via `requireName(...)` and bails out if any is missing. The worker variant assumes a CommonJS-capable `require` is in scope in the worker bundle. If `worker.js` is an ESM bundle (no synchronous `require`), the synchronous `require(\`node:fs\`)` call throws at runtime inside the open handler — which is exactly the path this patch is meant to fix. Worth confirming the worker bundle is CJS, or mirroring the main patch's alias-resolution approach so the injected code uses the bundle's own module bindings rather than assuming `require`.

avifenesh (commented):

---

Reviewed the worker `fileManager` open-target patch. The core approach is sound — `applyLinuxWorkerFileManagerPatch` reuses the established `findCallBlock` + idempotency-guard pattern, the descriptor is registered as `extracted-app`/`optional` with a unique id, and the worker bundle is found/written from `.vite/build/worker.js`. One robustness concern on the open() implementation is worth confirming.

scripts/patches/main-process/misc.js
ihipop: Updated in 28a1479. The worker patch now resolves the existing `node:fs` and `node:path` bundle aliases with `requireName(...)`

scripts/patches/main-process/misc.js
```diff
@@ -56,6 +56,84 @@ function applyLinuxFileManagerPatch(currentSource) {
...
+  if (!fs.existsSync(workerPath)) {
+    console.warn(
+      `WARN: Could not find worker bundle at ${workerPath} — skipping Linux Worker File Manager Patch`,
+    );
+    return { matched: 0, changed: 0, reason: "worker bundle not found" };
+  }
+
+  const source = fs.readFileSync(workerPath, "utf8");
+  const patchedSource = applyLinuxWorkerFileManagerPatch(source);
+  if (patchedSource === source) {
```
avifenesh, line 117: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

When `applyLinuxWorkerFileManagerPatch` fails for a recoverable-but-real reason, the returned `reason` here can mislead the report. Concretely: if the `id:`fileManager`` block exists but `requireName` can't resolve `node:fs`/`node:path` (or `}});` insertion point is missing), `applyLinuxWorkerFileManagerPatch` emits a `console.warn` and returns the source unchanged. In `patchLinuxWorkerFileManagerTarget`, `hasPatchableBlock` (= `findCallBlock(...) != null`) is then `true` and `hasLinuxTarget` is `false`, so this branch returns `reason: "fileManager target found but Linux worker patch was not applied"` — but the descriptor `status` callback in window-shell/patch.js keys off `warnings.length`/`result.reason` and the `console.warn` is captured by the engine, so it is reported as `skipped-optional` drift (correct). However, in the *normal already-applied* case (`block.text.includes("linux:{")` -> early return at line 67), the source is unchanged, `hasLinuxTarget` is `true`, `reason` is `null`, and no warning fires — so the descriptor status resolves to `already-applied` (correct). The edge to confirm: an upstream bundle where the worker `fileManager` block already contains some *other* `linux:{...}` key (not the File Manager target) would hit the line-67 early return, mark `already-applied`, and silently skip injecting the Linux target. The main-bundle `applyLinuxFileManagerPatch` has the identical `includes("linux:{")` guard, so this matches existing behavior — but worth a quick check that no real worker bundle ships an unrelated `linux:{` in that call block.

avifenesh (commented):

---

Reviewed the worker `fileManager` open-target patch (PR #546). The approach mirrors the existing `applyLinuxFileManagerPatch` and reuses `findCallBlock`/`requireName`, registering a new `extracted-app` descriptor `linux-worker-file-manager`. One status-semantics concern worth confirming on the warning path; the rest looks consistent with the framework.

ilysenko (approved):
Reviewed current head 28a1479 after the worker alias update. CI is green; local validation passed: node --test scripts/patch-linux-window-ui.test.js, git diff --check, node --check for changed patcher files, and real app.asar worker sanity check. No blockers found.

PR #547, opened 2026-06-22
fix(tray): handle role-only Linux quit items

pradigmaz:
## Summary

Fix Linux tray Quit when the upstream menu contains a role-only `{role: "quit"}` item.

Electron ignores `click` handlers when `role` is set, so the Linux explicit-quit marker was not applied for that tray menu shape. The app could then re-enter the close-to-tray path instead of exiting.

## Changes

- Replace role-only tray Quit items inside `getNativeTrayMenuItems()` with an explicit label/click handler.
- Scope the replacement to tray menu construction so helper uses of `Menu.buildFromTemplate([{ role: "quit" }])` keep working.
- Add regression coverage for the role-only tray Quit shape.

## Verified

- `node --test --test-name-pattern='role-only tray quit' scripts/patch-linux-window-ui.test.js`
- `node --test --test-name-pattern='upstream renames the quit label helper' scripts/patch-linux-window-ui.test.js`
- `node --test --test-name-pattern='explicit tray quit patching when minified aliases drift' scripts/patch-linux-window-ui.test.js`
- `node --check scripts/patches/main-process/quit-lifecycle.js`
- `git diff --check`

Full `node scripts/patch-linux-window-ui.test.js` still has two existing CLI `--enforce-critical` assertion failures; the same two fail with this diff reverted.

### DIFF
--- a/scripts/patches/main-process/quit-lifecycle.js
+++ b/scripts/patches/main-process/quit-lifecycle.js
@@ -1,5 +1,7 @@
 "use strict";
 
+const { findMatchingBrace, requireName } = require("../shared.js");
+
 function applyLinuxQuitGuardPatch(currentSource) {
   let patchedSource = currentSource;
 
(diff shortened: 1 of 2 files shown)


scripts/patches/main-process/quit-lifecycle.js
```diff
@@ -45,6 +47,44 @@ function linuxExplicitQuitExpression() {
...
+    }
+
+    const openIndex = methodIndex + methodNeedle.length - 1;
+    const closeIndex = findMatchingBrace(patchedSource, openIndex);
+    if (closeIndex === -1) {
+      break;
+    }
+
+    const body = patchedSource.slice(openIndex + 1, closeIndex);
+    const patchedBody = body.replace(roleOnlyTrayQuitRegex, () => {
```
avifenesh, line 70: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

Scoping note (please confirm against the real bundle): the replacement targets any `{role:`quit`}` token anywhere inside the `getNativeTrayMenuItems(){...}` body. The PR's stated intent ("Scope the replacement to tray menu construction so helper uses of `Menu.buildFromTemplate([{ role: \"quit\" }])` keep working") is satisfied only when the label-helper that calls `Menu.buildFromTemplate([{role:`quit`}])` is defined *outside* this method. The test at line 1617 places `mH` as a sibling function, which passes — but if upstream ever inlines such a helper (or any other `Menu.buildFromTemplate([{role:`quit`}])` call) *inside* `getNativeTrayMenuItems`, this regex would rewrite that template entry into a `{label:`Quit`,click:...}` item too, changing the helper's label-extraction result. Worth an assertion or a tighter anchor (e.g. requiring the token to be a top-level array element of the returned menu) if that shape is plausible in the bundle.

avifenesh (commented):

---

Reviewed the role-only tray Quit fix in `quit-lifecycle.js` plus its regression test. Per the always-on check for `scripts/patches/`, I verified the new `patchRoleOnlyTrayQuitItems` replaces ALL eligible occurrences (it uses the global `roleOnlyTrayQuitRegex` inside each method body and walks every `getNativeTrayMenuItems(){` via index-based cursor advancement, not a single `string.replace()`), and that idempotency holds (the patched output no longer contains `{role:`quit`}`, so a second pass is a no-op — confirmed by `applyPatchTwice`).

One scoping concern worth confirming below; the rest of the change looks consistent with the existing tray-quit patch helpers and the `electronVar` alias resolution.

PR #550, opened 2026-06-23
[codex] Fix updater daemon state reload

moxw:
## Summary
- reload persisted updater state from disk before daemon check and reconcile cycles
- keep auto-install state synchronized when another updater process writes a pending candidate
- bump codex-update-manager to 0.8.5

## Why
The daemon could keep stale in-memory state after another update/check process persisted `waiting_for_app_exit`. If the desktop app exited after that, the daemon might never notice the pending install state and would not advance the update.

## Source of truth
- `updater/src/app.rs`
- `updater/Cargo.toml`
- `Cargo.lock`

## Validation
- `cargo fmt --package codex-update-manager --check`
- `git diff --check`
- `CARGO_HOME=/home/mohit/.cache/cargo cargo check -p codex-update-manager`
- `CARGO_HOME=/home/mohit/.cache/cargo cargo test -p codex-update-manager`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -519,7 +519,7 @@ dependencies = [
 
 [[package]]
 name = "codex-update-manager"
-version = "0.8.4"
+version = "0.8.5"
 dependencies = [
  "anyhow",
  "chrono",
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed current head 67a31e3. CI is green; local validation passed with updater-focused checks and full updater suite using CODEX_UPDATE_MANAGER_SKIP_SYSTEM_CLI_LOOKUP=1. No blockers found.

PR #553, opened 2026-06-23
[codex] Harden remote download and update paths

NavesDev:
## Summary

This PR hardens several remote download and update paths that previously trusted HTTPS or configured remotes without an explicit integrity/policy gate.

Changes include:
- require SHA256 verification for downloaded or locally supplied Electron Linux runtime archives
- require SHA256 verification for 7zz bootstrap archives and replace `curl | sh` rustup with verified `rustup-init`
- gate remote-mobile-control runtime auto-install behind explicit opt-in
- require explicit opt-in, hashed Python requirements, and SHA256 checks for Read Aloud model assets
- restrict wrapper updates to the trusted upstream repository before detecting or applying updates
- require `CODEX_UPSTREAM_DMG_SHA256` for Codex.dmg, with an explicit local override for unverified manual use

## Security impact

These changes reduce supply-chain exposure from unauthenticated archive acceptance, implicit network installers, and arbitrary wrapper update remotes.

## Validation

Validated locally with shell syntax checks only:

```bash
bash -n scripts/install-deps.sh scripts/lib/dmg.sh scripts/lib/native-modules.sh linux-features/remote-mobile-control/cold-start-hook.sh linux-features/read-aloud/install-kokoro-runtime.sh
```

I did not run Node/Rust test suites because this review was performed in the context of auditing a suspicious checkout and those tests execute repository/dependency code.

### DIFF
--- a/linux-features/read-aloud/install-kokoro-runtime.sh
+++ b/linux-features/read-aloud/install-kokoro-runtime.sh
@@ -8,6 +8,37 @@ voices="${CODEX_LINUX_READ_ALOUD_KOKORO_VOICES:-$data_home/kokoro/voices-v1.0.bi
 model_url="${CODEX_LINUX_READ_ALOUD_KOKORO_MODEL_URL:-https://huggingface.co/zijuncheng/kokoro_model_v1.0/resolve/main/kokoro-v1.0.onnx}"
 voices_url="${CODEX_LINUX_READ_ALOUD_KOKORO_VOICES_URL:-https://huggingface.co/zijuncheng/kokoro_model_v1.0/resolve/main/voices-v1.0.bin}"
 
+truthy_env_value() {
+    case "${1:-}" in
+        1|true|TRUE|yes|YES|on|ON) return 0 ;;
+        *) return 1 ;;
+    esac
+}
+
+expected_sha_for_target() {
+    local target="$1"
+    case "$target" in
+        "$model") printf '%s\n' "${CODEX_LINUX_READ_ALOUD_KOKORO_MODEL_SHA256:-}" ;;
+        "$voices") printf '%s\n' "${CODEX_LINUX_READ_ALOUD_KOKORO_VOICES_SHA256:-}" ;;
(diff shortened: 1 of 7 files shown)


ilysenko:
Thanks for the hardening pass. I found several blockers before this can move forward:

- `scripts/install-deps.sh` now refuses to install Rust without `RUSTUP_INIT_SHA256` (`scripts/install-deps.sh:480`), and the 7zz bootstrap path will also refuse without `SEVENZIP_BOOTSTRAP_SHA256` (`scripts/install-deps.sh:417`). The install-deps workflow does not provide these values, so fresh Debian/Ubuntu setup now fails. Please wire the new policy into CI and the default helper flow, or provide a non-breaking distro/package-manager path.
- Nix package builds are broken by the Electron hash gate. `flake.nix` already fetches `electronZip` as a fixed-output source and passes it via `CODEX_ELECTRON_ZIP_SOURCE`, but `download_electron` still requires `CODEX_ELECTRON_ZIP_SHA256` and fails with `No SHA256 configured for Electron v42.1.0 linux-x64` (`scripts/lib/native-modules.sh:308`). Please integrate the existing Nix hash/source-of-truth with the installer check.
- Wrapper update detection now rejects a normal remote name like `origin` before resolving it to a URL (`updater/src/wrapper.rs:304`), so update detection regresses to `UnknownOffline` instead of `Aligned`, `DevMode`, or `UpdateAvailable`. This is why the updater tests are failing. Please resolve configured remote names before enforcing the trusted-remote policy, and add/update tests for trusted and untrusted remotes.
- The DMG policy is incomplete: downloaded/cached DMGs go through `verify_dmg_sha256_policy`, but `./install.sh path/to/Codex.dmg` bypasses that verification entirely (`install.sh:92`). If this PR is meant to require SHA256 for Codex.dmg, provided DMGs need the same explicit verify/override path.

I reproduced the updater failures locally with `cargo test -p codex-update-manager -- --nocapture` (5 failing tests). Please fix the failing CI/tests and add coverage for the new hash/trust policy paths.

PR #555, opened 2026-06-24
[codex] Suppress superseded install notifications

moxw:
## Summary
- distinguish exact pending-candidate recovery from cleanup where a newer installed package superseded the candidate
- skip the installed-package desktop notification when a stale pending candidate is only being cleared
- stop daemon/check-now startup normalization from emitting installed notifications for steady-state `Installed`
- bump codex-update-manager to 0.8.6

## Why
A machine can install a newer package while updater state still contains an older pending candidate. Reconciliation correctly clears that stale candidate, but it previously sent the installed notification using the newer installed version, which made it look like another package would be used on next app open even though the latest package was already active.

## Source of truth
- `updater/src/app.rs`
- `updater/Cargo.toml`
- `Cargo.lock`

## Validation
- `CARGO_HOME=/home/mohit/.cache/cargo cargo test -p codex-update-manager pending_install`
- `cargo fmt --package codex-update-manager --check`
- `git diff --check`
- `CARGO_HOME=/home/mohit/.cache/cargo cargo test -p codex-update-manager`
- `CARGO_HOME=/home/mohit/.cache/cargo cargo check -p codex-update-manager`

### DIFF
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -326,6 +326,23 @@ fn update_install_is_pending(status: &UpdateStatus) -> bool {
     )
 }
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+enum PendingInstallRecovery {
+    NoChange,
+    CandidateInstalled,
+    SupersededByInstalledVersion,
+}
+
+impl PendingInstallRecovery {
+    fn completed(self) -> bool {
+        !matches!(self, Self::NoChange)
+    }
+
+    fn should_notify_installed(self) -> bool {
+        matches!(self, Self::CandidateInstalled)
+    }
+}
+
 async fn run_daemon(
     config: &RuntimeConfig,
     state: &mut PersistedState,
@@ -338,7 +355,6 @@ async fn run_daemon(
     normalize_workspace_dir_and_persist(state, paths)?;
     maybe_prune_workspace_cache(&config.workspace_root, state);
     maybe_notify_cli_missing(state, paths, config.notifications)?;
-    maybe_notify_installed(state, paths, config.notifications)?;
     if packaged_runtime_removed(config) {
         info!("packaged app files are gone; stopping updater daemon");
         return Ok(());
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed locally; no blockers found. CI is green.

updater/src/app.rs
```diff
@@ -338,7 +355,6 @@ async fn run_daemon(
     normalize_workspace_dir_and_persist(state, paths)?;
     maybe_prune_workspace_cache(&config.workspace_root, state);
     maybe_notify_cli_missing(state, paths, config.notifications)?;
```
avifenesh, line 357: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This removes the startup `maybe_notify_installed` call from `run_daemon` (and the matching one in `run_check_now`). Per the PR intent this stops emitting an installed notification for a steady-state `Installed` status that was never the result of a fresh install in this run.

However, this also drops the notification in the case it previously served: when a `trigger_install` succeeded in a prior daemon lifetime but the process exited before `maybe_notify_installed` ran, the next daemon startup would surface the "Codex Desktop updated" notification. After this change, startup no longer recovers that notification — it only fires from `trigger_install` itself and from the `CandidateInstalled` recovery path. If the intent is purely to suppress the steady-state notification, consider gating on whether the recovery actually transitioned state (i.e., notify only when `complete_pending_install_if_already_installed` returned `CandidateInstalled`) rather than removing the startup call entirely, so a genuinely-completed-but-unnotified install is not silently dropped. Worth a maintainer confirming this is the intended trade-off.

avifenesh (commented):

---

## Review: Suppress superseded install notifications

The core change is sound: `complete_pending_install_if_already_installed` now returns a tri-state `PendingInstallRecovery` so callers can distinguish a genuine candidate install (notify) from a superseded-by-newer cleanup (no notify). Both `reconcile_pending_install` and `run_install_ready` correctly gate `maybe_notify_installed` on `should_notify_installed()`, and the tests cover the matching/superseded cases.

One behavioral concern worth confirming below regarding the removal of the startup `maybe_notify_installed` calls.

PR #557, opened 2026-06-24
feat: add opt-in persistent status panel feature

YudongN:
## Summary

- Add an opt-in `persistent-status-panel` Linux feature.
- Keep the Codex `/status` panel open across thread switches and app restarts until explicitly closed.
- Leave the existing status panel rendering unchanged.

## Behavior

The feature is disabled by default. When enabled, it stores the explicit open/closed preference in `localStorage` and applies it when the composer remounts.

The patch is fail-soft and idempotent. If the upstream composer bundle changes shape, it warns and leaves the bundle unchanged.

## Validation

Also checked broader suites locally:

These broader suites currently fail on `origin/main` as well, so the failures appear unrelated to this feature.

## Manual testing

Used locally with `persistent-status-panel` enabled to keep `/status` visible across Codex Desktop session switches and restarts.

### DIFF
--- a/linux-features/persistent-status-panel/patch.js
+++ b/linux-features/persistent-status-panel/patch.js
@@ -1,0 +1,61 @@
+"use strict";
+
+const STORAGE_KEY = "codex-linux-persistent-status-panel-open";
+
+const statusStatePattern =
+  /\{conversationId:[^,]+,threadId:[^,]+,rateLimit:[^,]+,onOpenChange:([A-Za-z_$][\w$]*)\}=e,([A-Za-z_$][\w$]*)=[A-Za-z_$][\w$]*\(\),\[([A-Za-z_$][\w$]*),([A-Za-z_$][\w$]*)\]=\(0,([A-Za-z_$][\w$]*)\.useState\)\(!1\),/;
+
+function applyPersistentStatusPanelPatch(source) {
+  if (source.includes(STORAGE_KEY)) {
+    return source;
+  }
+
+  if (!source.includes("composer.statusSlashCommand.description")) {
+    return source;
+  }
+
+  const match = source.match(statusStatePattern);
+  if (!match) {
+    console.warn("WARN: Could not find Codex status panel state - skipping persistent status panel patch");
+    return source;
+  }
+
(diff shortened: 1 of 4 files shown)


avifenesh (commented):

---

Reviewed the new opt-in `persistent-status-panel` feature. The feature framework integration is correct (disabled by default, unique id, `webview-asset` phase, fail-soft warnings). One correctness concern on the `String.replace`-based patch where the needle is not provably unique.

linux-features/persistent-status-panel/patch.js
```diff
@@ -0,0 +1,61 @@
...
+    .replace(stateNeedle, statePatch)
+    .replace(openNeedle, openPatch)
+    .replace(closeNeedle, closePatch);
+}
+
+const patches = [
+  {
+    id: "composer-status-state",
+    phase: "webview-asset",
+    order: 20_800,
```
avifenesh, line 48: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`source.replace(openNeedle, …)` / `source.replace(closeNeedle, …)` (and the `stateNeedle` replace) only patch the **first** occurrence. The needles are built purely from minifier-assigned identifiers (`setIsOpen`, `onOpenChange`) and small arrow-function shapes like `()=>{x(!1),o?.(!1)}` — these are not guaranteed unique across the composer bundle. Per the repo's ASAR-patch invariant (P1), webview-asset replacements must process every eligible occurrence (match-index based) or assert the needle is unique. If the same handler shape appears more than once, the wrong site gets patched and the persisted preference silently desyncs from the actual open/close handlers. The `STORAGE_KEY` idempotency guard does not protect against this — it only prevents re-patching, not patching the wrong occurrence on the first pass.

Recommend either asserting uniqueness (e.g. count occurrences and warn/skip if >1, like other features do) or switching to indexed replacement.

ilysenko (approved):
Approving with the known small blast-radius risk accepted: this is an opt-in Linux feature, disabled by default, CI is green, and the patch is fail-soft. A follow-up to make the handler replacements uniqueness-checked/index-scoped would still be welcome.

PR #558, opened 2026-06-24
fix: require unique persistent status panel needles

YudongN:
## Summary

- Make the persistent status panel patch fail-soft unless the state/open/close needles are unique.
- Add regression coverage for duplicate handler shapes.
- Add drift coverage for composer bundles that still contain the status i18n marker but no longer match the expected state shape.

## Context

Follow-up to #557. The original feature was merged as an opt-in, disabled-by-default Linux feature. This PR hardens the patch against ambiguous minified handler needles and upstream bundle shape drift.

## Validation

- `git diff --check`


ilysenko (approved):
Looks good. This addresses the ambiguous handler issue from #557 by requiring unique state/open/close needles and adding regression coverage for duplicate handler shapes and state-shape drift.

PR #561, opened 2026-06-25
[codex] Restore Linux Computer Use availability

moxw:
## Summary

- Restore the Linux Computer Use renderer availability patch for the current upstream `use-is-plugins-enabled` gate shape.
- Include `use-native-apps.electron-*` in the opt-in Computer Use UI availability patch so Linux native app candidates are queried when the UI is enabled.
- Add smoke coverage for the current renderer and native-apps minified snippets.

## Root Cause

The upstream renderer no longer includes the exact `featureName: windows_computer_use` context the Linux availability patch required before patching the `areRequiredFeaturesEnabled` object gate. As a result, `linux-computer-use-ui-availability` was recorded as skipped even though the bundled plugin, marketplace cache, backend binary, and local UI opt-in setting were present.

## Validation

- Targeted `test_linux_computer_use_ui_opt_in_smoke` by sourcing `tests/scripts_smoke.sh` without its full `main` runner
- `git diff --check`
- Minimal patch-report fixture using current `26.616.81150` webview assets confirmed `linux-computer-use-ui-availability` is now `applied`

### DIFF
--- a/scripts/patches/computer-use.js
+++ b/scripts/patches/computer-use.js
@@ -290,6 +290,7 @@ function applyLinuxComputerUseRendererAvailabilityPatch(currentSource) {
   let platformPredicateChanged = false;
   let availabilityChanged = false;
   let availabilityGateFound = false;
+  let nativeAppsGateChanged = false;
 
   const computerUseFeatureNeedle = "featureName:`computer_use`";
   const hasComputerUseAvailabilityGate = () =>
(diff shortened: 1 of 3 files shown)


ilysenko:
Looks good from my side, and the checks are green now. Since this PR is still marked as draft, I will treat it as not ready yet. Please mark it ready for review when you consider it ready for a final review pass.

PR #562, opened 2026-06-25
Fix Linux open target discovery

ihipop:
## Summary

The `open-target-discovery` Linux feature was only extending part of Codex Desktop's open-target pipeline. On current upstream builds, the main process, renderer bridge, and open-target selection bundle can each observe or enrich targets through different code paths. As a result, Linux targets discovered from `.desktop` files could appear in one place but be filtered out, marked unavailable, fail to launch, or lose their application icons in another.

This patch keeps the feature optional, but makes its Linux opener model consistent across the affected paths:

- registers discovered Linux open targets through the main open-target registry
- makes command lookup, bridge detection, and open execution consult that registry before falling back to upstream static target handling
- patches the open-target selection bundle so directory openers and native selectors receive the same Linux target set
- resolves `.desktop` launch and icon handling for both the top selector and native right-click Open with menu

### Before

Project-directory opener could collapse to only File Manager, so editors and terminals discovered on Linux were not available from the top selector.

img_v3_0212u_b3e84f08-9f03-4969-b690-491771eb79fg

File open selector listed some dynamic desktop-entry targets, but several icons fell back to generic file or terminal icons.

img_v3_0212u_ad5c5cee-7793-4d9d-96eb-e063454def0g

### After

Project-directory selector lists available Linux editors, terminals, File Manager, and discovered desktop-entry openers.

image

File open selector preserves application icons for built-in targets and discovered desktop entries.

image

Right-click Open with uses the same resolved targets and application icons.

image

## Verification

- `node --check linux-features/open-target-discovery/patch.js`
- Manually verified in the Linux desktop app:
  - project-directory opener lists discovered Linux targets
  - file open selector preserves application icons
  - right-click Open with menu preserves application icons

### DIFF
--- a/linux-features/open-target-discovery/feature.json
+++ b/linux-features/open-target-discovery/feature.json
@@ -4,6 +4,6 @@
   "description": "Adds opt-in Linux terminal and editor opener discovery, plus richer file manager launching.",
   "defaultEnabled": false,
   "entrypoints": {
-    "mainBundlePatch": "./patch.js"
+    "patchDescriptors": "./patch.js"
   }
 }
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed current head 784e12cba1fc57a28504c4cda7954bcc554b123f. No blockers found; focused feature tests and GitHub checks are passing.

PR #563, opened 2026-06-25
[codex] Add Linux native app mentions adapter

moxw:
## Summary
- Add a Linux main-process adapter for the Computer Use `native-desktop-apps` handler.
- Query the bundled `codex-computer-use-linux windows` backend, merge `.desktop` metadata, and return app mention objects with display name, app path, bundle id, pid, and window metadata.
- Add Linux native desktop app icon resolution through resolved desktop-entry icons.

## Tests
- `node --test scripts/patch-linux-window-ui.test.js --test-name-pattern "native desktop apps|Computer Use"`


scripts/patches/computer-use.js
```diff
@@ -507,11 +511,178 @@ function applyLinuxComputerUseInstallFlowPatch(currentSource) {
...
+  );
+  if (!nativeAppsReplacement.changed) {
+    console.warn(
+      "WARN: Could not wrap native-desktop-apps handler — skipping Linux native desktop apps patch",
+    );
+    return currentSource;
+  }
+  patchedSource = nativeAppsReplacement.source;
+
+  if (findHandlerValue(patchedSource, "computer-use-native-desktop-app-icon") == null) {
```
avifenesh, line 669: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

The fallback insertion of the `computer-use-native-desktop-app-icon` handler is order-dependent and idempotency-fragile. It re-runs `findHandlerValue(patchedSource, "native-desktop-apps")` after the native-apps value was already rewritten, then prepends the icon handler before that key. On a second pass through `applyPatchTwice`, the early guard `currentSource.includes("codexLinuxNativeDesktopApps(")` short-circuits, so idempotency holds for the bundle case — but note `findExpressionEnd` scans from the key forward and stops at the first top-level `,`/`)`/`}`/`]`. If the wrapped native-apps value (`async(...e)=>process.platform===\`linux\`?codexLinuxNativeDesktopApps(e[0]):await(${handler})(...e)`) is the LAST property in the handlers object (no trailing comma), `findExpressionEnd` will walk past the closing `}` of the object literal because the function arrow body increments/decrements depth but the terminating `}` of the object is consumed as the arrow expression's own scope. Worth confirming `findExpressionEnd` returns at the object's closing brace and not beyond when the handler is the final entry — the test fixtures all place a trailing object close immediately after, which may not match the real minified bundle layout.

scripts/patches/computer-use.js
```diff
@@ -507,11 +511,178 @@ function applyLinuxComputerUseInstallFlowPatch(currentSource) {
...
+    `function codexLinuxNativeDesktopAppsBackendPath(){let e=process.env.CODEX_LINUX_COMPUTER_USE_BACKEND_SOURCE?.trim(),t=process.env.CODEX_ELECTRON_RESOURCES_PATH||process.resourcesPath,n=process.env.CODEX_HOME||(codexLinuxNativeDesktopAppsHome()?${pathVar}.join(codexLinuxNativeDesktopAppsHome(),\`.codex\`):\`\`),r=[e,t&&${pathVar}.join(t,\`plugins\`,\`openai-bundled\`,\`plugins\`,\`computer-use\`,\`bin\`,\`codex-computer-use-linux\`),n&&${pathVar}.join(n,\`plugins\`,\`cache\`,\`openai-bundled\`,\`computer-use\`,\`latest\`,\`bin\`,\`codex-computer-use-linux\`),\`codex-computer-use-linux\`];for(let e of r)if(typeof e===\`string\`&&e.length>0&&codexLinuxNativeDesktopAppsCommandExists(e))return e;return null}`,
+    `function codexLinuxNativeDesktopAppsRun(e){let t=codexLinuxNativeDesktopAppsBackendPath();if(t==null)return null;try{let n=${childProcessVar}.spawnSync(t,e,{encoding:\`utf8\`,env:process.env,maxBuffer:1048576,timeout:2500});if(n.error||n.status!==0)return null;return JSON.parse(n.stdout||\`null\`)}catch{return null}}`,
```
avifenesh, line 596: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`codexLinuxNativeDesktopAppsRun` passes `env:process.env` to `spawnSync` of the backend with a 2500ms timeout and 1MB buffer — reasonable. Confirm the backend path resolution in `codexLinuxNativeDesktopAppsBackendPath` cannot pick up an attacker-influenced relative `codex-computer-use-linux` from CWD: the final candidate is the bare name `codex-computer-use-linux`, resolved via `which` (PATH lookup) in `codexLinuxNativeDesktopAppsCommandExists`, not CWD, so this looks safe. No change requested; flagging for maintainer awareness since the bundled/cache paths are tried first.

avifenesh (commented):

---

Review of the Linux native desktop apps adapter patch. One correctness concern around the icon-handler insertion path, plus a couple of notes for confirmation.

scripts/patches/computer-use.js
moxw: Addressed in the latest push. The patch now records the native-desktop-apps handler key position before wrapping, so fallback icon-handler insertion no longer depends on re-parsing the rewritten handler value. I also added regression coverage for the final-property/no-trailing-comma shape through the existing applyPatchTwice path.

scripts/patches/computer-use.js
moxw: Confirmed, and I tightened this further in the latest push. The bare fallback no longer shells out to which or returns a bare command; it scans PATH itself, ignores empty or relative PATH entries, and returns only an executable path from an absolute PATH directory. Added coverage that a relative PATH entry containing codex-computer-use-linux is not executed.

scripts/patches/core/all-linux/main-process/computer-use/patch.js
```diff
@@ -21,4 +22,12 @@ module.exports = [
     ciPolicy: "optional",
     apply: applyLinuxComputerUsePluginGatePatch,
   },
+  {
+    id: "linux-computer-use-native-desktop-apps",
+    phase: "main-bundle",
+    order: 150,
+    ciPolicy: "opt-in",
```
avifenesh, line 29: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`ciPolicy: "opt-in"` is used here but every other descriptor in this file uses `"optional"` or `"required-upstream"`. Confirm `"opt-in"` is a recognized policy value in the CI policy engine — if the engine only understands `required-upstream`/`optional`/`required` and treats an unknown value as a hard failure (or, conversely, silently ignores it so drift in the `native-desktop-apps` needle goes unreported), this descriptor's CI behavior will diverge from intent. Per repo invariant P2, a patch that fails to match must not be silently demoted; verify `opt-in` maps to a defined skip-without-failure semantics rather than an unhandled string.

avifenesh (commented):

---

Review of the Linux native desktop apps adapter patch.

The patch is well-structured: helper functions are factored into a single inserted block, the module dependency vars are looked up dynamically via `requireName`, the descriptor is `opt-in`/`enabled: enableComputerUseUi`, and the descriptor id (`linux-computer-use-native-desktop-apps`) is unique (confirmed against the grouping test). Idempotency is guarded by the `codexLinuxNativeDesktopApps(` early-return. Smoke and unit coverage exercise the linux/darwin paths and the icon handler.

One correctness concern around the descriptor `ciPolicy` is noted inline; the rest looks sound.

scripts/patches/core/all-linux/main-process/computer-use/patch.js
moxw: Addressed in the latest push. opt-in is intentional for Computer Use surfaces behind the local enable gate, but I made that contract explicit: descriptor normalization now accepts only required-upstream, optional, and opt-in, so a typo or unsupported policy cannot silently become optional. I also documented the supported policies and added coverage that opt-in is recognized as non-critical drift while an unsupported value is rejected.

ilysenko (approved):
Reviewed current head 909b3791f90b87b67701d7366b103910de200b7d. No blockers found; GitHub checks are green, local patch tests pass, and the opt-in Computer Use main-bundle patch applies to a real extracted app.asar.

PR #564, opened 2026-06-25
Remove zed-opener Linux feature

moxw:
## Summary
- remove the opt-in zed-opener Linux feature module
- keep stale zed-opener feature configs loadable by aliasing them to open-target-discovery
- update feature framework test coverage for the legacy id migration

## Rationale
zed-opener is a narrow editor-specific feature that now duplicates the broader open-target-discovery path. The feature directory can be removed without breaking existing users because configs preserved by local installs or update-builder bundles now normalize zed-opener to open-target-discovery before loading features or writing package metadata.

## Compatibility
Existing linux-features/features.json files that still contain "zed-opener" resolve to "open-target-discovery". Configs that contain both ids de-dupe to the maintained id.

## Validation
- node --test linux-features/example-feature/test.js
- node --test linux-features/open-target-discovery/test.js
- git diff --check


ilysenko:
Thanks for cleaning this up. One blocker before we can merge: deleting `zed-opener` outright can break existing users who already have it enabled in `linux-features/features.json`.

Those installs will now fail during feature loading with:

`Enabled Linux feature ids not found in this checkout: zed-opener`

This also affects auto-update/rebuild paths, because native packages preserve the enabled feature id list in the update-builder bundle. So a user who enabled `zed-opener` before may fail to rebuild/update after this PR.

Please add a compatibility alias/migration from `zed-opener` to `open-target-discovery`, with a small test covering an existing config that still contains `"zed-opener"`.

moxw:
Addressed in 2e9d0d2: stale `zed-opener` entries now normalize to `open-target-discovery` before feature loading and package/update-builder metadata writes, so existing configs keep rebuilding. Added coverage for a config containing only `zed-opener` and for a config containing both old and new ids.

ilysenko (approved):
Reviewed the updated head. The legacy zed-opener compatibility issue is addressed by normalizing it to open-target-discovery, and the focused local tests plus GitHub checks pass.

PR #565, opened 2026-06-25
Better proxy env handling

cozy-k42:
Totally vibe coded.

Tested with both HTTP proxy and HTTP proxy with basic authentication.

Hopefully this will be useful. If this kind of contribution isn't something you'd like to accept, feel free to close the PR. Alternatively, just let me know and I'll close it myself and avoid submitting similar contributions in the future.


ilysenko:
Thanks for working on this. The feature is useful, especially for users behind authenticated corporate proxies.

I would prefer not to merge it into core as-is, though. Core is already growing quite a bit, and this is a narrower workflow-specific integration. Could you please move it under `linux-features/` as an opt-in feature, for example `linux-features/authenticated-proxy/`?

Suggested shape:

- keep the main-process proxy auth patch as a feature patch descriptor via `entrypoints.patchDescriptors`
- keep the proxy-specific env parsing and Electron proxy setup inside the feature directory
- use the existing generic runtime hooks where possible (`env`, `electronArgs`, `prelaunch`)
- if those hooks are not enough, please add a small generic launcher extension point rather than a proxy-specific hook. For example, a feature-owned launcher hook that runs before Electron args are finalized and can contribute environment variables and Electron arguments. The core contract should stay generic; the proxy parsing/auth logic should live entirely inside the feature directory
- add a README and focused tests for enabling the feature and deriving proxy args from `CODEX_LINUX_PROXY_*` / standard proxy env vars

That would keep the baseline app simpler while still letting users who need authenticated proxy support enable it explicitly.

scripts/patches/main-process/misc.js
```diff
@@ -134,6 +135,97 @@ function patchLinuxWorkerFileManagerTarget(extractedDir) {
...
+    console.warn(
+      "WARN: Could not find Electron alias — skipping Linux proxy authentication patch",
+    );
+    return currentSource;
+  }
+
+  const appLoginHelper =
+    "function codexLinuxProxyAuthHost(e){return String(e??``).trim().replace(/^\\[|\\]$/g,``).toLowerCase()}" +
+    "function codexLinuxProxyAuthEntry(e=process.env){if(process.platform!==`linux`)return null;let t=codexLinuxProxyAuthHost(e.CODEX_LINUX_PROXY_AUTH_HOST),n=String(e.CODEX_LINUX_PROXY_AUTH_PORT??``).trim(),r=e.CODEX_LINUX_PROXY_USERNAME;if(!t||r==null||String(r).length===0)return null;return{host:t,port:n,username:String(r),password:String(e.CODEX_LINUX_PROXY_PASSWORD??``)}}" +
+    "function codexLinuxInstallProxyAuthHandler(e){let t=codexLinuxProxyAuthEntry();if(t==null)return;e.app.on(`login`,(n,r,i,a,o)=>{if(!a?.isProxy)return;let s=codexLinuxProxyAuthHost(a.host),l=String(a.port??``).trim();if(t.host!==s||t.port&&l&&t.port!==l)return;n.preventDefault(),o(t.username,t.password)})}";
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

The proxy-auth match keys on `host` and an optional `port`. `extract_proxy_auth_endpoint` in the launcher only sets `CODEX_LINUX_PROXY_AUTH_PORT` when the server URL contains an explicit `:port` (e.g. `http://proxy.example` with no port leaves it empty). When the env port is empty, this guard `if(t.host!==s||t.port&&l&&t.port!==l)` deliberately skips the port check — so a host-only match still works. That is fine.

The real risk: Electron's `login` event for a proxy reports `details.port` as a number (the test uses `port: 8080`), and the resolved port may differ from what the user wrote in the URL (e.g. user wrote `http://proxy.example` → Chromium uses default 80/3128). Since you only compare ports when *both* sides are non-empty, a host+explicit-port mismatch (user typed `:8080` but Electron reports the scheme-default) would silently refuse to supply credentials, producing the exact `ERR_NO_SUPPORTED_PROXIES`/auth-loop the feature is meant to fix. Consider matching on host only, or documenting that the port in `CODEX_LINUX_PROXY_SERVER` must equal the port Chromium actually dials. Worth verifying against a real authenticated proxy without an explicit port in the URL.

avifenesh (commented):

---

Reviewed the proxy-auth feature: launcher `configure_electron_proxy_from_env`, the `applyLinuxProxyAuthPatch` ASAR patch, the new descriptor module, and tests. One concern worth confirming around the auth host/port matching when the proxy URL omits an explicit port.

cozy-k42:
@ilysenko Ready for review.

I currently use the launcher script to handle/convert the proxy environment vars because Electron itself is not fully compatible with it. Making this an optional feature may not as "clean" as you expected, as I'd still need a way to "hack" the launcher to inject the required proxy configuration.

launcher/start.sh.template
```diff
@@ -2613,6 +2614,92 @@ load_feature_electron_args() {
...
+
+    name="${assignment%%=*}"
+    value="${assignment#*=}"
+    if [ "$name" = "$assignment" ] || [[ ! "$name" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]]; then
+        echo "Ignoring invalid Linux feature launcher hook env assignment: $assignment"
+        return 0
+    fi
+
+    printf -v "$name" '%s' "$value"
+    export "$name"
```
avifenesh, line 2630: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`apply_feature_launcher_hook_env` validates only that the name is a syntactically valid shell identifier, then `export`s it unconditionally. A launcher hook can therefore overwrite *any* launcher-internal variable that is read later in the same process — e.g. `SCRIPT_DIR`, `APP_STATE_DIR`, `CODEX_HOME`, `WARM_START`, `ELECTRON_OZONE_PLATFORM`, `LOG_FILE`. `run_feature_launcher_hooks` runs after `set_electron_defaults` but before `build_electron_launch_args` (line 2864-2865), and many of these are consumed downstream in `build_electron_launch_args` / `launch_electron`. Since hook stdout is parsed verbatim with no namespace restriction, a buggy or malicious hook silently corrupts launcher state rather than only contributing `CODEX_LINUX_PROXY_*` vars. Consider restricting accepted env names to a prefix (e.g. `CODEX_LINUX_PROXY_*` / `CODEX_*`) or rejecting names that collide with known launcher internals.

avifenesh (commented):

---

Reviewed the reworked authenticated-proxy feature (now an opt-in Linux feature with a generic `launcher` runtime-hook surface, the ASAR `patch.js`, the launcher hook, and tests). The launcher integration and feature wiring look correct: `defaultEnabled:false` is enforced, the descriptor is `ciPolicy: "optional"`, and hooks run before `build_electron_launch_args`. One concern around the launcher-hook env protocol that lets a hook overwrite any launcher variable.

Note: the prior review's host/port matching concern (proxy URL without explicit port) still applies to `patch.js` lines 18-21 — `t.port&&l&&t.port!==l` skips the port check when either side is empty, which is the intended host-only fallback, but a user-typed `:8080` that mismatches Chromium's dialed port would silently refuse credentials.

ilysenko (approved):
Reviewed the updated head. The proxy support is now an opt-in Linux feature backed by a generic launcher hook, local focused/smoke tests passed, and GitHub checks are green.

PR #571, opened 2026-06-26
Realign Linux patches to upstream 26.623 bundle (#567)

maplefater:
## Summary

Fixes #567.

Upstream Codex Desktop jumped from app version **26.616 → 26.623.31921**. vite re-chunked the webview (granular per-hook chunks were merged into shared `app-initial~app-main~…` bundles) and refactored several main-process code paths, so a batch of patch anchors no longer matched. `install.sh` aborted with the **5 required-patch failures** reported in #567, and silently skipped ~11 optional patches.

This PR re-anchors those patches to the 26.623 bundle. After it, `install.sh` completes with **0 failed patches**. Every pattern/anchor is **backward-compatible** with older DMGs (old chunk names are kept as alternatives).

## Critical fixes (the 5 `failed-required` patches in #567)

| Patch | Root cause in 26.623 | Fix |
|---|---|---|
| `linux-avatar-overlay-mouse-passthrough` + `main-process-ui` | `findAvatarOverlayClass` aborted its class scan when `findMatchingBrace` hit an unrelated regex-containing class (`class e extends Array{…}`, an SSH-config parser) **before** reaching the avatar class | Skip a class whose braces can't be balanced and keep scanning instead of bailing. The avatar class itself balances fine, so all avatar sub-patches re-apply. |
| `linux-owl-feature-binding-fallback` | Upstream now natively returns `null` on a missing owl binding and swallows `"No such binding was linked"` — the exact fallback this patch injected | Detect the native-safe loader shape and stand down silently → `already-applied`. |
| `linux-fast-mode-model-guard` | Upstream removed `additionalSpeedTiers` and moved to optional chaining; the crash the guard protected against no longer exists | Broaden the pattern to the merged bundle; it no-ops → `already-applied`. |
| `subagent-nickname-metadata-shape` | Content still present but relocated to the merged bundle | Repoint the pattern (backward-compatible) and gate the drift warning on `thread_spawn` so it doesn't false-fire on sibling chunks that also reference `agentNickname`. |

## Optional patches realigned to 26.623

- **`browser-use-node-repl-approval`** — the node_repl mcp-server config moved out of `main--*.js` into a sibling `.vite/build/src-*.js` chunk and was reshaped (`env` hoisted to a variable). Added an extracted-app scanner (mirroring the `owl-feature-binding` precedent) plus the new anchor; warnings are needle-gated so non-target chunks don't false-fail.
[...]

PR #572, opened 2026-06-26
Update nix dmg hash

bearpro:
Updates the hash expected by the Nix flake for `26.623.31921`

This also should fix current [pipeline fail](https://github.com/ilysenko/codex-desktop-linux/actions/runs/28226272535/job/83618885587) in main.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -81,10 +81,10 @@
 
         codexDmg = pkgs.fetchurl {
           url = "https://persistent.oaistatic.com/codex-app-prod/Codex.dmg";
-          hash = "sha256-S7VSvxwZBL1m4oTTScRYKcWREORuSDLF1ps++vkQIT8=";
+          hash = "sha256-/tf+iJCNU/85Voi8PikcZPapQAXt3zP2NdfiLXznpu0=";
         };
 
-        codexVersion = "26.616.81150";
+        codexVersion = "26.623.31921";
         electronVersion = "42.1.0";
         electronPlatform =
           {


scripts/patches/core/all-linux/main-process/browser-integrations/patch.js
```diff
@@ -17,10 +18,20 @@ module.exports = [
   },
   {
     id: "browser-use-node-repl-approval",
-    phase: "main-bundle",
+    phase: "extracted-app",
     order: 160,
     ciPolicy: "optional",
-    apply: applyBrowserUseNodeReplApprovalPatch,
+    apply: applyBrowserUseNodeReplApprovalAssets,
```
avifenesh, line 24: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`browser-use-node-repl-approval` moved from `phase: "main-bundle"` to `phase: "extracted-app"`, with `apply` switched to `applyBrowserUseNodeReplApprovalAssets`, which scans `.vite/build/*.js` for chunks containing `startup_timeout_sec:120` or `NODE_REPL_TRUSTED_BROWSER_CLIENT_SHA256S` and applies `applyBrowserUseNodeReplApprovalPatch` to each.

Worth confirming: the main bundle (`main-*.js`) is itself one of the `.vite/build/*.js` files this scanner walks, and `applyMainBundlePatches` (registry.js) no longer runs this patch since it's no longer `phase: "main-bundle"`. So if the node_repl config still lives in the main bundle on some build, it will only be patched here if `main-*.js` matches one of the two needles. The needle filter (`startup_timeout_sec:120` OR `NODE_REPL_TRUSTED_BROWSER_CLIENT_SHA256S`) should cover that, but please verify on a real 26.623 extract that the chunk(s) actually carrying the mcp config are matched — a missed chunk here fails silently as `skipped-optional` (`matched===0`) with no required-patch alarm, since this descriptor is `optional`.

avifenesh (commented):

---

## Summary

The PR's stated purpose is the `flake.nix` hash/version bump for `26.623.31921`, but it also pulls in a large set of ASAR patch adaptations (from #571) so the Nix derivation actually builds. The patch changes are mostly defensive (broadening filename patterns to also match the merged `app-initial~app-main~…` bundles, adding native-safe "stand down" detections, and re-chunking the node_repl approval patch to scan `.vite/build`).

I traced the framework wiring (`engine.js`, `registry.js`, `patch-report.js`, `shared.js`) and the touched apply functions. The two most sensitive areas — the `required-upstream` patches whose patterns were broadened (`fast-mode-guard`, `subagent-metadata`), and the `extracted-app` phase migration of the node_repl approval patch — appear correctly handled:

- The `browser-use-node-repl-approval` descriptor's `status` callback correctly maps `matched === 0` to `skipped-optional` (descriptor is `optional`, so this is appropriate, not a P2 demotion of a required patch).
- The `owl-feature-binding` descriptor remains `required-upstream` with `matched === 0 → failed-required`, so drift stays visible (P2 honored).
- The broadened `subagent-metadata`/`fast-mode-guard` patterns still no-op safely on non-matching chunks, and the `subagent` warning is now gated on the unique `thread_spawn` needle to avoid false drift alarms.

One thing worth a maintainer's confirmation is flagged inline (the node_repl approval phase change interacting with the main-bundle apply path).

Note: I was unable to execute the JS test suite in this sandbox (`node`/`cd` blocked), so the patch behavior was verified by source reading only, not by running `scripts/patch-linux-window-ui.test.js` / `browser.test.js`.

scripts/patches/core/all-linux/main-process/browser-integrations/patch.js
bearpro: I think this would be addressed in https://github.com/ilysenko/codex-desktop-linux/pull/571#discussion_r3479855019

ilysenko:
Thank you for the pull request. After the drift fixes were applied, our automation ran successfully. It runs every two hours via GitHub Actions. So I am closing this pull request.

Issue #569, opened 2026-06-26
Keyboard input not working — rendererWindowFocused=false on GNOME 49 (Fedora 43)


dojeee:
## Description
Codex Desktop window never receives keyboard focus. Mouse clicks work, but typing produces no input (English or any IME).

## Environment
- OS: Fedora 43
- Kernel: 6.17.8-300.fc43.x86_64  
- GNOME Shell: 49.7
- Mutter: 49.6
- Display: Wayland (also tested X11 via --x11)
- Input: fcitx5 5.1.19
- Codex Desktop: 26.623.31443 (`codex-desktop-2026.06.26.025202-1.fc43`)
- Electron: 42.1.0

## Evidence
Every `electron-message-handler` log entry shows `rendererWindowFocused=false` while `rendererWindowVisible=true`:
```
[electron-message-handler] ... rendererWindowFocused=false ... rendererWindowVisible=true
```
The window is visible and receives mouse events (can click), but keyboard focus is never granted by the compositor.

## What I tried (none helped)
- `--x11` (XWayland fallback)
- `--wayland` (native Wayland)
- `--safe-mode` (software rendering)
- `CODEX_FORCE_RENDERER_ACCESSIBILITY=0`
- Complete uninstall → clean `~/.config/Codex` → reinstall RPM

CLI `codex` from terminal works normally — issue is Desktop-specific.

wintopic:
## Keyboard input not working after update - rendererWindowFocused=false on
  Ubuntu 26.04 / GNOME 50

  ### Description

  After updating Codex Desktop, the app window no longer receives keyboard
  input.

  Mouse interaction still works: I can move the cursor over the composer/input
  box and the cursor changes as expected, and mouse clicks are recognized.
  However, typing does nothing. This happens with plain English input as well
  as with IME input.

  The Codex CLI in terminal works normally. The problem appears to be Desktop-
  specific.

  This looks very similar to #569.

  ### Environment

  - OS: Ubuntu 26.04 LTS
  - Kernel: `7.0.0-22-generic`
  - Desktop: GNOME Shell 50.1
  - Session: Wayland
  - `XDG_CURRENT_DESKTOP=ubuntu:GNOME`
  - `XDG_SESSION_TYPE=wayland`
  - Input method: fcitx5 `5.1.19-1`
  - Input env:
    - `GTK_IM_MODULE=fcitx`
    - `QT_IM_MODULE=fcitx`
[...]

PR #577, opened 2026-06-26
Block Wayland avatar overlay focus traps

ilysenko:
## Summary
- disable avatar overlay passthrough recovery on Linux backends where the overlay cannot safely use X11 input shaping
- gate new 26.623 avatar composition surface keyboard/focus paths behind the same X11 ozone backend check
- add regression coverage for Wayland vs explicit --ozone-platform=x11 overlay focus behavior

Follow-up for #569 and #575.

## Validation
- node --test --test-name-pattern 'avatar' scripts/patch-linux-window-ui.test.js
- ./install.sh ./Codex.dmg
- verified rebuilt app.asar contains the non-shape recovery guard and composition-surface focus guards
- patch report: 0 failed-required patches

### DIFF
--- a/scripts/patches/avatar-overlay.js
+++ b/scripts/patches/avatar-overlay.js
@@ -117,6 +117,10 @@ function avatarApplyInputShapePatch() {
   return "codexLinuxApplyAvatarInputShape(e){if(process.platform!==`linux`||e==null||e.isDestroyed()||typeof e.setShape!=`function`||typeof this.codexLinuxIsAvatarShapeBackend==`function`&&!this.codexLinuxIsAvatarShapeBackend())return!1;try{let t=this.codexLinuxBuildAvatarInputShape(e);if(t==null)return!1;let n=JSON.stringify(t);if(this.codexLinuxAvatarInputShapeKey===n)return!0;e.setShape(t),this.codexLinuxAvatarInputShapeKey=n;return!0}catch{this.codexLinuxAvatarInputShapeKey=null;return!1}}";
 }
 
+function avatarShapeBackendPredicate(electronVar) {
+  return `if(process.platform!==\`linux\`)return!1;let e=\`\`;try{e=${electronVar}.app.commandLine.getSwitchValue(\`ozone-platform\`)}catch{}return e===\`x11\`||e===\`\`&&!process.env.WAYLAND_DISPLAY`;
+}
+
 function patchAvatarOverlayWindowOptions(source) {
   const windowOptionsPatch =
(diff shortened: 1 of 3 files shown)


scripts/patches/avatar-overlay.js
```diff
@@ -154,6 +158,54 @@ function upgradeAvatarOverlayInjectedMethods(source, electronVar) {
...
+    }
+  }
+
+  if (!patchedSource.includes(`function ${helperName}()`)) {
+    return patchedSource;
+  }
+
+  patchedSource = patchedSource.replace(
+    "e.keyboardInteractive?(e.window.setFocusable(!0),e.window.setIgnoreMouseEvents(!1)):e.window.showInactive(),te(e)",
+    `e.keyboardInteractive&&${helperName}()?(e.window.setFocusable(!0),e.window.setIgnoreMouseEvents(!1)):(e.window.setFocusable(!1),e.window.showInactive()),te(e)`,
```
avifenesh, line 191: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This composition-surface focus patch relies on a fixed set of literal needles (the `e.keyboardInteractive?(...)`, `function te(e){...`, `function ce(e,t){...`, and `setFocusable(t.keyboardInteractive||n)` strings). Unlike the other avatar patches in this file, none of these `.replace()` calls record a `recordStrategy(...)` telemetry entry or emit a WARN when the needle is absent — `applyLinuxAvatarOverlayCompositionSurfaceFocusPatch` silently returns the source unchanged if the helper can't be inserted (line 185-187) or if any minified identifier drifts in a future upstream bundle. Per this repo's pattern around required-upstream patch visibility (P2), a focus-trap guard that silently no-ops on bundle drift would re-introduce the Wayland keyboard-focus trap without any signal in the patch report. Consider routing these through `recordStrategy`/a WARN so a missed match is observable rather than invisible.

avifenesh (commented):

---

Reviewed the Wayland avatar overlay focus-trap guards in `scripts/patches/avatar-overlay.js` and the corresponding test additions.

The change introduces `avatarShapeBackendPredicate()` as a shared helper for both `codexLinuxIsAvatarShapeBackend()` and the new composition-surface `codexLinuxAvatarOverlaySurfacesCanFocus()`, gates the new 26.623 composition-surface focus/keyboard paths behind the X11 ozone-backend check, and replaces the unconditional passthrough-recovery start with a backend-conditional policy. The string replacements all target literals that appear uniquely in the bundle (`function I(e){a?.orderSurfaces`, `function te(e){...`, `function ce(e,t){...`), and `applyPatchTwice` idempotency is covered by the tests, so the per-pattern checks (P1 single-occurrence, P3 module placement) hold.

One point worth confirming below.

PR #578, opened 2026-06-26
Fix X11 BrowserWindow option handling

### LINKED ISSUE #576: gui freeze,can not drag,minimize,keyboard input is broken
Image
system version:
6.8.0-124-generic #124~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue May 26 21:05:19 UTC  x86_64 x86_64 x86_64 GNU/Linux

with latest repo version.i get the issue above.
please take a look.THANKS!

Yo-DDV:
## Summary
- restore conditional BrowserWindow option spreading for parent, focusable, backgroundMaterial, and minimumSize in the current 26.623 window manager bundle
- keep the Linux icon window option patch active while applying the undefined-option cleanup independently
- add regression coverage for the current minified BrowserWindow option shape

Fixes #576.

## Validation
- make check
- built a 26.623.31921 lab app from the fixed branch
- verified the rebuilt app.asar uses conditional option spreads instead of explicit undefined BrowserWindow options
- verified on Zorin OS/X11 that the fixed lab window is managed by the window manager, exposes WM_STATE, and supports move/minimize/maximize actions

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -2607,6 +2607,24 @@ test("adds Linux window icon handling to current Linux autoHideMenuBar options",
   assert.match(patched, new RegExp(`icon:${escapeRegExp(iconPathExpression)}`));
 });
 
+test("omits undefined BrowserWindow options in the current window manager bundle", () => {
+  const iconAsset = "app-test.png";
+  const source = [
+    "let M=new a.BrowserWindow({width:b,height:x,...S===void 0||C===void 0?{}:{x:S,y:C},",
+    "title:n??a.app.getName(),backgroundColor:A,show:l,parent:p,focusable:m,",
+    "...process.platform===`win32`?{autoHideMenuBar:!0}:process.platform===`linux`?{icon:process.resourcesPath+`/../content/webview/assets/app-test.png`}:{},",
+    "backgroundMaterial:j??void 0,...D,minWidth:T?.width,minHeight:T?.height,webPreferences:k});",
+  ].join("");
+
+  const patched = applyPatchTwice(applyLinuxWindowOptionsPatch, source, iconAsset);
+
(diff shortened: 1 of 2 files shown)


avifenesh (commented):

---

Reviewed the BrowserWindow option-handling patch. The new `applyDefinedBrowserWindowOptionsPatch` uses a global (`/g`) regex `.replace` callback, which correctly handles all occurrences (P1), and is idempotent because the rewritten output (`...p==null?{}:{parent:p}`) no longer matches the `parent:p,focusable:m,` shape the regex requires. The `linux-window-options` descriptor retains `ciPolicy: "required-upstream"` (P2 OK), and the icon patch path is preserved. Tests cover the new shape and idempotency via `applyPatchTwice`. No correctness issues clearing the bar.

ilysenko:
Thank you for this pull request. Codex still couldn’t figure it out, and Claude was close to this conclusion, but your pull request came first.

Diegoalejandroingelec:
Confirming this fixes it on **Ubuntu 24.04.4 / GNOME 46 / Wayland / IBus** as well.

Before: window unmanaged (override-redirect under X11, `WM_STATE` absent), no keyboard focus (`document.hasFocus()` always false, zero key events reaching the renderer), couldn't drag/minimize — only mouse + right-click paste worked.

After building this branch (`fix/x11-browserwindow-options`) against upstream 26.623.31921: keyboard input works, the window is WM-managed, and drag/minimize/maximize work.

The root cause matches the PR description exactly — the 26.623 bundle constructs the primary `BrowserWindow` with optional fields passed as explicit `undefined` (`parent`, `focusable`, `backgroundMaterial:…??void 0`, `minWidth/minHeight`), which makes the window override-redirect/unmanaged. Restoring conditional option spreads (`...x==null?{}:{…}`) makes it managed again and keyboard focus returns. Verified the rebuilt `app.asar` uses the conditional spreads.

Thanks @Yo-DDV — please merge when you can; this is the actual fix for #569 / #576 too.

PR #582, opened 2026-06-26
fix: sync restored persistent status panel layout state

YudongN:
## Summary

Fixes a lifecycle edge case in the optional `persistent-status-panel` Linux feature.

With the current upstream composer lifecycle, restoring the status panel's local open state from `localStorage` is not enough to keep the parent composer layout marked as open. After switching threads, the status panel component can restore itself as open, while the parent layout state still treats the panel as closed, causing the panel to disappear.

This change adds a small restore-time effect after the patched status state declaration:

- restore `isOpen` from `localStorage` as before
- call `onOpenChange(true)` when the restored state is open
- keep the existing open / close handlers responsible for updating `localStorage`
- preserve fail-soft behavior if the upstream bundle shape drifts

## Validation

- rebuilt the Linux app with `persistent-status-panel` enabled
- rebuilt and installed local `.deb`
- verified the installed `/opt/codex-desktop` bundle contains the layout-sync patch
- manually verified on Ubuntu 22.04 / GNOME X11 using the `.deb` install path:
  - open `/status`
  - switch thread
  - status panel remains open

### DIFF
--- a/linux-features/persistent-status-panel/patch.js
+++ b/linux-features/persistent-status-panel/patch.js
@@ -46,7 +46,7 @@ function applyPersistentStatusPanelPatch(source) {
     return source;
   }
 
-  const [stateNeedle, onOpenChange, _intl, _isOpen, setIsOpen] = match;
+  const [stateNeedle, onOpenChange, _intl, isOpen, setIsOpen, reactVar] = match;
   const openNeedle = `async()=>{${setIsOpen}(!0),${onOpenChange}?.(!0)}`;
   const closeNeedle = `()=>{${setIsOpen}(!1),${onOpenChange}?.(!1)}`;
   if (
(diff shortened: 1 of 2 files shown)


avifenesh (commented):

---

Reviewed the `persistent-status-panel` layout-sync change.

One concern worth confirming before merge: the new restore effect inserts code immediately after the status-state declaration, but the insertion point is derived from the first `;` *after* `statePatch`. In the upstream bundle the destructured declaration is `{...onOpenChange:o}=e,s=Wt(),[c,l]=(0,Z.useState)(...),p;` — the `statePatch` only covers up through `.useState)(...),` (it does not include the trailing `,p`). So `indexOf(";", stateIndex + statePatch.length)` finds the `;` that terminates the whole `let ...,p;` statement, which is correct here. Confirmed against the test fixture and the matched pattern; no defect found on that path.

The match-index/uniqueness guards (`requireUniqueNeedle`) still gate the open/close needle replacements, fail-soft behavior is preserved on shape drift, and the patch remains idempotent via the `STORAGE_KEY` early return. No blocking issues identified.

YudongN:
Closing this draft for now.

After testing against the newer upstream Codex build around 26.623.41415, this feature appears to require patching minified webview bundle internals. The implementation can work for a specific upstream bundle, but it is too tightly coupled to chunk layout and React state structure, which makes it likely to drift with frequent upstream updates.

I do not want to add maintenance burden for a brittle optional feature, so I am withdrawing this follow-up PR for now. No maintainer action is needed.

PR #595, opened 2026-06-27
Fix upstream Computer Use install-flow drift

moxw:
## Summary
- retarget the Linux Computer Use install-flow patch at the current upstream shared plug chunk
- teach the install-flow patch to handle the current required-features gate shape
- remove the older install-flow filename matchers and legacy scalar-gate test cases
- keep the existing opt-in/non-failing CI policy for Computer Use UI patches

## Validation
- node --test scripts/patch-linux-window-ui.test.js --test-name-pattern "Computer Use install flow|current required-feature Computer Use gate|opt-in patch descriptors|criticalFailuresFromReport"
- CODEX_LINUX_ENABLE_COMPUTER_USE_UI=1 ./install.sh --inspect --report-dir dist-next/upstream-computer-use ./Codex.dmg
- node scripts/ci/validate-patch-report.js dist-next/upstream-computer-use/patch-report.json --profile upstream-build
- git diff --check

## Upstream check
The current upstream DMG with Linux Computer Use UI enabled now reports linux-computer-use-install-flow as applied instead of skipped-optional.


ilysenko:
Thanks, this looks directionally good.

One question before we merge: do you think we still need to keep the older install-flow matching paths?

My current preference is to keep these patches aligned with the latest upstream DMG and latest wrapper code only. If the new shared plug chunk is the current upstream shape, removing the legacy matcher/test cases would help avoid unnecessary patch growth and future confusion about what upstream shapes we support.

Does that sound reasonable to you?

moxw:
Yes, that sounds reasonable. I updated the branch to keep this aligned with the current upstream DMG only:

- removed the older `use-plugin-install-flow` / `plugins-availability` descriptor targets
- removed the legacy scalar feature-query matcher and related tests
- kept only the current shared `plug` chunk / required-features gate shape

I re-ran the focused patcher tests plus `CODEX_LINUX_ENABLE_COMPUTER_USE_UI=1 ./install.sh --inspect --report-dir dist-next/upstream-computer-use ./Codex.dmg`; the current upstream report still has `linux-computer-use-install-flow` as `applied`.

scripts/patches/core/all-linux/webview/computer-use-ui/patch.js
```diff
@@ -23,8 +23,8 @@ module.exports = [
     order: 1110,
     ciPolicy: "opt-in",
     enabled: (context) => context.enableComputerUseUi,
-    pattern: /^(use-plugin-install-flow|plugins-availability)-.*\.js$/,
-    missingDescription: "plugin install flow bundle",
+    pattern: /^app-initial~app-main~worktree-init-v2-page~remote-conversation-page~pull-requests-page~plug~.*\.js$/,
```
avifenesh, line 26: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

Retargeting this descriptor pattern (and rewriting `applyLinuxComputerUseInstallFlowPatch` to only handle the `areRequiredFeaturesEnabled:` gate) breaks the existing `test_linux_computer_use_ui_opt_in_smoke` smoke test in `tests/scripts_smoke.sh`, which is run in CI via `scripts/ci/container-entrypoint.sh` (`bash tests/scripts_smoke.sh`).

That test still:
- creates the install-flow asset as `use-plugin-install-flow-test.js` (`tests/scripts_smoke.sh:5529`), a filename the new pattern `^app-initial~app-main~…plug~.*\.js$` no longer matches, so `patchAssetFiles` skips it; and
- feeds the legacy gate body `...re=!ne.isLoading&&ne.enabled...` (`install_flow_body`, line 5558) and then asserts `assert_contains "$install_flow_asset" 'navigator.userAgent.includes(\`Linux\`)'` in branch 2 (line 5603) and branch 3 (line 5622).

With the new descriptor/patch, neither the filename matches nor would the new code emit `navigator.userAgent.includes(\`Linux\`)` for that body, so both `assert_contains` calls will `fail`. The smoke fixture needs the same retarget the Node test got: rename the asset to an `app-initial~app-main~…plug~` chunk, swap `install_flow_body` to the `areRequiredFeaturesEnabled:` shape, and update the branch-1/2/3 assertions to the new `areRequiredFeaturesEnabled:o===\`linux\`||...` markers.

avifenesh (commented):

---

## Stale CI smoke test against the retargeted install-flow patch

This PR retargets the `linux-computer-use-install-flow` patch at the current upstream `app-initial~app-main~…plug~` shared chunk and rewrites `applyLinuxComputerUseInstallFlowPatch` to operate on the `areRequiredFeaturesEnabled:` gate object, dropping the old `re=!ne.isLoading&&ne.enabled` (`navigator.userAgent.includes(\`Linux\`)`) shape. The Node tests and the descriptor were updated accordingly, but the bash smoke test that also exercises this descriptor was not — it still relies on the removed asset name and the removed patch output, so it will fail in CI.

See the inline comment for the specific assertions and why they break. Worth confirming `bash tests/scripts_smoke.sh` passes before merge (CI runs it via `scripts/ci/container-entrypoint.sh`).

ilysenko (approved):
Reviewed the updated head. The stale smoke fixture was fixed and all CI checks are green.

PR #596, opened 2026-06-27
[codex] fix open target discovery drift

moxw:
## Summary
- route Linux open-target command and availability checks through the main-process registry when upstream worker helper shapes drift
- patch the moved native selector chunk so directory/native open menus keep available editor and terminal targets
- add coverage for the latest upstream command, bridge, availability, directory-mode, and selector bundle shapes

## Root Cause
Upstream moved/renamed the open-target worker parameter helper and moved the native selector into route chunks. The old feature patch still let Linux targets be detected in the main bundle, but the enrichment path asked the worker for availability, which filtered Zed and discovered editors back out.

## Validation
- Fresh extracted current `app.asar` patch report shows both open-target descriptors as `applied`

### DIFF
--- a/linux-features/open-target-discovery/patch.js
+++ b/linux-features/open-target-discovery/patch.js
@@ -512,14 +512,11 @@ function applyOpenInTargetRegistryCommandPatch(currentSource) {
   }
 
   const insertionIndex = currentSource.indexOf("async function");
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed current head. No blockers found; CI is green.

PR #597, opened 2026-06-27
[codex] Guard remote mobile turn hydration

moxw:
## Summary

- Guard the `remote-mobile-control` hydration patch so ambiguous `turn/started` notifications are not treated as conversations.
- Resolve hydration from nested thread identifiers when available, and skip the read path when only a turn id is present.
- Add regression tests for both the ambiguous local-turn case and the valid nested-thread case.

## Why

The optional remote mobile hydration patch was using the unknown notification id as a conversation id. On current upstream bundles, that can be a turn id, which makes the Linux app call `thread/read` for a turn and surface a stray running `New chat` row in the sidebar.

## Source Of Truth

- `linux-features/remote-mobile-control/patch.js`
- `linux-features/remote-mobile-control/test.js`

## Validation

- `git diff --check`

## Notes

- This is scoped to the opt-in `remote-mobile-control` Linux feature.
- I did not run the full smoke suite because the change is limited to this feature patch and its targeted tests.


ilysenko (approved):
Reviewed current head including the normalizer helper fix. No blockers found; all CI checks are green.

avifenesh (commented):

---

Reviewed the remote-mobile-control hydration guard. The new needle correctly captures the upstream `let{threadId:e,turn:t}=n.params,r=I(e)` prefix and reuses the captured normalizer (`normalizerFn`) for both `l` and `u`, which the new test validates against the `I(`→`J(` rename. The ambiguous-turn guard (`d==null||u!=null&&d===normalizer(u)`) and the already-hydrated fast-path are sound, and the existing item/turn-completed queue patches are untouched. One observation below; no blocking correctness issue found that clears the bar.

PR #599, opened 2026-06-27
[codex] Fix app-server feature sync sanitization

moxw:
## Summary

Fix the Linux app-server feature enablement compatibility patch so current webview chunks stop leaking unsupported feature keys to the local app-server.

## What Changed

- Match current `app-initial~app-main~*.js` webview chunks for the app-server feature sync patch.
- Sanitize both `var featureList=[...]` and current assignment-style `featureList=[...]` arrays.
- Preserve supported feature keys such as `tool_search` and `mentions_v2` in the sanitizer and local bridge.
- Add regression tests for the current chunk pattern and assignment-style dynamic feature sync shape.

## Why

Current upstream webview assets can include `auth_elicitation` and `apps_mcp_path_override` in the feature sync array, but the local app-server rejects those feature keys. The existing Linux compatibility patch missed the current chunk name and assignment-style array, so unsupported keys were still sent and app-backed plugin/tool materialization could fail.

## Source Of Truth Files

- `scripts/patches/webview-assets.js`
- `scripts/patches/core/all-linux/webview/app-server-feature-enablement/patch.js`
- `scripts/patches/main-process/misc.js`

## Validation

- `node --check scripts/patches/main-process/misc.js`
- `node --check scripts/patches/core/all-linux/webview/app-server-feature-enablement/patch.js`

### DIFF
--- a/scripts/patches/webview-assets.js
+++ b/scripts/patches/webview-assets.js
@@ -631,19 +631,22 @@ function applyLinuxAppServerFeatureEnablementPatch(currentSource) {
 
   function sanitizeFeatureArrayDeclaration(source, arrayVar) {
     const arrayDeclarationRegex = new RegExp(
-      `((?:var\\s+)?${escapeRegExp(arrayVar)}=\\[)([^\\]]*?)(\\])`,
+      `(^|[^\\w$])((?:var\\s+)?${escapeRegExp(arrayVar)}=\\[)([^\\]]*?)(\\])`,
       "u",
     );
     const match = source.match(arrayDeclarationRegex);
     if (match == null) {
       return source;
     }
-    const [, prefix, featureArrayItems, suffix] = match;
+    const [, boundary, prefix, featureArrayItems, suffix] = match;
     const supportedFeatureArrayItems = sanitizeFeatureArrayItems(featureArrayItems);
     if (supportedFeatureArrayItems === featureArrayItems) {
       return source;
     }
-    return source.replace(arrayDeclarationRegex, `${prefix}${supportedFeatureArrayItems}${suffix}`);
+    return source.replace(
+      arrayDeclarationRegex,
+      `${boundary}${prefix}${supportedFeatureArrayItems}${suffix}`,
+    );
   }
 
   const featureArrayRegex =
--- a/scripts/patches/webview-assets.js
+++ b/scripts/patches/webview-assets.js
@@ -607,6 +607,7 @@ function applyLinuxAppServerFeatureEnablementPatch(currentSource) {
     "remote_control",
     "remote_plugin",
     "tool_call_mcp_elicitation",
+    "tool_search",
     "tool_suggest",
   ]);
   const defaultFeaturesMarker = "statsig_default_enable_features";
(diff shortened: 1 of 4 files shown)


avifenesh (commented):

---

Reviewed the app-server feature-sync sanitization changes.

The `sanitizeFeatureArrayDeclaration` regex now optionally matches a leading `var`, which is necessary to support the new assignment-style array (`k7=[...]` inside an IIFE). One concern worth verifying: the regex is unanchored on the left, so `(?:var\s+)?<arrayVar>=\[` can match a substring of a longer identifier when the array variable shares a suffix with another name (e.g. matching `7=[` if `arrayVar` were `7`). For the minified single-letter/short names this engine targets, this would mis-target if any preceding identifier ends with the array var's characters. The flagged case below is the more concrete risk.

scripts/patches/webview-assets.js
```diff
@@ -630,7 +631,7 @@ function applyLinuxAppServerFeatureEnablementPatch(currentSource) {
 
   function sanitizeFeatureArrayDeclaration(source, arrayVar) {
     const arrayDeclarationRegex = new RegExp(
-      `(var\\s+${escapeRegExp(arrayVar)}=\\[)([^\\]]*?)(\\])`,
+      `((?:var\\s+)?${escapeRegExp(arrayVar)}=\\[)([^\\]]*?)(\\])`,
```

---

Making the `var\s+` prefix optional removes the only left-boundary anchor on this match. Previously `var\s+${arrayVar}=\[` guaranteed the match started at a declaration; now `(?:var\s+)?${arrayVar}=\[` can match a trailing substring of a longer identifier when `arrayVar` is a short minified name. For example, if `arrayVar` is `k7` and the bundle also contains an identifier like `Xk7=[...]` earlier (or `j7`/`A7`-style siblings), the regex `k7=\[` will match inside `Xk7=[` and sanitize the wrong array. Consider re-anchoring the optional-`var` branch with a non-identifier left boundary, e.g. `(?:var\s+|[^\w$])` (capturing/preserving the boundary char), so the match still requires `arrayVar` to be a standalone token. The dynamic-builder path derives `arrayVar` from a regex on the bundle, so a sibling identifier sharing the suffix is plausible in real minified output.

moxw: Valid catch. I anchored the assignment-style match with a preserved non-identifier left boundary so short minified names like `k7` no longer match inside longer identifiers such as `Xk7`. I also added a regression test for that exact suffix-collision shape. Validation: `node --test scripts/patch-linux-window-ui.test.js` passes with 305/305, and `node --check scripts/patches/webview-assets.js` passes.

avifenesh (commented):

---

Reviewed the app-server feature-sync sanitization changes.

The core fix is sound: the descriptor pattern now also matches `app-initial~app-main~*.js` chunks, `tool_search` is added to the supported set, the assignment-style array (`(?:var\s+)?${arrayVar}=[`) is matched with a preserved left boundary that prevents matching inside longer identifiers, and the main-process handler feature key list is centralized and kept in sync with the sanitizer. Tests cover both the new chunk name and the suffix-collision case (`Xk7` vs `k7`). The previously raised concern about losing the left-boundary anchor was addressed with the `(^|[^\w$])` capture and a regression test.

One non-blocking note below on the boundary capture interacting with the global `String.replace` semantics. No correctness defect found that clears the bar — the boundary is re-emitted in the replacement so the match is preserved.

ilysenko (approved):
Reviewed current head e022e64; focused validation and GitHub CI passed.

PR #600, opened 2026-06-27
test: isolate check-now CLI environment

Yo-DDV:
## Summary

This isolates the `fresh_check_now_still_clears_stale_wrapper_candidate` updater test from the developer's real CLI environment.

A local global CLI install could be discovered during the test, which made it possible for the test to leave its temporary workspace and exercise the real global update path.

## Changes

- Capture and restore CLI-related environment variables mutated by the test.
- Force home and path lookup into the temporary test workspace.
- Disable system CLI lookup for this test.
- Keep the test compatible with the repository's clippy settings.

## Source of truth

- `updater/src/app.rs`

## Validation

- `cargo fmt --check`
- `cargo test -p codex-update-manager fresh_check_now_still_clears_stale_wrapper_candidate`
- `bash -n install.sh scripts/lib/*.sh launcher/start.sh.template scripts/install-deps.sh scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh`

### DIFF
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -2055,12 +2055,14 @@ mod tests {
             "HOME",
             "PATH",
             "NVM_DIR",
+            "XDG_CONFIG_HOME",
             "CODEX_CLI_PATH",
             "CODEX_UPDATE_MANAGER_SKIP_SYSTEM_CLI_LOOKUP",
         ]);
         std::env::set_var("HOME", temp.path());
         std::env::set_var("PATH", temp.path().join("missing-bin"));
         std::env::remove_var("NVM_DIR");
+        std::env::remove_var("XDG_CONFIG_HOME");
         std::env::remove_var("CODEX_CLI_PATH");
         std::env::set_var("CODEX_UPDATE_MANAGER_SKIP_SYSTEM_CLI_LOOKUP", "1");
 
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -2043,12 +2043,26 @@ mod tests {
         Ok(())
     }
 
-    #[tokio::test]
-    async fn fresh_check_now_still_clears_stale_wrapper_candidate() -> Result<()> {
+    #[test]
+    fn fresh_check_now_still_clears_stale_wrapper_candidate() -> Result<()> {
+        let _env_guard = crate::test_util::env_lock();
+        let runtime = tokio::runtime::Runtime::new()?;
         let temp = tempfile::tempdir()?;
         let paths = test_paths(temp.path());
         paths.ensure_dirs()?;
         let config = test_config(temp.path());
+        let _restore_env = crate::test_util::EnvRestoreGuard::capture(&[
+            "HOME",
+            "PATH",
+            "NVM_DIR",
+            "CODEX_CLI_PATH",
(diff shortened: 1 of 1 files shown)


avifenesh (commented):

---

One finding on the env-isolation completeness. The RAII guard mechanism itself is correct (Drop-based restore), but the capture/override set leaves an escape vector that defeats the PR's stated goal.

updater/src/app.rs
```diff
@@ -2043,12 +2043,26 @@ mod tests {
...
+            "PATH",
+            "NVM_DIR",
+            "CODEX_CLI_PATH",
+            "CODEX_UPDATE_MANAGER_SKIP_SYSTEM_CLI_LOOKUP",
+        ]);
+        std::env::set_var("HOME", temp.path());
+        std::env::set_var("PATH", temp.path().join("missing-bin"));
+        std::env::remove_var("NVM_DIR");
+        std::env::remove_var("CODEX_CLI_PATH");
+        std::env::set_var("CODEX_UPDATE_MANAGER_SKIP_SYSTEM_CLI_LOOKUP", "1");
```
avifenesh, line 2067: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

Isolation is incomplete: this overrides `HOME` but not `XDG_CONFIG_HOME`, so the test can still escape the temp workspace.

`run_check_now` calls `codex_cli::reconcile_if_present` (app.rs:414, before the `if_stale` early-return) → `resolve_cli_path` → `known_cli_locations` (codex_cli.rs:328-336), which derives nvm CLI candidates from `xdg_nvm_root`:

```rust
fn xdg_nvm_root(home: &Path) -> PathBuf {
    std::env::var_os("XDG_CONFIG_HOME")
        .map(PathBuf::from)
        .unwrap_or_else(|| home.join(".config"))
        .join("nvm")
}
```

When `XDG_CONFIG_HOME` is set in the developer/CI environment (common on Linux desktops), this resolves to the *real* `$XDG_CONFIG_HOME/nvm/...` instead of `$temp/.config/nvm`. `resolve_cli_path` can then discover a real nvm-managed `codex`, and `preflight` will execute it via `read_installed_version` — exactly the "leave the temporary workspace / exercise the real global path" leak this PR aims to close. Note `CODEX_UPDATE_MANAGER_SKIP_SYSTEM_CLI_LOOKUP=1` only suppresses `/usr/local/bin` and `/usr/bin`, not the XDG/nvm home locations.

The sibling `command_path_env` test captures and clears `XDG_CONFIG_HOME` for precisely this reason (codex_cli.rs:812-825). Suggest adding `"XDG_CONFIG_HOME"` to the capture list here and a matching `std::env::remove_var("XDG_CONFIG_HOME")` alongside the other overrides.

Yo-DDV: Good catch, thanks. Included in 9fb09a0: the test now captures and clears XDG_CONFIG_HOME as well, so the nvm lookup stays inside the temporary workspace.

ilysenko (approved):
Reviewed current head 9fb09a0de65428dc413fd3f4c9d46d8fbd1ceddb. The XDG_CONFIG_HOME review feedback is addressed and focused cargo test passed.

PR #602, opened 2026-06-27
Add Niri window backend for Linux Computer Use

benm-dev:
## Summary
- add a Niri window backend using `niri msg --json windows` and `niri msg action focus-window --id`
- hydrate `NIRI_SOCKET` for Codex Desktop sessions and expose Niri in doctor/capability output
- document Niri support in Linux Computer Use docs and changelog

## Verification
- `nix develop -c cargo test -p codex-computer-use-linux niri`
- `nix develop -c cargo test -p codex-computer-use-linux`
- `nix develop -c cargo build -p codex-computer-use-linux --bin codex-computer-use-linux`
- live Niri smoke: `target/debug/codex-computer-use-linux doctor` reports `window_control: ["niri"]`
- live Niri smoke: `target/debug/codex-computer-use-linux windows` reports backend `niri` and lists real windows


computer-use-linux/src/windowing/backends/niri.rs
```diff
@@ -0,0 +1,156 @@
...
+                .map(|[x, y]| (finite_rounded_i32(x), finite_rounded_i32(y)))
+                .unwrap_or((None, None));
+
+            Some(WindowBounds {
+                x,
+                y,
+                width,
+                height,
+            })
+        });
```
avifenesh, line 131: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`pos_in_scrolling_layout` is not a global screen coordinate — in niri's IPC schema it is the window's position *within the scrolling layout of its workspace/output*, not an absolute screen-space origin. These `WindowBounds.x`/`y` values flow directly into `window_crop_rect()` (server.rs:2454, used to crop screenshots) and into `apply_window_relative_click_coordinates()` (server.rs:2490, used as the absolute origin for relative clicks). With `can_exact_focus: true` and `requires_exact_focus` paths active, a relative click or window-cropped screenshot will be offset by the layout origin rather than the real screen position.

Note the test fixture itself shows the mismatch: `pos_in_scrolling_layout: [2.0, 1.0]` for a 952x1070 window yields `x: Some(2)` — a 2-pixel origin for a window that is clearly not at the top-left of the screen. Compare the Hyprland backend, which uses `at` (real layout coordinates) and i3 which uses `rect.x/y` (absolute). Consider deriving the absolute position from niri's per-output/workspace geometry, or omitting `x`/`y` (leaving bounds origin `None`) so the relative-click and crop paths fail safe instead of clicking the wrong location.

avifenesh (commented):

---

Reviewed the Niri window backend addition. The Rust backend follows the established pattern of the Hyprland/i3 backends closely (probe / list / activate, descriptor registration, capability map, diagnostics wiring), and the env hydration of `NIRI_SOCKET` is correctly threaded through `DESKTOP_ENV_KEYS`. The tray.js change in the second commit is unrelated to Niri but appears self-consistent (idempotent needle handling with old/new variants and matching test assertions).

One correctness concern on the window bounds mapping is flagged inline.

ilysenko:
Thanks for the work on this.

- The Niri backend currently maps `pos_in_scrolling_layout` to `WindowBounds.x/y`. In Niri IPC this is a column/tile index, not a screen position, so it would produce incorrect bounds for window screenshots and relative window clicks. Could you please either derive the origin from real layout coordinates, such as `tile_pos_in_workspace_view` where appropriate, or leave `x/y` unset when a reliable origin is not available? The parsing test should also cover that semantic difference.
- The PR currently leaves `tests/scripts_smoke.sh` failing in the Linux tray patch smoke. The new expected-zero occurrence check hits `grep -o` under `set -o pipefail`, so the helper exits before reporting the count when there are no matches. Could you please fix that helper or the assertion so the smoke suite passes?

One smaller process note: the tray activation fix seems separate from the Niri backend. It would be easier to review and maintain if that landed in a separate PR, unless there is a direct dependency here.

avifenesh:
@benm-dev @ilysenko note that niri is GPL3, I avoided it deliberately, and https://github.com/agent-sh/agent-workspace-linux is implementing IPC from scratch for the same reason. It won't be upstreamed (or downstreamed) to https://github.com/agent-sh/computer-use-linux and will create divergence.

ilysenko:
Thanks for putting this together and for the time spent on it.

We're going to close this for now because we're not ready to merge this direction at the moment. We may revisit it later if the project scope changes.

PR #603, opened 2026-06-28
Support fast mode selection when logged in with an API key

AHGGG:
## Summary

Adds an opt-in Linux feature, `api-key-service-tier`, that exposes the Codex Desktop Fast/service-tier selector for API-key authenticated OpenAI-compatible providers.

The feature patches the app main webview bundle to allow service-tier controls when the active host is logged in with an API key. It preserves the upstream ChatGPT/OAuth entitlement check for official Fast mode, and synthesizes a `fast` service tier option when model metadata does not provide `serviceTiers`.

This feature is disabled by default and must be enabled through `linux-features/features.json`.

## Why

Codex Desktop currently gates the Fast/service-tier UI behind the ChatGPT auth path and upstream account requirements. That is appropriate for official ChatGPT/OAuth usage, but it prevents API-key authenticated OpenAI-compatible providers from exposing a `serviceTier` setting even when the provider supports Codex's `serviceTier` request field.

This keeps official account behavior unchanged while allowing API-key provider users to opt in explicitly.

## Changed Files

- `linux-features/api-key-service-tier/feature.json`
  - Registers the disabled-by-default feature and its patch descriptor entrypoint.

- `linux-features/api-key-service-tier/patch.js`
  - Allows the service-tier UI gate for `authMethod === "apikey"`.
  - Keeps ChatGPT/OAuth hosts on the upstream `fast_mode` entitlement check.
  - Synthesizes a fallback `fast` service tier option when model metadata has no `serviceTiers`.

- `linux-features/api-key-service-tier/README.md`
  - Documents enablement, config shape, behavior, risks, and validation.

- `linux-features/api-key-service-tier/test.js`
  - Covers descriptor loading, opt-in behavior, target matching, auth gate patching, fallback tier synthesis, and patch idempotency.

- `README.md`
  - Adds the feature to the Linux Features table.

## Behavior

When disabled, there is no behavior change.

When enabled:

- API-key authenticated hosts can show service-tier controls.
- Models without `serviceTiers` metadata get a synthesized `fast` UI option.
- ChatGPT/OAuth hosts still use the upstream account `fast_mode` entitlement check.
- Providers must still support the `serviceTier` request setting for this to have an effect.

### DIFF
--- a/linux-features/api-key-service-tier/patch.js
+++ b/linux-features/api-key-service-tier/patch.js
@@ -1,0 +1,109 @@
+"use strict";
+
+const JS_IDENT = "[A-Za-z_$][\\w$]*";
+const PATCH_MARKER = "codexLinuxApiKeyFastTier";
+
+function warn(message, patchName) {
+  console.warn(`WARN: ${message} - skipping ${patchName}`);
+}
+
+function applyApiKeyServiceTierGatePatch(source) {
+  const gateNeedle = new RegExp(
+    `(${JS_IDENT})=(${JS_IDENT})\\?\\.authMethod===\\\`chatgpt\\\`,` +
+      `(${JS_IDENT})=\\2\\?\\.authMethod\\?\\?null([\\s\\S]{0,500}?),` +
+      `d=\\1&&!(${JS_IDENT})&&(${JS_IDENT})!=null&&\\6\\?\\.requirements\\?\\.featureRequirements\\?\\.fast_mode!==!1`,
+    "g",
+  );
+
+  const patched = source.replace(
+    gateNeedle,
+    (_match, isChatGptVar, hostVar, authMethodVar, middle, loadingVar, requirementsVar) =>
+      `${isChatGptVar}=${hostVar}?.authMethod===\`chatgpt\`,` +
+      `${authMethodVar}=${hostVar}?.authMethod??null${middle},` +
+      `d=!${loadingVar}&&(${isChatGptVar}?${requirementsVar}!=null&&${requirementsVar}?.requirements?.featureRequirements?.fast_mode!==!1:${authMethodVar}===\`apikey\`)`,
+  );
+
+  if (patched !== source || source.includes(`${authMethodVarName(source)}===\`apikey\``)) {
+    return patched;
+  }
+
+  if (source.includes("featureRequirements?.fast_mode") && source.includes("authMethod===`chatgpt`")) {
+    warn("Could not find service tier auth gate", "API key service tier gate patch");
+  }
+  return source;
+}
+
+function authMethodVarName(source) {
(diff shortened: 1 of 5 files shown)


linux-features/api-key-service-tier/patch.js
```diff
@@ -0,0 +1,109 @@
...
+
+  const optionsPatch = new RegExp(
+    `\\.\\.\\.\\((${JS_IDENT})\\?\\.serviceTiers\\?\\?\\[\\]\\)\\.map\\((${JS_IDENT})=>\\(\\{` +
+      `description:(${JS_IDENT})\\(\\2\\),iconKind:(${JS_IDENT})\\(\\2\\.id,\\2\\.name\\),` +
+      `label:(${JS_IDENT})\\(\\2\\),tier:\\2,value:\\2\\.id\\}\\)\\)`,
+    "g",
+  );
+  patched = patched.replace(
+    optionsPatch,
+    `...(($1?.serviceTiers?.length?$1.serviceTiers:[${PATCH_MARKER}($1)]).filter(Boolean)).map($2=>({description:$3($2),iconKind:$4($2.id,$2.name),label:$5($2),tier:$2,value:$2.id}))`,
```
avifenesh, line 103: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`applyFallbackFastTierPatch` rewrites the shared fast-tier resolver and the option list **unconditionally** — there is no `authMethod==="apikey"` scope check here, unlike `applyApiKeyServiceTierGatePatch`. The gate patch correctly preserves ChatGPT/OAuth behavior by branching on `isChatGptVar`, but this fallback synthesis (`codexLinuxApiKeyFastTier`) runs for *every* host, including ChatGPT/OAuth hosts whose model catalog has no `serviceTiers` metadata. That means a ChatGPT host that the upstream entitlement check would still gate could now see a synthesized `fast` option appear in the selector list, since the option-map at line 72-75 no longer depends on the gate.

The option list and resolver should be conditioned on the API-key context (or the synthesized tier filtered out for ChatGPT hosts) so the fallback only affects API-key providers, matching the feature's stated intent ("ChatGPT/OAuth hosts still use the upstream account `fast_mode` entitlement check"). As written, the fallback is broader than the gate.

avifenesh (commented):

---

## Review: api-key-service-tier feature

The feature is correctly disabled-by-default (`defaultEnabled: false`, validated by the framework), has a unique descriptor id, uses match-index/`replaceAll`-style replacement (the `optionsPatch`/`fastResolverPatch` use the `g` flag), and is not wired into the default ASAR patch set. One concern below regarding the unconditional rewrite of a shared resolver (P4-adjacent: the fallback applies regardless of auth context).

ilysenko (approved):
Reviewed current head 0bba9c7263debb283804bd612663375ee3cf3b6b. The feature is disabled by default, the synthetic Fast fallback is scoped through an API-key model marker, and focused feature tests passed.

PR #604, opened 2026-06-28
[codex] Fix Linux freeze paths in Chrome, mobile sync, and skills lists

moxw:
## Summary
- Patch every Chrome integration runtime resolver in the Electron 42 bundle, including the Chrome plugin app-server Codex CLI sync path.
- Keep the Linux runtime helper insertion idempotent while allowing multiple resolver patches in one bundle.
- Fix remote-mobile conversation hydration so turn/started can fall back to top-level threadId when the nested turn thread id is absent.

## Why
Two Linux freeze paths showed up in live logs.

The first was a repeated Chrome plugin app-server reconcile failure. The existing Linux Chrome runtime patch handled native-host sync, but a separate app-server resolver still required the bundled Electron Codex runtime that does not exist in the Linux package.

The second was a remote-mobile notification loop. A new thread could be unreadable briefly during session creation; after the earlier turn-id guard, the hydration patch skipped that case as ambiguous when only top-level threadId identified the real thread. Item notifications then fell through as unknown conversation updates and kept the renderer busy.

## Source Of Truth Files
- scripts/patches/chrome-plugin.js
- linux-features/remote-mobile-control/patch.js
- linux-features/remote-mobile-control/test.js

## User-visible Behavior
- Chrome plugin sync can use CODEX_CLI_PATH or codex from PATH on Linux instead of spinning on a missing bundled Electron runtime.
- Remote-mobile turn hydration can recover when upstream sends a real top-level threadId and no nested turn.threadId.
- The changes are source patches and tests only; generated app/package output is not committed.

## Validation
- git diff --check


ilysenko (approved):
Reviewed current head. CI and focused local checks passed; no code blockers found.

PR #606, opened 2026-06-28
Fix Linux open-target registry helper drift

joshyorko:
## Summary

- harden the Linux open-target registry helper when the upstream params helper returns a worker parameter object instead of the target list
- keep the newer `pP` lookup path when present, then fall back to `settingsStore.targets`
- make the existing Exec fallback test deterministic when host Linuxbrew helpers are available

## Validation

## Notes

This fixes the live Linux failure where file-viewer "Open in" could throw `find is not a function` before falling back to the feature's main-process target registry.

### DIFF
--- a/linux-features/open-target-discovery/patch.js
+++ b/linux-features/open-target-discovery/patch.js
@@ -507,6 +507,22 @@ function applyLinuxIconSummaryResolutionPatch(currentSource) {
 }
 
 function applyOpenInTargetRegistryCommandPatch(currentSource) {
+  const helper =
+    "function codexLinuxOpenTargetRegistryTargets(e){let t=null;try{t=typeof pP===`function`?pP(e):iP(e)}catch{}return Array.isArray(t)?t:Array.isArray(e?.targets)?e.targets:[]}" +
+    "async function codexLinuxOpenTargetRegistryCommand(e,t){if(process.platform!==`linux`)return;let n=codexLinuxOpenTargetRegistryTargets(e).find(e=>e.id===t),r=typeof ZN!==`undefined`?ZN:typeof IN!==`undefined`?IN:void 0;return typeof n?.detect===`function`?await n.detect(r):null}";
+  const legacyHelpers = [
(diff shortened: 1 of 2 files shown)


avifenesh (commented):

---

Reviewed the open-target registry helper hardening. The new `codexLinuxOpenTargetRegistryTargets` indirection cleanly handles the case where the upstream params helper (`iP`/`pP`) returns a worker-param object instead of a target list, with a sensible fallback to `e?.targets`. The legacy-helper rewrite list is idempotent (the `helper`-present early return guards re-application). The test change makes the Exec fallback deterministic by forcing the linuxbrew `gio`/`gtk-launch` paths to fail and asserting on `calls.at(-1)`, which avoids ordering dependence on whatever host helpers are present.

One observation worth confirming below.

linux-features/open-target-discovery/patch.js
```diff
@@ -507,15 +507,29 @@ function applyLinuxIconSummaryResolutionPatch(currentSource) {
 }
 
 function applyOpenInTargetRegistryCommandPatch(currentSource) {
+  const helper =
+    "function codexLinuxOpenTargetRegistryTargets(e){let t=null;try{t=typeof pP===`function`?pP(e):iP(e)}catch{}return Array.isArray(t)?t:Array.isArray(e?.targets)?e.targets:[]}" +
+    "async function codexLinuxOpenTargetRegistryCommand(e,t){if(process.platform!==`linux`)return;let n=codexLinuxOpenTargetRegistryTargets(e).find(e=>e.id===t),r=typeof ZN!==`undefined`?ZN:typeof IN!==`undefined`?IN:void 0;return typeof n?.detect===`function`?await n.detect(r):null}";
```
avifenesh, line 512: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`codexLinuxOpenTargetRegistryTargets` falls back to `iP(e)` (single-arg) when `pP` is absent. In the latest bundle shape `iP` is the worker-params builder `function iP(e,t){return{target:t}}` — calling it with one arg returns `{target:undefined}`, which is not an array, so the code correctly falls through to `Array.isArray(e?.targets)?e.targets:[]`. That relies on `e` (the settings store) actually exposing a `.targets` array. Worth confirming the runtime always passes the settings store (which has `.targets`) here and never the worker-params object itself — if the latter ever reaches this helper, both branches yield `[]` and every Linux target silently resolves as unavailable rather than throwing. The existing call sites (`getSettingsStore()` / `this.options.settingsStore`) look correct, so this is a defensive note rather than a confirmed break.

PR #607, opened 2026-06-28
Fix Read Aloud patching for shared app-main bundles

joshyorko:
## Summary

- update Read Aloud bundle matching for current shared `app-initial~app-main` webview chunks
- preserve the feature's settings/global-state patching on the newer split bundle layout
- add coverage for the current shared bundle filename shape

## Validation

## Notes

This was validated in a rebuilt Linux app where the Read Aloud plugin now appears and works.


linux-features/read-aloud/patch.js
```diff
@@ -200,6 +201,41 @@ function applySettingsPatch(source) {
...
+}
+
+function applyLinuxDesktopSettingsPatch(source) {
```
avifenesh, line 208: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`applyLinuxDesktopSettingsPatch` and the injected `LinuxReadAloudSettings` source assume the `linux-desktop-settings-linux.js` bundle exposes `LinuxSettingsSection`, `LinuxSettingsRow`, `React.Fragment`, and a `$.jsx(LinuxSettingsSection,{title:"Build",children:$.jsx(LinuxBuildInfoPanel,{})})` shape.

But this asset is generated by this repo in `scripts/patches/keybinds-settings.js::buildLinuxDesktopSettingsSource`, and the real generated output uses the *imported* aliases instead:
- `SettingsSection` / `SettingsRow` / `SettingsGroup` / `SettingsPage` (imported, e.g. `import{...as SettingsSection}from"./settings-group-*.js"`), not `LinuxSettingsSection` / `LinuxSettingsRow`.
- The Build section is `$.jsxs(SettingsSection,{className:"gap-2",children:[$.jsx(SettingsSection.Header,{title:"Build"}),$.jsx(SettingsSection.Content,{children:$.jsx(SettingsGroup,{children:$.jsx(LinuxBuildInfoPanel,{})})})]})` — not `$.jsx(LinuxSettingsSection,{title:"Build",children:$.jsx(LinuxBuildInfoPanel,{})})`.
- `grep` over the generated source shows `SettingsSection`/`SettingsRow`/`SettingsGroup`/`SettingsPage` only; there are zero occurrences of `LinuxSettingsSection`, `LinuxSettingsRow`, or `React.Fragment`.

Against the real bundle:
1. The Build-section `replace` needle (line 231) won't match, so `LinuxReadAloudSettings` is never wired into the rendered tree.
2. Even if injected, `LinuxReadAloudSettings` references `LinuxSettingsSection`/`LinuxSettingsRow`, which are undefined in the generated bundle → a ReferenceError when the section renders.

The new test at `test.js:1278` uses a hand-written fixture matching these `Linux*`/`React.Fragment` shapes rather than `buildLinuxDesktopSettingsSource` output, so the test passing does not demonstrate the patch works on a real build. Recommend regenerating the fixture from `buildLinuxDesktopSettingsSource` (or aligning the needles/source to the imported `SettingsSection`/`SettingsRow`/`SettingsSection.Header`+`.Content` shape) so coverage reflects the actual asset.

avifenesh (commented):

---

## Read Aloud — `linux-desktop-settings-linux.js` patch likely targets identifiers/shapes that don't exist in the real generated bundle

The new `applyLinuxDesktopSettingsPatch` is the substantive correctness concern here. The `linux-desktop-settings-linux.js` asset is **generated by this repo** in `scripts/patches/keybinds-settings.js::buildLinuxDesktopSettingsSource`, and the actual generated source does not use the identifiers/shapes the new patch (and its test fixture) assume. See inline comment for the specifics. The descriptor pattern broadening and the `app-initial~app-main~...` test addition look fine.

ilysenko:
Thanks for working on this.

The blocker is that `applyLinuxDesktopSettingsPatch()` does not match the real generated `linux-desktop-settings-linux.js` shape from `scripts/patches/keybinds-settings.js`. It adds the Read Aloud settings function, but does not wire it into the rendered Linux desktop settings tree.

Please update the patch and fixture against the real generated asset shape.

ilysenko (approved):
Reviewed the current head. The previous Linux desktop settings blocker is addressed: the patch now targets the real generated SettingsSection/SettingsRow/SettingsGroup shape and the focused Read Aloud tests pass.

PR #608, opened 2026-06-28
Add Linux workspace root Open In targets

joshyorko:
## Summary

- add a core Linux patch for workspace-root "Open in" targets
- inject VS Code, VS Code Insiders, Zed, terminal, and file-manager project menu entries from the Linux open-target registry
- tolerate aliased callback handlers and current shared app-main project index chunks

## Validation

- `node --test scripts/workspace-root-open-targets-patch.test.js`

## Notes

This brings the project/workspace-root Open In menu in line with the file-level Open In Linux integration.

### DIFF
--- a/scripts/patches/core/all-linux/extracted-app/workspace-root-open-targets/patch.js
+++ b/scripts/patches/core/all-linux/extracted-app/workspace-root-open-targets/patch.js
@@ -8,6 +8,8 @@ const {
 } = require("../../../../shared.js");
 
 const PATCH_MARKER = "codexLinuxWorkspaceRootOpenTarget";
+const MISSING_FILE_MANAGER_ACTION_REASON =
+  "Could not find workspace-root File Manager open action while Linux targets are enabled";
 
 function warn(message) {
   console.warn(`WARN: ${message} - skipping Linux workspace-root open targets patch`);
@@ -315,7 +317,12 @@ function patchWorkspaceRootOpenTargets(extractedDir) {
   }
 
   if (matched === 0) {
-    return { matched, changed };
+    return {
+      matched,
+      changed,
+      status: "failed-required",
+      reason: MISSING_FILE_MANAGER_ACTION_REASON,
+    };
   }
   return { matched, changed };
 }
--- a/scripts/patches/core/all-linux/extracted-app/workspace-root-open-targets/patch.js
+++ b/scripts/patches/core/all-linux/extracted-app/workspace-root-open-targets/patch.js
@@ -1,0 +1,300 @@
+"use strict";
+
+const fs = require("node:fs");
+const path = require("node:path");
+
+const {
+  findMainBundle,
+} = require("../../../../shared.js");
+
+const PATCH_MARKER = "codexLinuxWorkspaceRootOpenTarget";
+
+function warn(message) {
+  console.warn(`WARN: ${message} - skipping Linux workspace-root open targets patch`);
+}
+
+function findMatching(source, openIndex, openChar, closeChar) {
+  let depth = 0;
+  const stack = [{ type: "code" }];
+
+  for (let index = openIndex; index < source.length; index += 1) {
+    const char = source[index];
+    const top = stack[stack.length - 1];
+
+    if (top.type === "string") {
+      if (top.escaped) {
+        top.escaped = false;
+      } else if (char === "\\") {
+        top.escaped = true;
+      } else if (char === top.quote) {
+        stack.pop();
+      }
+      continue;
+    }
+
+    if (top.type === "template") {
+      if (top.escaped) {
+        top.escaped = false;
+      } else if (char === "\\") {
+        top.escaped = true;
+      } else if (char === "`") {
+        stack.pop();
+      } else if (char === "$" && source[index + 1] === "{") {
+        stack.push({ type: "templateExpression", depth: 1 });
+        index += 1;
+      }
+      continue;
+    }
+
+    if (char === "\"" || char === "'") {
+      stack.push({ type: "string", quote: char, escaped: false });
+      continue;
+    }
+    if (char === "`") {
+      stack.push({ type: "template", escaped: false });
+      continue;
+    }
+
+    if (top.type === "templateExpression") {
+      if (char === "{") {
+        top.depth += 1;
+      } else if (char === "}") {
+        top.depth -= 1;
+        if (top.depth === 0) {
+          stack.pop();
+        }
+      }
+      continue;
+    }
+
+    if (char === openChar) {
+      depth += 1;
+    } else if (char === closeChar) {
+      depth -= 1;
+      if (depth === 0) {
+        return index;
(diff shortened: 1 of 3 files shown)


scripts/patches/core/all-linux/extracted-app/workspace-root-open-targets/patch.js
```diff
@@ -0,0 +1,300 @@
...
+  if (currentSource.includes(PATCH_MARKER)) {
+    return currentSource;
+  }
+  if (targets.length === 0) {
+    warn("Could not find Linux editor or terminal open targets in the main bundle");
+    return currentSource;
+  }
+
+  const openCallPattern = /([A-Za-z_$][\w$]*)\(\{path:([A-Za-z_$][\w$]*),cwd:([A-Za-z_$][\w$]*),target:`fileManager`,openFile:([A-Za-z_$][\w$]*)\.mutate\}\)/u;
+  const openCallMatch = currentSource.match(openCallPattern);
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`applyWorkspaceRootOpenTargetsPatch` only ever patches the **first** `target:`fileManager`` site: `currentSource.match(openCallPattern)` returns a single match, and `findItemAssignment` / `onSelectNameCandidates` then operate around that one location. The file filter at lines 270-277 explicitly accepts both `app-main-*` and the shared `app-initial~app-main~...projects-index*` chunk, and the workspace-root File Manager action can plausibly appear more than once within a single chunk (e.g. current-workspace menu + projects-index list menu colocated in the shared chunk). When it does, only the first occurrence gets the VS Code/Zed/Terminal entries; the rest are silently left as File-Manager-only.

This is exactly the single-occurrence-replacement failure mode (every patch helper must process every eligible occurrence, not just the first). Recommend looping over all `openCallPattern` matches (match-index based) and applying the menu-item rewrite to each, or asserting the needle is unique per chunk. The idempotency guard at line 176 (`includes(PATCH_MARKER)`) does not help here — once the first match is patched the marker is present, so a second un-patched occurrence in the same file is skipped entirely on every subsequent run.

avifenesh (commented):

---

## Review: Linux workspace-root Open In targets

This patch injects editor/terminal/file-manager "Open in" menu entries into the workspace-root dropdown by string-surgery on minified app-main chunks. Two correctness concerns worth a maintainer's attention; both relate to multi-occurrence robustness and the patch's CI policy.

scripts/patches/core/all-linux/extracted-app/workspace-root-open-targets/patch.js
```diff
@@ -0,0 +1,300 @@
...
+    return { matched, changed };
+  }
+  return { matched, changed };
+}
+
+module.exports = {
+  id: "linux-workspace-root-open-targets",
+  phase: "extracted-app",
+  order: 2060,
+  ciPolicy: "optional",
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`ciPolicy: "optional"` means that if the upstream bundle shape drifts (the `target:`fileManager`` action, the `()=>{}` callback shape, or the `.Item` assignment pattern stop matching), every `warn()` early-return at lines 180/187/199/207/214/230 leaves the chunk unpatched and the build still passes green — the regression is invisible. The matched file(s) at line 278 already contained `target:`fileManager`` (the feature the user expects to extend), so a no-op here is a genuine feature regression, not a benign skip. Consider whether this should be `required-upstream` so drift surfaces as `failed-required` rather than `skipped-optional`, or at minimum distinguish "no candidate chunk found" (optional skip) from "candidate chunk found but patch failed to apply" (which should be loud).

ilysenko:
Thanks for working on this.

Two blockers before this can merge:
- `node --test scripts/patch-linux-window-ui.test.js` fails because the new `linux-workspace-root-open-targets` core descriptor is missing from the expected descriptor inventory.
- The PR says VS Code Insiders is included, but `enabledWorkspaceRootTargets()` only adds VS Code, Zed, and Terminal.

Please fix the descriptor test and either add VS Code Insiders support or remove that claim.

scripts/patches/core/all-linux/extracted-app/workspace-root-open-targets/patch.js
```diff
@@ -0,0 +1,345 @@
...
+    if (patched !== source) {
+      fs.writeFileSync(filePath, patched, "utf8");
+      changed += 1;
+    }
+  }
+
+  if (matched === 0) {
+    return { matched, changed };
+  }
+  return { matched, changed };
```
avifenesh, line 327: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

Drift in the `target:`fileManager`` action becomes invisible here. When no asset chunk matches the `fileManager` open action (`matched === 0`), `patchWorkspaceRootOpenTargets` returns `{matched:0, changed:0}` with no `status` and no warnings. The engine's `applyExtractedAppPatchDescriptors` then calls this descriptor's `status(result, warnings)`: `result.status` is null, `result.changed` is falsy, and `warnings.length === 0`, so it falls through to `return "already-applied"` (line 341). `already-applied` is in `SUCCESS_STATUSES`, so `criticalFailuresFromReport` never flags it even though this is a `required-upstream` patch.

This is exactly the P2 failure mode: an upstream bundle that no longer contains a workspace-root `target:`fileManager`` site (e.g. the chunk was renamed or the action restructured) produces a green build with the Open In menu silently un-patched. Contrast with the `apply` body, which already `warn()`s when `matchedOpenCall` is false inside `applyWorkspaceRootOpenTargetsPatch` — but that path is only reached when a candidate chunk is read; if no candidate chunk passes the `target:`fileManager`` filter at all (line 306), `matched` stays 0 and no warning is emitted.

Consider returning a `failed-required`/drift status (or emitting a warning) when `matched === 0` after the main bundle reported enabled targets, so required-upstream drift is caught.

avifenesh (commented):

---

## Review: Linux workspace-root Open In targets

This descriptor is `ciPolicy: "required-upstream"`, so it must surface as `failed-required` when the upstream bundle shape drifts (per the CI gate in `scripts/ci/validate-patch-report.js`, which only catches non-success, applicable statuses, plus the missing-from-report case). The patch's own `status()` callback contains a hole where upstream drift is silently demoted to a success status. One inline note below.

The prior P1 (single-occurrence) and P2 (`ciPolicy: optional`) concerns appear addressed: the open-call replacement now iterates `matchAll` and patches every site, and `ciPolicy` is `required-upstream` with the descriptor wired into the expected inventory in `patch-linux-window-ui.test.js`.

ilysenko (approved):
Reviewed the current head after the maintainer fix. The required-upstream drift case now fails explicitly when Linux targets are enabled but the workspace-root File Manager chunk is missing, and focused tests cover that behavior.

PR #610, opened 2026-06-28
Fix Linux desktop settings for current webview bundles

colonelpanic8:
## Summary
- restore Linux desktop settings injection for current webview route-map bundles
- detect the current Codex request helper shape outside setting-storage chunks
- generate fallback settings layout components when upstream no longer ships the previous settings row/layout assets

## Validation
- node --check scripts/patches/shared.js
- node --check scripts/patches/keybinds-settings.js
- node --check scripts/patch-linux-window-ui.test.js
- node --test --test-name-pattern "Linux desktop" scripts/patch-linux-window-ui.test.js
- git diff --check
- smoke-tested against installed Nix webview bundle; keybinds-settings report is applied and linux-desktop-settings-linux.js is generated

### DIFF
--- a/scripts/patches/keybinds-settings.js
+++ b/scripts/patches/keybinds-settings.js
@@ -524,9 +524,13 @@ function collectLinuxDesktopRouteAndNavigationPatches(extractedDir) {
     throw new Error(`Required Keybinds settings patch failed: missing webview assets directory ${webviewAssetsDir}`);
   }
 
+  // Newer builds split the lazy settings route map out of `app-main-*.js`/`index-*.js`
+  // into hashed concatenation chunks named `app-initial~app-main~*.js` (e.g. the
+  // automations-page chunk), while the icon/navigation metadata stays in
+  // `settings-page-*.js`. Scan all three so the route map is always discoverable.
   const candidates = fs
     .readdirSync(webviewAssetsDir)
-    .filter((name) => /^(?:app-main|index)-.*\.js$/.test(name) || /settings-page.*\.js$/.test(name))
+    .filter((name) => /^(?:(?:app-main|index)-|app-initial~app-main~).*\.js$/.test(name) || /settings-page.*\.js$/.test(name))
     .sort();
 
   let routeMatched = false;
@@ -903,9 +907,13 @@ function applyKeybindsSettingsIndexPatch(currentSource) {
 }
 
 function isSettingsRouteBundleSource(currentSource) {
+  // Only treat a chunk as the route map when general-settings is wired to a lazy
+  // page component (or the Linux desktop route is already present). The previous
+  // `"general-settings":ID,...,"keyboard-shortcuts":ID` heuristic also matched the
+  // icon map (slug -> SVG component), which caused the route patch to inject the
+  // settings page component as a navigation icon and render a broken nav entry.
   return currentSource.includes(linuxDesktopSettingsAsset)
-    || /"general-settings":\(0,[A-Za-z_$][\w$]*\.lazy\)\(\(\)=>[A-Za-z_$][\w$]*\(/.test(currentSource)
-    || /"general-settings":[A-Za-z_$][\w$]*,[^}]*"keyboard-shortcuts":[A-Za-z_$][\w$]*/.test(currentSource);
+    || /"general-settings":\(0,[A-Za-z_$][\w$]*\.lazy\)\(\(\)=>[A-Za-z_$][\w$]*\(/.test(currentSource);
 }
 
 function isSettingsSectionsMetadataBundleSource(currentSource) {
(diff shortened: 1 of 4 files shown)


Yo-DDV:
The fix looks relevant to me, and the added tests are meaningful. The only blocker I see right now is that this PR conflicts with `main` after #609 was merged, so it probably needs a rebase on top of `0abd25f` before it can be reviewed or merged cleanly.

avifenesh (commented):

---

## Review notes

Reviewed the settings-injection rework for current webview bundles. The route/nav/icon disambiguation and the split-chunk handling are well-tested. Two concerns worth a look around the new fallback-component path and the broadened request-asset scan.

### 1. `writeLinuxSettingsFallbackComponents` always writes the fallback assets, even when native upstream components exist

In `resolveSettingsAssetDependencies`, `writeLinuxSettingsFallbackComponents()` is called unconditionally (line 370) before checking whether the upstream `settings-row-*`, `settings-content-layout-*`, `settings-group-*`, `settings-surface-*` assets exist. The fallback files (`linux-settings-row-linux.js`, etc.) are therefore always written to the extracted webview dir, even on bundles where they aren't used. That's mostly harmless leftover, but worth confirming it doesn't break any "matched assets" accounting or leave orphaned `*-linux.js` files that later patch passes scan over (e.g. `collectOptionalMatchingAssetPatches` iterates all `.js`). See inline.

### 2. `findCodexRequestWebviewAsset` now scans every `.js` asset for `vscode://codex/`

The widened scan (shared.js:193-198) iterates all `.js` files after `setting-storage-*`. `findCodexRequestExportName` gates on a precise regex, so false positives are unlikely, but the first asset that both contains `vscode://codex/` and matches the wrapper regex wins by sort order. If more than one asset matches, the selection is order-dependent. See inline.

Neither is a confirmed defect — flagging for author confirmation.

scripts/patches/keybinds-settings.js
```diff
@@ -236,15 +367,19 @@ function resolveSettingsAssetDependencies(extractedDir, { includeHotkeySettings
         "hotkey settings asset",
       )
     : null;
-  const toggleAsset = findRequiredWebviewAsset(webviewAssetsDir, /^toggle-.*\.js$/, null, "toggle asset");
-  const settingsRowAsset = findRequiredWebviewAsset(webviewAssetsDir, /^settings-row-.*\.js$/, null, "settings row asset");
+  const fallbackComponents = writeLinuxSettingsFallbackComponents(webviewAssetsDir, {
+    jsxRuntimeAsset,
+    jsxRuntimeExportName,
+  });
```
avifenesh, line 364: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`writeLinuxSettingsFallbackComponents` is called unconditionally and always writes the four `linux-settings-*-linux.js` files into `webview/assets`, before the `tryFindRequiredWebviewAsset` calls below determine whether upstream equivalents exist. On bundles that ship native `settings-row-*`/`settings-content-layout-*`/etc., these fallback files are written but unused.

Two things to confirm:
1. Orphaned `*-linux.js` files persist in the extracted tree. `collectOptionalMatchingAssetPatches` (called later in the same `patchKeybindsSettingsAssets` run) iterates *all* `.js` assets and runs the section/shared predicates against them — verify these generated files can never match `isSettingsSectionsMetadataBundleSource` / `isSettingsSharedMetadataBundleSource` (they don't today, but it's a latent coupling).
2. Consider writing the fallbacks lazily only when the corresponding upstream asset lookup returns null, to avoid shipping dead assets.

scripts/patches/shared.js
```diff
@@ -173,10 +186,16 @@ function findCodexRequestWebviewAsset(webviewAssetsDir) {
...
+  const settingStorageCandidates = fs
     .readdirSync(webviewAssetsDir)
     .filter((name) => regexpTest(/^setting-storage-.*\.js$/, name))
     .sort();
+  const allRequestCandidates = fs
+    .readdirSync(webviewAssetsDir)
+    .filter((name) => regexpTest(/\.js$/, name))
+    .sort()
+    .filter((name) => !settingStorageCandidates.includes(name));
+  const modernCandidates = [...settingStorageCandidates, ...allRequestCandidates];
```
avifenesh, line 198: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

The candidate list now includes every `.js` asset (not just `setting-storage-*`) and the first one that contains `vscode://codex/` AND passes `findCodexRequestExportName` wins, ordered by filename sort. If two assets satisfy both conditions the selection is purely sort-order dependent and could bind `__post` to the wrong module. Given P1 (single-occurrence/ambiguous-match) concerns in this repo, consider asserting uniqueness or preferring the most specific match rather than relying on lexical sort order.

avifenesh, line 198: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`findCodexRequestWebviewAsset` now scans **every** `.js` asset and returns the first one whose source contains `vscode://codex/` *and* whose body matches the request-wrapper regex in `findCodexRequestExportName`. The old code restricted modern matching to `setting-storage-*.js`. The new fallback list (`allRequestCandidates`) is `.sort()`ed alphabetically, so among multiple assets that both contain `vscode://codex/` and an exported request wrapper, selection is purely lexical. If a non-canonical chunk (e.g. a re-export shim or a test/demo bundle) sorts before the real request helper, `__post` in the generated settings asset will be imported from the wrong module. The `setting-storage-*` prefix is kept first, which mitigates the common case, but the broadened tail has no uniqueness assertion. Consider asserting the wrapper is unique among candidates, or narrowing the fallback to a known chunk-name shape (e.g. `shared-app-*`/`shared-runtime-*`) rather than all `.js`.

scripts/patches/keybinds-settings.js
avifenesh, line 364: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`writeLinuxSettingsFallbackComponents` is invoked unconditionally inside `resolveSettingsAssetDependencies`, so it always writes the four `linux-settings-*-linux.js` files into `webview/assets` even when the upstream `settings-row-*`, `settings-content-layout-*`, `settings-group-*`, and `settings-surface-*` assets are all present and used (the `tryFind... ?? fallback` lines below mean the fallbacks are typically discarded). This leaves orphaned, unreferenced assets in the bundle on the common path and runs the writes before the route/nav patches that can still throw and abort the whole patch — leaving stray files behind on failure. Consider writing fallbacks lazily only when a required asset is actually missing, so the no-fallback path produces no orphaned files.

avifenesh (commented):

---

Reviewed the settings-injection rework for the current webview bundles. Most of the change is sound: the route/icon disambiguation (`isSettingsRouteBundleSource`, the navigation-bundle guard in `applyLinuxDesktopSettingsRoutePatch`) directly fixes the broken-nav-icon bug, and the split-route chunk handling has matching test coverage. Two correctness concerns below regarding the broadened candidate scans in `shared.js` and the always-on fallback writes.

Per repo policy I do not approve/request-changes — flagging for human judgment.

ilysenko (approved):
Looks good. Validated locally and CI is green.

PR #620, opened 2026-06-29
Add optional UI tweaks feature

ilysenko:
## Summary
- Add a disabled-by-default `ui-tweaks` Linux feature for small UI customization patches.
- Add the first modular tweak, `sidebar.projectName`, with CSS defaults in `feature.json` and local overrides from gitignored `linux-features/features.json` settings.
- Extend the Linux feature config flow so per-feature settings are passed to patches and preserved through bootstrap edits and update-builder packaging.
- Document available sub-tweaks, settings, drift behavior, and how future tweaks should be registered.

## Tests/checks run

## Notes
- The sidebar project-name default is `font-weight: 700 !important; padding-top: 0.25rem;` and does not force a color.
- The patch is fail-soft: drift or invalid style config emits `WARN` and leaves the app bundle unchanged or falls back to the default style.
- Local ignored `linux-features/features.json` is still only for testing/user preferences and is not part of this PR.

### DIFF
--- a/linux-features/ui-tweaks/patches/sidebar-project-name.js
+++ b/linux-features/ui-tweaks/patches/sidebar-project-name.js
@@ -1,9 +1,12 @@
 "use strict";
 
 const DEFAULT_PROJECT_NAME_STYLE = "font-weight: 700 !important; padding-top: 0.25rem;";
+const PROJECTS_SIDEBAR_ASSET_PATTERN =
+  /^app-initial~app-main~remote-conversation-page~projects-index-page-[^.]+\.js$/;
 const PROJECT_NAME_SELECTOR = ".group\\/folder-row .min-w-0.truncate.pr-1";
 const STYLE_ID = "codex-linux-ui-tweaks-sidebar-project-name-style";
 const RUNTIME_MARKER = "codexLinuxUiTweaksSidebarProjectNameStyleRuntime";
+const UNSAFE_PROJECT_NAME_STYLE_PATTERN = /[{}@<>]|\r|\n|\/\*|\*\/|\burl\s*\(/i;
 
 const SIDEBAR_PROJECT_NAME_MARKERS = [
   "sidebarElectron.projectsNavLink",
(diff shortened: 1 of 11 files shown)


linux-features/ui-tweaks/patches/sidebar-project-name.js
```diff
@@ -0,0 +1,123 @@
...
+    ...(defaults != null && typeof defaults === "object" && !Array.isArray(defaults) ? defaults : {}),
+    ...(settings != null && typeof settings === "object" && !Array.isArray(settings) ? settings : {}),
+  };
+}
+
+function normalizedProjectNameStyle(context) {
+  const config = sidebarProjectNameConfig(context);
+  if (config.enabled === false) {
+    return { enabled: false, style: DEFAULT_PROJECT_NAME_STYLE };
+  }
```
avifenesh, line 53: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`sidebarProjectNameConfig` reads defaults from `context.feature.manifest.tweaks...`, but `loadEnabledLinuxFeatures` only attaches `settings` to the feature object (`{ ...feature, settings: ... }` in linux-features.js:319). The `feature` object's `manifest` is the normalized manifest, which does include the raw `tweaks` key (spread via `...manifest` in normalizeLinuxFeatureManifest), so the manifest default path works at runtime.

However, in the real load path the per-tweak `enabled` default lives only in `feature.json`'s `tweaks` and is reached through `context.feature.manifest`. The test at test.js:154 calls `descriptor.apply(projectBundleFixture(), {})` and relies on this — confirm that `wrapFeaturePatchDescriptor` passes the full `feature` (with `.manifest`) into `featureContext`. It does (linux-features.js:441/405-407), so this is fine. No change required; flagging only to confirm the manifest-default wiring is intentional and exercised end-to-end rather than just via hand-built contexts.

avifenesh (commented):

---

Reviewed the new `ui-tweaks` Linux feature and the feature-settings plumbing. The patch correctly uses `context.feature.manifest` / `context.feature.settings` for config, is fail-soft, and injects idempotently (RUNTIME_MARKER/STYLE_ID guards). One concern about how the manifest defaults reach the patch via `context.feature.manifest`.

Note on always-on checks: this patch appends a `<style>` runtime to the asset (not a `string.replace()`), so P1 (single-occurrence replacement) does not apply; webview-asset injection is sidebar-list scoped (P4) and the feature is `defaultEnabled: false` with a unique descriptor id. Those gates pass.

PR #624, opened 2026-06-29
Preserve user Node path in desktop terminal

### LINKED ISSUE #623: Local node version is override by codex-desktop node runtime from codex terminal
## Summary

In codex terminal, node version is different from local path.

## Environment

- Distro/version: Arch Linux rolling
- Package format: pacman package

## Reproduction

In Codex Desktop terminal,
```bash
which node
# /opt/codex-desktop/resources/node-runtime/bin/node
node -v
# v22.22.2
```

In local terminal,
```bash
which node
# /usr/bin/node
node -v
# v20.20.2
```

ilysenko:
## Summary
- Stop prepending the packaged managed Node runtime to the global launcher PATH inherited by Electron.
- Keep managed Node available through CODEX_MANAGED_NODE_RUNTIME_DIR and explicit Browser Use / node_repl env vars.
- Add a smoke guard so the launcher does not leak resources/node-runtime/bin ahead of the user PATH again.

Fixes #623

## Tests/checks run
- git diff --check
- bash tests/scripts_smoke.sh

## Notes
- This is a draft PR for review; it intentionally does not change managed Node packaging or Browser Use runtime resolution.

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -540,13 +540,9 @@ export_packaged_runtime_env() {
     fi
 }
 
-prepend_managed_node_runtime_to_path() {
+export_managed_node_runtime_env() {
     [ -x "$MANAGED_NODE_BIN_DIR/node" ] || return 0
 
-    case ":$PATH:" in
-        *":$MANAGED_NODE_BIN_DIR:"*) ;;
-        *) export PATH="$MANAGED_NODE_BIN_DIR:$PATH" ;;
-    esac
     export CODEX_MANAGED_NODE_RUNTIME_DIR="$SCRIPT_DIR/resources/node-runtime"
 }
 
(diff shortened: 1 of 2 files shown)


Yo-DDV:
Hummm I'm late =(

Yo-DDV:
Thanks for picking this up. I can reproduce #623 against the current packaged Linux install: the launcher prepends `resources/node-runtime/bin` before Electron starts, and the terminal backend forwards `process.env` into `node-pty`, so `which node` resolves to the packaged runtime.

I checked this branch's approach against the relevant runtime paths. Keeping the managed runtime out of the launcher `PATH` while leaving `CODEX_BROWSER_USE_NODE_PATH` / `NODE_REPL_NODE_PATH` explicit is the right shape for the fix.

Local checks run from my side:
- `git diff --check`

Looks good to me.

PR #625, opened 2026-06-29
Keep integrated terminals on the user PATH

### LINKED ISSUE #623: Local node version is override by codex-desktop node runtime from codex terminal

Yo-DDV:
## Summary
- Preserve the launcher-managed Node runtime for internal Linux app flows.
- Capture the pre-managed-runtime user `PATH` before the launcher prepends `resources/node-runtime/bin`.
- Restore that user `PATH` only for local integrated terminal sessions before passing the environment to `node-pty`.

Fixes #623

## Root cause
The Linux launcher prepends the packaged managed Node runtime to `PATH` before Electron starts. The terminal backend then builds local terminal environments from `process.env`, so the integrated terminal resolves `node` to `/opt/codex-desktop/resources/node-runtime/bin/node` instead of the user's normal terminal `node`.

## Notes
This keeps the managed runtime available to updater and internal app paths while narrowing the behavior change to local terminal sessions.

## Checks
- `node --check scripts/patches/main-process/misc.js && node --check scripts/patch-linux-window-ui.js`
- `node --test scripts/patch-linux-window-ui.test.js --test-name-pattern 'terminal|default core patch descriptors'`
- `node scripts/patch-linux-window-ui.js --enforce-critical --report-json ...` against a copied extracted app
- `timeout 900 bash tests/scripts_smoke.sh`


ilysenko:
Thanks, could we make this patch optional instead of `required-upstream`? It touches the main app bundle, so on future upstream changes it seems safer for this terminal PATH fix to fail soft and not apply than to risk breaking the whole app/update.

Yo-DDV:
Good point, agreed. I'll make this fail soft so future upstream drift reports as optional instead of blocking the app/update path, while keeping the current behavior covered by tests. I'll update the branch shortly.

ilysenko (approved):
Reviewed current head b224f0dcdb3a3869d61be4e3f35f2f9b39ef9e9d. The terminal PATH patch is now optional/fail-soft, local validation passed, and GitHub checks are green.

PR #628, opened 2026-06-29
Fix Linux build information action layout

PinguuSS:
Summary
- Render Linux build information actions inline with their related metadata rows instead of in the SettingsRow control column.
- Keep Copy/Open actions next to the source commit, Refresh next to Generated, and Details next to Metadata file.
- Add regression coverage so the fallback Linux desktop settings page does not recreate the detached side action group.

Validation
- node --check scripts/patches/keybinds-settings.js
- node --test --test-name-pattern "keeps Linux desktop toggles visible" scripts/patch-linux-window-ui.test.js
- git diff --check
- CODEX_INSTALL_DIR=<tmp>/codex-app make build-app, then inspected generated linux-desktop-settings-linux.js
- bash tests/scripts_smoke.sh


PinguuSS:
I opened this follow-up PR because my previous PR was marked ready for review while I was still treating it as a draft and finishing one last layout fix plus validation.

This PR only contains that final focused change: moving the Linux build information buttons back inline with their related metadata rows, and adding regression coverage so they do not end up detached in the SettingsRow control column again.

ilysenko:
Thanks for opening the follow-up. This one is on Codex: I marked and merged the previous PR a bit too quickly because many contributors leave PRs in draft even after they are effectively ready.

I’ll review this follow-up separately and make sure the final layout fix gets checked properly before merging.

PR #629, opened 2026-06-30
[codex] Make workspace-root open targets optional on drift

gingygames:
## Summary
- make the workspace-root open-targets patch optional when upstream no longer contains the File Manager insertion point
- return skipped-target/skipped-optional instead of failing the full app patch flow
- update descriptor and focused patch tests for the new optional drift behavior

## Root cause
Latest upstream Codex Desktop builds can omit the File Manager workspace-root action chunk. The Linux wrapper treated that as required upstream drift, which blocked building otherwise usable Linux apps even though the open-targets enhancement can be skipped safely.

## Validation
- `node --test scripts/workspace-root-open-targets-patch.test.js scripts/patch-linux-window-ui.test.js`
- `git diff --check`

### DIFF
--- a/scripts/patches/core/all-linux/extracted-app/workspace-root-open-targets/patch.js
+++ b/scripts/patches/core/all-linux/extracted-app/workspace-root-open-targets/patch.js
@@ -9,7 +9,7 @@ const {
 
 const PATCH_MARKER = "codexLinuxWorkspaceRootOpenTarget";
 const MISSING_FILE_MANAGER_ACTION_REASON =
-  "Could not find workspace-root File Manager open action while Linux targets are enabled";
+  "Workspace-root File Manager open action is not present in this upstream build";
 
 function warn(message) {
   console.warn(`WARN: ${message} - skipping Linux workspace-root open targets patch`);
@@ -320,7 +320,7 @@ function patchWorkspaceRootOpenTargets(extractedDir) {
     return {
       matched,
       changed,
-      status: "failed-required",
+      status: "skipped-target",
       reason: MISSING_FILE_MANAGER_ACTION_REASON,
     };
   }
@@ -331,7 +331,7 @@ module.exports = {
   id: "linux-workspace-root-open-targets",
   phase: "extracted-app",
   order: 2060,
-  ciPolicy: "required-upstream",
+  ciPolicy: "optional",
   apply: patchWorkspaceRootOpenTargets,
   status(result, warnings) {
     if (result?.status != null) {
@@ -339,11 +339,11 @@ module.exports = {
     }
     if (result?.changed) {
       return warnings.length > 0
-        ? { status: "failed-required", reason: warnings[0] }
+        ? { status: "applied-with-warnings", reason: warnings[0] }
         : "applied";
     }
     if (warnings.length > 0) {
-      return { status: "failed-required", reason: warnings[0] };
+      return { status: "skipped-optional", reason: warnings[0] };
     }
     return "already-applied";
   },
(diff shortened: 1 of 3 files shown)


scripts/patches/core/all-linux/extracted-app/workspace-root-open-targets/patch.js
```diff
@@ -320,7 +320,7 @@ function patchWorkspaceRootOpenTargets(extractedDir) {
     return {
       matched,
       changed,
-      status: "failed-required",
+      status: "skipped-target",
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

When the File Manager action is genuinely absent from upstream (`matched === 0`), this returns `skipped-target`. But `skipped-target` is in `NOT_APPLICABLE_STATUSES` in `scripts/lib/patch-report.js`, so it is filtered out of *both* `criticalFailuresFromReport` and `optionalDriftFromReport`. The result: real upstream drift (the enhancement silently dropped because the insertion point disappeared) is never surfaced — not even in the optional-drift WARN summary.

`skipped-target` is intended for "not applicable here" cases (platform targeting / no enabled targets), which is correct for the `targets.length === 0` branch at line 295. But here Linux targets *are* enabled and the patch *would* apply if upstream had the chunk — that is exactly optional drift.

Compare the sibling optional File Manager patches: `linux-worker-file-manager` (window-shell/patch.js:99-107) and `linux-app-updater-bridge` (updater/patch.js:14-18) both return `skipped-optional` (not `skipped-target`) when `matched === 0`, precisely so the drift stays visible. Recommend returning `skipped-optional` here too, with the `status()` `result.status` passthrough adjusted accordingly.

scripts/workspace-root-open-targets-patch.test.js
```diff
@@ -217,16 +217,16 @@ test("workspace root open targets patch fails required when Linux targets are en
...
 
     const result = patchWorkspaceRootOpenTargets(root);
-    const expectedReason = "Could not find workspace-root File Manager open action while Linux targets are enabled";
+    const expectedReason = "Workspace-root File Manager open action is not present in this upstream build";
 
     assert.deepEqual(result, {
       matched: 0,
       changed: 0,
-      status: "failed-required",
+      status: "skipped-target",
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This test now locks in `skipped-target` for the absent-File-Manager case, which makes the drift invisible to `optionalDriftFromReport` (see comment on patch.js). If the status is changed to `skipped-optional` to keep drift visible, this assertion (and line 229) should be updated to match.

avifenesh (commented):

---

## Review: workspace-root open targets optional drift

One concern about the chosen status for missing-upstream drift. The descriptor is now `ciPolicy: "optional"` (matching P2: required→optional is the intended behavioral change here, so the demotion itself is fine and explicitly tested). My concern is the *status* used when the File Manager chunk is absent.

When `matched === 0` (File Manager action genuinely missing from upstream), the patch now returns `status: "skipped-target"`. Per `scripts/lib/patch-report.js`, `skipped-target` is in `NOT_APPLICABLE_STATUSES`, so it is excluded from **both** `criticalFailuresFromReport` and `optionalDriftFromReport`. That means upstream removing the File Manager insertion point becomes completely invisible — it won't even surface in the optional-drift warning summary (`scripts/lib/asar-patch.sh`, `scripts/ci/validate-patch-report.js`).

That contradicts the PR's own stated intent ("return skipped-target/skipped-optional instead of failing") and the established convention for sibling optional `extracted-app` File Manager patches, which surface absent targets as drift, not as not-applicable. See inline comment.

ilysenko (approved):
Reviewed the updated head after fixing the optional drift status. Checks are green.

PR #630, opened 2026-06-30
Update authenticated proxy patch

cozy-k42:


linux-features/authenticated-proxy/patch.js
```diff
@@ -84,15 +91,22 @@ function applyAuthenticatedProxyPatch(currentSource) {
...
+    [
+      `let l=${electronVar}.net.request({method:n,url:i,headers:t}),u=-1,d=()=>{let e=l.getUploadProgress();!e.started||e.current===u||(u=e.current,r({loaded:e.current,total:e.total}))}`,
+      `let l=${electronVar}.net.request({method:n,url:i,headers:t});codexLinuxAttachProxyAuthToRequest(l);let u=-1,d=()=>{if(r==null)return;let e=l.getUploadProgress();!e.started||e.current===u||(u=e.current,r({loaded:e.current,total:e.total}))}`,
+    ],
+    [
+      `let u=${electronVar}.net.request({method:n,url:i,headers:t,useSessionCookies:s}),d=-1,f=()=>{let e=u.getUploadProgress();!e.started||e.current===d||(d=e.current,r({loaded:e.current,total:e.total}))}`,
+      `let u=${electronVar}.net.request({method:n,url:i,headers:t,useSessionCookies:s});codexLinuxAttachProxyAuthToRequest(u);let d=-1,f=()=>{if(r==null)return;let e=u.getUploadProgress();!e.started||e.current===d||(d=e.current,r({loaded:e.current,total:e.total}))}`,
+    ],
+  ];
+  const requestReplacement = requestReplacements.find(([needle]) => patchedSource.includes(needle));
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`find()` applies only the first matching `[needle, replacement]` pair, so each fetch/request site is patched at most once per `applyAuthenticatedProxyPatch` call. This matches P1 (process all eligible occurrences) only if each needle is guaranteed to appear at most once in the bundle. The old code used the same single `string.replace()` semantics, so this is not a regression — but if upstream ever emits the same `performProgressRequest` shape at two call sites, only the first would be patched and the second would silently route around proxy auth without triggering the drift warning (the `else if` guard checks for `codexLinuxAttachProxyAuthToRequest(` anywhere in the source, so a second unpatched occurrence is masked once the first is patched). Worth confirming these needles are unique in the real bundle, or switching to an index-based replace-all.

avifenesh (commented):

---

Review of authenticated-proxy patch update. The change generalizes the fetch/request needle matching to support both the old and new upstream bundle shapes by iterating over candidate `[needle, replacement]` pairs and applying the first match. The descriptor remains `ciPolicy: "optional"` (P2 not applicable). The drift-detection `else if` warnings are preserved, and the `requestReplacements` warning guard was correctly broadened to `codexLinuxAttachProxyAuthToRequest(` (now that two variable names — `l` and `u` — are possible). One observation below.

ilysenko:

- This drift fix keeps the old fetch/request needles as fallback candidates. Per CONTRIBUTING, drift fixes should target only the latest `CODEX.DMG` and remove legacy DMG shapes in the same PR.
- Please drop the old candidate needles/tests and keep only the current upstream shape, then rerun the focused authenticated-proxy tests.

ilysenko (approved):
Reviewed the updated head after removing the legacy drift paths. Checks are green.

PR #633, opened 2026-06-30
[codex] Block Ctrl-W closing Linux windows

SuroshAhmadZobair:
## Summary
- Add a Linux main-process patch that prevents Ctrl+W from closing Electron windows through the `before-input-event` layer.
- Use `removeMenu()` for Linux menu removal and normalize older menu snippets that only hid the menu bar.
- Cover the patch descriptor, Ctrl+W insertion shapes, menu normalization, and full main-bundle patch behavior in the existing patch suite.

## Validation
- `node --test scripts/patch-linux-window-ui.test.js` (322/322 passing)
- `git diff --check`


Yo-DDV:
Thanks for the patch. I gave this a deeper pass, and I think it needs another iteration before it is ready.

The current branch still leaves `Rust and Smoke Tests` red. I can reproduce the same failure locally with `timeout 300 bash tests/scripts_smoke.sh`: `tests/scripts_smoke.sh:4537` still expects the old Linux `setMenuBarVisibility(!1)` snippet, while this patch now emits `removeMenu()`.

The bigger concern is the new `before-input-event` guard. It prevents the default action for every plain `Ctrl+W` in every Linux `webContents`. That should stop the close accelerator, but it also prevents renderer key handling, so it can break legitimate `Ctrl+W` usage in editors, text fields, terminals, or other embedded UI. I would rather avoid a global input-layer block here. If the issue is the window/menu accelerator, the fix should be scoped to menu shortcuts/window chrome, or at least prove that editable/content shortcuts still work.

There is also an interaction with `frameless-titlebar`: after the new core `linux && removeMenu()` snippet, the feature patch does not recognize it as already handled and inserts a second Linux menu branch with `setMenuBarVisibility(!1)`. I reproduced that with the core menu patch followed by `applyFramelessTitlebarMainPatch`. It should treat the new core snippet as already applied/upgradable and cover that in the feature test.

Minor robustness point: the Ctrl+W insertion regex misses ``const $a=require(`electron`)`` / ``var $a=require(`electron`)`` even though `requireName()` accepts `$` aliases. Not blocking for the current bundle, but worth hardening since the descriptor is `required-upstream`.

For reference, the unit tests pass for me (`node --test scripts/patch-linux-window-ui.test.js`, `node --test linux-features/frameless-titlebar/test.js`, selected feature tests, and `cargo check -p codex-update-manager`). I also applied the patch to a real extracted `app.asar` with `--enforce-critical`, and the required descriptor applies cleanly. So the patch is not drifting on the current bundle; the remaining blockers are the red smoke assertion and the global Ctrl+W behavior.

PR #635, opened 2026-07-01
Sanitize Linux external open environment

KaiStarkk:
## Summary

This is a small Linux packaging fix for external URL opens from the desktop app.

The packaged app can run with wrapper-specific environment variables, including `LD_LIBRARY_PATH`. When the app delegates a URL to the desktop browser, that environment can leak into the external process and make browser launch/default-app handling fail on some Linux setups.

## User-visible behavior

External links opened from the app should now be handed to the desktop via `xdg-open` with the wrapper environment scrubbed first.

## Scope

This only changes Linux external URL handoff from the app. It does not try to solve broader desktop association, portal, or file-manager configuration issues.

## Implementation notes

The patch wraps Electron require initializers in the main bundle and patches `shell.openExternal()` for Linux string URLs. It removes `LD_LIBRARY_PATH`, `LD_PRELOAD`, and Node/Electron/Codex wrapper variables before spawning `xdg-open`.

I marked the patch `required-upstream` because silently skipping this on bundle-shape drift would leave the packaged app in the broken state again.

## Validation

- `git diff --check`
- Ad hoc runtime smoke with fake Electron/`xdg-open`


ilysenko:
Thanks for working on this. The direction makes sense, but there are a few blockers before this can merge:

- CI is currently failing in `tests/scripts_smoke.sh` because the new helper adds a second `process.platform!==\`linux\`` occurrence. The smoke assertion should be updated to check the tray patch more narrowly, not just relaxed blindly.
- The patch now hard-depends on `xdg-open`, but native package deps do not include `xdg-utils` for deb/rpm/pacman. On minimal installs, external links may stop opening.
- The wrapper intercepts all Linux string URLs and drops the `shell.openExternal(url, options)` options argument. Please either preserve/delegate options or only use the sanitized `xdg-open` path when it is safe to do so.
- Please add direct test coverage for the new behavior: env scrubbing, non-string delegation, options/fallback handling, and spawn error handling.

KaiStarkk:
I should have caught these before putting it up as draft, sorry didn't mean to waste time with an early check. Will convert to ready after fixing these though, thanks for reviewing so fast.

@ilysenko EDIT: So I've fixed the rookie errors like the API contract miss in a second commit.

I'd missed the smoke issue because they already failed on the shebang, and I ignored it, oopsie. It failed cause I'm on nix which has bash at /run/current-system/sw/bin/bash and no /bin/bash - I think this might be useful in the repo if you're open to it so I added a commit for 'improved portability of generated wrappers'. Lmk if you want the PR split.

About that native package deps issue:

I had seen that Nix already has xdg-utils which is why I went ahead with it initially, but didn't notice that DEB [control (line 7)](/tmp/codex-desktop-linux-pr/packaging/linux/control:7), Pacman [PKGBUILD.template (line 9)](/tmp/codex-desktop-linux-pr/packaging/linux/PKGBUILD.template:9), and RPM [spec (line 12)](/tmp/codex-desktop-linux-pr/packaging/linux/codex-desktop.spec:12) don't declare the provider -- my bad. Just not sure what the best way to tackle this is?

I've proposed one approach in the third commit: keeps it as native package metadata rather than anything clever at runtime. I also kept the behavior defensive: if sanitized xdg-open fails, it falls back to Electron so non-native/minimal setups aren't worse than before. Let me know if this doesn't look right though. Doesn't feel entirely clean to me but couldn't think of a better way.

scripts/patches/main-process/browser.js
```diff
@@ -318,9 +318,56 @@ function applyLinuxChromeExtensionStatusPatch(currentSource) {
...
+    `function codexLinuxOpenExternalWithFallback(__codexOriginalOpenExternal,__codexUrl){return codexLinuxLaunchExternalUrl(__codexUrl).catch(()=>__codexOriginalOpenExternal(__codexUrl))}` +
+    `function codexLinuxPatchExternalOpen(__codexElectron){if(process.platform!==\`linux\`||__codexElectron?.shell==null||typeof __codexElectron.shell.openExternal!==\`function\`)return __codexElectron;if(__codexElectron.shell.openExternal.__codexLinuxExternalOpenPatched)return __codexElectron;let __codexOriginalOpenExternal=__codexElectron.shell.openExternal.bind(__codexElectron.shell);async function __codexOpenExternal(__codexUrl,__codexOptions){if(typeof __codexUrl===\`string\`&&__codexOptions==null)return codexLinuxOpenExternalWithFallback(__codexOriginalOpenExternal,__codexUrl);return __codexOriginalOpenExternal(__codexUrl,__codexOptions)}__codexOpenExternal.__codexLinuxExternalOpenPatched=!0,__codexElectron.shell.openExternal=__codexOpenExternal;return __codexElectron}`
+  );
+}
+
+function applyLinuxExternalOpenEnvPatch(currentSource) {
```
avifenesh, line 332: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`applyLinuxExternalOpenEnvPatch` is registered with `ciPolicy: "required-upstream"`, but its no-match path silently returns `currentSource` without emitting a warning when the helper is already present, and — more importantly — when the Electron require initializer regex fails to match on a fresh bundle it only `console.warn`s and returns unchanged. In the engine (`applyMainBundlePatchDescriptors` -> `patchStatusFromDescriptorChange`), a required-upstream patch that produces no change AND emits a warning becomes `failed-required` (good), but if `patchedAnyElectronRequire` is false and `hasHelper` is false the warning IS emitted, so that case is covered.

The real gap: the regex `/([A-Za-z_$][\w$]*=)require\((`|'|")electron\2\)/g` only matches `IDENT=require("electron")` assignment initializers. Upstream commonly emits `require("electron")` in other shapes (e.g. destructured `let{shell:x}=require("electron")`, or `var e=t("electron")` after bundler interop where the call is wrapped). If the bundle shape drifts to any non-`ident=require(...)` form, `patchedAnyElectronRequire` stays false and — since `hasHelper` is also false — the WARN fires and it correctly becomes `failed-required`. So drift is visible. Please confirm this is the intended detection path; if any upstream chunk legitimately contains `codexLinuxPatchExternalOpen(` from a prior run but no matchable require (partial re-chunk), `hasHelper` short-circuits the warning and the patch is silently recorded as `already-applied`/`applied` with no change, hiding drift. Recommend keying the warning suppression on actual patched state rather than mere helper presence.

avifenesh (commented):

---

Reviewed the Linux external-open sanitization patch. One correctness concern worth confirming; the rest of the diff (packaging deps, shebang normalization, descriptor wiring, tests) looks consistent with existing conventions.

scripts/patches/main-process/browser.js
```diff
@@ -318,9 +318,59 @@ function applyLinuxChromeExtensionStatusPatch(currentSource) {
   return currentSource.slice(0, blockStart) + replacement + currentSource.slice(blockEnd);
 }
 
+function buildLinuxExternalOpenHelpers() {
+  return (
+    `function codexLinuxExternalOpenEnv(){let __codexEnv={...process.env};` +
+    `for(let __codexKey of[\`LD_LIBRARY_PATH\`,\`LD_PRELOAD\`,\`NODE_OPTIONS\`,\`NODE_PATH\`,\`NODE_REPL_EXTERNAL_MODULE\`,\`ELECTRON_RUN_AS_NODE\`,\`ELECTRON_NO_ASAR\`,\`ELECTRON_ENABLE_LOGGING\`,\`VSCODE_NODE_OPTIONS\`,\`VSCODE_NODE_REPL_EXTERNAL_MODULE\`,\`npm_config_node_options\`,\`NPM_CONFIG_NODE_OPTIONS\`,\`CHROME_DESKTOP\`,\`ELECTRON_RENDERER_URL\`,\`CODEX_ELECTRON_RESOURCES_PATH\`,\`CODEX_ELECTRON_USER_DATA_DIR\`,\`CODEX_LINUX_APP_ID\`,\`CODEX_LINUX_APP_DISPLAY_NAME\`,\`CODEX_LINUX_WEBVIEW_PORT\`])delete __codexEnv[__codexKey];` +
```
avifenesh, line 324: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`codexLinuxExternalOpenEnv()` strips `PATH`-affecting/`NODE_*`/`CODEX_LINUX_*` wrapper vars but does not include `LD_AUDIT` or the glibc `LD_*` tunables (`LD_BIND_NOW`, etc.). The stated goal is to prevent wrapper library-injection env from leaking into `xdg-open`; `LD_AUDIT` is the most common `LD_PRELOAD` sibling that alters dynamic-linker behavior of the spawned browser process. If the wrapper only ever sets `LD_LIBRARY_PATH`/`LD_PRELOAD` this is fine — please confirm no other `LD_*` variable is set by the launcher, otherwise add `LD_AUDIT` to the strip list.

avifenesh (commented):

---

Reviewed the Linux external-open sanitization patch. The prior drift-detection concern (commit `5cd06ea`) is addressed: `applyLinuxExternalOpenEnvPatch` now suppresses the WARN only when the helper AND an actually-patched `codexLinuxPatchExternalOpen(require("electron"))` are both present, so partial re-chunk drift still emits a warning and, under `required-upstream`, resolves to `failed-required`. Packaging dep additions (`xdg-utils`), wrapper shebang portability, and the smoke-test occurrence-count update all look consistent.

One correctness concern on the sanitizer's env-var scope is worth confirming before merge (inline).

scripts/patches/main-process/browser.js
KaiStarkk: These auto-reviews are getting good. Pushed `5cd06ea` to make the already-applied check require both the helper and the wrapped Electron require, with regression coverage for the clean and helper-only cases. CI is green.

scripts/patches/main-process/browser.js
KaiStarkk: @avifenesh / @ilysenko I've had a look, the launcher doesn't set LD_AUDIT or any other additional LD_* variables. The only LD_* env handled by the wrapper path is the existing LD_LIBRARY_PATH / LD_PRELOAD sanitizer scope, so no code change needed here. But it also couldn't hurt. Happy to add more scope to the sanitizer for completeness' sake if desired.

ilysenko:
Thanks for the follow-up fixes. One blocker remains: the new external-open descriptor is marked `required-upstream`.

That policy is reserved for patches whose absence makes the app fail to launch or core-unusable. External URL sanitization is important, but if this needle drifts in a future DMG, install/rebuild should warn and continue rather than fail the build by default. Please make this descriptor `optional`, unless there is a concrete core-unusable failure mode to document and justify.

ilysenko (approved):
Reviewed current head 1fb75de. The external-open patch now remains fail-soft as optional drift, local validation and GitHub checks are green.

KaiStarkk:
@ilysenko noted, my mistake. I see that this was merged so let me know if any change is still needed there otherwise I'll hold off.

PR #636, opened 2026-07-01
ci(nix): skip unchanged refresh verification

Yo-DDV:
## Summary
- skip package-output verification during the Nix hash refresh when the refresh leaves pin files unchanged
- keep full Nix verification when the refreshed version pins or Codex.dmg hash change
- cover unchanged and changed refresh paths in script smoke tests

## Validation
- bash -n scripts/ci/update-nix-hashes.sh
- git diff --check
- timeout 300 bash tests/scripts_smoke.sh

### DIFF
--- a/scripts/ci/update-nix-hashes.sh
+++ b/scripts/ci/update-nix-hashes.sh
@@ -18,6 +18,12 @@ PACKAGE_OUTPUTS=(
     ".#installer"
 )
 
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head. No blockers found; GitHub checks and local smoke validation passed.

PR #638, opened 2026-07-01
Fix Linux remote control availability patches

riccardo-algorime:
## Summary
- skip Linux remote-control status reads for `remote-control:` hosts, matching the existing remote SSH guard
- split the enable-for-host params patch into its own webview descriptor so it applies to current app-main bundle shapes
- update the status guard descriptor for the current pull-requests bundle shape

## Test


ilysenko (approved):
Reviewed current head 2408adc61fe8cd4bcf7e8ba2cacd917208d1004d. No blockers found; local feature tests passed and GitHub checks are green after rerunning the transient Pacman failure.

Issue #639, opened 2026-07-01
TESTNG: Track remote mobile hydration latency on unstable networks


Yo-DDV:
Hi,

I'm opening this to collect feedback around the remote mobile path on Linux.

I often leave the Linux desktop app running on my computer and connect back to an existing conversation from the mobile app. On long sessions, opening the thread could take a long time before the conversation and active tool updates became usable.

I tested a local build that deduplicates in-flight conversation hydration reads in the Linux remote mobile patch. In my case it is a clear improvement: opening an existing conversation from mobile is noticeably faster, and tool-call updates appear much more quickly.

Network context for this test:

- I tested from France over a French 4G/5G mobile network, not from a stable wired LAN.
- The desktop-side capture was taken on an interface with MTU 1420 towards the Cloudflare TLS endpoint used during the session.
- The local 60-second capture had 6,805 packets, 0 tcpdump/kernel drops, no TCP RST, no zero-window condition, and 16 duplicate data segments.
- `ss -tinp` showed retrans counters and a transient send queue/notsent state that later drained.

Local app-side observations after restart:

- `remoteControl/status/read`: up to 1 ms
- `thread/read`: up to 40 ms
- `thread/list`: up to 504 ms
- no new `refresh_remote_control_for_local_identity_failed` entries after restart

I am not attaching the raw pcap here because it contains local connection metadata, but these are the relevant capture notes.

I plan to send a small PR that keeps the current queueing behaviour but avoids launching several concurrent `readThread` hydrations for the same missing thread. It would be useful to hear whether other users see the same improvement, especially on long threads or unstable mobile networks.

Useful feedback would be:

- distribution and package format
- whether `remote-mobile-control` is enabled
- mobile network type, Wi-Fi, or wired baseline
- rough before/after time to open a long thread
- logs around `turn/started`, `thread/read`, `thread/list`, and remote-control status calls

Yo-DDV:
This issue needs your input to gather feedback from real-world experience. Thanks to everyone who takes part!

ilysenko:
thanks I merged it. If anything goes wrong, please let me know, and I’ll revert it.

joshyorko:
Just saw this,  will report back with some notes I have some

Yo-DDV:
Hi, thanks, that would be useful.

[...]

PR #640, opened 2026-07-01
fix(remote-mobile): dedupe in-flight hydration reads

Yo-DDV:
Hi,

This is the small patch I mentioned in #639.

On my machine, mobile remote was noticeably slow when I reopened an existing long conversation from the mobile app. The awkward part was the first hydration step: while the desktop app still did not know the conversation locally, repeated `turn/started` notifications could start repeated `readThread` calls for the same thread.

This patch keeps the existing queue, but adds one in-flight hydration guard per thread. New notifications for the same missing thread are still queued. They just do not start another `readThread` while the first one is already running.

I also renamed a couple of local generated key variables in the same file. Behaviour is unchanged there; this only keeps the public leak check quiet without changing the persisted record shape.

Local test context from #639:

- tested from France over an unstable 4G/5G mobile path
- desktop-side capture on an MTU 1420 interface towards the Cloudflare TLS endpoint used by the session
- 60-second capture: 6,805 packets, 0 tcpdump/kernel drops, no TCP RST, no zero-window condition, 16 duplicate data segments
- `ss -tinp` showed retrans counters and a transient send queue/notsent state that drained later
- after restart, local app calls stayed low in the logs: `remoteControl/status/read` up to 1 ms, `thread/read` up to 40 ms, `thread/list` up to 504 ms

This is one local measurement, not a broad benchmark. The user-visible result was still clear here: opening an existing conversation from mobile became much faster, and tool-call updates appeared sooner on the phone.

Tests run locally:

- `node --test linux-features/remote-mobile-control/test.js` 68/68 pass
- `node --test linux-features/*/test.js` 358/358 pass
- `node --test scripts/patch-linux-window-ui.test.js --test-name-pattern 'remote mobile|remote-control|hydration|thread'` 320/320 pass
- `make test` 186/186 pass
- `make check`
- `git diff --check`
- `/home/rootroot/.codex/scripts/github-public-leak-check.py --scope publish /tmp/mobile-remote-hydration-pr --assume-public-github`

Thanks.

Related to #639.


ilysenko (approved):
Reviewed locally against current main. The merge is clean with #638, and the relevant test suites pass.

PR #641, opened 2026-07-01
Remove Linux window menus

Yo-DDV:
Hi,

This follows up on #633 after it was closed. That PR pointed at the right Linux UX problem, and this keeps the work moving with a narrower implementation.

On Linux we were only hiding the BrowserWindow menu with `setMenuBarVisibility(false)`. Hidden menus can still leave accelerators active even though the menu is not visible, which is the behavior users can hit with accidental window closes. This changes the Linux patch to remove the window menu with `removeMenu()`, matching the existing Windows path.

I also updated the `frameless-titlebar` feature patch so it recognizes the new core Linux `removeMenu()` snippet and does not re-add the older `setMenuBarVisibility(false)` branch.

This does not carry over the global `before-input-event` Ctrl+W guard from #633. The menu path is narrower, and it avoids blocking renderer or editor shortcuts.

Validation run locally:

- `node --test linux-features/frameless-titlebar/test.js`
- `timeout 300 bash tests/scripts_smoke.sh`
- `make check && make test`

### DIFF
--- a/linux-features/frameless-titlebar/patch.js
+++ b/linux-features/frameless-titlebar/patch.js
@@ -78,35 +78,32 @@ function applyFramelessTitlebarOverlaySyncPatch(currentSource) {
 
 function applyFramelessTitlebarMenuPatch(currentSource) {
   const menuRegex = /process\.platform===`win32`&&([A-Za-z_$][\w$]*)\.removeMenu\(\),/g;
-  let patchedAny = false;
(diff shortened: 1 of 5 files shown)


ilysenko (approved):
Reviewed the current head f41514a. The change is narrow, tests/checks are green, and I did not find blockers.

PR #642, opened 2026-07-01
fix(updater): include required bundle assets

patelvyom:
- Linux auto-update was downloading and rebuilding the app successfully, then failing before install during package creation:
  `Missing icon: .../builder/assets/codex-linux.png`
- The installed update-builder had the file, but the updater’s per-update workspace copy whitelist omitted it.

**Change Summary**
- Fixed updater rebuild workspace copying so pacman auto-update packaging includes all required repo assets.
- Added `record-replay-linux` and `assets/codex-linux.png` to `REQUIRED_BUNDLE_FILES` in `updater/src/builder.rs`.
- Expanded updater builder tests to verify both files are copied into temporary update workspaces.

### DIFF
--- a/updater/src/builder.rs
+++ b/updater/src/builder.rs
@@ -14,11 +14,12 @@ use std::{
 use tokio::process::Command;
 use tracing::info;
 
-const REQUIRED_BUNDLE_FILES: [(&str, &str); 17] = [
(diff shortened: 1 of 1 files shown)


ilysenko (approved):
Reviewed current head; no blockers found and checks are green.

PR #643, opened 2026-07-02
Add Linux Chronicle and Record & Replay parity

joshyorko:
## Summary

This adds the Linux Chronicle/Skysight perception substrate needed for Record & Replay to compile real desktop workflows into usable skills without guessing.

The branch ports the missing Linux bridge around the upstream Chronicle/Sky product shape:

- adds Linux Skysight recent-activity capture with Chronicle-compatible segment metadata and artifacts
- records structured desktop/window/browser evidence, including focused browser title and URL context when supplied by browser evidence
- mirrors dictation/finalized speech into Record & Replay `speech_context` and transcript files
- integrates browser trace, desktop snapshots, native audio evidence, and privacy exclusions into recording bundles
- bridges Chronicle settings/tray controls to Linux Skysight start/stop/status handlers
- hardens GNOME extension screenshot capture, stale helper cleanup, daemon status semantics, and generated prompt drafting
- adds a compact upstream DMG probe helper for the Record & Replay/Sky/Chronicle contract surface

## Why

Record & Replay only works if the generated bundle tells the next agent what actually happened. On Linux, the previous bundle could miss transcript and browser/window context, leaving the LLM to infer user actions. This makes that evidence first-class and reusable by Chronicle/Skysight and Record & Replay.

## Validation

Validated on the clean PR branch/worktree:

```bash
git diff --check upstream/main...HEAD

docker run --rm \
  -v "$PWD":/workspaces/codex-desktop-linux \
  -w /workspaces/codex-desktop-linux \
  codex-desktop-linux-devcontainer:local \
  bash -lc 'node --test linux-features/record-and-replay/test.js scripts/patch-linux-window-ui.test.js && cargo fmt --check -p codex-record-replay-linux && cargo test -p codex-record-replay-linux && cargo fmt --check -p codex-computer-use-linux && cargo test -p codex-computer-use-linux && bash tests/scripts_smoke.sh'
```

Additional latest upstream DMG validation, with `CODEX_DMG_REFRESH_MODE=auto`:

- `./install.sh --inspect --report-dir dist-next/latest-dmg-inspect`
- `scripts/dev/probe-codex-dmg-record-replay.sh --dmg Codex.dmg --work-dir dist-next/latest-dmg-record-replay-probe --keep`
- feature-enabled lab build from the refreshed DMG with `record-and-replay`, `conversation-mode`, and `read-aloud` enabled

Observed latest-DMG evidence:

- refreshed DMG detected Electron `42.1.0`
- upstream probe passed Record & Replay, SkyComputerUseClient, Chronicle, event-stream, and Sky client string contracts
- lab build staged the Record & Replay plugin and `codex-record-replay-linux` backend
- patch report showed all Record & Replay feature descriptors applied


joshyorko:
after this one i feel like openai might know about us

ilysenko:
Thanks for pushing this forward. A few things stood out that seem worth thinking through before merge:

1. For Skysight exclusions, I’m a bit worried that `domain`/`urlDomain` still reads like a privacy boundary, but the screenshot/accessibility suppression path only has window title/app_id/wm_class to match against. That seems like it could still let excluded browser content land in screenshot artifacts unless the domain happens to appear in window metadata.

2. The native audio path also feels worth making more explicit. Since audio capture defaults on, this could start pw-record/parecord/ffmpeg/arecord and persist audio artifacts even though the Linux-specific permission/consent story is not very visible yet. It might be safer if audio were opt-in or gated on an explicit permission state.

3. The Skysight/audio PID handling looks like it may be relying on `/proc/<pid>` existence as proof that the process is still ours. If a stale status file meets PID reuse, status/stop could point at an unrelated same-user process. Some process identity/start-time/session-token check would probably make this lifecycle much safer.

One smaller cleanup: the browser trace bridge writes temp JSON files and appears to leave them behind after import. It may be worth deleting them in a finally path, or having the Rust side consume and remove them.

record-replay-linux/src/skysight.rs
```diff
@@ -0,0 +1,1988 @@
...
+}
+
+fn active_status_pid(paths: &SkysightPaths) -> Option<u32> {
+    read_status(paths)
+        .ok()
+        .and_then(|status| status.pid)
+        .filter(|pid| process_is_alive(*pid))
+}
+
+fn process_is_alive(pid: u32) -> bool {
```

---

PID-only liveness check is unsafe against PID reuse. `active_status_pid`/`process_is_alive` treat `/proc/<pid>` existence as proof the stored PID is still the Skysight daemon. If the daemon died and the OS reused that PID for an unrelated same-user process, `skysight_status` will report `running`, and `stop_skysight` (line 535-543) will send `SIGTERM` to that unrelated process via `request_process_stop`. The status file is persisted across restarts (`from_env` reuses `CODEX_SKYSIGHT_RUNTIME_DIR`), so a stale `status.json` surviving a crash + PID reuse is a realistic path. Recommend pinning process identity (e.g. compare `/proc/<pid>/stat` start-time or an argv/session token) before treating a PID as the live daemon and before signaling it. The same pattern exists in `audio.rs` `terminate_process`/`process_is_alive` (lines 350-391).

record-replay-linux/src/audio.rs
```diff
@@ -0,0 +1,428 @@
...
+                    "-t".to_string(),
+                    "wav".to_string(),
+                    output.to_string(),
+                ]
+            },
+        },
+    ]
+}
+
+fn audio_capture_enabled() -> bool {
```
avifenesh, line 271: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

Native audio capture defaults ON (`audio_capture_enabled()` returns true when `CODEX_RECORD_REPLAY_AUDIO` is unset), and `start_session` passes `include_audio: !args.no_audio` / `params.include_audio.unwrap_or(true)`. That means starting any recording spawns pw-record/parecord/ffmpeg/arecord and persists a `recording.wav` artifact with no explicit Linux consent/permission gate visible in this path — unlike screenshots/accessibility which have suppression logic. Microphone capture is materially more sensitive than the screen/window evidence around it. Consider making audio opt-in (default false) or gating on an explicit permission/consent state before spawning a recorder.

avifenesh (commented):

---

Reviewed the Linux Chronicle / Record & Replay parity work. The ASAR patch changes match-index-based safeguards, secure_fs usage, and the GNOME extension screenshot-path allowlist all look sound. Two evidence-backed concerns below; both were also flagged by the maintainer's pre-merge comment, so I'm anchoring them to the specific source lines.

record-replay-linux/src/skysight.rs
joshyorko: Addressed in the current commit. Skysight now records and verifies Linux process identity via `/proc/<pid>/stat` start time before treating a stored PID as live or before signaling it, and the audio recorder path uses the same identity check before termination. A reused PID no longer satisfies status/stop liveness.

record-replay-linux/src/audio.rs
joshyorko: Addressed in the current commit. Native audio is now opt-in at every entrypoint: the CLI only records audio with `--audio` and `--no-audio` wins, MCP defaults `include_audio` to false, and `start_audio_capture` still requires an affirmative `CODEX_RECORD_REPLAY_AUDIO` value before spawning any recorder. Without both explicit request and env permission, no `pw-record`/`parecord`/`ffmpeg`/`arecord` process is started.

joshyorko:
Follow-up pushed and CI is rerunning. Mapping the maintainer/review points to the current branch:

- `domain`/`urlDomain` exclusions now suppress browser observations and skip screenshot/accessibility evidence when excluded browser content is visible, or when browser URL/domain cannot be verified while domain exclusions are active.
- Native audio is opt-in only: CLI requires `--audio`, MCP defaults `include_audio` to false, and recorder spawn still requires affirmative `CODEX_RECORD_REPLAY_AUDIO`.
- Skysight and audio liveness now verify process identity/start time before treating a stored PID as live or before signaling it, so stale PID reuse fails closed.
- Browser trace temp JSON cleanup now runs through the bridge cleanup path after import.

The PR branch remains one commit; the latest CI run has App-against-DMG, Rust/Smoke, Debian, RPM, and Pacman green, with Nix still running.

record-replay-linux/src/skysight.rs
```diff
@@ -0,0 +1,2250 @@
...
+        is_running: true,
+        paused: false,
+        pause_reason: None,
+        pid: Some(pid),
+        started_at: Some(now_timestamp()),
+        end_reason: None,
+        message: Some("Skysight daemon started".to_string()),
+    })?;
+    write_status(paths, &status)?;
+    Ok(status)
```
avifenesh, line 293: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`start_skysight` spawns the daemon and immediately writes `state: "running"` with `pid: Some(pid)` and `process_start_time_ticks` computed from that pid via `status_value` -> `process_start_time_ticks(pid)`. The pid here is the reaped child returned by `spawn_reaped`, which is the process that runs `skysight daemon`. That is correct, but note the daemon loop itself (`run_skysight_daemon`, line 296) never rewrites the status with its own `std::process::id()` at startup — it only does so on pause (line 324). On the normal running path the daemon never re-asserts identity, so the parent-recorded `pid`/`process_start_time_ticks` is the source of truth. Confirm the spawned child's pid is stable (it is, since `spawn_reaped` hands the same `Child` to the reaper and returns `child.id()`), otherwise `active_status_pid`/`stop_skysight` liveness checks would race against a still-initializing child. This looks correct as written; flagging only to confirm the start-time snapshot is taken after the child is live enough that `/proc/<pid>/stat` exists — if the child exits before `status_value` reads its start time, `process_start_time_ticks` returns `None` and the status is persisted as running-with-no-start-time, which would make every future `process_is_alive` check return false (`process_matches_start_time` returns false when `expected` is `None`), silently marking a live daemon as stopped.

avifenesh (commented):

---

Reviewed the Linux Chronicle / Record & Replay parity work. The ASAR patch changes are match-index/regex-based with idempotence guards and unit coverage, secure_fs is used for private files, process identity is verified via `/proc/<pid>/stat` start time (addressing the earlier PID-reuse concern), and audio capture is now opt-in at every entrypoint. Packaging changes for `CODEX_DMG_REFRESH_MODE=pinned` are guarded and don't touch the `ls | sort | tail` anti-pattern. One correctness concern on the daemon start path is noted inline.

ilysenko (approved):
Reviewed current head; the earlier privacy and process-lifecycle blockers look addressed, and CI is green.

record-replay-linux/src/skysight.rs
joshyorko: Confirmed. `spawn_reaped` returns the `child.id()` from the same spawned child handed to the reaper, so the parent-recorded PID is the daemon process PID. On Linux, `/proc/<pid>/stat` is available for a live child immediately after `Command::spawn`; if the child has already exited before `status_value` reads it, the missing start time makes later liveness fail closed rather than matching an unrelated process. That can at worst make status report stopped for a daemon that failed during startup, not kill a reused PID.

PR #645, opened 2026-07-02
Implement Linux Chronicle sidecar parity

joshyorko:
## Summary
- Add a Chronicle-compatible Linux Skysight scheduler/status contract, including running/paused state, last and next capture timestamps, rollup timing, PID path, raw screen buffer paths, and Chronicle memory resource locations.
- Write Chronicle-shaped PID, raw screen, segment, and memory-resource artifacts so the Linux sidecar has inspectable state instead of a fire-and-forget snapshot path.
- Wire Chronicle Settings/permissions to start or resume Linux Skysight with the summary agent enabled when the user turns Chronicle on.
- Add summary-agent policy precedence: `CODEX_SKYSIGHT_SUMMARY_AGENT` override first, runtime setting second, `$CODEX_HOME/config.toml` `[features].chronicle = true` third, default disabled.
- Keep `CODEX_SKYSIGHT_SUMMARY_AGENT=disabled` as the emergency kill switch even when runtime settings or config enable Chronicle.
- Keep capture on the normal 60-second cadence while guarding model calls: 10-minute summaries run at most once per 10 minutes, 6-hour rollups follow the rollup cadence, and non-due captures still write local fallback resources.

## Runtime verification
- Rebuilt and launched the packaged Linux app locally for the base Skysight/Chronicle sidecar.
- `/run/user/1000/skysight/status.json` reported `is_running: true`, `is_paused: false`, `last_capture_at`, `next_capture_at`, and `next_6h_rollup_at`.
- `/tmp/codex_chronicle/chronicle-started.pid` pointed at a live installed Skysight daemon process.
- `/tmp/chronicle/screen_recording/` contained rolling `.jpg`, `.capture`, `.capture.json`, and `.ocr.jsonl` artifacts.
- `$CODEX_HOME/memories/extensions/chronicle/resources/` contained generated 10-minute and 6-hour Linux activity resources.
- The new summary-agent settings bridge still needs one post-rebuild runtime acceptance pass to confirm Settings shows enabled status and `next_summary_agent_run_at`.

## Validation
- `docker run --rm -v "$PWD":/workspaces/codex-desktop-linux -w /workspaces/codex-desktop-linux codex-desktop-linux-devcontainer:local bash -lc 'node --test linux-features/record-and-replay/test.js && cargo test -p codex-record-replay-linux'`
- `docker run --rm -v "$PWD":/workspaces/codex-desktop-linux -w /workspaces/codex-desktop-linux codex-desktop-linux-devcontainer:local bash -lc 'cargo fmt --check -p codex-record-replay-linux'`
- `git diff --check`

## Known limits
- Linux OCR is currently represented as unavailable in `.ocr.jsonl`; this PR does not implement a native OCR engine.
- Summary-agent model work is user/config enabled and rate guarded; it is not run on every capture.


ilysenko:
- The default Chronicle memory path migration drops existing `exclusions.json`, so users can lose exclusion rules right before this PR starts persisting raw screen buffer frames. Please migrate or preserve existing exclusions when moving from `memories_extensions/chronicle` to `memories/extensions/chronicle`.
- `capture_skysight_snapshot` always sets `next_capture_at`, even for stopped one-shot snapshots, so status can report a future scheduled capture while the daemon is not running. Please keep it unset unless Skysight is actually running.

joshyorko:
Fixed in the squashed PR commit `770d8d7`.

- Preserves legacy exclusions by migrating `memories_extensions/chronicle/exclusions.json` into the new `memories/extensions/chronicle/exclusions.json` path before reads/writes, without deleting the legacy file.
- Keeps `next_capture_at` unset for stopped one-shot snapshots; only a running daemon reports the next scheduled capture.

Validation rerun in Docker before the squash: `node --test linux-features/record-and-replay/test.js && cargo test -p codex-record-replay-linux && cargo fmt --check -p codex-record-replay-linux`.

record-replay-linux/src/skysight.rs
```diff
@@ -1508,16 +1824,71 @@ fn ensure_parent_dirs(paths: &SkysightPaths) -> Result<()> {
...
+    };
+    if legacy_path == paths.exclusions_path || !legacy_path.exists() {
+        return Ok(());
+    }
+    let raw = fs::read_to_string(&legacy_path)
+        .with_context(|| format!("failed to read legacy exclusions {}", legacy_path.display()))?;
+    crate::secure_fs::write_private_file(&paths.exclusions_path, raw).with_context(|| {
+        format!(
+            "failed to migrate exclusions to {}",
+            paths.exclusions_path.display()
```

---

`migrate_legacy_exclusions_if_needed` only fires for the *default* layout (`.../memories/extensions/chronicle/exclusions.json`), because `legacy_default_exclusions_path` derives the legacy path by walking exactly `chronicle/extensions/memories/<code_home>` and rebuilding `<code_home>/memories_extensions/chronicle/exclusions.json`.

But `exclusions_path` is overridable via `CODEX_SKYSIGHT_EXCLUSIONS_PATH` and via `CODEX_SKYSIGHT_MEMORY_EXTENSION_DIR` / `CODEX_CHRONICLE_MEMORY_EXTENSION_DIR` (see `from_env`, lines 289-306). When any of those env vars is set — which is exactly the case for the spawned daemon (`start_skysight` passes `CODEX_SKYSIGHT_MEMORY_EXTENSION_DIR` and `CODEX_SKYSIGHT_EXCLUSIONS_PATH` to the child, lines 349-353) — the parent-chain check (`file_name != "chronicle"`/`"extensions"`/`"memories"`) fails and `legacy_default_exclusions_path` returns `None`, so migration is silently skipped. In that configuration the new empty `exclusions.json` is created and the legacy rules under `memories_extensions/chronicle/` are never migrated.

Since the earlier blocker was specifically about not dropping legacy `exclusions.json`, please confirm the migration also covers the custom-path/daemon configuration (or that the daemon inherits the already-migrated file from the parent's `ensure_layout` before writing).

avifenesh (commented):

---

Review of the Linux Chronicle sidecar parity changes. One correctness concern worth flagging around the exclusions migration invariant that a maintainer raised earlier; the rest of the diff looks internally consistent.

ilysenko:
- `chronicle-permissions` now starts/resumes Skysight with the summary agent enabled, but the current Settings UI polls this endpoint as a permissions/status query. Opening Settings can therefore start background screen capture before the user explicitly enables Chronicle. Please keep `chronicle-permissions` side-effect-free and move start/resume behind the explicit enable/toggle flow.

joshyorko:
Fixed in the squashed PR commit `af060ff`.

- `chronicle-permissions` is now side-effect-free: it only reads `skysight status` through `codexLinuxChronicleSidecarControlStateAsync()`. Start/resume with `--summary-agent enabled` remains behind the explicit Chronicle toggle/start path.
- Legacy exclusion migration now also checks the path derived from `memory_extension_dir`, so daemon/custom `CODEX_SKYSIGHT_EXCLUSIONS_PATH` destinations still migrate the old `memories_extensions/chronicle/exclusions.json` before creating a new exclusions file.

Regression coverage added for both cases. Docker validation passed: `node --test linux-features/record-and-replay/test.js && cargo test -p codex-record-replay-linux && cargo fmt --check -p codex-record-replay-linux`.

record-replay-linux/src/skysight.rs
joshyorko: Fixed in af060ff. Migration now considers both the destination exclusions path and the path derived from memory_extension_dir, so the daemon/custom exclusions path still picks up the legacy memories_extensions/chronicle/exclusions.json before creating a new file. Added skysight_migrates_legacy_exclusions_with_daemon_path_overrides to cover this exact shape.

ilysenko (approved):
Reviewed current head af060ff; previous blockers are addressed and all required checks are passing.

record-replay-linux/src/skysight.rs
```diff
@@ -1676,13 +2175,383 @@ fn write_6h_rollup(paths: &SkysightPaths) -> Result<PathBuf> {
...
+}
+
+fn codex_config_path() -> Option<PathBuf> {
+    let code_home = env::var_os("CODEX_HOME")
+        .map(PathBuf::from)
+        .or_else(|| env::var_os("HOME").map(|home| PathBuf::from(home).join(".codex")))?;
+    Some(code_home.join("config.toml"))
+}
+
+fn summary_agent_skip_until(paths: &SkysightPaths, level: &str) -> Option<String> {
```
avifenesh, line 2274: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

The 10-minute rate limit is keyed on `summary_agent_last_run_at`, which `write_resource_with_summary_agent` sets to `ran_at` on **every** run outcome, including `failed` (see the `Err(error)` and empty-output branches at lines 2350-2375 that return `ran_at: Some(ran_at)`). Combined with `summary_agent_skip_until` gating on that same field, a single failed summary-agent invocation now suppresses all further 10-minute summary attempts for a full `TEN_MINUTE_WINDOW_SECONDS`.

That matches the test `summary_agent_10min_work_is_rate_limited`, which asserts the second call is `scheduled` (skipped) purely because the first call *failed* against a missing CLI. If the intent is "summaries run at most once per 10 minutes" that is fine, but a transient failure (CLI momentarily unavailable, spawn error) will not be retried until the window elapses — the daemon will keep writing fallback resources for up to 10 minutes even though the failure was immediate. Consider only advancing the rate-limit clock on a *successful* run (or tracking last-success separately from last-attempt) so transient failures retry on the next capture rather than being throttled as if a real summary had been produced.

avifenesh (commented):

---

Review of the Linux Chronicle sidecar parity changes. One correctness concern in the summary-agent rate-limit path is worth confirming; the rest of the diff (env-var precedence, path migration fixes, secure_fs idempotency, screenshot payload handling) looks sound.

PR #647, opened 2026-07-02
Use update-builder manifest for rebuild workspace

### LINKED ISSUE #646: Auto-update rebuild fails on Arch: updater workspace misses record-replay-linux and assets/codex-linux.png
## Summary

On Arch Linux, the background `codex-update-manager` detects a new upstream DMG and starts a local pacman rebuild, but the rebuild fails.

At first I suspected a `kdialog` / privilege escalation issue, but the failure actually happens earlier, during the local rebuild step, before `pkexec` is involved.

## Environment

- Distro: Arch Linux
- Installed package: `codex-desktop 2026.06.30.071530-1`
- Updater service: `codex-update-manager.service` (user service)
- `kdialog` is installed and working

## Actual behavior

`~/.local/state/codex-update-manager/state.json` ends up with:

- `status: "failed"`
- `error_message: ".../builder/scripts/build-pacman.sh failed during local rebuild"`

The failed workspace was:

- `~/.cache/codex-update-manager/workspaces/2026.07.02.184148+55540dbb`

Relevant log excerpts:

### `logs/install.log`

```text
error: failed to load manifest for workspace member `.../builder/record-replay-linux`
referenced by workspace at `.../builder/Cargo.toml`

Caused by:
  failed to read `.../builder/record-replay-linux/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```

### `logs/build-package.log`

```text
[ERROR] Missing icon: .../builder/assets/codex-linux.png
```

## Expected behavior

The updater should be able to rebuild the Arch package from the downloaded DMG and either:

- mark the update as `ReadyToInstall`, or
- proceed to the `pkexec` installation step.

## Root cause

It looks like the temporary updater workspace is missing files that are required by the rebuild.

In `updater/src/builder.rs`, `REQUIRED_BUNDLE_FILES` does not include:

- `record-replay-linux`
- `assets/codex-linux.png`

But those are required by the packaged update-builder flow.

For example:

- `Cargo.toml` already includes `record-replay-linux` in `[workspace].members`
- `scripts/build-pacman.sh` expects `assets/codex-linux.png`
- `scripts/lib/package-common.sh` already copies both of these into the packaged `update-builder`

So the installed bundle under `/opt/codex-desktop/update-builder` is complete, but the temporary rebuild workspace created by the updater is incomplete.

## Proof / workaround

I manually patched the failed workspace by copying:

- `/opt/codex-desktop/update-builder/record-replay-linux`
- `/opt/codex-desktop/update-builder/assets/codex-linux.png`

into:

- `~/.cache/codex-update-manager/workspaces/2026.07.02.184148+55540dbb/builder/`

After that, rerunning `build-pacman.sh` succeeded and produced:

- `codex-desktop-2026.07.02.185650-1-x86_64.pkg.tar.zst`

So this does not appear to be a privilege escalation / `kdialog` problem. The rebuild can succeed once those missing files are present.

## Suggested fix

Update `updater/src/builder.rs` so the temporary builder bundle also copies:

- `record-replay-linux`
- `assets/codex-linux.png`

It would also be good to add/update tests to ensure the temporary builder bundle stays in sync with the packaged update-builder contents.

joshyorko:
## Summary
- add a packaged update-builder manifest so rebuild workspaces copy the staged bundle from one generated source of truth
- keep the existing hardcoded copy list as the fallback for packages built before the manifest existed
- validate the manifest path safety and smoke-test that optional feature assets/resources stay in the rebuild bundle

Fixes #646.

## Validation
- Docker devcontainer: `cargo fmt --check -p codex-update-manager && cargo test -p codex-update-manager builder::tests`
- Docker devcontainer: `bash tests/scripts_smoke.sh`
- Host: `git diff --check`

### DIFF
--- a/tests/scripts_smoke.sh
+++ b/tests/scripts_smoke.sh
@@ -472,6 +472,7 @@ test_update_builder_preserves_enabled_linux_features_config() {
     local staged_config="$root/opt/codex-desktop/update-builder/linux-features/features.json"
     local staged_local_manifest="$root/opt/codex-desktop/update-builder/linux-features/local/local-tool/feature.json"
     local source_info="$root/opt/codex-desktop/update-builder/.codex-linux/source-info.json"
+    local update_builder_manifest="$root/opt/codex-desktop/update-builder/.codex-linux/update-builder-manifest.txt"
 
     mkdir -p "$workspace"
     make_fake_app "$app_dir"
@@ -523,12 +524,16 @@ JSON
 
     assert_file_exists "$staged_config"
     assert_file_exists "$staged_local_manifest"
+    assert_file_exists "$update_builder_manifest"
     assert_contains "$staged_config" "example-feature"
     assert_contains "$staged_config" "local-tool"
     assert_contains "$staged_config" "tweaks"
     assert_contains "$staged_config" "mode"
     assert_not_contains "$staged_config" "localComment"
     assert_not_contains "$staged_config" "disabled-feature"
+    assert_contains "$update_builder_manifest" "record-replay-linux/Cargo.toml"
+    assert_contains "$update_builder_manifest" "assets/codex-linux.png"
+    assert_not_contains "$update_builder_manifest" "^node-runtime/"
 
     node - "$staged_config" <<'NODE' || fail "Expected staged Linux features config to be sanitized"
 const fs = require("node:fs");
--- a/scripts/lib/package-common.sh
+++ b/scripts/lib/package-common.sh
@@ -646,6 +646,18 @@ fs.writeFileSync(infoFile, `${JSON.stringify(info, null, 2)}\n`, "utf8");
 NODE
 }
 
+write_update_builder_manifest() {
+    local update_builder_root="$1"
+    local manifest="$update_builder_root/.codex-linux/update-builder-manifest.txt"
+    (
+        cd "$update_builder_root"
+        find . -mindepth 1 -type f \
+            ! -path './node-runtime/*' \
+            ! -path './.codex-linux/update-builder-manifest.txt' \
+            -printf '%P\n' | LC_ALL=C sort > "$manifest"
+    )
+}
+
 stage_common_package_files() {
     local root="$1"
     local app_root="$root/opt/$PACKAGE_NAME"
@@ -787,6 +799,7 @@ stage_update_builder_bundle() {
     cp "$REPO_DIR/assets/codex.png" "$update_builder_root/assets/codex.png"
     cp "$REPO_DIR/assets/codex-linux.png" "$update_builder_root/assets/codex-linux.png"
     stage_update_builder_source_info "$update_builder_root"
+    write_update_builder_manifest "$update_builder_root"
     if [ -d "$node_runtime_source" ]; then
         cp -a "$node_runtime_source" "$update_builder_root/node-runtime"
     else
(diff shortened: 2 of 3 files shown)


tests/scripts_smoke.sh
```diff
@@ -523,12 +524,16 @@ JSON
...
+    assert_file_exists "$update_builder_manifest"
     assert_contains "$staged_config" "example-feature"
     assert_contains "$staged_config" "local-tool"
     assert_contains "$staged_config" "tweaks"
     assert_contains "$staged_config" "mode"
     assert_not_contains "$staged_config" "localComment"
     assert_not_contains "$staged_config" "disabled-feature"
+    assert_contains "$update_builder_manifest" "record-replay-linux/Cargo.toml"
+    assert_contains "$update_builder_manifest" "assets/codex-linux.png"
+    assert_not_contains "$update_builder_manifest" "^node-runtime/"
```
avifenesh, line 536: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`assert_not_contains` calls `grep -q -- "$pattern"`, so the pattern `^node-runtime/` is interpreted as a basic regex anchored at line start. But `write_update_builder_manifest` emits paths via `find ... -printf '%P\n'` with no leading `./`, so a node-runtime entry (if the exclusion regressed) would appear as `node-runtime/...` — which this anchored pattern would correctly catch. However, note the assertion here is somewhat weak: the fixture's node-runtime lives under `resources/node-runtime` in the *app*, not under the `update_builder_root`, so no `node-runtime/` entry could ever be produced in this manifest regardless of the exclusion logic. This test therefore does not actually exercise the `! -path './node-runtime/*'` filter — consider staging a `node-runtime/` dir directly under the update-builder root to make the exclusion assertion meaningful.

avifenesh (commented):

---

## Review: update-builder manifest for rebuild workspace

One correctness concern with the manifest-based copy path, plus a test-assertion note. Overall the approach (generate a manifest at packaging time, use it as source of truth with a hardcoded fallback) is sound and the path-safety validation is a good addition.

### Manifest copy omits `node-runtime`, which the rebuild requires

`write_update_builder_manifest` deliberately excludes `node-runtime/*` from the manifest (`scripts/lib/package-common.sh:655`). On the Rust side, `copy_builder_bundle_from_manifest` copies only the entries listed in the manifest and then returns — so the `node-runtime` directory is never copied into the workspace `bundle_dir`.

In the legacy fallback path this was fine because `node-runtime` was never in `REQUIRED_BUNDLE_FILES` either, but the managed node source is resolved separately (`builder.rs:109-113`) via `config.builder_bundle_root.join("node-runtime")` and passed to `install.sh` as `CODEX_MANAGED_NODE_SOURCE`. Since `build_command_path` and `managed_node_source` read from `builder_bundle_root` (the packaged install root), not `bundle_dir` (the copied workspace), this is consistent with the old behavior and node-runtime is intentionally sourced from outside the workspace. Confirming this is the intent — the manifest exclusion mirrors the fallback and node-runtime is provided out-of-band — so no functional regression. Flagging for the author to double-check that assumption holds on the deb/rpm/pacman rebuild paths.

See inline note on the smoke-test assertion.

ilysenko (approved):
Reviewed the current head; no blockers found. Local focused builder tests and smoke tests passed.

PR #650, opened 2026-07-03
Sync computer-use-linux to v0.3.0

avifenesh:
Ports the standalone [computer-use-linux v0.3.0](https://github.com/agent-sh/computer-use-linux/releases/tag/v0.3.0) release into the vendored crate, keeping all codex-specific glue intact (identity, `CODEX_*` env aliases, chrome-extension host, `CaptureScreenshot` extension method). Closes the standing computer-use-sync reminder for the 0.2.10 → 0.3.0 gap.

## What's in v0.3.0

- **AT-SPI connect survives AppArmor-confined a11y peers.** The atspi crate's default `p2p` feature enumerates every a11y-bus peer during connect; a single confined peer (e.g. snap-packaged Slack) denying `GetInterfaces` aborted the whole `AccessibilityConnection`, which surfaced as "failed to connect to AT-SPI bus" for every app. The crate now depends on `atspi-connection` directly with `p2p` off (P2P routing was already deliberately unused).
- **Focused-element feedback after targeted keyboard input.** `press_key`/`type_text` results report the AT-SPI focused element (role, name, editable) and warn when no editable element holds focus — previously keystrokes could silently land nowhere while the result reported verified window focus.
- **Off-screen detection.** Screenshot, click, and input results warn when the target window or coordinate is partially/fully off-screen. Window bounds are checked against the extension's logical monitor layout, click coordinates against physical capture space — the two spaces are never mixed, so fractional scaling cannot produce false positives.
- **New `move_window` / `resize_window` tools** (GNOME Shell extension v2, which also adds `GetMonitorLayout`). Unmaximize handles both mutter API generations (pre-49 `get_maximized` + flags, 49+ `is_maximized` + flagless). Old installed extensions degrade with an explicit re-run-setup hint.
- **Compact `get_app_state`.** Readiness block by default; `verbose=true` opts into the full diagnostics dump.
- **`scroll` window targeting + relative coordinates** (parity with `click`); `press_key` documents its full key grammar and never silently drops unknown keys.

## Validation

- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --all-targets` — all green (130 vendored-crate tests incl. codex-specific ones).
- Release binary builds: `codex-computer-use-linux`, `codex-chrome-extension-host`, `codex-computer-use-cosmic`.
- The standalone release this ports was itself live-verified against a real GNOME 50.1 Wayland desktop and pre-release-reviewed (the review caught and fixed a GNOME 49/50 `get_maximized` removal and a HiDPI logical/physical coordinate-space bug before publish).

Crate version: `0.2.10-linux-alpha1` → `0.3.0-linux-alpha1` (parity signal with the standalone repo).


ilysenko:
Thinking out loud on the new scroll targeting path: I think there is one case we should tighten before merging.

`scroll(window_id=...)` currently focuses the requested window, but if the caller does not also pass `x/y` or `element_index`, the scroll path keeps `target_point = None`. Both the portal path and the ydotool path then send the wheel/axis event without moving the pointer first, so the call can report success while scrolling whatever is currently under the pointer rather than the requested window.

Could we either require a target point for window-targeted scroll, or default to a safe point inside the resolved target window? A regression test for `scroll(window_id=...)` without coordinates would be great too. This probably should also cover the off-screen coordinate warning for scroll, since that helper says click/scroll but the warning is only appended for click right now.

avifenesh:
good catch - that path was real: focus never moves the pointer, so a window-targeted scroll without a point wheeled whatever sat under the cursor and still reported success.

went with the safe-default option: when a window target is given and x/y/element_index are all absent, the point now defaults to the centre of the resolved target window (apply_window_center_scroll_point). if bounds can't be resolved it errors with a pass-x/y hint instead of silently scrolling. also wired the off-screen coordinate warning into scroll's success paths - you're right that the helper text said click/scroll while only click appended it.

fixed upstream first as agent-sh/computer-use-linux v0.3.1 (https://github.com/agent-sh/computer-use-linux/releases/tag/v0.3.1), then ported here in b21a19a - vendored crate bumped to 0.3.1-linux-alpha1 to keep the parity signal. regression tests cover the centre default and the missing-bounds error.

computer-use-linux/src/server.rs
```diff
@@ -761,17 +808,90 @@ impl ComputerUseLinux {
...
+                if let Err(message) = apply_window_relative_scroll_coordinates(&mut params, focus) {
+                    return Json(ActionOutput {
+                        ok: false,
+                        implemented: true,
+                        action: "scroll".to_string(),
+                        message,
+                        received,
+                    });
+                }
+            } else if params.x.is_none() && params.y.is_none() && params.element_index.is_none() {
```
avifenesh, line 869: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

Minor asymmetry with the relative-scroll and relative-click validation paths: `apply_window_relative_scroll_coordinates` rejects coordinates outside the window bounds, but the window-center default branch here is only entered when `x`, `y`, AND `element_index` are all `None`. If the caller supplies a window target plus a single coordinate (e.g. only `x`) without `relative=true`, neither branch fires, and `resolve_optional_target_point` returns `None` for the partial `x.zip(y)` — so the scroll silently falls back to "whatever is under the pointer," the exact failure this PR set out to fix. Consider treating a partial coordinate under a window target as an error (like relative mode does) rather than silently discarding it.

avifenesh (commented):

---

Reviewed the computer-use-linux v0.3.0 sync. The changes are well-considered: coordinate-space separation (logical window bounds vs physical capture coords) is consistently honored, the `atspi`/`atspi-connection` p2p-off split matches the documented rationale, version strings are consistent between Cargo.toml and the MCP `version` literal (both `0.3.1-linux-alpha1`), the GNOME extension bumps `metadata.json` version to 2 with an outdated-method hint, scroll now enforces window-target parity with click, and the new `desktop_size` field is covered by `#[derive(Default)]`. One narrow edge-case worth confirming below; not a blocker.

ilysenko:
Thanks, the center-default path fixes the main no-coordinate window-targeted scroll case. I agree the remaining partial-coordinate case (`x` without `y`, or `y` without `x`) is technically the same class of silent fallback, but it looks much closer to a malformed tool call than a normal expected usage path.

I would still tighten it in a follow-up by rejecting partial coordinate pairs, ideally with a small regression test, but I do not think that edge case needs to block this PR now that the normal `scroll(window_id=...)` path is safe.

ilysenko (approved):
Reviewed current head b21a19a. The original scroll-targeting blocker is fixed, CI is green, and the remaining partial-coordinate edge case is acceptable as a follow-up.

Issue #649, opened 2026-07-03
alway_on_top lead to computer stack


lfxx:
When I used the “Always on Top” option from the right-click menu to try to keep codex-desktop on top, my entire computer froze the moment I clicked the button.

I could see that the task inside Codex was still running, but I could no longer interact with it. The keyboard stopped responding, mouse clicks did not work, and no key combinations had any effect. The system only recovered after I rebooted the machine.

<img width="1059" height="623" alt="Image" src="https://github.com/user-attachments/assets/f2832b17-9feb-4581-92fa-a14fe89b5d6b" />

ilysenko:
Thanks for the report. Could you share a bit more detail so we can reproduce this safely?

- Which distro/version, desktop environment or window manager, and Wayland vs X11 session are you using?
- Which install/package method and Codex Desktop Linux app/upstream version is this?
- Where exactly did you click Always on Top: the window titlebar/menu, taskbar/dock menu, tray menu, or inside the app?
- Does the freeze happen every time, and does it happen on a fresh launch with no active task?
- After reboot, are there relevant entries in ~/.cache/codex-desktop/launcher.log or in journalctl around the freeze time, especially compositor/GPU/Electron errors?

lfxx:
> Thanks for the report. Could you share a bit more detail so we can reproduce this safely?
> 
> * Which distro/version, desktop environment or window manager, and Wayland vs X11 session are you using?
> * Which install/package method and Codex Desktop Linux app/upstream version is this?
> * Where exactly did you click Always on Top: the window titlebar/menu, taskbar/dock menu, tray menu, or inside the app?
> * Does the freeze happen every time, and does it happen on a fresh launch with no active task?
> * After reboot, are there relevant entries in ~/.cache/codex-desktop/launcher.log or in journalctl around the freeze time, especially compositor/GPU/Electron errors?

1.ubutun22.04，x11
2.make install-native,26.623.101652
3.the window titlebar/menu

<img width="2136" height="290" alt="Image" src="https://github.com/user-attachments/assets/8775e251-699b-4643-b366-cb28ec9cf8a3" />
4.i try twice.They all forzen.

5.Maybe have.if you need,i can email those files to you for debug purpose

Yo-DDV:
Hi, thanks again for the details.

[...]

PR #652, opened 2026-07-03
Remove Linux application menu from patched windows

Yo-DDV:
Hi,

This follows #649. I could not reproduce the full machine freeze locally, but the report points at the window titlebar/menu path on Ubuntu 22.04/X11.

The exact 26.623.101652 bundle still installed an Electron application menu after the Linux window `removeMenu()` patch. On Linux, that menu can still become the top window menu shown in the report.

This patch suppresses that application menu only on Linux with `Menu.setApplicationMenu(null)`, while keeping the existing menu behavior on macOS and Windows.

Tested with:

- `node --test linux-features/frameless-titlebar/test.js`
- `make check`
- `make test`
- `./install.sh --inspect` against the reported 26.623.101652 DMG
- X11/GNOME nested smoke check with `_NET_WM_STATE_ABOVE`


ilysenko (approved):

PR #654, opened 2026-07-03
fix: harden AppShots FIFO setup

Yo-DDV:
Hi,

This tightens the AppShots bare modifier helper by avoiding `mktemp -u` for the event FIFO path. The helper now creates a private temp directory with `mktemp -d`, places the FIFO inside it, and removes the directory during cleanup.

That avoids the small race where a path is printed but not reserved before `mkfifo` runs. The AppShots test now keeps this tempdir/FIFO contract covered.

Checks run locally:

- `node --check` on 110 JS files
- `node --test` on 24 Node test files, 717 tests passed
- `make check`
- `make test`
- `cargo test --workspace --all-targets` with isolated runtime dirs
- `cargo clippy --workspace --all-targets -- -D warnings` with isolated runtime dirs
- `bash -n` on the shell entrypoints and the AppShots helper
- `shellcheck -S error` on the shell scripts and the AppShots helper

### DIFF
--- a/linux-features/appshots/bin/bare-modifier-monitor
+++ b/linux-features/appshots/bin/bare-modifier-monitor
@@ -156,8 +156,8 @@ if [ -z "$keyboard_ids" ]; then
     exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


ilysenko (approved):

PR #655, opened 2026-07-03
fix: suppress launcher notice fd warnings

Yo-DDV:
Hi,

This fixes the launcher notice fallback that ShellCheck reports as SC2261. The previous redirection order sent stdout to the duplicated early-stderr FD before applying `2>/dev/null`, so a stale or closed FD could still print a redirection warning from what is meant to be a best-effort fallback.

The launcher now applies the stderr suppression before redirecting stdout to the saved FD. Valid-FD behavior stays the same, and the stale-FD path stays quiet. I also added a smoke guard so the template keeps that order.

Checks run locally:

- `shellcheck -s bash -S error launcher/start.sh.template`
- `shellcheck -S error tests/scripts_smoke.sh`
- `git diff --check`
- launcher FD behavior harness, valid FD matches and stale FD is suppressed
- `node --check` on 110 JS files
- `node --test` on 24 Node test files, 717 tests passed
- `make check`
- `make test`, 189 tests passed
- `cargo test --workspace --all-targets` with isolated runtime dirs
- `cargo clippy --workspace --all-targets -- -D warnings` with isolated runtime dirs

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -300,7 +300,7 @@ launcher_notice() {
 
     echo "$message"
     if [ -n "${LAUNCHER_EARLY_STDERR_FD:-}" ]; then
-        printf '%s\n' "$message" >&"$LAUNCHER_EARLY_STDERR_FD" 2>/dev/null || true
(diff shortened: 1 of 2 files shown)


ilysenko (approved):

PR #656, opened 2026-07-04
Add Skysight OCR with RapidOCR fallback

joshyorko:
Summary
- Adds local Skysight OCR evidence with a backend-neutral OCR result contract.
- Uses RapidOCR plus ONNXRuntime as the preferred optional backend when available.
- Preserves Tesseract CLI as fallback and keeps OCR non-fatal unless required.

Validation
- cargo fmt --check -p codex-record-replay-linux
- cargo test -p codex-record-replay-linux
- node --test linux-features/record-and-replay/test.js
- Real RapidOCR smoke in devcontainer with rapidocr, onnxruntime, pillow, and libgl1 recognized a generated Codex RapidOCR image.

### DIFF
--- a/record-replay-linux/src/ocr.rs
+++ b/record-replay-linux/src/ocr.rs
@@ -1,0 +1,1543 @@
+use anyhow::{Context, Result};
+use serde::{Deserialize, Serialize};
+use serde_json::{json, Value};
+use std::{
+    fs,
+    path::{Path, PathBuf},
+    process::{Child, Command, Output, Stdio},
+    thread,
+    time::{Duration, Instant, SystemTime, UNIX_EPOCH},
+};
+
+const DEFAULT_TESSERACT: &str = "tesseract";
+const DEFAULT_RAPIDOCR_PYTHON: &str = "python3";
+const DEFAULT_LANGUAGE: &str = "eng";
+const DEFAULT_RAPIDOCR_LANGUAGE: &str = "ch";
+const DEFAULT_PSM: &str = "11";
+const DEFAULT_TIMEOUT_MS: u64 = 10_000;
+const TESSERACT_BACKEND_NAME: &str = "tesseract-cli";
+const RAPIDOCR_BACKEND_NAME: &str = "rapidocr-python";
+const AUTO_BACKEND_NAME: &str = "auto";
+const RAPIDOCR_JSON_PREFIX: &str = "__CODEX_RAPIDOCR_JSON__";
+const MAX_OCR_OBSERVATIONS: usize = 200;
+const MAX_OCR_NORMALIZED_TEXT_BYTES: usize = 32 * 1024;
+const MAX_OCR_CANDIDATE_TEXT_BYTES: usize = 512;
+
+const RAPIDOCR_CHECK_SCRIPT: &str = r#"
+import importlib.metadata as metadata
+import json
+import sys
+
+try:
+    import onnxruntime  # noqa: F401
+    from rapidocr import RapidOCR
+
+    lang_type = sys.argv[1] if len(sys.argv) > 1 else "ch"
+    engine = RapidOCR(params={"Rec.lang_type": lang_type})
+    payload = {
+        "rapidocr": metadata.version("rapidocr"),
+        "onnxruntime": metadata.version("onnxruntime"),
+        "lang_type": str(getattr(engine.cfg.Rec, "lang_type", lang_type)),
+    }
+    print("__CODEX_RAPIDOCR_JSON__" + json.dumps(payload, ensure_ascii=False))
+except Exception as exc:
+    payload = {"error": f"{type(exc).__name__}: {exc}"}
+    print("__CODEX_RAPIDOCR_JSON__" + json.dumps(payload, ensure_ascii=False))
+    raise
+"#;
+
+const RAPIDOCR_RUN_SCRIPT: &str = r#"
+import json
+import sys
+
+from rapidocr import RapidOCR
+
+def as_list(value):
+    if value is None:
+        return []
+    if hasattr(value, "tolist"):
+        return value.tolist()
+    return list(value)
+
+if len(sys.argv) != 3:
+    raise SystemExit("expected rapidocr language and image path arguments")
+lang_type = sys.argv[1]
+image_path = sys.argv[2]
+engine = RapidOCR(params={"Rec.lang_type": lang_type})
+result = engine(image_path)
+payload = {
+    "boxes": as_list(getattr(result, "boxes", None)),
+    "txts": as_list(getattr(result, "txts", None)),
+    "scores": as_list(getattr(result, "scores", None)),
+    "elapse": getattr(result, "elapse", None),
+    "lang_type": str(getattr(engine.cfg.Rec, "lang_type", lang_type)),
+}
+print("__CODEX_RAPIDOCR_JSON__" + json.dumps(payload, ensure_ascii=False))
+"#;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum OcrMode {
+    Auto,
+    Enabled,
+    Required,
+    Disabled,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum OcrBackendPreference {
+    Auto,
+    RapidOcr,
+    Tesseract,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct OcrPolicy {
+    pub(crate) mode: OcrMode,
+    pub(crate) backend_preference: OcrBackendPreference,
+    tesseract_executable: PathBuf,
+    rapidocr_python: PathBuf,
+    pub(crate) language: String,
+    pub(crate) rapidocr_language: String,
+    pub(crate) page_segmentation_mode: String,
+    pub(crate) timeout: Duration,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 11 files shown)


record-replay-linux/src/ocr.rs
```diff
@@ -0,0 +1,1515 @@
...
+from rapidocr import RapidOCR
+
+def as_list(value):
+    if value is None:
+        return []
+    if hasattr(value, "tolist"):
+        return value.tolist()
+    return list(value)
+
+lang_type = sys.argv[1] if len(sys.argv) > 2 else "ch"
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`lang_type` selection is guarded by `len(sys.argv) > 2` but assigns `sys.argv[1]`. The run script is invoked with exactly two args (`rapidocr_language` then `frame_path`) — see `recognize_frame_rapidocr` which pushes `.arg(&policy.rapidocr_language).arg(frame_path)`, giving `len(sys.argv) == 3` at runtime (script path + 2). So this happens to work, but the guard is off-by-one relative to the value it reads: `image_path = sys.argv[2] if len(sys.argv) > 2 else sys.argv[1]` on line 63 uses the same `> 2` guard yet reads index 2 in the true branch. If the process is ever invoked with only the image path (1 arg), line 62 would read `sys.argv[1]` (the image path) as `lang_type`. The condition on line 62 should be `> 1` to match its `sys.argv[1]` access, mirroring the check script's `> 1` on line 35.

record-replay-linux/src/skysight.rs
```diff
@@ -769,6 +789,8 @@ fn provider_readiness_event(
     source: &str,
     diagnostics: &codex_computer_use_linux::diagnostics::DoctorReport,
 ) -> Value {
+    let ocr_policy = crate::ocr::OcrPolicy::from_env();
+    let ocr_readiness = ocr_policy.readiness();
```

---

`OcrPolicy::from_env()` + `readiness()` is invoked in three separate hot paths per capture (`provider_readiness_event` here, `status_value`, and again inside `write_chronicle_screen_recording_artifacts_inner`). Each `readiness()` call spawns a subprocess (`python3 -c` for RapidOCR check and/or `tesseract --version`) with up to a 10s timeout. For the auto backend that's two probes per call. During a single snapshot this can spawn the readiness probe multiple times, and `recognize_frame` calls `readiness()` yet again internally (ocr.rs:642). Consider computing readiness once and threading it through, so a slow/hanging python or tesseract can't multiply the per-frame latency.

avifenesh (commented):

---

Review of Skysight OCR (RapidOCR + Tesseract fallback). The OCR-after-privacy-gate ordering, exclusion text suppression, persistence limits, timeouts, and readiness contract all look correct and well tested. Two concerns worth a look, one MEDIUM and one lower-confidence.

joshyorko:
> 
> *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*
> 
> ---
> 
> Review of Skysight OCR (RapidOCR + Tesseract fallback). The OCR-after-privacy-gate ordering, exclusion text suppression, persistence limits, timeouts, and readiness contract all look correct and well tested. Two concerns worth a look, one MEDIUM and one lower-confidence.

What things lol. You didn't finish revauto. Come on feed me bro

record-replay-linux/src/ocr.rs
```diff
@@ -0,0 +1,1534 @@
...
+
+def as_list(value):
+    if value is None:
+        return []
+    if hasattr(value, "tolist"):
+        return value.tolist()
+    return list(value)
+
+lang_type = sys.argv[1] if len(sys.argv) > 1 else "ch"
+image_path = sys.argv[2] if len(sys.argv) > 2 else sys.argv[1]
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`recognize_frame_rapidocr` always spawns Python with exactly two positional args (`lang_type` at argv[1], `frame_path` at argv[2]), so `len(sys.argv) > 2` is true and `image_path` correctly resolves to argv[2]. Fine for the production path. But the `image_path = ... else sys.argv[1]` branch is dead for the real invocation and would silently OCR the *language string as a path* if the run script were ever called with a single arg — worth an assertion or removing the fallback so a future caller regression fails loudly instead of producing empty/garbage OCR text.

record-replay-linux/src/skysight.rs
```diff
@@ -2027,6 +2165,22 @@ fn status_value(input: StatusValueInput<'_>) -> Result<SkysightStatus> {
...
             .and_then(|status| status.summary_agent_last_error.clone()),
+        ocr_enabled: ocr_readiness.enabled,
+        ocr_available: ocr_readiness.available,
+        ocr_mode: Some(ocr_policy.mode_name().to_string()),
+        ocr_status: Some(ocr_readiness.status),
+        ocr_backend: Some(ocr_readiness.backend),
+        ocr_backend_version: ocr_readiness.version,
+        ocr_language: Some(ocr_readiness.language),
+        ocr_dependency_hint: ocr_readiness.dependency_hint,
+        ocr_last_run_at: existing
```
avifenesh, line 2206: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`ocr_last_run_at` is only ever populated by carrying forward the *existing* status value — no code path in this PR ever sets it after a successful OCR run (`recognize_frame_with_readiness` results in `write_chronicle_screen_recording_artifacts_inner` are not threaded back into the status record). As written the field will always be `None` in practice, so `skysight status` reports OCR readiness but never a last-run timestamp, and `ocr_last_error` (line 2181) will surface the readiness-probe error rather than an actual recognition error. Confirm this is intended vs. a missing wire-up.

avifenesh (commented):

---

Review of Skysight OCR (RapidOCR + Tesseract fallback). Two correctness concerns worth confirming before merge; inline. Overall the privacy-gate ordering (OCR downstream of screenshot suppression), exclusion-text stripping, and persistence bounds look sound.

record-replay-linux/src/skysight.rs
joshyorko: Good catch, fixed. OCR execution now threads back into status, and readiness errors no longer populate ocr_last_error.

record-replay-linux/src/skysight.rs
```diff
@@ -1434,6 +1570,18 @@ fn write_chronicle_screen_recording_artifacts_inner(
     crate::secure_fs::write_private_file(&latest_frame_path, &jpeg_bytes)?;
     crate::secure_fs::write_private_file(&sparse_frame_path, &jpeg_bytes)?;
     crate::secure_fs::write_private_file(&capture_marker_path, "active\n")?;
+    let exclusion_values = exclusions
+        .iter()
+        .map(|rule| rule.value.clone())
+        .collect::<Vec<_>>();
+    let ocr_result = crate::ocr::recognize_frame_with_readiness(
```
avifenesh, line 1577: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

Readiness and recognition can select different backends within one capture. `ocr_readiness` is computed once per snapshot in `capture_skysight_snapshot` (`policy.readiness()`), and `recognize_frame_with_readiness` dispatches on `readiness.backend`. In `auto` mode, `readiness()` probes RapidOCR then Tesseract and returns a concrete backend name (`rapidocr-python`/`tesseract-cli`). That's consistent here.

But note the `auto` readiness probe spends a full RapidOCR Python spawn + a Tesseract `--version` spawn on *every* snapshot before recognition itself spawns a third process — three subprocess launches per capture. Since readiness is now recomputed per snapshot (not cached across the daemon loop), a slow/hanging Python import under `auto` blocks each capture for up to `timeout` (default 10s) even when RapidOCR is not installed, before falling back. Consider caching readiness for the daemon session or short-circuiting `auto` when the RapidOCR python path is the bare `python3` default and import fails once.

record-replay-linux/src/ocr.rs
```diff
@@ -0,0 +1,1543 @@
...
+    } else {
+        image_height as f64
+    };
+    let x1 = x1.clamp(0.0, max_x);
+    let y1 = y1.clamp(0.0, max_y);
+    let x2 = x2.clamp(x1, max_x);
+    let y2 = y2.clamp(y1, max_y);
+    let pixel_x = x1.floor() as u32;
+    let pixel_y = y1.floor() as u32;
+    let pixel_width = x2.ceil().saturating_sub_f64(x1.floor());
```
avifenesh, line 1096: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`saturating_sub_f64` computes width/height from `x2.ceil() - x1.floor()`, so a zero-area RapidOCR box (x1==x2) yields a pixel width of at least 1 after ceil/floor rounding, while `observation_from_words` (tesseract path) uses `x2.saturating_sub(x1)` and can legitimately produce 0. The two backends therefore emit different pixel_bounding_box conventions for degenerate boxes. Minor, but worth confirming the downstream Chronicle consumer tolerates both.

avifenesh (commented):

---

Review of Skysight OCR (RapidOCR + Tesseract fallback). One correctness concern on the `auto` backend path around required-mode + runtime backend routing; inline. The earlier readiness/`ocr_last_run_at` issue and the RapidOCR argv issue appear addressed in this head.

ilysenko (approved):

joshyorko:
Follow-up for the two post-merge OCR review notes is open as #658. It caches OCR readiness for the running Skysight daemon so `auto` mode does not re-probe every capture, and it aligns RapidOCR zero-area pixel boxes with the existing Tesseract convention. Focused Docker validation passed on that branch.

PR #657, opened 2026-07-04
Add upstream DMG intelligence tooling

joshyorko:
Summary
- Adds upstream DMG intelligence tooling for protected-surface drift checks.
- Adds a one-command devcontainer wrapper for repeatable candidate-vs-baseline inspection.
- Keeps reports compact while surfacing blockers, structural drift, review-only drift, and protected surface coverage.

Validation
- node --test scripts/dev/upstream-dmg-intel.test.js
- bash -n scripts/dev/upstream-dmg-intel-devcontainer

### DIFF
--- a/scripts/lib/upstream-dmg-intel.js
+++ b/scripts/lib/upstream-dmg-intel.js
@@ -1,0 +1,1659 @@
+"use strict";
+
+const crypto = require("node:crypto");
+const fs = require("node:fs");
+const os = require("node:os");
+const path = require("node:path");
+const { spawnSync } = require("node:child_process");
+
+const TEXT_FILE_PATTERN = /\.(cjs|css|html|js|json|mjs|md|text|ts|tsx|txt|xml|yml|yaml)$/i;
+const NATIVE_FILE_PATTERN = /(^|\/)(codex_chronicle|SkyComputerUseClient|sky\.node|node_repl|node|[^/]+\.(node|dylib))$/i;
+const DEFAULT_MAX_TEXT_BYTES = 2_500_000;
+const DEFAULT_MAX_INVENTORY_HASH_BYTES = 10_000_000;
+const DRIFT_PATH_SAMPLE_LIMIT = 8;
+const DRIFT_CHANGED_SAMPLE_LIMIT = 12;
+const MARKDOWN_PATH_SAMPLE_LIMIT = 3;
+const ACTION_PLAN_PATH_SAMPLE_LIMIT = 3;
+const SUCCESSFUL_PATCH_STATUSES = new Set(["applied", "already-applied", "skipped-target", "skipped-disabled"]);
+const BLOCKING_PATCH_STATUSES = new Set(["failed-required"]);
+const ACTIONABLE_CLASSIFICATIONS = new Set([
+  "MOVED",
+  "RENAMED",
+  "PAYLOAD_CHANGED",
+  "REMOVED",
+  "NEW_UPSTREAM_CAPABILITY",
+  "PATCH_BROKEN",
+  "PATCH_REVIEW",
+  "LINUX_SUBSTRATE_GAP",
+  "PROTECTED_SURFACE_PARTIAL",
+  "PROTECTED_SURFACE_MISSING",
+]);
+const STRING_LITERAL_PATTERN = /`([^`\\]*(?:\\.[^`\\]*)*)`|"([^"\\]*(?:\\.[^"\\]*)*)"|'([^'\\]*(?:\\.[^'\\]*)*)'/g;
+const BRIDGE_CHANNEL_TERM_PATTERN =
+  /browser[-_]trace|chrome|chronicle|computer[-_]use|desktop[-_]snapshot|dictation|event[-_]stream|focused[-_]window|global[-_]dictation|nativeMessaging|record[-_ ]?(?:and[-_ ]?)?replay|skysight|speech[-_]context|window[-_]metadata/i;
+const LOWER_BRIDGE_CHANNEL_PATTERN = /^[a-z][a-z0-9_.:-]{2,119}$/;
+const CAMEL_BRIDGE_CHANNEL_PATTERN = /^(?:browserTrace|computerUse|eventStream|focusedWindow|nativeMessaging|recordAndReplay|speechContext|windowMetadata)$/;
+const ASSET_EXTENSION_PATTERN = /\.(?:css|dylib|exe|gif|ico|jpeg|jpg|js|json|md|node|png|svg|ts|tsx|txt|wasm|yml|yaml)$/i;
+const commandPathCache = new Map();
+
+function sha256(buffer) {
+  return crypto.createHash("sha256").update(buffer).digest("hex");
+}
+
+function readJson(filePath) {
+  return JSON.parse(fs.readFileSync(filePath, "utf8"));
+}
+
+function writeJson(filePath, value) {
+  fs.mkdirSync(path.dirname(filePath), { recursive: true });
+  fs.writeFileSync(filePath, `${JSON.stringify(value, null, 2)}\n`, "utf8");
+}
+
+function normalizePath(value) {
+  return String(value).replace(/\\/g, "/").replace(/^\/+/, "");
+}
+
+function toRegex(pattern) {
+  return pattern instanceof RegExp ? pattern : new RegExp(pattern, "i");
+}
+
+function matchAny(patterns, value) {
+  return (patterns ?? []).some((pattern) => toRegex(pattern).test(value));
+}
+
+function textSnippet(text, needle) {
+  const index = text.toLowerCase().indexOf(String(needle).toLowerCase());
+  if (index < 0) {
+    return null;
+  }
+  const start = Math.max(0, index - 80);
+  const end = Math.min(text.length, index + String(needle).length + 80);
+  return text.slice(start, end).replace(/\s+/g, " ").trim();
+}
+
+function includesNeedle(value, needle) {
+  const valueText = String(value);
+  const needleText = String(needle);
+  if (/^[A-Za-z0-9_]+$/.test(needleText) && needleText.length <= 4) {
+    const escaped = needleText.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+    return new RegExp(`(^|[^A-Za-z0-9_])${escaped}([^A-Za-z0-9_]|$)`, "i").test(valueText);
+  }
+  return valueText.toLowerCase().includes(needleText.toLowerCase());
+}
+
+function printableStrings(buffer, minLength = 4) {
+  const strings = new Set();
+  let current = "";
+  for (const byte of buffer) {
+    if (byte >= 32 && byte <= 126) {
+      current += String.fromCharCode(byte);
+    } else {
+      if (current.length >= minLength) {
+        strings.add(current);
+      }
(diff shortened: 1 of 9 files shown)


scripts/lib/upstream-dmg-intel.js
```diff
@@ -0,0 +1,1603 @@
...
+    repoRoot,
+  });
+  const reportDir =
+    outputDir ??
+    path.join(
+      repoRoot,
+      "reports/upstream-dmg",
+      timestamp ?? new Date().toISOString().replace(/[:.]/g, "-"),
+    );
+  const candidateInventory = createInventory({ registry, sourcePath: candidatePath });
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`createInventory` is invoked here (and for the baseline at line 1535) without a `workDir`, so when the candidate/baseline is a `.dmg`, `createInventory` allocates its own `mkdtempSync` scratch dir and removes it in its `finally` (lines 342-345, 365-368). That is fine functionally, but note that `extractProtectedSurfaces` / `createNativeBinaryMap` run *after* `createInventory` returns and rely on `file.absolutePath` to shell out to `file`, `nm`/`llvm-nm`, `otool` (lines 328, 868-890). For a DMG source those absolute paths point into the scratch dir that `createInventory` has already `rmSync`-ed by the time `extractProtectedSurfaces` runs — so `nativeSymbolInventory`/`nativeLinkedLibraries`/`runFileCommand` will silently get `null` (the spawn fails on the deleted path) and native symbol/library intelligence is quietly dropped for DMG inputs. The fixtures only exercise `.app` directory sources (`createFixtureApp`), so the tests never catch this. Consider passing a shared `workDir` that outlives the protected-surface extraction (or extracting native metadata during inventory while the scratch dir still exists).

avifenesh (commented):

---

Reviewed the new upstream DMG intelligence tooling. One correctness issue worth flagging in `buildIntelReports`; the rest of the diff (Makefile targets, devcontainer wrapper, registry, tests) looks consistent.

scripts/lib/upstream-dmg-intel.js
```diff
@@ -0,0 +1,1619 @@
...
+    const candidateInventory = createInventory({
+      registry,
+      sourcePath: candidatePath,
+      workDir: path.join(scratchRoot, "candidate"),
+    });
+    const candidateProtected = extractProtectedSurfaces({ inventory: candidateInventory, registry, repoRoot });
+    const baselineInventory =
+      resolvedBaselinePath == null
+        ? null
+        : createInventory({
```
avifenesh, line 1584: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`createInventory` is invoked here for the candidate (and at line 1544 for the baseline) with a `workDir` of `path.join(scratchRoot, "candidate")` / `"baseline"`, but the directory is never created before use. `createInventory` only creates a scratch dir itself when `workDir == null` (see line 342-345); when a `workDir` is passed it goes straight to `extractDmgToApp`, which does `fs.mkdirSync(path.join(workDir, "dmg-extract"), { recursive: true })`. Since `recursive: true` creates intermediate dirs, the extract subdir is fine — but if the source is a **directory/.app** (non-dmg) kind, the passed `workDir` is simply ignored, which is harmless. So this specific path is OK for dmg. The real gap: when both candidate and baseline are `.dmg`, both extract into distinct subdirs of the same `scratchRoot`, which is fine. No action needed — retracting; see the sibling comment for the actual concern.

scripts/lib/upstream-dmg-intel.js
```diff
@@ -0,0 +1,1619 @@
...
+    source: entry.candidate.source ?? entry.baseline.source,
+    type: entry.candidate.type ?? entry.baseline.type,
+  });
+}
+
+function normalizedHashedAssetPath(entryPath) {
+  const normalized = normalizePath(entryPath);
+  const directory = path.posix.dirname(normalized);
+  const basename = path.posix.basename(normalized);
+  const match = basename.match(/^(.+?)-[A-Za-z0-9_-]{6,}(\.[A-Za-z0-9.]+)$/);
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`normalizedHashedAssetPath` uses the regex `^(.+?)-[A-Za-z0-9_-]{6,}(\.[A-Za-z0-9.]+)$` to detect content-hashed asset names. The hash character class includes `-`, and `.+?` is non-greedy, so for a filename like `main-CNod9zFW.js` it matches, but for names containing multiple hyphenated segments (e.g. `record-and-replay-CNod9zFW.js`) the non-greedy `.+?` will stop at the first hyphen it can, capturing `record` as the stem and treating `and-replay-CNod9zFW` as the hash. This makes `hashed_asset_churn` classification collapse distinct assets that share only a first segment, producing false `hashed_asset_churn` verdicts (masking genuine `protected_path_changed` drift). Consider anchoring the hash segment to a fixed hash-length/charset or excluding `-` from the hash class so multi-segment stems are preserved.

avifenesh (commented):

---

Reviewed the upstream DMG intelligence tooling. Two correctness issues worth flagging in the library entry point; the rest of the diff (dev tooling, registry, Makefile targets, docs) is scoped to a gitignored `reports/` tree and does not touch the ASAR patch set, launcher, or updater.

scripts/lib/upstream-dmg-intel.js
```diff
@@ -0,0 +1,1619 @@
...
+  }
+  if (buffer != null && (type === "text" || type === "json") && enriched.size <= maxTextBytes) {
+    enriched.text = buffer.toString("utf8");
+  }
+  if (buffer != null && type === "native") {
+    enriched.nativeStrings = printableStrings(buffer).slice(0, 5000);
+    enriched.fileCommand = runFileCommand(file.absolutePath);
+  }
+
+  return enriched;
```
avifenesh, line 358: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`enrichInventoryFile` returns a new `enriched` object that never carries `absolutePath` forward (it copies `mode`, `relativePath`, `size`, `source`, `type`, and conditionally `sha256`/`text`/`nativeStrings`/`fileCommand`, but not `absolutePath`). `collectInventoryFiles` maps every file through `enrichInventoryFile`, so the inventory `files` that `createNativeBinaryMap` consumes have `absolutePath === undefined`.

As a result `nativeSymbolInventory` (line 868) and `nativeLinkedLibraries` (line 883) both hit the `file.absolutePath == null` guard and return `null` for *every* native binary — for `.app`, `.dmg`, and directory sources alike, not just DMG inputs. `nm`/`llvm-nm`/`otool` symbol and linked-library intelligence is therefore silently never populated (`symbols`/`linkedLibraries` are always `null` in `native-binary-map.json`).

The tests only assert on `protectedStringHits` (which comes from `nativeStrings`, computed inside `enrichInventoryFile` while `absolutePath` is still in scope via the `file` param at line 328), so this gap is invisible to the suite. To restore native symbol/library extraction, propagate `absolutePath` onto `enriched` (and rely on `publicInventory` at line 1476, which already deletes it before serialization).

avifenesh (commented):

---

Reviewed the upstream DMG intelligence tooling. The three concerns from the prior reviews appear addressed at head (`buildIntelReports` now passes an explicit `workDir` so the scratch dir survives protected-surface extraction; the hashed-asset regex is greedy). One remaining correctness gap worth flagging in the library entry point around `enrichInventoryFile` / native metadata extraction.

ilysenko:
- `enrichInventoryFile()` drops `absolutePath`, so `createNativeBinaryMap()` cannot populate native symbols or linked libraries; both stay `null` for every native binary. Please preserve `absolutePath` on the internal enriched inventory object and keep stripping it only in `publicInventory()`.

joshyorko:
Fixed in `5e815201`: internal inventory now preserves `absolutePath` for native symbol/library extraction while `publicInventory()` still strips it before JSON output. I also replaced shell-based PATH lookup with direct PATH resolution so agent/tool shims work reliably, and added a regression that proves `native-binary-map.json` gets symbol output from a fake `llvm-nm` receiving an existing file path. Validated in Docker: focused DMG intel test plus the full `scripts/dev/upstream-dmg-intel.test.js` suite.

scripts/lib/upstream-dmg-intel.js
```diff
@@ -0,0 +1,1644 @@
...
+    return ["NEW_UPSTREAM_CAPABILITY"];
+  }
+  return ["UNCHANGED"];
+}
+
+function patchFailuresBySurface(patchReport, surfacesById = {}) {
+  const map = new Map();
+  const surfaces = Object.values(surfacesById);
+  for (const patch of patchReport?.patches ?? []) {
+    if (SUCCESSFUL_PATCH_STATUSES.has(patch.status)) {
```
avifenesh, line 965: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`patchFailuresBySurface` treats any patch status not in `SUCCESSFUL_PATCH_STATUSES` as a failure, but the drift classification `PATCH_BROKEN` is emitted uniformly regardless of whether the failing patch was `required-upstream` vs `skipped-optional`. A `skipped-optional` patch (see the test at upstream-dmg-intel.test.js:369-374, which uses `status: "skipped-optional"`) is folded into `PATCH_BROKEN` and then into `BLOCKING_CLASSIFICATIONS` in the CLI. Per the repo invariant that required-upstream failures must remain distinguishable from skipped-optional (so upstream drift stays visible and optional skips don't produce false blockers), consider keying the classification/severity off the patch's `ciPolicy`/status rather than collapsing all non-success statuses into a single blocking `PATCH_BROKEN`.

avifenesh (commented):

---

Reviewed the new upstream DMG intelligence tooling. The prior review concerns (`absolutePath` preservation for native symbol extraction, `publicInventory()` stripping) appear addressed at head. One correctness issue remains in the P2-adjacent classification of failed required-upstream patches; the rest of the diff is dev-only tooling with no production wiring. Flagging one item worth a look.

Note: this is dev-only tooling (gitignored `reports/upstream-dmg/`, no ASAR patch-set wiring, no launcher/updater changes), so most repo-specific invariants (P1–P33) don't apply.

ilysenko:
Thanks for the update. I rechecked the current head, and the blocker still looks real:

- `skipped-optional` patch-report entries are still classified as `PATCH_BROKEN`, and `PATCH_BROKEN` is treated as an acceptance blocker by `--fail-on-blockers`.
- This can falsely block a new DMG even when the protected surface is still fully present and the skipped patch is optional.
- Please distinguish required-upstream failures from optional skipped/warning statuses, and keep optional drift as review-only rather than blocking acceptance.

avifenesh:
@joshyorko @ilysenko Sorry for the amount. I don't remember changing the policy to per push, revuto should review once each PR once. If it's not a bother, I'd leave it, if it's spamming, I'll limit it.

joshyorko:
Fixed in 9022853. Optional skipped/warning patch statuses now emit PATCH_REVIEW and stay review-only under --fail-on-blockers; required failed patches still emit blocking PATCH_BROKEN. Added regressions for both paths.

ilysenko (approved):
Reviewed the current head after the requested fixes. The previous blocker is resolved, optional patch drift is review-only, and focused validation passes.

PR #659, opened 2026-07-04
[codex] localize Linux desktop settings metadata

pradigmaz:
## Summary

This localizes the generated Linux desktop settings page through the upstream renderer message path, adds shared Linux target detection for packaging/setup flows, and keeps Read Aloud compatible with the current generated Linux desktop page.

## Changes

- Add Linux desktop renderer message ids and Russian locale entries for the custom settings page and build information panel.
- Share Linux target detection between shell helpers and JS build-info generation.
- Keep the Read Aloud settings injector matching the current upstream-message `Build` section.
- Make patcher and smoke tests portable in restricted local environments without skipping normal coverage.
- Keep the build-info locale table scoped to 60 locales with the same build-info key set in every locale.

## Testing

- `bash -n tests/scripts_smoke.sh scripts/bootstrap-wizard.sh scripts/install-deps.sh scripts/lib/linux-target-detect.sh && git diff --check`
- `node --check linux-features/codex-wrapper-updater/patch.js && node --check linux-features/read-aloud/patch.js && node --check linux-features/read-aloud/test.js && node --check scripts/lib/build-info.js && node --check scripts/lib/linux-target-context.js && node --check scripts/patch-linux-window-ui.js && node --check scripts/patch-linux-window-ui.test.js && node --check scripts/patches/keybinds-settings.js && node --check scripts/patches/main-process/tray.js && node --check scripts/patches/registry.js && node --check scripts/patches/linux-build-info-i18n.js && node --check scripts/patches/linux-build-info-locales.js`


Yo-DDV (changes requested):
Hi, thanks for working on this.

I pulled the PR merge commit locally because the GitHub `Rust and Smoke Tests` job is red. That failure looks transient to me: the job stops while Cargo downloads `futures-lite` with a curl `unexpected eof`, and the same merge commit passes locally.

I did find one issue worth fixing before merge. The new Fedora Atomic profile says the package manager is `rpm-ostree`, but the shared target detection still reports `unknown` on an Atomic Fedora host when only `rpm-ostree` is present. That also shows up in the build-info UI and detail dialog because those rows use `linuxTarget.packageManager ?? packageProfile.packageManager`, so the string `"unknown"` is treated as a value and prevents the profile fallback. It also makes `DETECT_ONLY=1 bash scripts/install-deps.sh` print `Detected dependency profile: unknown` for the same host shape.

Minimal repro I used:

- Fedora 44 os-release fixture
- `OSTREE_BOOTED_FILE` present
- `PATH` containing only a fake executable `rpm-ostree`

Result:

- JS target: `target.atomic === true`, `target.packageFormat === "rpm"`, `target.packageManager === "unknown"`, `packageProfile(target).packageManager === "rpm-ostree"`
- shell helper: `atomic=yes`, `format=rpm`, `manager=unknown`

I would either make the shared JS/shell detector understand `rpm-ostree` for Atomic Fedora, or keep the dependency installer conservative but make the build-info display fall back to the profile when `linuxTarget.packageManager` is `unknown`. Either way, it needs a direct test for Atomic Fedora with `rpm-ostree` present and no `dnf`/`dnf5`, so the profile and the visible metadata cannot drift again.

Verified locally on the PR merge commit:

- `make check`
- `bash scripts/ci/run-node-checks.sh` (724 passed)
- `make test` (189 passed)

tests/scripts_smoke.sh
```diff
@@ -1940,6 +1940,78 @@ test_fedora_dependency_bootstrap_installs_rpmbuild() {
...
+            source "$HELPER_PATH"
+            OS_RELEASE_ID="$(os_release_field ID)"
+            OS_RELEASE_ID_LIKE="$(os_release_field ID_LIKE 2>/dev/null || true)"
+            OS_RELEASE_VERSION_ID="$(os_release_field VERSION_ID)"
+            printf "id=%s\nid_like=%s\nmanager=%s\nformat=%s\natomic=%s\n" \
+                "$OS_RELEASE_ID" \
+                "$OS_RELEASE_ID_LIKE" \
+                "$(detect_package_manager)" \
+                "$(detect_package_format)" \
+                "$([ linux_target_is_atomic ] && echo yes || echo no)"
```
avifenesh, line 1988: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This assertion doesn't actually exercise `linux_target_is_atomic`. `[ linux_target_is_atomic ]` is a one-argument `test`, which is true whenever the single argument is a non-empty string — and the literal word `linux_target_is_atomic` is always non-empty, so the bracket expression is unconditionally true and the function is never called.

As a result the `atomic=yes` check on line 1995 passes regardless of what the helper does: it would still pass if `linux_target_is_atomic` were hard-coded to return false. Since there's no negative-case assertion either, this test gives no real coverage of the atomic-detection helper it was added for.

Production callers use the correct form (e.g. `bootstrap-wizard.sh`: `if linux_target_is_atomic; then`). Drop the brackets here to match:

```sh
"$(linux_target_is_atomic && echo yes || echo no)"
```

Consider also adding a companion assertion with `OSTREE_BOOTED_FILE` unset / pointing at a nonexistent path expecting `atomic=no`, so a regression in the helper is actually caught.

avifenesh (commented):

---

One issue in the new smoke-test coverage for the shared Linux target detection helper. The rest of the change (locale table key-set consistency across all 60 locales, renderer message wiring, `wayland`/`x11` mutual-exclusivity, the read-aloud/wrapper-updater regex replacements, and the tray i18n helper scoping) checked out.

ilysenko:
Thanks for the work here. This PR should not move forward in its current shape.

Localization is not a priority for this project right now, and this PR also mixes several separate concerns: Linux desktop settings localization, Build information UI changes, shared Linux target detection, and Read Aloud compatibility. Some of those changes touch core setup/install/build-info behavior, not just localization, so reviewing and maintaining this as one combined PR would be risky.

Current blockers:
- The required Rust and Smoke Tests check is red. The log may be a transient Cargo registry EOF, but the required check still needs to be green before merge.
- Fedora Atomic/rpm-ostree detection is inconsistent. The package profile says `rpm-ostree`, but both JS and shell detection can still report `manager=unknown` when only `rpm-ostree` is present.
- The new smoke test does not actually call `linux_target_is_atomic`: `[ linux_target_is_atomic ]` is always true because it tests a non-empty string.

If there are useful non-localization fixes here, please split them into separate focused PRs:
- Read Aloud compatibility with the current Linux Desktop settings Build section, if that fix is still needed.
- Shared Linux target detection, with direct Fedora Atomic/rpm-ostree coverage in both JS and shell paths.
- Build information UI/diagnostics, if it can be kept separate from localization.

Keeping these separate will make the changes easier to review and much safer to maintain later.

PR #660, opened 2026-07-04
Cache immutable webview assets

Yo-DDV:
Hi,

This keeps the local webview server from forcing `no-store` on hashed assets.
HTML and other non-asset paths still stay uncached, but `/assets/*` is now served with `Cache-Control: public, max-age=31536000, immutable` and normal conditional requests are preserved.

In a Chrome benchmark against the packaged webview content, four repeated loads went from:

- `index.html`: 293 requests / 43.32 MiB to 77 requests / 10.87 MiB
- `avatar-overlay-composition-surface.html`: 313 requests / 43.72 MiB to 82 requests / 10.97 MiB

I also added a smoke test covering immutable asset caching, HTML no-store behavior, non-asset no-store behavior, conditional `304` responses, and a normalized `/assets/../index.html` path.

Tests:

- `python3 -m py_compile launcher/webview-server.py`
- `git diff --check`
- `make check`
- `make test`

### DIFF
--- a/launcher/webview-server.py
+++ b/launcher/webview-server.py
@@ -4,8 +4,10 @@ import ctypes.util
 import functools
 import http.server
 import os
+import posixpath
 import signal
 import sys
+import urllib.parse
 
 
 def _install_parent_death_signal():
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed current head. Webview cache policy validation and required checks are green.

PR #662, opened 2026-07-04
Detect Fedora Atomic rpm-ostree targets

Yo-DDV:
Hi,

This follows up on the useful target-detection part from #659 after that PR was closed, without carrying over the localization, Read Aloud, or build-info UI changes.

The installer stays conservative on rpm-ostree systems: it detects the profile and prints the manual layering command, but it does not run rpm-ostree automatically. The shell setup paths now share the same target detector, so setup-native and install-deps report the same host shape. The JS build metadata path also reports Fedora Atomic as an rpm-ostree target.

What changed:
- detect ostree-booted Fedora hosts and report rpm-ostree in the JS target context and build metadata
- share shell distro and package detection through scripts/lib/linux-target-detect.sh
- add direct JS and shell smoke coverage for Fedora Atomic with rpm-ostree present and no dnf or dnf5
- cover invalid atomic override fallback and explicit atomic opt-out

Tested:
- bash -n scripts/lib/linux-target-detect.sh scripts/install-deps.sh scripts/bootstrap-wizard.sh
- node --check scripts/lib/linux-target-context.js && node --check scripts/lib/build-info.js && node --check scripts/patch-linux-window-ui.test.js
- node --test scripts/patch-linux-window-ui.test.js --test-name-pattern 'Fedora Atomic|package profile|Linux target context'
- bash scripts/ci/run-node-checks.sh
- bash tests/scripts_smoke.sh
- make check
- make test
- cargo clippy --workspace --all-targets -- -D warnings

Thanks.

### DIFF
--- a/scripts/bootstrap-wizard.sh
+++ b/scripts/bootstrap-wizard.sh
@@ -6,6 +6,8 @@ SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 REPO_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
 FEATURES_ROOT="${CODEX_LINUX_FEATURES_ROOT:-$REPO_DIR/linux-features}"
 PACKAGE_NAME="${PACKAGE_NAME:-codex-desktop}"
+# shellcheck disable=SC1091
+source "$SCRIPT_DIR/lib/linux-target-detect.sh"
 SETUP_ERROR_REPORTED=0
 COLOR_RESET=""
 COLOR_BOLD=""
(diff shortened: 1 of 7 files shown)


ilysenko:
- `install_rpm_ostree()` always exits with an error, so after the user follows the manual `rpm-ostree install ...` instruction and reboots, rerunning `scripts/install-deps.sh` hits the same error again.
- Please make the rpm-ostree path able to complete after dependencies are already layered, and keep it aligned with the dnf5 managed-Node behavior instead of requiring system `nodejs/npm`.

Yo-DDV:
Hi, thanks for catching that.

You were right: I had covered the target-detection path, but missed the rerun-after-layering case. I pushed an update so the rpm-ostree path now checks whether the layered build tools are already available and completes in that state. When something is still missing, it keeps the manual `rpm-ostree install ...` guidance and lists the missing tools.

I also aligned the Node handling with the dnf5 path: rpm-ostree no longer requires system `nodejs/npm`, because `install.sh` provides the managed runtime.

I added smoke coverage for both states:
- before layering, it still stops with the manual rpm-ostree instruction
- after layering, it completes without system node/npm on PATH

Local checks are green on my side, and the GitHub checks are running again now.

ilysenko (approved):
Reviewed the updated head after the rpm-ostree fix. The previous rerun blocker is resolved, managed Node behavior is preserved, and focused plus smoke validation passes.

PR #663, opened 2026-07-04
perf(remote-mobile): coalesce duplicate pending turn starts

Yo-DDV:
Hi,

This follows #639 and builds on #640.

The first remote-mobile hydration fix avoids duplicate `readThread` calls while a missing thread is being loaded. While testing longer conversations from a French 4G/5G connection, I still saw one noisy case: repeated `turn/started` notifications for the same turn can remain queued and then replayed once hydration completes.

This keeps the existing queue and narrows the extra coalescing to that duplicate case only:

- same pending `turn/started`
- same normalized turn id
- same thread hydration already in progress

Events that may carry distinct content, including `item/*`, `turn/completed`, and `turn/started` without a turn id, are still queued and replayed in order. When a duplicate is coalesced, the latest payload is retained.

In the synthetic burst check, 100 duplicate pending `turn/started` notifications for the same turn now become 1 queued and replayed notification. On my mobile test path from France, long conversations that could previously take close to a minute to open before this series are usually around 4 to 8 seconds now.

Tests:

- `node --test linux-features/remote-mobile-control/test.js --test-name-pattern 'hydration|remote mobile'`
- `cargo +1.96.0 fmt --check`
- `git diff --check`
- fork CI passed: https://github.com/Yo-DDV/codex-desktop-linux/actions/runs/28707674099

Thanks.

Related to #639.


ilysenko (approved):
Looks good after rebasing onto the new patch descriptor flow. CI is green on the updated head.

Issue #631, opened 2026-06-30
Investigate Codex hardware / Work Louder control surface support on Linux


joshyorko:
## Summary

OpenAI Developers teased a Codex-related hardware release for July 15 with the line:

> Your favorite Codex shortcuts are getting an upgrade.

The teaser appears to be an OpenAI × Work Louder device, likely a programmable macro/control surface rather than a new model or general-purpose AI device.

Source:
- https://www.theverge.com/ai-artificial-intelligence/959174/openai-codex-hardware-work-louder

This issue is **not** about tracking every hardware accessory. It is about tracking whether this launch reveals a first-party **Codex desktop control-surface contract** that `codex-desktop-linux` should support.

## Why this matters for Linux

Most of the new Codex desktop primitives are currently macOS-first because OpenAI appears to be building and dogfooding there first:

- Record & Replay
- Computer Use helpers
- desktop plugin shells
- shortcut/control affordances
- app-specific workflow surfaces

That does not necessarily mean the underlying product contract is macOS-only. It may simply mean the first official helper implementation is macOS-only. Linux can often implement the same contract with different lower-level primitives.

For this hardware teaser, the important question is:

> Is this just a macro pad sending keyboard shortcuts, or does the Codex desktop app expose a deeper shortcut/action/control protocol?

If it is only HID keyboard/macro behavior, Linux support may be mostly documentation, keymap profiles, and action mapping.

If it talks to the Codex desktop app through a local protocol, plugin hook, SDK surface, WebSocket, IPC, HID vendor protocol, or app action registry, then `codex-desktop-linux` should investigate implementing the same contract.

## Hypothesis

This may be a Work Louder-style programmable control surface for Codex, potentially exposing buttons/actions for first-party Codex workflows such as:

- start/stop Record & Replay
- mark a recording step
- launch Skills / Plugins
- accept or reject patches
- run tests
- switch threads / goals / repositories
- stop/cancel an active task
- push-to-talk or voice context
- open terminal/browser/computer-use affordances

The launch may also reveal which Codex actions OpenAI considers flagship workflows. Record & Replay is especially likely to be part of the shortcut surface.

### Before July 15

Inspect the current Codex desktop app plumbing for signs that a shortcut/control-surface contract already exists:

- action registries or command palettes
- shortcut maps / keybinding maps
- global shortcut registration
- plugin action hooks
- Record & Replay activation hooks
- thread/goal start-stop action names
[...]

PR #665, opened 2026-07-04
Track Work Louder control surface in upstream DMG intel

Yo-DDV:
Hi,

This adds the Work Louder control-surface hooks to the upstream DMG protected-surface registry.

The current upstream bundle already carries the Work Louder device kits, the micro service, Project2077 references, and node-hid. Tracking those pieces gives #631 a reportable drift check instead of relying on manual bundle inspection.

This does not enable hardware support or add Linux HID packaging. It only classifies whether the upstream bundle still exposes the same surface, with tests for the complete case and for the HID runtime going missing.

Testing:
- node --test scripts/dev/upstream-dmg-intel.test.js
- bash scripts/ci/run-node-checks.sh
- git diff --check
- real bundle scan: work_louder_control_surface PRESENT with 5 evidence paths and no missing anchors

### DIFF
--- a/scripts/dev/upstream-dmg-intel.test.js
+++ b/scripts/dev/upstream-dmg-intel.test.js
@@ -53,6 +53,38 @@ const registry = {
       },
     },
     {
+      id: "work_louder_control_surface",
+      title: "Work Louder control-surface bundle hooks",
+      category: "native",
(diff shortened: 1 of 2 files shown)


ilysenko (approved):

PR #666, opened 2026-07-04
fix(remote-mobile): recover stale stream state

Yo-DDV:
Hi,

This is a small follow-up to #663 and the remote mobile feedback tracked in #639.

While testing long remote mobile sessions, I found two stale stream states that can leave the Android view behind the finished conversation state:

- a completed stream item can be dropped if the local turn state lost the matching started item
- the remote composer can keep showing a running state when the thread runtime status is still active, but the response is no longer rendering and there are no active runtime flags

The patch keeps the existing remote hydration flow, but adds two fail-soft webview patches:

- recover the completed item payload when it is safe to do so
- treat an empty active runtime status as stale only when no response is in progress, while preserving real waits, approvals, and unknown status shapes

Tests:

- `node --test --test-name-pattern 'completed stream|empty active runtime|remote terminal status' scripts/patch-linux-window-ui.test.js`
- `node --test scripts/patch-linux-window-ui.test.js` - 334 pass / 0 fail
- `node --test linux-features/*/test.js` - 385 pass / 0 fail
- `make check`
- `make test` - 189 pass / 0 fail
- patcher dry-run against a copy of the currently installed extracted bundle: first run `applied`, second run `already-applied` for both new descriptors

Related to #639.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -148,6 +148,8 @@ const {
   applyLocalEnvironmentActionModalDraftPatch,
   applyPersistentRateLimitFooterPatch,
   applyLinuxAppServerBackfillWaitPatch,
+  applyLinuxCompletedItemRecoveryPatch,
+  applyLinuxRemoteTerminalStatusRecoveryPatch,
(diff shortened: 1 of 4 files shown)


ilysenko (approved):

PR #669, opened 2026-07-05
Fix standalone Codex CLI update path

### LINKED ISSUE #668: Codex Update Manager tries to reinstall instead of updating the official standalone CLI via npm
When the Codex CLI is installed using the official OpenAI standalone installer, the Linux update manager correctly detects the existing installation but still attempts to reinstall it using npm. This causes the update to fail.

**Environment**
Linux Mint 22.3 (Ubuntu 24.04)
Node.js 22.23.0
npm 10.9.8
Git 2.43.x

**Steps to reproduce**
Install the official Codex CLI:
curl -fsSL https://chatgpt.com/codex/install.sh | sh
Verify the CLI:
which codex
codex --version

**Output**:

/home/user/.local/bin/codex
codex-cli 0.141.0

The executable is a symlink to:

/home/user/.codex/packages/standalone/current/bin/codex

**Update error:**
The update manager correctly detects the existing standalone installation:

{
  "cli_path": "/home/user/.local/bin/codex",
  "cli_installed_version": "0.141.0",
  "cli_latest_version": "0.142.5",
  "cli_status": "update_required"
}

However, instead of updating the existing standalone installation, it executes:

/usr/bin/npm install -g @openai/codex@0.142.5

This fails with:

EACCES: permission denied, mkdir '/usr/lib/node_modules/@openai'

The update manager then falls back to:

npm install -g --prefix ~/.local @openai/codex@0.142.5

which fails with:

npm ERR! EEXIST
File exists: ~/.local/bin/codex

because that file already exists as the symlink created by the official standalone installer.

The service then exits and systemd restarts it, resulting in repeated desktop notifications.

ilysenko:
## Summary
- Detect official standalone Codex CLI installs under `~/.codex/packages/standalone` during updater preflight.
- Update standalone installs through the official standalone installer with the resolved latest version instead of falling back to npm.
- Keep npm-managed and missing CLI install flows on the existing npm path, and document the behavior.

Fixes #668

## Tests
- `cargo fmt --check`
- `cargo test -p codex-update-manager codex_cli`

## Notes / Risks
- The updater still uses `npm view @openai/codex version` for latest-version discovery.
- Standalone updates require `curl` or `wget` so the official installer can be downloaded.

### DIFF
--- a/updater/src/codex_cli.rs
+++ b/updater/src/codex_cli.rs
@@ -8,7 +8,7 @@ use anyhow::{anyhow, Context, Result};
 use chrono::{Duration, Utc};
 use semver::Version;
 use std::{
-    ffi::OsString,
+    ffi::{OsStr, OsString},
     fs,
     os::unix::fs::PermissionsExt,
     path::{Path, PathBuf},
@@ -17,6 +17,7 @@ use std::{
 use tracing::{info, warn};
 
 const CLI_PACKAGE_NAME: &str = "@openai/codex";
+const STANDALONE_INSTALLER_URL: &str = "https://chatgpt.com/codex/install.sh";
 const CLI_NOT_INSTALLED_MESSAGE: &str =
     "Codex CLI is required but not currently installed. Open the app to retry the automatic install flow, or install it manually with npm.";
 const CLI_VERSION_CHECK_TTL: Duration = Duration::hours(1);
@@ -116,7 +117,7 @@ pub fn preflight(
 
     state.cli_status = CliStatus::Updating;
     persist_state(paths, state)?;
-    install_latest_cli(&latest_version)?;
+    update_existing_cli(&cli_path, &latest_version)?;
 
     let (refreshed_path, refreshed_version) = if let Some(updated_cli) =
         resolve_cli_path_with_version(requested_path, &latest_version)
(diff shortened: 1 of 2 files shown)


updater/src/codex_cli.rs
```diff
@@ -498,6 +499,159 @@ fn read_latest_version() -> Result<String> {
...
+}
+
+fn update_standalone_cli(install: &StandaloneCliInstall, latest_version: &str) -> Result<()> {
+    let downloader = standalone_installer_downloader()?;
+    let mut command = Command::new("sh");
+    command
+        .arg("-c")
+        .arg(downloader.shell_script())
+        .env(
+            "PATH",
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

For standalone installs we discover the downloader in `standalone_installer_downloader()` using the current `command_path_env()` (lines 629-635), but then run a bare `curl`/`wget` after changing `PATH` here to prepend the user-writable `install_dir`. If `~/.local/bin` (or a custom install dir) contains a different/broken `curl`, the update will execute that instead of the downloader we just selected, so an otherwise valid standalone CLI update can fail or run the wrong program. Please carry the resolved downloader path through and invoke that exact path (or avoid prepending `install_dir` until after the installer download).

avifenesh (commented):

PR #670, opened 2026-07-05
Restore Linux desktop settings route scan

colonelpanic8:
## Summary
- broaden the Linux Desktop settings route scan to include hashed settings-page chunks such as `app-initial~settings-page-*.js`
- add regression coverage for that route-map location while preserving the separate navigation/icon patch behavior

## Root cause
The Linux Desktop settings page patch only scanned legacy app-main/index chunks, `app-initial~app-main~*` chunks, and direct `settings-page*` bundles. If upstream hoists the lazy route map into a hashed settings-page concatenation chunk without the app-main prefix, the optional patch can miss the route map and skip restoring the Linux Desktop settings entry.

## Validation
- `nix build .#codex-desktop --no-link --print-out-paths`
- inspected the Nix output for `linux-desktop-settings-linux.js` and `keybinds-settings` patch report status `applied`


ilysenko (approved):
Reviewed current head locally; focused patch looks safe and CI is green.

PR #671, opened 2026-07-05
Fix read-aloud warning for normalizer bundles

Yo-DDV:
## Summary

- avoid reporting read-aloud assistant render drift on turn-normalizer bundles that only build assistant-message items
- keep the warning for actual render candidates that still expose `assistantCopyText`
- add regression coverage for both the normalizer and render-drift cases

## Validation

- `make check`
- `make test`


ilysenko (approved):
Reviewed: small opt-in read-aloud warning fix with focused tests; no blockers found.

PR #673, opened 2026-07-05
Prefer newest local CLI when launching Desktop

### LINKED ISSUE #672: Desktop can't send messages — "Tool permission stream closed before response received" — when the app resolves an older codex-cli from PATH
## Summary

On a fresh install, the desktop app failed to send **any** message. Every turn ended immediately with:

```
[external_agent_tool_result: error]
Tool permission request failed: Error: Tool permission stream closed before response received
[/external_agent_tool_result]
```

Root cause turned out to be a **`codex-cli` version mismatch**: the GUI-launched app spawned `codex app-server` using an **older** `codex-cli` that happened to be first on `PATH`. Pointing the app at a newer `codex` binary via `CODEX_CLI_PATH` fixed it completely.

## Environment

- OS: Ubuntu 24.04, GNOME Shell 46, X11
- Desktop app: appVersion `26.623.101652`, port `0.8.6` (build commit `0f4c722`), Electron `42.1.0`
- Two `codex-cli` installs present on the machine:
  - `/usr/bin/codex` → **0.120.0** (older, root/global npm install, first on the GUI session `PATH`)
  - `~/.npm-global/bin/codex` → **0.142.5** (newer, user npm prefix, NOT on the GUI `PATH`)

## What happens

The app's `resolveCodex` picks up `/usr/bin/codex` (0.120.0) because that's what the GUI session's `PATH` resolves. Confirmed with:

```
$ pgrep -af "codex app-server"
node /usr/bin/codex app-server --analytics-default-enabled
.../@openai/codex-linux-x64/vendor/x86_64-unknown-linux-musl/codex app-server ...
```

The session rollout (`~/.codex/sessions/.../rollout-*.jsonl`, `originator: Codex Desktop`, `cli_version: 0.120.0`) contains the `Tool permission stream closed before response received` error. It looks like the app↔agent tool-permission / approval bridge is incompatible with that older core, so the permission stream closes before the turn can proceed — and no message can be sent.

Note: the exact same account/config sends fine via the **standalone CLI** (`codex exec "..."` with `cli_version 0.142.5`), so the model backend was never the problem — it's purely which `codex-cli` the desktop app drives.

## Fix / workaround

Point the app at the newer `codex` binary. Verified working immediately (`pgrep -af "codex app-server"` then shows the new path, and messages send normally):

- Ad hoc:
  ```bash
  CODEX_CLI_PATH=/home/<user>/.npm-global/bin/codex /opt/codex-desktop/start.sh
  ```
- Persistent, launcher-independent, no sudo (systemd user env, applies to icon/taskbar/CLI launches after next login):
  ```bash
  mkdir -p ~/.config/environment.d
  echo 'CODEX_CLI_PATH=/home/<user>/.npm-global/bin/codex' > ~/.config/environment.d/codex-cli-path.conf
  ```
- Alternatively: make sure the `codex` that is first on `PATH` is a recent version.

## Suggested improvements

- Log the resolved `codex` path **and version** at startup (would have made this a 1-minute diagnosis).
- Warn (or refuse) when the resolved `codex-cli` is below a known-good minimum for the current app build.
- Document `CODEX_CLI_PATH` and the minimum supported `codex-cli` version in the README / troubleshooting.
- Consider preferring the newest `codex` found, or shipping/pinning a known-good core.

Possibly related to CLI-version handling in #668.

Yo-DDV:
## Summary
- Resolve executable `codex` candidates from the GUI `PATH` and known user install locations, then choose the newest readable version instead of the first `PATH` hit.
- Keep the resolver local-only and bound version probes so a broken candidate cannot stall launcher startup.
- Log the selected CLI path and version, document the behavior, and keep `CODEX_CLI_PATH` as the explicit override.

Fixes #672

## Tests
- Red check: `origin/main` selects the older `PATH` CLI in the reported fixture.
- `node --test scripts/patch-linux-window-ui.test.js` (335 pass / 0 fail)
- `node --test linux-features/*/test.js` (385 pass / 0 fail)
- `make check`
- `make test` (192 pass / 0 fail)

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -1362,11 +1362,228 @@ resolve_notification_icon() {
     echo "$APP_NOTIFICATION_ICON_NAME"
 }
 
-find_codex_cli() {
-    if command -v codex >/dev/null 2>&1; then
-        command -v codex
+codex_cli_version_probe() {
+    local output_file="$1"
+    shift
+    local probe_pid attempt wait_status
+
+    "$@" >"$output_file" 2>/dev/null &
+    probe_pid=$!
+    for ((attempt = 0; attempt < 50; attempt++)); do
+        if ! kill -0 "$probe_pid" 2>/dev/null; then
+            break
+        fi
+        sleep 0.02
+    done
+    if kill -0 "$probe_pid" 2>/dev/null; then
+        kill "$probe_pid" 2>/dev/null || true
+        for ((attempt = 0; attempt < 10; attempt++)); do
+            if ! kill -0 "$probe_pid" 2>/dev/null; then
+                break
+            fi
+            sleep 0.02
+        done
+        if kill -0 "$probe_pid" 2>/dev/null; then
+            kill -9 "$probe_pid" 2>/dev/null || true
+        fi
+        wait "$probe_pid" 2>/dev/null || true
+        return 124
+    fi
+    wait_status=0
+    wait "$probe_pid" 2>/dev/null || wait_status=$?
+    return "$wait_status"
+}
+
+codex_cli_version() {
+    local cli_path="$1"
+    local output version probe_file probe_status
+
+    probe_file="$(mktemp "${TMPDIR:-/tmp}/codex-cli-version.XXXXXX")" || return 1
+    probe_status=0
+    codex_cli_version_probe "$probe_file" "$cli_path" --version || probe_status=$?
+    if [ "$probe_status" -ne 0 ]; then
+        [ "$probe_status" -ne 124 ] || {
+            rm -f "$probe_file"
+            return 1
+        }
+        probe_status=0
+        codex_cli_version_probe "$probe_file" "$cli_path" version || probe_status=$?
+        [ "$probe_status" -eq 0 ] || {
+            rm -f "$probe_file"
+            return 1
+        }
+    fi
+    output="$(cat "$probe_file" 2>/dev/null || true)"
+    rm -f "$probe_file"
+    version="$(printf '%s\n' "$output" | sed -nE 's/.*(^|[[:space:]])v?([0-9]+([.][0-9]+)+(-[0-9A-Za-z-]+([.][0-9A-Za-z-]+)*)?([+][0-9A-Za-z-]+([.][0-9A-Za-z-]+)*)?)($|[^0-9A-Za-z._+-]).*/\2/p' | head -n 1)"
+    [ -n "$version" ] || return 1
+    echo "$version"
+}
+
+codex_cli_version_core() {
+    local version="${1%%+*}"
+    echo "${version%%-*}"
+}
+
+codex_cli_version_prerelease() {
+    local version="${1%%+*}"
+
+    case "$version" in
+        *-*) echo "${version#*-}" ;;
+        *) echo "" ;;
+    esac
+}
+
+codex_cli_version_core_compare() {
+    local left="$1"
+    local right="$2"
+    local left_core right_core left_part right_part i max
+    local -a left_parts right_parts
+
+    left_core="$(codex_cli_version_core "$left")"
+    right_core="$(codex_cli_version_core "$right")"
+    IFS=. read -r -a left_parts <<< "$left_core"
+    IFS=. read -r -a right_parts <<< "$right_core"
+    max="${#left_parts[@]}"
+    [ "${#right_parts[@]}" -gt "$max" ] && max="${#right_parts[@]}"
+
+    for ((i = 0; i < max; i++)); do
+        left_part="${left_parts[$i]:-0}"
+        right_part="${right_parts[$i]:-0}"
+        [[ "$left_part" =~ ^[0-9]+$ ]] || return 1
+        [[ "$right_part" =~ ^[0-9]+$ ]] || return 1
+        if ((10#$left_part > 10#$right_part)); then
+            echo 1
+            return 0
+        fi
+        if ((10#$left_part < 10#$right_part)); then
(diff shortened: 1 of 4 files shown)


launcher/start.sh.template
```diff
@@ -1362,11 +1362,79 @@ resolve_notification_icon() {
...
+
+    probe_file="$(mktemp "${TMPDIR:-/tmp}/codex-cli-version.XXXXXX")" || return 1
+    probe_status=0
+    codex_cli_version_probe "$probe_file" "$cli_path" --version || probe_status=$?
+    if [ "$probe_status" -ne 0 ]; then
+        if [ "$probe_status" -ne 124 ]; then
+            codex_cli_version_probe "$probe_file" "$cli_path" version || true
+        fi
+    fi
+    output="$(cat "$probe_file" 2>/dev/null || true)"
```
avifenesh, line 1416: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This still parses whatever the probe wrote even when `codex_cli_version_probe` returned `124`. A broken candidate that prints a high version and then hangs (for example, `echo 'codex-cli 9.999.0'; sleep 30`) will be killed by the timeout but then selected as the newest CLI from this partial output, defeating the “skip a blocked candidate” behavior. Please discard the output/return failure on the timeout path before parsing it.

avifenesh (commented):

---

I found one resolver edge case where a timed-out version probe can still win selection.

launcher/start.sh.template
Yo-DDV: Hi, thanks for catching that.

I updated the resolver so timed-out probes are treated as failed before their output is parsed. I also added smoke coverage for both cases: a partial `--version` response that hangs and the same pattern through the `version` fallback.

ilysenko:
Thanks for working on this. The direction makes sense: asking users to force-install the latest CLI is not enough when the GUI session can still resolve an older `/usr/bin/codex` before a newer user install.

There is one blocker before merge:

- `codex_cli_version()` still parses the probe output after `codex_cli_version_probe` returns `124`. A broken candidate can print a high version and then hang, for example `echo "codex-cli 9.999.0"; sleep 30`, and still be selected as the newest CLI after the timeout. Please discard probe output on timeout and add a regression test for that exact case.

One non-blocking follow-up: the launcher now probes the selected CLI again only to log the version. If possible, avoid that extra probe by carrying the selected version out of the resolver, or keep the logging best-effort so startup cost stays predictable.

launcher/start.sh.template
```diff
@@ -1362,11 +1362,86 @@ resolve_notification_icon() {
...
+    [ -n "$version" ] || return 1
+    echo "$version"
+}
+
+codex_cli_version_gte() {
+    local left="$1"
+    local right="$2"
+
+    [ "$left" = "$right" ] && return 0
+    [ "$(printf '%s\n%s\n' "$left" "$right" | sort -V | tail -n 1)" = "$left" ]
```

---

This comparison admits prerelease-style versions (the extractor above allows `-`/alphanumeric segments), but `sort -V` orders `0.142.0` before `0.142.0-alpha.1`, so `tail -n 1` treats the prerelease as newer than the stable release. Since `find_codex_cli` uses this helper to choose `best_candidate`, a local `0.142.0-alpha.1` can win over `0.142.0`, which is the opposite of SemVer precedence (and the updater's Rust side uses `semver::Version` for installed/latest comparisons). Please use a SemVer-aware comparison here or reject/normalize prerelease suffixes before sorting.

avifenesh (commented):

---

Found one resolver comparison edge case that can pick the wrong CLI when prerelease versions are present.

ilysenko:
Thanks for digging into this. After looking at the resolver behavior more closely, this may be better handled with a simpler policy.

The launcher probably should not try to choose the “newest” `codex` binary automatically. Users can install or wrap the CLI in many ways: npm, standalone, nvm, bun, pnpm, Nix, distro packages, custom scripts, or paths outside the common locations. Once the launcher starts ranking candidates, it becomes responsible for guessing user intent across all of those setups, and that is likely to create long-term support edge cases.

A narrower fix would be easier to maintain:

- Keep `CODEX_CLI_PATH` as the explicit source of truth when users need a specific CLI.
- Keep normal PATH lookup as a fallback instead of trying to rank all candidates.
- Log the resolved CLI path and version at startup.
- If the resolved CLI is clearly too old or known-incompatible, show a warning that points users to set `CODEX_CLI_PATH=/path/to/codex`.
- Document this in README/troubleshooting.

That still fixes the main diagnosability problem from #672 without turning the launcher into a CLI version selector.

launcher/start.sh.template
Yo-DDV: Hi, thanks for the extra catch.

I pushed a follow-up that removes `sort -V` from the resolver comparison and compares the numeric core version plus prerelease precedence explicitly. The smoke coverage now includes `0.142.0-alpha.1` vs `0.142.0`, a newer prerelease core, the partial-timeout cases, and the fallback `version` path.

I also carried the selected version out of the resolver, so the selected CLI is not probed again just for logging.

ilysenko:
I opened a narrower PR implementing the simpler policy from review: no automatic newest-CLI selection, `CODEX_CLI_PATH` remains the override, launcher logs resolved path/version. Superseding PR: https://github.com/ilysenko/codex-desktop-linux/pull/674

Yo-DDV:
Hi @ilysenko, thanks for opening #674.

I agree the narrower policy is a better fit than this branch now that the launcher should not rank installed CLIs. I am closing this PR to avoid keeping two competing approaches open, and I will follow the replacement PR there.

PR #674, opened 2026-07-05
Simplify launcher CLI resolution policy

ilysenko:
## Summary
- Keep launcher CLI resolution simple: no newest-CLI ranking.
- Keep `CODEX_CLI_PATH` as the explicit override, then use the existing PATH/nvm/known-location fallback order.
- Log the resolved CLI path plus a bounded best-effort version probe, including a pinning hint when the version is unknown.
- Document the policy in README/updater docs.

## Tests/checks run
- `make check`

## Notes
Supersedes the broader newest-CLI ranking approach in #673 with the narrower policy requested during review.

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -1408,24 +1408,45 @@ find_codex_cli() {
     return 1
 }
 
+pid_parent_matches() {
+    local pid="$1"
+    local expected_parent="$2"
+    local stat_line
+    local rest
+
+    [ -r "/proc/$pid/stat" ] || return 1
+    stat_line=$(< "/proc/$pid/stat") || return 1
+    rest=${stat_line##*) }
+    set -- $rest
+    # $1 = state, $2 = ppid
+    [ "${2:-}" = "$expected_parent" ]
+}
+
 codex_cli_version_probe() {
     local output_file="$1"
     shift
     local marker="${output_file}.timed-out"
     local probe_pid
     local watchdog_pid
+    local self_pid=$$
     local status=0
 
     rm -f "$marker"
-    "$@" >"$output_file" 2>/dev/null &
+    ( trap - EXIT
+      { exec 9>&-; } 2>/dev/null || true
+      exec "$@" >"$output_file" 2>/dev/null
+    ) &
     probe_pid="$!"
     ( trap - EXIT
+      { exec 9>&-; } 2>/dev/null || true
       sleep 1
-      if kill -0 "$probe_pid" 2>/dev/null; then
+      if pid_parent_matches "$probe_pid" "$self_pid"; then
           : > "$marker"
           kill "$probe_pid" 2>/dev/null || true
           sleep 0.05
-          kill -9 "$probe_pid" 2>/dev/null || true
+          if pid_parent_matches "$probe_pid" "$self_pid"; then
+              kill -9 "$probe_pid" 2>/dev/null || true
+          fi
       fi
     ) &
     watchdog_pid="$!"
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -1408,6 +1408,98 @@ find_codex_cli() {
     return 1
 }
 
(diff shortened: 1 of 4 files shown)


launcher/start.sh.template
```diff
@@ -1408,6 +1408,98 @@ find_codex_cli() {
...
+    shift
+    local marker="${output_file}.timed-out"
+    local probe_pid
+    local watchdog_pid
+    local status=0
+
+    rm -f "$marker"
+    "$@" >"$output_file" 2>/dev/null &
+    probe_pid="$!"
+    ( trap - EXIT
```

---

This probe is reached after the cold-start path has acquired `launcher.lock` (the `acquire_launcher_lock` block near the end of the file runs before `log_codex_cli_path`). Both the CLI process and the watchdog subshell therefore inherit fd 9, even though the launcher explicitly documents that fd 9 must not survive into child processes and closes it before exec'ing Electron. A pinned/wrapper CLI that starts a background child can keep the flock alive after `release_launcher_lock`, making later launches block or time out. Please close fd 9 in these spawned subshells (or run the version probe before taking the launcher lock).

launcher/start.sh.template
```diff
@@ -1408,6 +1408,98 @@ find_codex_cli() {
...
+      sleep 1
+      if kill -0 "$probe_pid" 2>/dev/null; then
+          : > "$marker"
+          kill "$probe_pid" 2>/dev/null || true
+          sleep 0.05
+          kill -9 "$probe_pid" 2>/dev/null || true
```

---

This watchdog sends SIGTERM/SIGKILL based only on `kill -0` after a sleep. If the probe exits just as the watchdog wakes, the parent can reap it and the PID can be recycled before these kills run, so an unrelated same-user process could be signalled. The existing `/dev/tcp` watchdog below this function guards the same race by checking `/proc/$probe_pid/stat` still reports this launcher as the parent before killing; the version probe should carry over that guard.

avifenesh (commented):

---

I found two correctness issues in the new launcher-side CLI version probe.

Yo-DDV (commented):
Hi @ilysenko, I checked #674 because it supersedes #673. The simpler policy looks like the right direction to me.

For the fd 9 case, I could reproduce the issue locally with a wrapper CLI that starts a background child. After the parent launcher closes fd 9, the child still has `/proc/<pid>/fd/9` and a second `flock -n launcher.lock` fails until that child exits, so the first review point looks valid.

The `/proc/$probe_pid/stat` parent check from `webview_port_is_open()` also looks worth carrying over before killing the version probe. In this branch the version watchdog still uses `kill -0` only, while the webview watchdog already guards PID reuse by checking the probe's parent.

I did not find another blocker around the simplified lookup/docs path. Once those two probe details are tightened, this seems cleaner than maintaining automatic newest-CLI selection.

launcher/start.sh.template
```diff
@@ -1408,6 +1408,119 @@ find_codex_cli() {
...
+    [ "${2:-}" = "$expected_parent" ]
+}
+
+codex_cli_version_probe() {
+    local output_file="$1"
+    shift
+    local marker="${output_file}.timed-out"
+    local probe_pid
+    local watchdog_pid
+    local self_pid=$$
```

---

`codex_cli_version_probe` is reached through `log_codex_cli_path`, which captures `codex_cli_version` with command substitution (`version="$(codex_cli_version ...)"` below). In Bash, `$$` inside that command-substitution subshell is still the original launcher PID, while the probe process is parented by the substitution shell (the value exposed as `$BASHPID`). That makes `pid_parent_matches "$probe_pid" "$self_pid"` false on the real logging path, so the watchdog never marks/kills a hung `codex --version` probe and the subsequent `wait "$probe_pid"` can block the launcher while it is holding `launcher.lock`. Use the current shell PID for the guard (e.g. `$BASHPID` in this Bash launcher) or avoid running the probe under command substitution.

avifenesh (commented):

---

I found one remaining correctness issue in the updated CLI version watchdog path.

PR #676, opened 2026-07-05
Add HiDPI scaling overrides and --diagnose-scaling to the Linux launcher

djvdorp:
## What changed and why

On HiDPI / fractional-scaling setups (reported on Ubuntu GNOME under Wayland), the Codex UI can render massively oversized — or tiny and blurry — inside its window while every other app scales normally. Chromium computes its device scale factor differently per Ozone backend: native Wayland uses the compositor's monitor scale, while X11/XWayland derives it from `Xft.dpi / 96`, `GDK_SCALE`, and `GDK_DPI_SCALE`. With GNOME fractional scaling (`scale-monitor-framebuffer`) or GNOME 47+ `xwayland-native-scaling` enabled, the compositor scales the window buffer *and* Chromium applies its own scale on top, so the UI ends up double-scaled. The launcher had no clean knob to correct this short of hand-editing flags.

This PR adds session-level overrides plus a diagnostics command. I didn't find an existing issue for this (searched for scaling/HiDPI/oversized/dpi); happy to file one with full reproduction details if you'd like it tracked separately.

**Defaults are unchanged** — both env vars are inert unless explicitly set, so working DPI setups on every desktop environment behave exactly as before. What's new:

- `CODEX_OZONE_PLATFORM=x11|wayland|auto` — selects the Ozone backend per launch. Explicit `--x11` / `--wayland` / `--ozone-platform*` flags still win (a notice is logged when the env var is ignored); invalid values warn and keep the default `auto` hint. The only behavioral nuance: `CODEX_OZONE_PLATFORM=auto` overrides the ChromeOS/Sommelier X11 fallback (#95) — intentionally, and only when an operator explicitly sets it.
- `CODEX_FORCE_DEVICE_SCALE_FACTOR=N` — passes a validated `--force-device-scale-factor=N` to Electron (e.g. `1`, `1.25`, `2`); a user-supplied flag wins over the env var, and invalid/zero values are rejected with a warning.
- `./start.sh --diagnose-scaling` — prints session type, scaling-related env vars, GNOME `scaling-factor` / `text-scaling-factor` / mutter experimental features, `Xft.dpi`, monitor layout, Electron version, and the exact computed launch flags, then exits. Output goes to the terminal instead of the launcher log (this required restoring fd 2, which the early-stderr capture leaves on `/dev/null` when the log redirect is skipped).
- `electron-flags.conf` template and `--help` document the new options; launch-mode log lines now record `forced_scale=` for triage.

One-line workarounds this enables:

    CODEX_FORCE_DEVICE_SCALE_FACTOR=1 codex-desktop
    CODEX_OZONE_PLATFORM=x11 codex-desktop

Why core launcher rather than a `linux-features/` module: this is required Linux compatibility glue for the standard upstream UI on common GNOME configurations (same category as the existing `--x11`/`--wayland`/`CODEX_LINUX_RENDERING_MODE` handling), and it's a no-op unless the user opts in via env var or flag.

- `launcher/start.sh.template` — env overrides, `--force-device-scale-factor` arg tracking, `--diagnose-scaling`, help/flags-template text
- `tests/scripts_smoke.sh` — 10 new launcher-probe assertions (both env vars, precedence over defaults and the Sommelier fallback, explicit flags winning, invalid/zero values ignored)
- `docs/troubleshooting.md`, `README.md` — new "Oversized or blurry UI (HiDPI / fractional scaling)" section (root cause, one-liners, persistent flags, `.desktop` `Exec=env …` editing, Ubuntu GNOME notes) and table rows

The generated `codex-app/start.sh` was regenerated locally only to validate; it is not committed.

## Impact on supported paths

No packaging, installer, Nix, or updater changes are needed — every path consumes the template as-is:

- native packages: `/usr/bin/codex-desktop` execs `start.sh "$@"` (`write_launcher_stub`), so env vars and the new option pass through on `.deb`/`.rpm`/pacman
- AppImage: `AppRun` execs `start.sh "$@"`
- Nix: builds the same generated launcher
- updater rebuilds: the update-builder bundle already copies `launcher/start.sh.template` (`stage_update_builder_bundle`)
- updater crate untouched → no version bump per the versioning rules

## How it was tested

Smoke-test assertions were added alongside the change and exercised via the existing launcher rendering probe. On a 4K X11 GNOME machine (Xft.dpi 192), with `codex-app/start.sh` regenerated from the template and Electron's real scale read via `--remote-debugging-port` + CDP:

- baseline launch: `devicePixelRatio: 2` (correct, unchanged defaults; `ozone_hint=auto`, `forced_scale=none` logged)
- `CODEX_FORCE_DEVICE_SCALE_FACTOR=1`: `devicePixelRatio: 1`, `forced_scale=1` logged
- `CODEX_OZONE_PLATFORM=x11`: launches with `--ozone-platform=x11`, correct scale
- webview serves HTTP 200, app-server traffic flows, warm-start IPC handoff still works; `--help` and `--diagnose-scaling` behave as documented
- generated launcher inspected after regeneration

Validation commands run: `bash -n launcher/start.sh.template` and `bash tests/scripts_smoke.sh` (pass, including after rebasing onto current `main` with the patch-architecture refactor). Not run because those areas are untouched: `cargo check`/`cargo test -p codex-update-manager` and the `.deb`/`.rpm`/pacman package builds (wrapper pass-through verified in `scripts/lib/package-common.sh` instead).

## Limitations and follow-ups

- I could not reproduce the Wayland fractional-scaling double-scale on this X11 box; both workarounds are proven to reach Electron and take effect, and `--diagnose-scaling` output from affected GNOME Wayland machines will show which backend misbehaves.
- Possible follow-up (intentionally out of scope): once affected users share `--diagnose-scaling` output, consider a smarter default for fractional-scaling GNOME Wayland sessions instead of an opt-in override.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,7 @@ not download or extract the DMG themselves. See
 | Blank window or splash stuck | Check `~/.cache/codex-desktop/launcher.log` and whether port `5175` is already in use |
 | `CODEX_CLI_PATH` or CLI install error | Check `~/.cache/codex-desktop/launcher.log`, set `CODEX_CLI_PATH=/path/to/codex` to pin a binary, or install `@openai/codex` manually |
 | Wayland / GPU / Vulkan hang | Try `CODEX_LINUX_RENDERING_MODE=wayland-gpu ./codex-app/start.sh` or persistent launch flags |
+| UI oversized or blurry (HiDPI / fractional scaling) | Try `CODEX_FORCE_DEVICE_SCALE_FACTOR=1 ./codex-app/start.sh` or `CODEX_OZONE_PLATFORM=x11 ./codex-app/start.sh`; see `./codex-app/start.sh --diagnose-scaling` |
 | Resize ghosting or stale frame trails | Try `CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=1 ./codex-app/start.sh` or `--disable-gpu-compositing` |
 | Computer Use UI is hidden | Enable the UI opt-in; account/server rollouts may still hide upstream-gated parts |
 | Computer Use has no input backend | Check `/dev/uinput`, portal support, or `ydotoold` / `ydotool.service` |
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -11,6 +11,7 @@
 | `gh auth status` works in terminal but fails inside Codex Desktop | See [GitHub CLI auth in app-launched shells](github-cli-auth.md) |
 | Electron hangs while CLI is outdated | Re-run the launcher and check `~/.cache/codex-desktop/launcher.log` plus `~/.local/state/codex-update-manager/service.log` |
 | GPU / Vulkan / Wayland errors | Try `CODEX_LINUX_RENDERING_MODE=wayland-gpu ./codex-app/start.sh` or persistent launch flags below |
+| UI massively oversized, tiny, or blurry | See [Oversized or blurry UI](#oversized-or-blurry-ui-hidpi--fractional-scaling); quick fix: `CODEX_FORCE_DEVICE_SCALE_FACTOR=1 ./codex-app/start.sh` |
 | Window flickering, resize ghosting, or stale frame trails | Try `CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=1 ./codex-app/start.sh`, then `./codex-app/start.sh --disable-gpu` if needed |
 | Transparent or dark left sidebar | Check whether the Linux opaque-window patch was applied, then rebuild with a current checkout |
 | Sandbox errors | The launcher already sets `--no-sandbox` |
@@ -52,6 +53,77 @@ For native Wayland IME setups, try:
 
 Restart Codex Desktop after changing this file. Warm-start launches reuse the
 running Electron process and will not pick up new flags.
+
+## Oversized Or Blurry UI (HiDPI / Fractional Scaling)
+
+If the whole Codex UI renders far too large (or too small/blurry) inside its
+window while other apps scale normally, Electron picked a wrong device scale
+factor for your display setup. Chromium computes the scale differently per
+backend: under native Wayland it uses the compositor's monitor scale, while
+under X11/XWayland it derives the scale from `Xft.dpi` (dpi / 96), `GDK_SCALE`,
+and `GDK_DPI_SCALE`. On GNOME Wayland sessions with fractional scaling or
+XWayland native scaling enabled, those two views can disagree — the compositor
+scales the window buffer and Chromium applies its own scale on top, so the UI
+ends up double-scaled (oversized) or unscaled (tiny/blurry).
+
+First inspect what the launcher and your session report:
+
+```bash
+./codex-app/start.sh --diagnose-scaling          # local build
+/opt/codex-desktop/start.sh --diagnose-scaling   # native package install
+```
+
+It prints the session type, scaling-related environment variables, GNOME
+`scaling-factor` / `text-scaling-factor`, `Xft.dpi`, monitor layout, and the
+exact Electron flags a launch would use.
+
+One-line workarounds (quit the app fully first — warm starts reuse the
+running process and ignore new flags):
+
+```bash
+# Force a specific device scale factor (1 = unscaled; 1.5, 2, ... also work)
+CODEX_FORCE_DEVICE_SCALE_FACTOR=1 codex-desktop
+
+# Force the X11/XWayland backend instead of native Wayland
+CODEX_OZONE_PLATFORM=x11 codex-desktop
+
+# Force native Wayland (best for fractional scaling on current GNOME)
+CODEX_OZONE_PLATFORM=wayland codex-desktop
+```
+
+For a local self-build, replace `codex-desktop` with `./codex-app/start.sh`.
+Explicit launcher flags (`--x11`, `--wayland`, `--ozone-platform=*`,
+`--force-device-scale-factor=*`) always win over these environment variables.
+
+To make the fix persistent, uncomment the matching flag in
+`~/.config/codex-desktop/electron-flags.conf`:
+
+```text
+--force-device-scale-factor=1
+```
+
+or edit the desktop launcher: copy
+`/usr/share/applications/codex-desktop.desktop` to
+`~/.local/share/applications/` and prepend the variable to the `Exec` line:
+
+```text
+Exec=env CODEX_FORCE_DEVICE_SCALE_FACTOR=1 BAMF_DESKTOP_FILE_HINT=... /usr/bin/codex-desktop %u
+```
+
+Ubuntu GNOME notes:
+
+- With plain 100% or 200% scaling in Settings → Displays, the defaults work;
+  do not force anything.
+- Fractional scaling (125%/150%/175%) is a Mutter experimental feature
+  (`scale-monitor-framebuffer`). If the UI is oversized there, try
+  `CODEX_OZONE_PLATFORM=wayland` first; if it is blurry instead, try
+  `CODEX_FORCE_DEVICE_SCALE_FACTOR` matching your monitor scale (e.g. `1.5`).
+- On GNOME 47+ the `xwayland-native-scaling` experimental feature changes how
+  XWayland apps are scaled; if you enabled it and Codex looks double-scaled
+  under `CODEX_OZONE_PLATFORM=x11`, either disable that feature or run the
+  app on native Wayland.
+- "Large Text" (accessibility) only sets `text-scaling-factor` and affects
+  fonts, not the window scale; `--diagnose-scaling` shows both values.
 
 ## Authenticated HTTP Proxies
 
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -229,12 +229,19 @@ Options:
   --enable-gpu                Re-enable Electron GPU acceleration
   --x11                       Force X11/XWayland
   --wayland                   Force native Wayland
+  --diagnose-scaling          Print display/session scaling diagnostics and exit
 
 Default launch keeps Electron GPU enabled and lets Electron choose the platform.
 Extra flags are passed directly to Electron.
 Persistent launch flags: $USER_ELECTRON_FLAGS_FILE
 
 Environment:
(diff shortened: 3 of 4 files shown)


Yo-DDV:
Hi Daniel, thanks for working on this. I found one edge case around the launcher feature extension point.

The static `runtimeHooks.electronArgs` path seems fine, but `runtimeHooks.launcher` can add explicit Electron args after `set_electron_defaults()` has already applied `CODEX_OZONE_PLATFORM` or `CODEX_FORCE_DEVICE_SCALE_FACTOR`. That can leave conflicting flags in the final Electron argv.

I reproduced it on this PR head with a launcher hook that prints `electron-arg --ozone-platform=x11` while `CODEX_OZONE_PLATFORM=wayland`; the final args include both `--ozone-platform=wayland` and `--ozone-platform=x11`. Same pattern with `CODEX_FORCE_DEVICE_SCALE_FACTOR=1` plus `electron-arg --force-device-scale-factor=2`.

Since `runtimeHooks.launcher` is documented as running before final Electron launch args are built, I think the env-derived values should be cleared or suppressed after launcher output is scanned, and the smoke probe should add regression coverage for those two combinations.

Everything else I checked looked good: upstream CI is green, and locally `bash -n launcher/start.sh.template tests/scripts_smoke.sh`, `node --test scripts/patch-linux-window-ui.test.js`, `bash tests/scripts_smoke.sh`, `make check`, and `make test` all pass.

ilysenko:
Hi Daniel, thanks again for the focused HiDPI work. I rechecked the current head and the runtimeHooks.launcher precedence issue is still reproducible.

A launcher hook that emits `electron-arg --ozone-platform=x11` while `CODEX_OZONE_PLATFORM=wayland` leaves both `--ozone-platform=wayland` and `--ozone-platform=x11` in the final Electron argv. The same happens with `CODEX_FORCE_DEVICE_SCALE_FACTOR=1` plus a hook-emitted `--force-device-scale-factor=2`.

Could you clear or suppress the env-derived launcher values after launcher hook output is scanned, and add smoke coverage for both late-hook cases?

djvdorp:
Thanks both — @ilysenko you were most likely still rechecking 041b6a5; the fix is now pushed as a49e855, so the current head should reproduce clean. Should address this properly now, also thanks @Yo-DDV for the quick feedback!

Root cause was ordering: runtimeHooks.launcher runs after set_electron_defaults() has already applied the CODEX_OZONE_PLATFORM / CODEX_FORCE_DEVICE_SCALE_FACTOR env overrides, so a hook-emitted --ozone-platform / --force-device-scale-factor duplicated the env-derived flag. The launcher now tracks whether those values are env-derived and, in build_electron_launch_args() (after the hooks are scanned), drops the env value once an explicit arg is present, logs a notice, and lets the hook's flag win — which also stops a cleared env Wayland platform from adding a stray WaylandWindowDecorations. Pass-through CLI args were already suppressed inside set_electron_defaults, so only the late-hook path changes.

Added smoke-probe regression coverage for both cases: CODEX_OZONE_PLATFORM=wayland + hook --ozone-platform=x11, and CODEX_FORCE_DEVICE_SCALE_FACTOR=1 + hook --force-device-scale-factor=2.

Validation (all green): bash -n launcher/start.sh.template tests/scripts_smoke.sh, node --test scripts/patch-linux-window-ui.test.js (335 pass), bash tests/scripts_smoke.sh, make check, make test (192 pass).

Yo-DDV:
Hi Daniel, thanks for the detailed follow-up.

I rechecked `a49e855`. The two env-derived cases now look fixed on my side: `CODEX_OZONE_PLATFORM=wayland` with a launcher hook emitting `--ozone-platform=x11` leaves only the hook arg, and `CODEX_FORCE_DEVICE_SCALE_FACTOR=1` with a hook-emitted `--force-device-scale-factor=2` behaves the same way.

There is still an adjacent late-hook case when the platform comes from a rendering profile or fallback instead of directly from `CODEX_OZONE_PLATFORM`.

With `CODEX_LINUX_RENDERING_MODE=wayland-gpu` plus a launcher hook emitting `--ozone-platform=x11`, the final argv still contains launcher args for Wayland, including `--ozone-platform=wayland` and `WaylandWindowDecorations`, while the hook adds `--ozone-platform=x11`.

I see the same x11/wayland conflict with `CODEX_LINUX_RENDERING_MODE=wslg` plus a hook-emitted `--ozone-platform=wayland`, and with the Sommelier fallback plus the same hook override.

So I think the suppression should cover launcher-computed Ozone defaults as well as env-derived defaults. The standard local checks are otherwise green here: `bash -n launcher/start.sh.template tests/scripts_smoke.sh`, `node --test scripts/patch-linux-window-ui.test.js`, and `bash tests/scripts_smoke.sh`.

djvdorp:
Good catch on the profile/fallback cases @Yo-DDV, apologies for not catching these earlier — should be fixed in 0040497.

Instead of special-casing each origin I generalized it: I enumerated every place the launcher sets ELECTRON_OZONE_PLATFORM (CODEX_OZONE_PLATFORM, the wayland-gpu/wslg profiles, the Sommelier fallback, and the explicit --x11/--wayland/--ozone-platform=/--safe-mode flags) and now drop the launcher-computed --ozone-platform in build_electron_launch_args() whenever an explicit --ozone-platform already reached the Electron args — the same ELECTRON_OZONE_SWITCH_IN_ARGS guard the --ozone-platform-hint emission already uses. That also clears the WaylandWindowDecorations rider, and the same treatment applies to --force-device-scale-factor. Pass-through CLI args are scanned before those defaults are computed and were already suppressed, so only the late-hook path changes and the hook flag wins cleanly.

Added smoke coverage for the hook override across all four platform sources (CODEX_OZONE_PLATFORM, wayland-gpu, wslg, Sommelier) plus the scale case. bash -n launcher/start.sh.template tests/scripts_smoke.sh, node --test scripts/patch-linux-window-ui.test.js, bash tests/scripts_smoke.sh, make check, and make test all pass.

Yo-DDV:
Hi Daniel, thanks for the quick update. I rechecked 0040497.

The launcher-computed cases now look fixed on my side: CODEX_OZONE_PLATFORM, wayland-gpu, wslg, Sommelier, and the env scale override all drop the computed value before the final argv. Local checks are green here too.

I found one adjacent case that still leaves duplicate final Electron args. If an explicit Electron arg is already in ELECTRON_ARGS and a runtimeHooks.launcher hook emits a conflicting arg later, apply_feature_launcher_hook_electron_arg() appends the hook arg without replacing the earlier one.

For example:
- start.sh -- --ozone-platform=x11 plus a launcher hook emitting electron-arg --ozone-platform=wayland ends with electron=<--ozone-platform=x11><--ozone-platform=wayland>.
- start.sh -- --force-device-scale-factor=1 plus a hook emitting electron-arg --force-device-scale-factor=2 ends with both scale flags.

I also see the same scale-factor duplication when the first scale flag comes from the persistent flags file or runtimeHooks.electronArgs.

So the new build_electron_launch_args() cleanup covers ELECTRON_OZONE_PLATFORM / ELECTRON_FORCED_SCALE_FACTOR, but not pre-existing entries already stored in ELECTRON_ARGS. I think the hook path should replace an earlier ozone/scale arg in ELECTRON_ARGS, or add smoke coverage if duplicate argv is considered intentional.

djvdorp:
Hi @Yo-DDV - should be fixed in 80589cd (new commit on top, so the delta is easy to re-review).

You're right that the build_electron_launch_args() cleanup only covered launcher-computed values, not conflicting entries already collected in ELECTRON_ARGS. Since hooks are the last source applied and Chromium resolves repeated switches last-wins, apply_feature_launcher_hook_electron_arg() now removes any earlier occurrence of the same switch before appending — generically for every switch, not just ozone/scale, with --ozone-platform / --ozone-platform-hint treated as one family (an explicit platform supersedes a hint). Each drop is logged. Effective Chromium behavior is unchanged; the argv just stops being contradictory. As a side effect this also covers a hook --use-gl superseding the WSLg profile's --use-gl=angle, and later hooks superseding earlier ones.

To be explicit about scope: duplicates among the static sources themselves (feature electron-args → flags file → CLI pass-through) keep their existing append-order last-wins behavior — the current smoke suite explicitly asserts that ordering (e.g. <--use-gl=angle><--use-gl=desktop>), so I treated it as intentional and didn't change it. Happy to revisit in a follow-up if you'd rather dedupe those too.

Added smoke coverage for hook-vs-pass-through ozone platform, hook-vs-pass-through ozone hint, and hook-vs-{pass-through, persistent flags file, feature electron-args} scale factor. bash -n launcher/start.sh.template tests/scripts_smoke.sh, node --test scripts/patch-linux-window-ui.test.js (335 pass), bash tests/scripts_smoke.sh, make check, and make test (192 pass) all green.

Yo-DDV (commented):
Thanks for the follow-up. I rechecked `80589cd`; the duplicate argv cases covered by the new smoke assertions now pass on my side, including hook-vs-pass-through platform and scale, persistent flags, and feature electron args.

I found one remaining adjacent case that still seems worth addressing. The Ozone platform/hint family is now deduped in the final Electron argv, but replacing an earlier `--ozone-platform=wayland` with a later hook-provided `--ozone-platform-hint=auto` can leave the internal effective-platform state from the removed platform.

Repro shape:
- `CODEX_LINUX_RENDERING_MODE=wayland-gpu`
- start with `-- --ozone-platform=wayland`
- launcher hook emits `electron-arg --ozone-platform-hint=auto`

The final argv is clean, `electron=<--ozone-platform-hint=auto>`, but `renderer_accessibility=0`. A direct `-- --ozone-platform-hint=auto` under the same profile gives `renderer_accessibility=1`, so the hook replacement path is not equivalent to the final argv it produces.

I think `remove_superseded_electron_args()` needs to clear or recompute `ELECTRON_OZONE_PLATFORM_ARG` when it drops `--ozone-platform`, or the scan state needs to be rebuilt after hook replacement. A small smoke case for hook-hint replacing platform should cover it.

For reference, the existing checks are green here: `bash -n launcher/start.sh.template tests/scripts_smoke.sh`, `node --test scripts/patch-linux-window-ui.test.js`, `bash tests/scripts_smoke.sh`, `make check`, and `make test`.

ilysenko (approved):
Reviewed current head 80589cd34153bea432435af09be3dec5bfd33f81; no blockers found.

PR #677, opened 2026-07-05
fix(remote-mobile): hydrate late app-server events

Yo-DDV:
Refs #639.

This fixes another stale remote mobile state case on Linux.

When the mobile view has lost the local conversation entry for a thread, the app-server can still send late final events such as `turn/completed` or `item/completed`. The current Linux bundle treats those as unknown conversations and drops them. On Android this can leave the thread looking like it is still working until the user leaves and reopens the conversation, because that forces a fresh thread read.

The patch keeps this recovery in the core Linux webview path, so it applies to standard installs as well as opt-in remote-mobile-control builds. For missing conversations it hydrates the thread, queues late notifications while the read is in flight, then replays them once the thread is available.

I also covered a partial queue state during review: if a late-event queue already exists but no hydration read is in flight, the next late event restarts the thread read instead of leaving the queue stuck.

Validation:

- `node --test --test-name-pattern "pending queue exists without an in-flight|restarts when a pending queue" scripts/patch-linux-window-ui.test.js linux-features/remote-mobile-control/test.js`: 2 pass / 0 fail
- `node --test scripts/patch-linux-window-ui.test.js`: 339 pass / 0 fail
- `node --test linux-features/remote-mobile-control/test.js`: 76 pass / 0 fail
- `node --test linux-features/*/test.js`: 390 pass / 0 fail
- `bash tests/scripts_smoke.sh`: OK
- `make check`: OK
- `make test`: 192 pass / 0 fail
- descriptor dry-run against a copy of the current installed remote conversation bundle: matched and applied


ilysenko (approved):
Reviewed current head. No blockers found; focused fail-soft Linux webview hydration fix with passing checks.

Issue #496, opened 2026-06-15
Right-clicking Codex title bar causes desktop input lockup on Zorin OS (X11)


kadsyy:
## Environment

* OS: Zorin OS 18.1
* Desktop: GNOME (X11 session)
* Installation method: `make bootstrap-native`

## Description

Right-clicking the Codex window title bar causes the desktop to become unresponsive to mouse clicks and keyboard input.

The mouse cursor continues to move and existing applications continue running (e.g. YouTube videos keep playing), but I cannot click anywhere on the desktop or interact with any windows. The issue occurs immediately when opening the title bar context menu. 

* Recovery requires switching to a TTY (`Ctrl+Alt+F3`) and terminating Codex.

## Additional Notes

The issue appears specific to Codex Desktop.

I tested the same title bar right-click behavior with other applications and could not reproduce the problem.

As a workaround, I can successfully make the window visible on all workspaces using:

```bash
wmctrl -i -r <window-id> -b add,sticky
```

This suggests the workspace/window manager functionality itself is working, but the title bar context menu interaction may be triggering an X11 input grab or related window-management issue.

ilysenko:
Thanks for the detailed report. A couple of questions to narrow this down:

1. Do you have any optional Linux features enabled, especially `frameless-titlebar`? If possible, please share the contents of `linux-features/features.json` from your checkout, or `/opt/codex-desktop/.codex-linux/linux-features-staged.json` from the installed app.

2. Does the same lockup happen if you focus the Codex window and press `Alt+Space` to open the window-manager menu, instead of right-clicking the title bar?

My current suspicion is an X11 input grab issue around the GNOME/Zorin titlebar context menu interacting with Codex's Linux titlebar overlay, but the feature/config check should tell us whether this is the default titlebar path or the optional frameless path.

kadsyy:
> Thanks for the detailed report. A couple of questions to narrow this down:
> 
>     1. Do you have any optional Linux features enabled, especially `frameless-titlebar`? If possible, please share the contents of `linux-features/features.json` from your checkout, or `/opt/codex-desktop/.codex-linux/linux-features-staged.json` from the installed app.
> 
>     2. Does the same lockup happen if you focus the Codex window and press `Alt+Space` to open the window-manager menu, instead of right-clicking the title bar?
> 
> 
[...]

PR #678, opened 2026-07-06
docs: add GNOME/X11 titlebar lockup troubleshooting

Yo-DDV:
Hi,

This adds a troubleshooting path for the GNOME/X11 title-bar right-click lockup reports in #496 and #649.

The patch does not change the default window behavior. It documents the immediate recovery path with `Esc` / `Alt+Space`, points users at the existing opt-in `frameless-titlebar` feature for a targeted retest, and lists the environment and build details needed when reporting whether the issue still reproduces.

I kept this as documentation because my local Zorin GNOME/X11 repro setup could trigger the title-bar menu/input grab, but it did not reproduce the persistent desktop-wide lock reported on Ubuntu 22.04/X11. Physical clicks outside the menu recovered in one or two clicks locally, and `Esc` always recovered focus. So I do not think a default titlebar behavior change is justified from my side yet.

Validation:
- `node --test linux-features/frameless-titlebar/test.js` - 5 pass / 0 fail
- `node --test scripts/patch-linux-window-ui.test.js` - 339 pass / 0 fail
- `make check`
- `make test` - 192 pass / 0 fail
- GNOME/X11 local retest on Zorin 18.1 / GNOME Shell 46.0 / X11:
  - titlebar menu/input grab observed
  - persistent desktop-wide lock not reproduced
  - `Esc` recovered focus in every run
  - physical clicks outside the menu recovered in one or two clicks
  - no leftover nested GNOME/X11 test processes after cleanup

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -12,6 +12,7 @@
 | Electron hangs while CLI is outdated | Re-run the launcher and check `~/.cache/codex-desktop/launcher.log` plus `~/.local/state/codex-update-manager/service.log` |
 | GPU / Vulkan / Wayland errors | Try `CODEX_LINUX_RENDERING_MODE=wayland-gpu ./codex-app/start.sh` or persistent launch flags below |
 | Window flickering, resize ghosting, or stale frame trails | Try `CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=1 ./codex-app/start.sh`, then `./codex-app/start.sh --disable-gpu` if needed |
+| Right-clicking the title bar leaves GNOME/X11 input stuck | Press `Esc` first, or use `Alt+Space` for the window menu. If the lockup is repeatable, test the optional `frameless-titlebar` feature below and include your distro, GNOME version, X11/Wayland session, package method, and `.codex-linux/linux-features-staged.json` when reporting it |
 | Transparent or dark left sidebar | Check whether the Linux opaque-window patch was applied, then rebuild with a current checkout |
 | Sandbox errors | The launcher already sets `--no-sandbox` |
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Docs-only change looks good after resolving the current main conflict.

docs/troubleshooting.md
```diff
@@ -127,6 +128,49 @@ Ubuntu GNOME notes:
...
+On some GNOME/X11 setups, right-clicking the Codex title bar can leave the
+desktop input focused on the window-manager menu. Try `Esc` first. `Alt+Space`
+opens the same window menu through the keyboard path and can be a safer
+workaround while debugging.
+
+If the issue is repeatable, verify whether the installed app is using optional
+Linux features:
+
+```bash
+if [ -f /opt/codex-desktop/.codex-linux/linux-features-staged.json ]; then
```
avifenesh, line 142: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This manifest doesn't actually verify which optional feature IDs are enabled. `writeStagedFeatureManifest` only writes the declarative staging plan (`resources` and `runtimeHooks`) into `.codex-linux/linux-features-staged.json` (`scripts/lib/linux-features.js` lines 653-669), while `frameless-titlebar/feature.json` only has `entrypoints.patchDescriptors`, so enabling the exact diagnostic feature still leaves this manifest with no mention of it. The enabled IDs are recorded in build info under `linuxFeatures.enabled` (`scripts/lib/build-info.js` lines 311-313); can this snippet read `.codex-linux/build-info.json` / `resources/codex-linux-build-info.json` instead, or make clear that this file only reports declarative staged artifacts?

avifenesh (commented):

---

Found one docs accuracy issue around which file actually records enabled feature IDs.

PR #679, opened 2026-07-06
Launcher performance: gate --disable-dev-shm-usage and renderer accessibility on actual need

djvdorp:
## What changed

- `--disable-dev-shm-usage` is now passed only when `/dev/shm` is missing, not writable,
  or below 1 GiB (the container case the flag exists for — Docker defaults to 64 MiB).
  New override: `CODEX_ELECTRON_DISABLE_DEV_SHM_USAGE=auto|0|1`, mirroring
  `CODEX_ELECTRON_DISABLE_GPU_COMPOSITING`.
- `--force-renderer-accessibility` is now added only when an assistive technology is
  detected (Orca/brltty running, the GNOME screen-reader setting, or the
  `GNOME_ACCESSIBILITY` / `QT_LINUX_ACCESSIBILITY_ALWAYS_ON` / `ACCESSIBILITY_ENABLED`
  env markers). `CODEX_FORCE_RENDERER_ACCESSIBILITY=1|0` still wins in both directions.
- Launch-mode diagnostics now log `dev_shm_usage_disabled=`, and a new decision record
  `docs/launcher-performance.md` documents these defaults plus the alternatives that were
  reviewed and deliberately left unchanged.

## Why

A side-by-side comparison with another Electron 42 app on the same X11 GNOME 4K host
showed every Codex process mapping Chromium shared memory from disk-backed
`/tmp/.org.chromium.*` (the other app maps RAM-backed `/dev/shm`), and the Chromium
accessibility engine forced on with no assistive technology present, which serializes the
accessibility tree on every DOM update. The WSLg and wayland-gpu profiles already skip
forced accessibility for performance; this extends the same policy behind detection.

- `tests/scripts_smoke.sh`
- `docs/troubleshooting.md`, `docs/launcher-performance.md` (new)
- `README.md`, `AGENTS.md`, `CHANGELOG.md`

## How it was tested

- TDD via the launcher rendering probe in `tests/scripts_smoke.sh`; new tests were
  written first and confirmed failing:
  - `df` stubs for large / small / unreadable `/dev/shm`
  - PATH stubs for Orca and `gsettings` screen-reader detection
  - env-marker detection cases
  - `1` / `0` / invalid override values across the default, wayland-gpu, and WSLg profiles
- Full `bash tests/scripts_smoke.sh` passes.
- `bash -n` on `install.sh`, `scripts/lib/*.sh`, the package builders, and the template.
- `codex-app/start.sh` regenerated from the template and probed on a real 4K X11 host:
  both flags are correctly omitted there, and all overrides are honored.

## Limitations

- No Rust or package-payload changes, so `cargo check`/`cargo test` and the package
  builds were not run.

## Known risks / follow-ups

- An assistive technology the detection misses (or one started after the app launches)
  needs `CODEX_FORCE_RENDERER_ACCESSIBILITY=1`; this is documented in troubleshooting
  and worth a reviewer's eye.
- Hosts with `/dev/shm` between 64 MiB and 1 GiB keep the old conservative behavior.
- Intentionally out-of-scope items (sandbox flags, Wayland compositing workaround,
  webview server model, startup ordering, upstream RPC latency) are recorded in
  `docs/launcher-performance.md`.

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -2791,29 +2791,27 @@ assistive_technology_detected() {
 # AT-SPI state that `codex-computer-use-linux setup` enables (org.a11y.Status
 # IsEnabled via busctl, with org.gnome.desktop.interface toolkit-accessibility
 # as its gsettings fallback). gsettings/busctl can hang on a broken session
-# bus, so the probes run in a subshell killed by the same ppid-guarded
-# watchdog webview_port_is_open uses (/usr/bin/timeout intentionally avoided;
-# see that function), capped at 0.5 s. A killed or failing probe counts as
-# "not detected" and never blocks launch.
-session_bus_assistive_technology_detected() {
-    local probe_pid kill_pid rc=1 self_pid=$$
+# bus, so each external helper runs under a watchdog capped at 0.5 s. When
+# setsid is available the watchdog kills the helper process group, which keeps
+# shell stubs and helper descendants from surviving launch. A killed or failing
+# probe counts as "not detected" and never blocks launch.
+SESSION_BUS_PROBE_OUTPUT=""
 
-    ( trap - EXIT
-      if command -v gsettings >/dev/null 2>&1; then
-          case "$(gsettings get org.gnome.desktop.a11y.applications screen-reader-enabled 2>/dev/null || true)" in
-              true) exit 0 ;;
-          esac
-          case "$(gsettings get org.gnome.desktop.interface toolkit-accessibility 2>/dev/null || true)" in
-              true) exit 0 ;;
-          esac
-      fi
-      if command -v busctl >/dev/null 2>&1; then
-          case "$(busctl --user get-property org.a11y.Bus /org/a11y/bus org.a11y.Status IsEnabled 2>/dev/null || true)" in
-              "b true") exit 0 ;;
-          esac
-      fi
-      exit 1 ) 2>/dev/null &
+session_bus_probe_command() {
+    local output_file probe_pid kill_pid rc=1 self_pid use_process_group=0
+
+    SESSION_BUS_PROBE_OUTPUT=""
+    output_file="$(mktemp "${TMPDIR:-/tmp}/codex-session-bus-probe.XXXXXX")" || return 1
+    self_pid=${BASHPID:-$$}
+
+    if command -v setsid >/dev/null 2>&1; then
+        setsid "$@" >"$output_file" 2>/dev/null &
+        use_process_group=1
+    else
+        "$@" >"$output_file" 2>/dev/null &
+    fi
     probe_pid=$!
+
     ( trap - EXIT
       sleep 0.5
       [ -r "/proc/$probe_pid/stat" ] || exit 0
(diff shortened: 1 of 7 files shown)


Yo-DDV:
Hi Daniel, thanks for working on this. The direction makes sense to me, but I think two startup-path edge cases are worth tightening before merge.

First, the new accessibility auto-detection does not include the accessibility state that the bundled Linux Computer Use backend enables. `setup_accessibility_report()` first sets `org.a11y.Status IsEnabled=true`, then falls back to `org.gnome.desktop.interface toolkit-accessibility=true`; the launcher only checks Orca/brltty, `org.gnome.desktop.a11y.applications screen-reader-enabled`, and env markers. I reproduced the launcher probe with `toolkit-accessibility=true` and `screen-reader-enabled=false`; it still produced `renderer_accessibility=0`, so a user who prepared Computer Use but is not running a screen reader loses the previous `--force-renderer-accessibility` default. Since the backend notes that Electron apps need this flag for an AT-SPI tree, I would include `org.a11y.Status IsEnabled` and/or `org.gnome.desktop.interface toolkit-accessibility` as positive signals, or make the Computer Use limitation explicit in troubleshooting.

Second, the new `gsettings get ... screen-reader-enabled` call is unbounded on the launch path. With a slow `gsettings` in `PATH`, the launcher probe stayed blocked past one second before Electron args were built. Most startup probes in this launcher are deliberately capped, so this one should probably be capped as well, or made best-effort in some other way so a broken session bus cannot delay every launch.

The `/dev/shm` gating itself looks well covered by the new smoke cases.

djvdorp:
Thanks for the careful review here @Yo-DDV — both points were real, and both are addressed in f202df2.

**Computer Use accessibility state:** detection now includes exactly what
`setup_accessibility_report()` enables — `org.a11y.Status IsEnabled` (busctl) and the
`org.gnome.desktop.interface toolkit-accessibility` gsettings fallback. I verified both
are `false` on stock Ubuntu 24.04 GNOME, so they only fire when something actually
enabled AT-SPI, and I reproduced your exact scenario (`toolkit-accessibility=true`,
`screen-reader-enabled=false`) — the launcher now produces `renderer_accessibility=1`.
New smoke probes cover both signals. One note: if Computer Use setup runs while the app
is already open, the flag applies on next launch — same restart semantics the backend
already documents for target apps.

**Unbounded gsettings:** all session-bus probes (gsettings + busctl) now run in a
subshell killed by the same ppid-guarded watchdog pattern `webview_port_is_open` uses,
capped at 0.5 s — `/usr/bin/timeout` avoided per the existing uutils cold-start note. A
killed or failing probe counts as "not detected" and never blocks launch. A new smoke
test uses a gsettings stub that sleeps 5 s and asserts the probe neither forces the flag
nor takes more than 3 s (measured ~90 ms warm on a healthy bus).

Full `tests/scripts_smoke.sh` is green; docs/changelog updated to match.

avifenesh (commented):

launcher/start.sh.template
```diff
@@ -2648,6 +2701,75 @@ effective_electron_ozone_platform() {
...
+# bus, so the probes run in a subshell killed by the same ppid-guarded
+# watchdog webview_port_is_open uses (/usr/bin/timeout intentionally avoided;
+# see that function), capped at 0.5 s. A killed or failing probe counts as
+# "not detected" and never blocks launch.
+session_bus_assistive_technology_detected() {
+    local probe_pid kill_pid rc=1 self_pid=$$
+
+    ( trap - EXIT
+      if command -v gsettings >/dev/null 2>&1; then
+          case "$(gsettings get org.gnome.desktop.a11y.applications screen-reader-enabled 2>/dev/null || true)" in
```

---

This watchdog only bounds the wrapper shell, not the external probe that can hang. The `gsettings`/`busctl` calls are run inside command substitutions; when the watchdog later sends `kill -9 "$probe_pid"`, that PID is the background shell running this block, while the command-substitution child (and the `gsettings` process blocked on the session bus) can be reparented and keep running. That means the broken-bus case no longer blocks launch, but it can leak a hung helper on every launch. The launcher’s other bounded external-command probe avoids this by `exec`ing the probed command as the watched PID and even asserts the hanging process is gone in the smoke test, so please make the watchdog own/kill the actual session-bus helper (or its process group), not just this wrapper.

ilysenko (approved):
Looks good after the session-bus watchdog cleanup and smoke coverage update.

PR #682, opened 2026-07-06
Detach CLI version probe watchdog from the caller pipe (1s cold-start stall)

djvdorp:
## What changed

- The CLI version probe's watchdog subshell in `launcher/start.sh.template` now runs
  with stdout/stderr detached (`>/dev/null 2>&1`), so its `sleep 1` child can no longer
  hold the caller's command-substitution pipe after the watchdog is killed.

## Why

`log_codex_cli_path` reads `codex_cli_version` via command substitution on the launch
path, immediately before Electron is spawned. The watchdog subshell inherited the
substitution pipe; its sleep child survives the watchdog kill and holds the pipe open,
and command substitution waits for pipe EOF. Result: every cold launch stalled ~1 s
even with a CLI that answers `--version` in ~50 ms. Measured on a live cold start, the
`launch_state_refreshed_under_lock` → `electron_launch` phase gap drops from ~1010 ms
to ~74 ms, so the window and GNOME's startup feedback (dock icon / launch spinner) show
up about a second sooner. Same bug class as the session-bus probe cleanup fixed in
1a63ae5; the decision record now documents the pattern for future bounded probes.

- `launcher/start.sh.template`, `tests/scripts_smoke.sh`
- `CHANGELOG.md`, `docs/launcher-performance.md`

## How it was tested

- TDD: a new timing regression test in `test_launcher_cli_resolution_policy` reads a
  fast stub CLI through command substitution and asserts completion well under the
  watchdog second — it fails at ~1010 ms before the fix and passes at ~60 ms after.
- Full `bash tests/scripts_smoke.sh` passes; `bash -n` on the template.
- Regenerated launcher measured on a real cold start (phase log confirms the gap).

## Known risks / follow-ups

- None functional: the watchdog's output was never meaningful, and probe/timeout
  semantics are unchanged. `webview_port_is_open`'s watchdog shares the shape but is
  only used in status contexts (never captured), so it was left untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Cold launches no longer stall for a full second between acquiring the
+  launcher lock and spawning Electron. The CLI version log line reads the
+  probe result through command substitution, and the probe's watchdog
+  subshell inherited that pipe: its `sleep 1` child survived the watchdog
+  kill and held the pipe open until the sleep expired, so even a CLI that
+  answers `--version` in ~50 ms blocked the launch path for ~1 s. The
+  watchdog now runs detached from the caller's stdout/stderr, cutting that
+  launch phase from ~1010 ms to ~74 ms and making the window (and GNOME's
+  startup feedback) appear about a second sooner on every cold start.
+
 ### Changed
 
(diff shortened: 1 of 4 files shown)


ilysenko (approved):
Looks good. I verified the current head is focused on detaching the CLI version probe watchdog from the command-substitution pipe, preserves the probe timeout semantics, and has green CI.

PR #687, opened 2026-07-06
Add MCP helper reaper feature

joshyorko:
## Summary

Adds an opt-in Linux feature that conservatively cleans up stale duplicate MCP helper generations for Codex Desktop and Codex CLI sessions.

- adds the `mcp-helper-reaper` Linux feature with a self-contained Rust helper under `linux-features/mcp-helper-reaper/reaper`
- keeps the root Cargo workspace and root `Cargo.lock` untouched when the feature is disabled
- dedupes helper roots only under the same live Codex parent PID, keeping independent CLI/Desktop sessions separate
- keeps the newest helper generation and reaps older duplicate roots plus descendants
- supports conservative orphan cleanup for configured/app-scoped helper roots with Codex-origin proof
- preserves existing `SessionStart` hooks while installing its own hook idempotently
- uses generic config/plugin/app-path/MCP convention detection rather than matching specific helper or provider names
- keeps `node-repl-reaper` compatible with wrapped `resources/node_repl` entrypoints

## Validation

- `cargo fmt --manifest-path linux-features/mcp-helper-reaper/reaper/Cargo.toml -- --check`
- `cargo test --manifest-path linux-features/mcp-helper-reaper/reaper/Cargo.toml`
- `cargo clippy --manifest-path linux-features/mcp-helper-reaper/reaper/Cargo.toml --all-targets -- -D warnings`
- `node --test linux-features/mcp-helper-reaper/test.js linux-features/node-repl-reaper/test.js`
- `bash -n linux-features/mcp-helper-reaper/stage.sh linux-features/mcp-helper-reaper/cleanup.sh linux-features/mcp-helper-reaper/install-session-hook.sh linux-features/mcp-helper-reaper/cold-start-hook.sh linux-features/mcp-helper-reaper/after-exit-hook.sh linux-features/mcp-helper-reaper/node-repl-wrapper.sh linux-features/node-repl-reaper/reaper.sh linux-features/node-repl-reaper/stage.sh linux-features/node-repl-reaper/cold-start-hook.sh linux-features/node-repl-reaper/after-exit-hook.sh`
- `git diff --check`


ilysenko:
Thanks for the PR. I understand why this is implemented as a Rust helper: process-tree inspection, `/proc` parsing, and careful SIGTERM/SIGKILL handling are much safer to express and test in Rust than in shell.

That said, because this is an optional Linux feature, I want to keep its maintenance footprint contained. Adding `mcp-helper-reaper` as a root Cargo workspace member means the feature is no longer isolated: it affects `Cargo.lock`, CI, package/update-builder payloads, Nix/package maintenance, and future auto-update rebuilds even when the feature is disabled by default.

So before we can consider this, please either:

- keep the Rust helper but fully integrate it as a supported workspace component, including update-builder packaging and tests; or
- keep the implementation self-contained inside `linux-features/mcp-helper-reaper/` so the optional feature does not expand the global workspace maintenance surface.

There are also a few safety blockers that need fixing either way:

- `codex-linux-sandbox` must not be treated as a helper candidate.
- orphan cleanup needs tighter proof that a process was launched by Codex before it is terminated.
- PID identity should be revalidated immediately before signaling, including descendants.
- the interaction with `node-repl-reaper` should be handled explicitly, either by declaring a conflict or making both features compatible.

joshyorko:
TL;DR: addressed in `13ff0fe`. The helper is now self-contained under `linux-features/mcp-helper-reaper/reaper` with no root Cargo workspace/lock diff; sandbox exclusions, orphan origin proof, descendant PID identity checks, and `node-repl-reaper` compatibility are tightened and tested.

Cargo.toml
```diff
@@ -1,3 +1,9 @@
 [workspace]
-members = ["computer-use-linux", "read-aloud-linux", "updater", "record-replay-linux"]
+members = [
+  "computer-use-linux",
+  "mcp-helper-reaper",
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

Adding this helper as a root workspace member breaks the packaged update-builder path as currently staged. `scripts/lib/package-common.sh` copies this root `Cargo.toml`/`Cargo.lock` into `/opt/.../update-builder`, but its bundle list only copies the existing Rust crates (`computer-use-linux`, `record-replay-linux`, `read-aloud-linux`, `updater`), and `updater/src/builder.rs` has the same required bundle list. The packaged rebuild then sees this workspace member but no `mcp-helper-reaper/` directory, so Cargo fails before it can build the usual Rust components. Please either include this crate in the update-builder bundle/tests, or keep the helper source self-contained under `linux-features/mcp-helper-reaper/`.

linux-features/mcp-helper-reaper/stage.sh
```diff
@@ -0,0 +1,88 @@
...
+install -m 0755 "$feature_dir/after-exit-hook.sh" "$codex_linux_dir/after-exit.d/mcp-helper-reaper"
+
+if [ ! -e "$node_repl" ]; then
+    echo "mcp-helper-reaper staged: reaper installed; resources/node_repl not present to wrap" >&2
+    exit 0
+fi
+
+if [ ! -e "$original_node_repl" ]; then
+    mv "$node_repl" "$original_node_repl"
+elif ! grep -q "mcp-helper-reaper-node-repl-wrapper" "$node_repl" 2>/dev/null; then
```
avifenesh, line 82: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This makes the feature non-idempotent after any reinstall/update that refreshes `resources/node_repl`. On a normal rebuild, `install_browser_use_node_repl_resource` writes a fresh real `resources/node_repl` while the old `node_repl.codex-linux-original` backup may still exist from the prior enabled install; this branch then sees a non-wrapper current entrypoint and aborts staging. The feature needs to tolerate that state (for example, replace a stale backup or restore/refresh it) rather than failing the next rebuild.

linux-features/mcp-helper-reaper/stage.sh
```diff
@@ -0,0 +1,88 @@
...
+    echo "mcp-helper-reaper: refusing to overwrite resources/node_repl because original backup already exists and current entrypoint is not this feature's wrapper" >&2
+    exit 1
+fi
+
+install -m 0755 "$feature_dir/node-repl-wrapper.sh" "$node_repl"
```
avifenesh, line 87: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

Wrapping `resources/node_repl` changes the actual helper argv0 to `resources/node_repl.codex-linux-original` after the wrapper `exec`s. The existing `node-repl-reaper` still matches only `$APP_DIR/resources/node_repl`, so enabling both features silently stops Browser Use leak cleanup for these helpers. Please declare a `conflicts` relationship in `feature.json` or update `node-repl-reaper` to match the backup path as well.

linux-features/mcp-helper-reaper/reaper/src/lib.rs
```diff
@@ -0,0 +1,1636 @@
...
+                if dry_run { "would reap" } else { "reaping" },
+                candidate.stale_pid,
+                candidate.keep_pid,
+                candidate.signature
+            ),
+        );
+        if dry_run {
+            continue;
+        }
+        signal_tree(candidate.stale_pid, libc::SIGTERM, processes);
```
avifenesh, line 523: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`signal_tree` is called using the stale `/proc` snapshot collected before planning, so the first SIGTERM can hit a recycled PID (root or descendant) if the process exits between `all_processes()` and this line. The later SIGKILL path revalidates only the root after the grace period, but the initial signal needs the same PID identity check immediately before killing, and descendants should be derived/revalidated from a fresh snapshot of that same root.

avifenesh (commented):

---

I found a few integration/safety issues in the new feature path that look blocking before this can be safely enabled.

ilysenko:
Thanks, this is much better now. The root workspace/update-builder issue, sandbox exclusion, PID identity revalidation, and node-repl-reaper compatibility all look addressed.

One staging blocker remains: mcp-helper-reaper/stage.sh is still not idempotent across non-fresh rebuilds. If node_repl.codex-linux-original remains from a previous enabled install and install.sh refreshes resources/node_repl, the stage hook exits instead of refreshing the backup/reinstalling the wrapper. Please make that state safe and add a focused test for it.

p.s. This is the tone selected by the codex; it is not my choice. :)

linux-features/mcp-helper-reaper/stage.sh
joshyorko: Fixed in `d208f00`: non-fresh rebuild staging now refreshes the stale backup from the newly generated `resources/node_repl` instead of aborting. Added a focused test for this exact state.

Cargo.toml
joshyorko: Addressed in current HEAD: the helper source is self-contained under `linux-features/mcp-helper-reaper/reaper`, root `Cargo.toml`/`Cargo.lock` are no longer changed, and the feature has its own local lockfile.

linux-features/mcp-helper-reaper/stage.sh
joshyorko: Addressed in current HEAD: `node-repl-reaper` now matches both `resources/node_repl` and `resources/node_repl.codex-linux-original`, with coverage for the wrapped helper path.

joshyorko:
Fixed in `d208f00`. `stage.sh` now refreshes a stale `node_repl.codex-linux-original` backup from the newly staged `resources/node_repl` on non-fresh rebuilds, then reinstalls the wrapper. Added a focused test for that state; validation is green locally.

linux-features/mcp-helper-reaper/reaper/src/lib.rs
joshyorko: Addressed in current HEAD: signaling revalidates PID identity for each root/descendant immediately before sending the signal, with a unit test covering skipped recycled descendants.

ilysenko:

- `install-session-hook.sh` still overwrites invalid or unreadable `hooks.json` with a new file. Runtime hook installation should fail soft and leave user hooks untouched when they cannot be parsed.
- Orphan cleanup still accepts inherited env keys such as `CODEX_HOME` or `CODEX_CLI_PATH` as Codex-origin proof before signaling. For SIGTERM/SIGKILL, please require a stronger marker or skip orphan killing.

ilysenko (approved):
Reviewed current head; previous blockers are fixed.

PR #688, opened 2026-07-06
Overlap webview startup and run plugin cache syncs concurrently

djvdorp:
## What changed

- `ensure_webview_server` is split into `start_webview_server` (spawn) and
  `await_webview_server_ready` (readiness + origin verification). The Python webview
  server now boots while the launcher does CLI lookup and plugin cache syncs; the wait
  runs after them — still strictly before Electron, so the verified-origin invariant is
  unchanged, and PR_SET_PDEATHSIG ties the server to the launcher on early-exit paths.
- The five bundled plugin cache syncs run concurrently and are each awaited before
  cold-start hooks dispatch, keeping every existing phase marker. The shared bundled
  marketplace metadata they previously each rewrote (identical rm+cp of the same
  marketplace.json, five times) is staged exactly once beforehand
  (`stage_bundled_marketplace_metadata`) — a DRY cleanup that is also what makes the
  concurrency safe; the syncs otherwise touch only disjoint per-plugin cache dirs.

## Why

Cold starts serialized independent work before Electron could spawn. Measured on a
warm-cache cold start: sync block ~285 ms → ~110 ms, webview wait ~150 ms → ~35 ms,
Electron spawn ~330 ms earlier, window mapped at ~2.6 s instead of ~3.1 s.

- `launcher/start.sh.template`, `tests/scripts_smoke.sh`
- `CHANGELOG.md`, `docs/launcher-performance.md` (startup-ordering decision record updated)

## How it was tested

- TDD: structural checks written first and confirmed failing — spawn-before-syncs /
  await-after-syncs ordering, single marketplace staging site, per-sync waits before
  cold-start hooks, preserved phase markers, cold-start-only scoping.
- Full `bash tests/scripts_smoke.sh` and `bash tests/webview_probe_equivalence.sh` pass;
  `bash -n` on the template.
- Live cold start against the regenerated launcher verified the phase timeline,
  `marketplace.json` staging, and all plugin cache symlinks.
- `cargo check`/`cargo test` and the package builds were not run: no Rust or
  package-payload changes (launcher template ships identically to all formats).

## Known risks / follow-ups

- Sync log lines can now interleave in the launcher log (single-line O_APPEND writes,
  so no corruption). Per-sync phase timestamps reflect completion order under
  concurrency. Launching Electron in parallel remains rejected — the renderer needs a
  verified origin (documented in `docs/launcher-performance.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Changed
 
+- Cold starts overlap the webview server boot with the rest of launcher
+  startup and run the five bundled plugin cache syncs concurrently. The
+  launcher now spawns the Python webview server, does CLI lookup and cache
+  syncs while it boots, and only then waits for readiness and verifies the
+  origin — still strictly before Electron launches. The shared bundled
+  marketplace metadata is staged exactly once before the concurrent syncs
+  instead of being rewritten inside each sync. Measured on a warm-cache cold
+  start, the sync block drops from ~285 ms to ~110 ms and the webview wait
+  from ~150 ms to ~35 ms, putting Electron spawn ~330 ms earlier.
 - The launcher now passes `--disable-dev-shm-usage` to Electron only when
   `/dev/shm` is missing, not writable, or smaller than 1 GiB (the container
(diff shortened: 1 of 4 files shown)


ilysenko (approved):

PR #689, opened 2026-07-06
Restore Linux desktop settings section metadata

colonelpanic8:
## Summary
- compose keybinds/Linux desktop webview asset patches per file so route/navigation writes do not overwrite section metadata writes
- tighten Linux desktop section metadata idempotence so split section-order and slug-list shapes are both verified
- add regression coverage for the current `app-initial~app-main~automations-page` bundle carrying both route and section metadata

## Validation
- `PATH=/run/current-system/sw/bin:$PATH node --test scripts/patch-linux-window-ui.test.js --test-name-pattern 'Linux desktop|Keyboard Shortcuts|settings route|section'`
- `PATH=/run/current-system/sw/bin:$PATH node --test scripts/patch-linux-window-ui.test.js`
- `PATH=/run/current-system/sw/bin:$PATH nix build .#codex-desktop-computer-use-ui-remote-mobile-control --no-link --print-out-paths`
- inspected generated package `/nix/store/97xy72qqhz4xjhiyljwzxlkx2c387s10-codex-desktop-computer-use-ui-remote-mobile-control-26.623.101652`: webview assets include `slug:\`linux-desktop\`` in `app-initial~app-main~automations-page-Bl6HoLGr.js` and still reference `linux-desktop-settings-linux.js`

### DIFF
--- a/scripts/patches/impl/keybinds-settings.js
+++ b/scripts/patches/impl/keybinds-settings.js
@@ -577,6 +577,7 @@ function collectOptionalAssetPatches(extractedDir, filenamePattern, patchFn) {
         filePath,
         currentSource,
         patchedSource: patchFn(currentSource),
+        patchFn,
       });
     } catch (error) {
       const message = error instanceof Error ? error.message : String(error);
@@ -610,6 +611,7 @@ function collectOptionalMatchingAssetPatches(extractedDir, predicate, patchFn) {
         filePath,
         currentSource,
         patchedSource: patchFn(currentSource),
+        patchFn,
       });
     } catch (error) {
       const message = error instanceof Error ? error.message : String(error);
@@ -651,7 +653,21 @@ function collectLinuxDesktopRouteAndNavigationPatches(extractedDir) {
       patchedSource = applyLinuxDesktopSettingsNavigationPatch(patchedSource);
     }
     if (patchedSource !== currentSource) {
-      patches.push({ filePath, currentSource, patchedSource });
+      patches.push({
+        filePath,
+        currentSource,
+        patchedSource,
+        patchFn(source) {
+          let nextSource = source;
+          if (isSettingsRouteBundleSource(nextSource)) {
+            nextSource = applyLinuxDesktopSettingsRoutePatch(nextSource);
+          }
+          if (isSettingsNavigationBundleSource(nextSource)) {
+            nextSource = applyLinuxDesktopSettingsNavigationPatch(nextSource);
+          }
+          return nextSource;
+        },
+      });
     }
   }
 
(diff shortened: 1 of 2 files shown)


Yo-DDV:
Thanks Ivan, this looks good on my side.

I rechecked the patch against the Linux desktop settings route/section cases, including the same-asset composition case this PR is fixing. The behavior I see in the tests matches the improvement you are making: the route patch and the section metadata patch now compose cleanly instead of one write losing the other.

The checks I ran are green on the PR head:

- `node --test --test-name-pattern 'Linux desktop|Keyboard Shortcuts|settings route|section' scripts/patch-linux-window-ui.test.js`
- `node --check` on the touched JS files

scripts/patches/impl/keybinds-settings.js
```diff
@@ -777,30 +816,36 @@ function applyKeybindsSettingsSectionsPatch(currentSource) {
...
   }
 
-  if (!/=\[`general-settings`,`linux-desktop`/.test(patchedSource)) {
-    const orderPattern = /([A-Za-z_$][\w$]*=\[`general-settings`,)(?!`linux-desktop`)/;
-    if (orderPattern.test(patchedSource)) {
-      patchedSource = patchedSource.replace(orderPattern, "$1`linux-desktop`,");
+  if (!patchedArrayOrderPattern.test(patchedSource)) {
+    const arrayInsertPattern = /([A-Za-z_$][\w$]*=\[`general-settings`,)(?!`linux-desktop`)/;
+    if (arrayInsertPattern.test(patchedSource)) {
+      patchedSource = patchedSource.replace(arrayInsertPattern, "$1`linux-desktop`,");
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This patch helper still applies the section metadata insertions with one-shot `replace()` calls (same pattern below for the split order and slug list). The repo invariant for ASAR bundle patches is that helpers either process every eligible occurrence or explicitly assert the needle is unique: these webview chunks are concatenated, so once the first occurrence is patched the later `patched*Pattern` checks become true and any second unpatched section-order/slug-list occurrence in the same asset is silently left behind. Please iterate/count matches (or assert uniqueness for each shape) before returning.

avifenesh (commented):

---

Found one ASAR patch-helper invariant issue.

ilysenko:
Thanks for the fix. One blocker is still present: `applyLinuxDesktopSettingsSectionsPatch` still uses one-shot `replace()` calls and the final patched checks only prove that at least one matching section-order / slug-list shape was patched.

For concatenated webview chunks, a second eligible occurrence in the same asset can remain unpatched while the helper returns success. I reproduced this on the current PR head with duplicate array-order and object slug-list occurrences: the first occurrence gets `linux-desktop`, but the second remains unchanged.

Please either iterate/count and patch every eligible occurrence for each supported shape, or explicitly assert uniqueness before returning success, and add regression coverage for duplicate occurrences in one asset.

scripts/patches/impl/keybinds-settings.js
colonelpanic8: Addressed in `89b6be4`.

`applyLinuxDesktopSettingsSectionsPatch` now uses global unpatched-shape matchers for the array order, split order, and object slug-list forms, so every eligible occurrence in a concatenated asset is patched before success is returned. I also added regression coverage with duplicate occurrences of all three supported shapes in one asset.

Validation:
- `PATH=/run/current-system/sw/bin:$PATH node --test scripts/patch-linux-window-ui.test.js --test-name-pattern 'Linux desktop|Keyboard Shortcuts|settings route|section'`
- `PATH=/run/current-system/sw/bin:$PATH node --test scripts/patch-linux-window-ui.test.js` (`342` passing)
- `PATH=/run/current-system/sw/bin:$PATH nix build .#codex-desktop-computer-use-ui-remote-mobile-control --no-link --print-out-paths`
- inspected `/nix/store/7fqxp6dqgnxh0gb93swfa3g8hl63xcig-codex-desktop-computer-use-ui-remote-mobile-control-26.623.101652`: generated assets include `slug:\`linux-desktop\`` and the `linux-desktop-settings-linux.js` route ref.

colonelpanic8:
Updated in `89b6be4` to address the duplicate-occurrence blocker: the Linux desktop section metadata patch now patches every eligible array-order, split-order, and object slug-list occurrence in a concatenated asset, with regression coverage for duplicate shapes in one source.

Validation rerun:
- `node --test scripts/patch-linux-window-ui.test.js --test-name-pattern 'Linux desktop|Keyboard Shortcuts|settings route|section'`
- `node --test scripts/patch-linux-window-ui.test.js` (`342` passing)
- `nix build .#codex-desktop-computer-use-ui-remote-mobile-control --no-link --print-out-paths`
- inspected generated package output and confirmed `slug:\`linux-desktop\`` plus the `linux-desktop-settings-linux.js` route reference.

ilysenko (approved):
Reviewed current head. The duplicate section metadata blocker is addressed, and the focused/local validation plus CI checks pass.

PR #690, opened 2026-07-06
feat(linux): add CODEX_LINUX_DISABLE_EXTERNAL_OPEN_PATCH env var

kkimdev:
## Summary

Adds a runtime environment variable `CODEX_LINUX_DISABLE_EXTERNAL_OPEN_PATCH=1` that skips the Linux external open environment sanitization patch, falling back to the original Electron `shell.openExternal` behavior.

## Motivation

The Linux external open env patch (`linux-external-open-env`) sanitizes 19 Electron-specific env vars before spawning `xdg-open` for `shell.openExternal(url)` calls. While this prevents Electron env pollution from breaking `xdg-open` on standard desktops, it can break URL forwarding on Crostini and other specialized environments (e.g. Flatpak, containers) where env vars like `DBUS_SESSION_BUS_ADDRESS` or `XDG_CURRENT_DESKTOP` are critical for correct garcon/portal routing.

Previously, users on Crostini had no recourse short of forking the repo or hacking the build. This PR gives them an opt-out at runtime.

## Implementation

Follows the established `CODEX_LINUX_MULTI_LAUNCH` precedent — a single env var check in the injected `codexLinuxPatchExternalOpen` function at runtime:

```js
if(process.env.CODEX_LINUX_DISABLE_EXTERNAL_OPEN_PATCH===`1`)return __codexElectron;
```

The patcher (`applyLinuxExternalOpenEnvPatch`) is unchanged — it always wraps `require(\"electron\")`. The env var is checked only when the wrapped function executes, so existing builds are unaffected.

## Usage

```bash
CODEX_LINUX_DISABLE_EXTERNAL_OPEN_PATCH=1 codex-desktop
```

Or set it in the Electron launcher wrapper / cros-garcon service override.

## Testing

- All 339 existing tests pass (unchanged).
- 4 new tests in `scripts/patches/impl/main-process/browser.test.js`:
  - Verifies `require(\"electron\")` is wrapped correctly
  - Verifies the env var guard is present in the injected helper
  - Verifies idempotency (double-application is a no-op)
  - Verifies warning when no electron require is found


Yo-DDV:
Thanks for putting this together.

I tested the runtime path locally and the implementation behaves as intended from what I can see: with `CODEX_LINUX_DISABLE_EXTERNAL_OPEN_PATCH=1`, the patched helper delegates to Electron's original `shell.openExternal` and does not call `xdg-open`; with any other value, the sanitized `xdg-open` path still runs.

The one thing I would still adjust is the regression coverage. The new test only checks that the env var string is present in the generated helper. For this flag, the important contract is the runtime behavior, especially because it is an explicit escape hatch around the env sanitization. I think it would be safer to add a test next to the existing `evaluatePatchedExternalOpen` cases in `scripts/patch-linux-window-ui.test.js`, asserting that `CODEX_LINUX_DISABLE_EXTERNAL_OPEN_PATCH=1` delegates to the original Electron call and never invokes `xdg-open`.

Checks I ran on the PR head:

- `node --test scripts/patches/impl/main-process/browser.test.js`
- `node --check` on the touched JS files
- manual VM-harness check for the enabled and disabled runtime paths

kkimdev:
@Yo-DDV Thanks for the review! Added two tests!

ilysenko (approved):
Looks good. The runtime escape hatch is narrow, CI is green, and the added tests cover both the disabled delegation path and the normal sanitized xdg-open path.

PR #691, opened 2026-07-07
Prevent summary-only app-server thread hydration

sanshi42:
## Summary

- request turns when hydrating unknown app-server conversations before upserting them into the local conversation list
- skip/retry summary-only thread reads so threads without local rollout turns are not shown as resumable conversations
- add regression coverage for the summary-only thread response in both the core webview patch and the remote mobile feature patch

## Why

When a late app-server event arrives for a conversation that is not in the local sidebar state, the Linux hydration patch currently calls `readThread(..., { includeTurns: false })` and treats any returned thread summary as a valid local conversation. In practice, the app server can return a thread summary for a thread id that has no local rollout file. That inserts a ghost "New conversation" into the sidebar; selecting it then fails with an error like:

```text
no rollout found for thread id ...
```

Requiring a non-empty `turns` payload makes the hydration path distinguish a resumable local conversation from a summary-only app-server thread. Summary-only reads keep the existing retry/skip behavior instead of being upserted into the local conversation map.

## Validation

- Manual packaged install smoke: rebuilt and reinstalled the Linux package locally; the ghost conversation no longer appears.


ilysenko:
Please make sure you remove the draft status so the pull request can proceed.

sanshi42:
> Please make sure you remove the draft status so the pull request can proceed.

Thanks for the heads-up. I’ve marked the PR as ready for review.

Yo-DDV:
Thanks for picking this up. The direction makes sense to me and it lines up with the remote-mobile hydration thread rather than contradicting it.

One thing I would still check before this moves forward: this updates the fresh patch output, but an already-patched bundle still looks like `already-applied` because the guard only checks for the existing `codexLinuxRemoteMobileNotificationQueue` marker. I tested this against a copied installed bundle that already had the previous hydration patch: the report returned `linux-app-server-conversation-hydration: already-applied`, but the bundle still contained `this.readThread(d,{includeTurns:!1})` and no new turns-array guard in that hydration block.

Fresh DMG rebuilds should be fine, and the CI covers that path, so this may be acceptable if only fresh rebuild/reinstall is supported. If rerunning the patcher over an existing installed content tree is meant to be supported, it would be worth adding a small migration branch/test for the older marker shape so the new behavior is applied there too.

ilysenko (approved):
Looks good. I rechecked the current head and the focused tests pass. The fresh DMG/rebuild path is covered here; old already-patched generated bundles are not the durable source of truth for this repository.

sanshi42:
I checked the maintainer follow-up. Since already-patched generated bundles are not treated as the durable source of truth here, I’ll leave the merged PR as-is. The fresh rebuild path is covered by the added tests and local package reinstall validation.

PR #693, opened 2026-07-07
Fix Linux Computer Use install flow asset drift

joshyorko:
## Summary

- widen the Linux Computer Use install-flow asset matcher for current app-initial route chunks
- add a regression fixture for the current `remote-conversation-page~new-thread-panel-page~onboarding-page~appgen` bundle shape

## Verification

- `node --test --test-name-pattern='Computer Use|patchExtractedApp scans apps bundles' scripts/patch-linux-window-ui.test.js`
- `node --test linux-features/x11-ewmh-computer-use/test.js`
- direct descriptor probe against the installed 26.623.141536 asset: matcher returns true, patch changes source, Linux gate is present


ilysenko:

- This fixes upstream asset drift, but the new matcher keeps broad legacy bundle shapes (`plugin-detail-page`, `pull-requests-page`, `plug`, `use-plugin-install-flow`, `plugins-availability`) instead of targeting the current `CODEX.DMG` shape.
- Per CONTRIBUTING.md, drift fixes need to remove old compatibility paths in the same PR; please keep the current bundle match and drop the legacy fallback shapes.

PR #698, opened 2026-07-07
Add DMG integrity guard for Linux settings patches

joshyorko:
## Summary

- add a DMG-intel acceptance blocker for unresolved Linux settings patch symbols
- run that integrity scan inside the patcher and carry findings through `patch-report.json`, so `Codex.dmg + patch-report` flows see post-patch webview failures
- distinguish real `var` / `let` / `const` declarators from bare assignment expressions in Linux settings icon detection
- update Agent Workspaces settings icon patching and Browser Use node_repl trust detection for current upstream chunk shapes

## Verification

- `docker run --rm -v "$PWD":/workspaces/codex-desktop-linux -w /workspaces/codex-desktop-linux codex-desktop-linux-devcontainer:local node --test scripts/dev/upstream-dmg-intel.test.js scripts/patch-linux-window-ui.test.js` (362/362)
- `docker run --rm -v "$PWD":/workspaces/codex-desktop-linux -w /workspaces/codex-desktop-linux codex-desktop-linux-devcontainer:local node --test scripts/patch-linux-window-ui.test.js linux-features/agent-workspace/test.js` (378/378, run before the review-fix follow-up; the follow-up did not touch `linux-features/agent-workspace`)

### DIFF
--- a/scripts/lib/upstream-dmg-intel.js
+++ b/scripts/lib/upstream-dmg-intel.js
@@ -941,12 +941,67 @@ function createNativeBinaryMap(inventory, registry = null) {
 
 const LINUX_SETTINGS_PATCH_SYMBOL_PATTERN = /\bcodexLinux[A-Za-z0-9_$]*SettingsIcon\b/g;
 
+function segmentDeclaresSymbol(segment, symbol) {
+  const escaped = escapeRegExp(symbol);
+  return new RegExp(`^\\s*${escaped}(?![A-Za-z0-9_$])`).test(segment);
+}
+
+function hasVariableDeclarator(source, symbol) {
+  const keywordPattern = /\b(?:var|let|const)\b/g;
+  let match;
+  while ((match = keywordPattern.exec(source)) != null) {
+    let segmentStart = keywordPattern.lastIndex;
+    let depth = 0;
+    let quote = null;
+    let escaped = false;
+    let terminated = false;
+    for (let index = segmentStart; index < source.length; index += 1) {
+      const char = source[index];
+      if (quote != null) {
+        if (escaped) {
+          escaped = false;
+        } else if (char === "\\") {
+          escaped = true;
+        } else if (char === quote) {
+          quote = null;
+        }
+        continue;
+      }
+      if (char === '"' || char === "'" || char === "`") {
+        quote = char;
+        continue;
+      }
+      if (char === "(" || char === "[" || char === "{") {
+        depth += 1;
+        continue;
+      }
+      if (char === ")" || char === "]" || char === "}") {
+        depth = Math.max(0, depth - 1);
+        continue;
+      }
+      if (depth === 0 && (char === "," || char === ";")) {
+        if (segmentDeclaresSymbol(source.slice(segmentStart, index), symbol)) {
+          return true;
+        }
+        if (char === ";") {
+          terminated = true;
+          break;
+        }
+        segmentStart = index + 1;
+      }
+    }
+    if (!terminated && segmentDeclaresSymbol(source.slice(segmentStart), symbol)) {
+      return true;
+    }
+  }
+  return false;
+}
+
 function hasLocalPatchSymbolDeclaration(source, symbol) {
   const escaped = escapeRegExp(symbol);
   return (
     new RegExp(`\\bfunction\\s+${escaped}\\s*\\(`).test(source) ||
-    new RegExp(`\\b(?:var|let|const)\\s+[^;]*\\b${escaped}\\s*=`).test(source) ||
-    new RegExp(`[;,]\\s*${escaped}\\s*=`).test(source)
+    hasVariableDeclarator(source, symbol)
   );
 }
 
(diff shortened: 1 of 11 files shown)


scripts/lib/upstream-dmg-intel.js
```diff
@@ -682,6 +687,7 @@ function extractProtectedSurfaces({ inventory, registry, repoRoot = process.cwd(
   const bridgeMap = createBridgeMap(inventory);
   const pluginMap = createPluginMap(inventory);
   const nativeBinaryMap = createNativeBinaryMap(inventory, registry);
+  const postPatchIntegrity = findPostPatchIntegrityFindings(inventory);
```
avifenesh, line 690: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This integrity scan is wired to the same inventory that is built from `--candidate`, which in the normal flow is the raw upstream `Codex.dmg` (the CLI usage and this PR's verification pass `--candidate ./Codex.dmg --patch-report ...`). That pre-patch DMG cannot contain the Linux-injected `codexLinux...SettingsIcon` symbols, and the patch report is only folded in later as status metadata, so `PATCH_INTEGRITY_BROKEN` will never fire for the real failure mode: a patched webview asset with an unresolved Linux settings icon. This needs to scan the patched app/extracted ASAR/installed `content/webview` (or run inside the patcher while `$WORK_DIR/app-extracted` is available) rather than the upstream candidate inventory.

scripts/lib/upstream-dmg-intel.js
```diff
@@ -932,6 +939,39 @@ function createNativeBinaryMap(inventory, registry = null) {
   return { binaries };
 }
 
+const LINUX_SETTINGS_PATCH_SYMBOL_PATTERN = /\bcodexLinux[A-Za-z0-9_$]*SettingsIcon\b/g;
+
+function hasLocalPatchSymbolDeclaration(source, symbol) {
+  const escaped = escapeRegExp(symbol);
+  return (
+    new RegExp(`\\bfunction\\s+${escaped}\\s*\\(`).test(source) ||
+    new RegExp(`\\b(?:var|let|const)\\s+[^;]*\\b${escaped}\\s*=`).test(source) ||
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This treats any matching assignment within the current `var` statement text as a declaration, so it can suppress the exact unresolved-symbol class this guard is meant to catch. The existing Read Aloud patch has a dedicated regression test for `codexLinuxReadAloudSettingsIcon=e=>...` in an imported asset requiring an explicit `var codexLinuxReadAloudSettingsIcon;` before the assignment; without that declaration the module can throw. With `[^;]*\b${symbol}\s*=` (and the `[;,] ${symbol} =` case below), a bare assignment inside a callback/minified statement is classified as a local declaration and the integrity finding is skipped. The check should distinguish actual declarators from expression assignments.

avifenesh (commented):

---

Found two issues with the new DMG integrity guard wiring/detection that could let the intended blocker miss real post-patch failures.

ilysenko:
Thanks for the PR. I found two blockers in the new integrity guard:

1. The scan is wired to the raw `--candidate` inventory, so the normal `Codex.dmg + patch-report` flow cannot see Linux-injected settings symbols from the patched webview assets. It needs to inspect the patched app/extracted ASAR/installed webview assets, or run from the patcher while those files are available.

2. `hasLocalPatchSymbolDeclaration()` treats bare assignments as declarations, so shapes like `;codexLinuxReadAloudSettingsIcon=e=>...` or `foo(),codexLinuxReadAloudSettingsIcon=e=>...` are skipped even though they have no local declarator. Please distinguish real declarators from expression assignments and add regression coverage for that shape.

Yo-DDV (commented):
Thanks for the quick follow-up. Small correction from my side: `97a2c01` landed while I was finishing my previous review, so my first note was based on `9a06ba9`.

I rechecked the new head. The bare-assignment reproducer now behaves as expected:
- unresolved reference: finding
- real `const` declaration: no finding
- `;codexLinuxReadAloudSettingsIcon=e=>e`: finding
- `foo(),codexLinuxReadAloudSettingsIcon=e=>e`: finding

Local checks are green on `97a2c01`:
- `node --test scripts/dev/upstream-dmg-intel.test.js scripts/patch-linux-window-ui.test.js` — 362 pass
- `git diff --check`

So the declaration blocker I flagged is fixed in the latest commit. I will leave the broader call to Ilya and CI, but this follow-up looks good for that specific point.

joshyorko:
@ilysenko addressed both blockers:

- The patcher now scans the patched extracted app while those webview assets exist and writes post-patch integrity findings into `patch-report.json`; DMG-intel folds those findings into `PATCH_INTEGRITY_BROKEN` even when `--candidate` is the raw DMG.
- `hasLocalPatchSymbolDeclaration()` now only accepts real function/variable declarators. Bare assignments like `codexLinuxReadAloudSettingsIcon=e=>...` and `foo(),codexLinuxReadAloudSettingsIcon=e=>...` are covered as failures.

Verified in the devcontainer with `node --test scripts/dev/upstream-dmg-intel.test.js scripts/patch-linux-window-ui.test.js` passing 362/362.

ilysenko:
- This PR touches drift-sensitive patch logic, but it still carries legacy upstream compatibility in the touched paths, including the old agent-workspace Settings visibility fallbacks and older Browser Use node_repl builder handling.
- This repo targets only the latest `CODEX.DMG`, so please remove those old DMG shapes/fallback paths in the same PR instead of carrying them forward.

ilysenko (approved):
Reviewed current head d783b6a after removing legacy upstream drift fallbacks and fixing the current DMG quit guard prelude. GitHub checks are green.

PR #699, opened 2026-07-07
Fix Linux Computer Use settings plugin metadata

joshyorko:
## Summary
- include the current computer-use-settings webview asset in the Linux Computer Use availability patch
- keep Computer Use settings availability enabled on Linux and synthesize the settings plugin card only when missing
- avoid emitting the invalid synthetic openai-bundled/plugins/computer-use marketplace path, which cannot be read by plugin detail loading

## Tests
- bash tests/scripts_smoke.sh


ilysenko:
- This PR updates drift-sensitive Computer Use webview patching for the current `computer-use-settings` bundle, but it keeps legacy bundle patterns/fallback paths (`use-model-settings`, `use-is-plugins-enabled`, `use-in-app-browser-use-availability`, `use-native-apps.electron`, and old synthetic marketplace-path cleanup). This repo targets only the latest `CODEX.DMG`, so please remove older-upstream compatibility in the same PR.

ilysenko (approved):
Reviewed current head after legacy drift cleanup; checks are green and no blockers remain.

PR #700, opened 2026-07-07
fix(remote-mobile): clear stale waiting input state

Yo-DDV:
## Summary

- tighten the Linux webview recovery around restored `waitingOnUserInput` runtime state
- keep that state active only when a real pending user-input request is still present
- keep permission waits, real user-input waits, malformed active states, and already-patched bundles behaving as before

## Why

While testing #639, I could reproduce one stale shape in the Linux webview patch path: a restored thread can still report `waitingOnUserInput` after the request list no longer contains a matching input request. In that state, remote mobile can keep looking busy even though the conversation is ready for the next message.

The permission wait path already checks the request list before surfacing that state. This adds the same kind of guard for user-input waits, while keeping unknown active shapes on the safe loading side.

Refs #639.

## Tests

- `node --check scripts/patches/impl/webview/index.js && node --check scripts/patch-linux-window-ui.test.js`
- `git diff --check`
- `node --test --test-name-pattern "remote status|empty active runtime status|upgrades already-patched" scripts/patch-linux-window-ui.test.js`
- `make check && make test`
- patcher dry-run twice against a copied installed content tree; the webview bundle contains the new helper, call-site wiring, stale pending guard, and safe optional flag handling after patching

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -5398,14 +5398,19 @@ test("recovers completed stream items that arrive after local state lost their s
 
 test("treats empty active runtime status as stale once response rendering has completed", () => {
   const source =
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Looks good. The fix is narrow, keeps real pending input/permission waits intact, and CI is green.

PR #702, opened 2026-07-08
fix(launcher): stabilize GNOME Wayland multi-monitor scaling

sanshi42:
## Summary

This PR stabilizes Codex Desktop startup on GNOME Wayland sessions with multiple monitors by selecting the X11/XWayland Ozone backend automatically when the launcher is in its default `auto` rendering mode.

## Root Cause

On GNOME Wayland multi-monitor setups, Electron 42 can pick the native Wayland backend through `--ozone-platform-hint=auto`. In that backend, moving pointer focus across displays can trigger display/scale related window state changes. On an installed package this showed up as a maximized Codex window suddenly shrinking/rescaling when the pointer moved to a side monitor, while other desktop apps remained stable.

The installed launcher already had manual escape hatches (`CODEX_OZONE_PLATFORM=x11`, `CODEX_FORCE_DEVICE_SCALE_FACTOR`, and explicit `--x11` / `--wayland` flags), but the default path still left backend selection to Electron on the affected GNOME Wayland multi-monitor layout.

## Fix

- Detect GNOME Wayland multi-monitor sessions in the launcher via connected DRM connector state under `/sys/class/drm`, avoiding any undeclared runtime tool dependency.
- In the default `auto` rendering mode, force `--ozone-platform=x11` for that layout instead of passing `--ozone-platform-hint=auto`.
- Preserve explicit user choices: `--wayland`, `--ozone-platform=*`, `CODEX_OZONE_PLATFORM=*`, and `CODEX_LINUX_RENDERING_MODE=default|wayland-gpu|wslg` still win.
- Document the fallback and the opt-out path in troubleshooting docs.
- Add a smoke test that reproduces the GNOME Wayland two-monitor launcher decision.

## Generality

This is intentionally broader than one machine but still scoped. The fallback applies to the class of GNOME Wayland sessions with more than one monitor, which is where Electron's native Wayland backend is most likely to encounter monitor-dependent scale/maximize drift. It does not change single-monitor sessions, WSLg, non-GNOME desktops, or any launch where the user explicitly requested a backend.

If future Electron/GNOME versions make native Wayland reliable for this layout, users can already opt back in with `CODEX_OZONE_PLATFORM=wayland` or `CODEX_LINUX_RENDERING_MODE=default`, and the condition can be relaxed later.

## Validation

- `bash -n install.sh && bash -n scripts/lib/*.sh && bash -n launcher/start.sh.template && bash -n scripts/build-deb.sh && bash -n scripts/build-rpm.sh && bash -n scripts/build-pacman.sh && bash -n scripts/build-appimage.sh`
- `git diff --check`
- Local package rebuild and runtime validation on a GNOME Wayland dual-monitor setup: the previous shrink/rescale behavior disappeared after the launcher selected X11/XWayland.

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -2433,6 +2433,39 @@ is_wslg_session() {
     [ -n "${DISPLAY:-}" ] && [ -e /mnt/wslg ]
 }
 
+gnome_desktop_detected() {
+    case "${XDG_CURRENT_DESKTOP:-}:${XDG_SESSION_DESKTOP:-}:${DESKTOP_SESSION:-}" in
+        *GNOME*|*gnome*) return 0 ;;
+        *) return 1 ;;
+    esac
+}
+
+connected_drm_monitor_count() {
+    local drm_root="${CODEX_DRM_CLASS_ROOT:-/sys/class/drm}"
+    local status_file status count=0
+
+    [ -d "$drm_root" ] || return 1
+
+    for status_file in "$drm_root"/*/status; do
+        [ -f "$status_file" ] || continue
+        IFS= read -r status < "$status_file" || continue
+        [ "$status" = "connected" ] || continue
+        count=$((count + 1))
+    done
+
+    [ "$count" -gt 0 ] || return 1
+    printf '%s\n' "$count"
+}
+
+gnome_wayland_multi_monitor_detected() {
+    local count
+
+    [ "${XDG_SESSION_TYPE:-}" = "wayland" ] || return 1
+    gnome_desktop_detected || return 1
+    count="$(connected_drm_monitor_count)" || return 1
+    [ "$count" -gt 1 ]
+}
+
 normalize_linux_rendering_mode() {
     case "${CODEX_LINUX_RENDERING_MODE:-auto}" in
         auto|default|wslg|wayland-gpu)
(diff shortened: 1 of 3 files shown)


launcher/start.sh.template
```diff
@@ -2433,6 +2433,44 @@ is_wslg_session() {
...
+        *GNOME*|*gnome*) return 0 ;;
+        *) return 1 ;;
+    esac
+}
+
+xrandr_monitor_count() {
+    local output first_line count
+
+    [ -n "${DISPLAY:-}" ] || return 1
+    command -v xrandr >/dev/null 2>&1 || return 1
```

---

This makes the new default launch decision depend on an `xrandr` executable being present, but the packaged artifacts don't currently guarantee that CLI: the Debian control file only depends on `libxrandr2`, the RPM/Arch metadata only list the RandR libraries, and the Nix wrapper's `launcherPath` doesn't include `xorg.xrandr`. On a minimal packaged/Nix install `command -v xrandr` will fail, `gnome_wayland_multi_monitor_detected` returns false, and a GNOME Wayland dual-monitor session still gets `--ozone-platform-hint=auto`—the path this PR is trying to avoid. Please add the appropriate xrandr runtime dependency/wrapper entry or use a signal that is already guaranteed to be available.

avifenesh (commented):

---

Found one issue in the new launcher auto-detection path: it depends on an undeclared runtime tool, so the GNOME fallback can silently never run on some installs.

sanshi42:
Updated the launcher auto-detection path to avoid the undeclared `xrandr` runtime dependency. The GNOME Wayland multi-monitor fallback now detects connected displays through `/sys/class/drm/*/status`, which is available from the kernel DRM sysfs state instead of a packaged CLI. If DRM state is unavailable or unreadable, the probe remains fail-soft and does not force the fallback.

I also added a smoke test using `CODEX_DRM_CLASS_ROOT` with a fake two-monitor DRM tree to verify that the auto profile selects `--ozone-platform=x11` and does not leave Electron on `--ozone-platform-hint=auto`.

ilysenko (approved):
Looks good. The updated launcher fallback is scoped to GNOME Wayland multi-monitor auto mode, preserves explicit rendering choices, removes the undeclared xrandr dependency, and the focused smoke coverage plus CI are green.

PR #704, opened 2026-07-08
feat(updater): add runtime diagnostics command

GTanger:
## Summary
- add `codex-update-manager diagnose` with text and `--json` output for post-update support reports
- report updater state, installed app executable, launcher pid files, local webview health, warm-start socket, build metadata, and relevant runtime paths
- route `diagnose` before updater runtime writes so the command does not create/update state or logs

## Validation
- `cargo fmt -p codex-update-manager --check`
- `cargo run -p codex-update-manager -- diagnose`
- `cargo run -p codex-update-manager -- diagnose --json`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -536,7 +536,7 @@ dependencies = [
 
 [[package]]
 name = "codex-update-manager"
-version = "0.8.6"
+version = "0.9.0"
 dependencies = [
  "anyhow",
  "chrono",
--- a/docs/updater.md
+++ b/docs/updater.md
@@ -31,9 +31,16 @@ need to pin a particular binary from a GUI-launched session.
 ```bash
 systemctl --user status codex-update-manager.service
 codex-update-manager status --json
+codex-update-manager diagnose --json
 sed -n '1,160p' ~/.local/state/codex-update-manager/state.json
 sed -n '1,160p' ~/.local/state/codex-update-manager/service.log
 ```
+
+`diagnose` is read-only and intended for post-update support reports. It checks
+the persisted updater state, installed app executable, launcher `app.pid` and
+`webview.pid`, local webview HTTP endpoint, warm-start handoff socket, and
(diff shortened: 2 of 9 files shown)


ilysenko:
Thanks for adding the diagnostics command. This needs a follow-up before merge:

1. `diagnose` currently reads `app.pid` through `liveness::app_pid_file()`, which is hard-coded to `~/.local/state/codex-desktop/app.pid`. The launcher writes pid files under the resolved app id and, for `--new-instance`, under `instances/<id>/`, so `diagnose` can report false missing/invalid pid warnings for supported runtime layouts.
2. The webview probe uses only `CODEX_WEBVIEW_PORT`/default `5175`, but the launcher uses `CODEX_LINUX_WEBVIEW_PORT` and non-default app ids default to `5176`. This can probe the wrong port in side-by-side installs.
3. This adds a new user-facing updater CLI command, so `updater/Cargo.toml` should get the appropriate version bump per `CONTRIBUTING.md`.

The command shape is useful; it just needs to share the launcher/runtime path resolution instead of partially reimplementing it.

GTanger:
Thanks for the review. I pushed `5ed7432` to address the follow-up items:

- `diagnose` now uses the same resolved app state layout as the launcher, including non-default app ids and `instances/<id>/` for multi-launch pid files.
- The webview probe now follows launcher port precedence: `CODEX_WEBVIEW_PORT`, then `CODEX_LINUX_WEBVIEW_PORT`, then the default 5175/5176 based on app id.
- Bumped `codex-update-manager` from `0.8.6` to `0.9.0` for the new user-facing CLI command.

Validation is green locally (`fmt`, `clippy -D warnings`, `cargo check`, full updater tests, script smoke tests, and a live `diagnose` run) and the GitHub CI checks are passing.

ilysenko (approved):
Looks good. The follow-up addresses the runtime-layout blockers and CI is green.

Issue #283, opened 2026-05-22
Codex CLI auto-upgrade can report failure after npm installs the new version because the updater rechecks a persisted /usr/bin/codex path


ddy314:
## Environment

- OS: Arch Linux
- Package: `codex-desktop`
- Installed desktop version: `2026.05.21.165124-1`
- Bundled updater: `codex-update-manager`
- System Codex CLI package: `openai-codex 0.132.0-1` at `/usr/bin/codex`
- npm-installed Codex CLI after the updater attempt: `0.133.0` at `~/.npm-global/bin/codex`

## Problem

The updater attempts to upgrade the Codex CLI from `0.132.0` to `0.133.0`, npm installs the new version into the user npm prefix, but the updater still reports that the installed version is unchanged.

The updater state persists the CLI path as the system binary:

```json
{
  "cli_path": "/usr/bin/codex",
  "cli_installed_version": "0.132.0",
  "cli_latest_version": "0.133.0"
}
```

After the updater attempt, the user npm prefix contains the updated CLI:

```text
~/.npm-global/bin/codex -> ../lib/node_modules/@openai/codex/bin/codex.js
```

and running the shell-resolved CLI reports:

```text
codex-cli 0.133.0
```

while `/usr/bin/codex` is still owned by the system package `openai-codex 0.132.0-1`.

## Logs

The user service reports:

```text
Error: Codex CLI upgrade finished but the installed version is still 0.132.0 instead of 0.133.0
```

`~/.local/state/codex-update-manager/service.log` also shows the attempted prelaunch upgrade:

```text
INFO codex_update_manager::codex_cli: Codex CLI is outdated; attempting prelaunch upgrade installed_version="0.132.0" latest_version="0.133.0"
```

## Expected behavior

If npm successfully installs the requested Codex CLI version into the user prefix, the updater should resolve and validate the upgraded CLI path instead of reporting a failed upgrade because the old system-managed `/usr/bin/codex` binary did not change.

## Suspected cause

The updater persists `cli_path` as `/usr/bin/codex` and uses that persisted path again after `npm install -g`. On systems where the old CLI is package-managed and npm falls back to or uses a user prefix, the new CLI lands at a different executable path, but the post-upgrade version check still reads the old system binary.

ilysenko:
[...]

PR #706, opened 2026-07-08
[codex] Handle pacman-managed CLI separately from npm updates

### LINKED ISSUE #705: Updater should not treat distro-managed Codex CLI like an npm-managed install
### Summary

Related to Issue #283 and the follow-up fix in PR #284, but this is a separate behavior issue.

The current updater/launcher CLI preflight still treats the official `npm` release stream as the primary update model even when the resolved `codex` binary is a distro-managed system install such as `/usr/bin/codex` on Arch Linux.

### Background

Issue #283 and PR #284 fixed an important path-recovery problem:

- `npm` could install a newer Codex CLI into a user prefix
- the updater could keep rechecking the old persisted `/usr/bin/codex`
- this could report a false CLI upgrade failure

That fix makes sense for `npm`-managed/user-prefix installs.

However, it still leaves a separate policy problem for system-package-managed installs.

### Problem

The following assumes Arch Linux, but the idea applies to other distros.
For a `pacman`-managed CLI, there are two distinct version signals:

1. the latest official `@openai/codex` `npm` release
2. the latest version currently available through `pacman` for the package that owns `/usr/bin/codex`

Those can legitimately diverge.

If the updater only treats `npm` upstream as the source of truth, it can still classify a `pacman`-managed CLI as outdated even when `pacman` itself does not currently offer any newer package. In that situation, the system package manager and `npm` are different ownership/update models, and the updater should not implicitly treat them as interchangeable.

### Expected behavior

When the resolved CLI is a pacman-managed system binary:

- do not auto-update it through `npm`
- detect ownership with `pacman -Qo`
- distinguish between:
  - the latest official `npm` version
  - the latest `pacman` package version / actionable `pacman` upgrade state
- if `pacman` offers a newer package, tell the user to update through `pacman`
- if `pacman` does not currently offer a newer package but `npm` upstream is newer, do not mark the CLI as requiring an update; instead show an informational message so the user can decide whether to stay on the `pacman`-managed package or switch installation channels
- if ownership cannot be determined, skip `npm` auto-update and tell the user to inspect the CLI source manually

### Why this matters

`npm install -g` and a distro package manager are different update/ownership channels.

A `pacman`-managed `/usr/bin/codex` should not be treated as if it were just another `npm`-managed CLI binary.

VirgilMing:
## Summary

- detect Arch pacman-managed Codex CLI installs and query pacman for ownership and latest package status
- keep npm-managed and standalone CLI behavior unchanged while skipping npm auto-upgrades for pacman-managed system binaries
- separate `cli_official_latest_version` from `cli_package_manager_latest_version` in updater state, status output, and docs

## Context

- Fixes #705
- Related to #283 and #284, which fixed npm path recovery after user-prefix installs; this PR addresses the separate ownership/update-model problem for pacman-managed `/usr/bin/codex`

## Scope

This change is intentionally scoped to pacman-managed CLI installs on Arch-like systems. Other distro package manager ownership/update models are left unchanged in this PR.

## Behavior

- if pacman has a newer package, `cli_status` becomes `UpdateRequired` with pacman guidance
- if pacman does not have a newer package but npm upstream is newer, `cli_status` stays `UpToDate` with informational channel-mismatch guidance
- if `pacman -Qo` cannot determine ownership, the updater skips npm auto-update and reports `Unknown`

## Validation

- `cargo fmt --all`

## Not run

- Manual Arch smoke test not run; covered by updater unit tests that simulate `pacman -Qo`, `pacman -Si`, and `pacman -Qu`
- Native package builds not run because this change is confined to updater logic and documentation

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -7,6 +7,8 @@
 | `ERR_CONNECTION_REFUSED` on the webview port | Ensure `python3` works and the configured port is free |
 | Stuck on Codex logo splash | Check `~/.cache/codex-desktop/launcher.log`; another process may be serving the webview port |
 | `CODEX_CLI_PATH` error | Reopen the app to retry automatic CLI install, or install manually with `npm i -g @openai/codex` / `npm i -g --prefix ~/.local @openai/codex` |
+| `codex-update-manager status --json` shows `cli_status: "update_required"` for `/usr/bin/codex` on Arch | Pacman itself has a newer package for the installed CLI. Update through pacman instead of npm, for example `sudo pacman -Syu`; pacman-managed CLI installs are intentionally not auto-updated through npm |
+| `codex-update-manager status --json` shows `/usr/bin/codex` with `cli_status: "up_to_date"` but `cli_official_latest_version` is newer than `cli_package_manager_latest_version` | The distro package is behind the official npm release, but pacman does not currently offer a newer package. Codex Desktop will not auto-switch channels; read `cli_error_message` and decide whether to stay on the distro-managed CLI or replace it with another install method |
 | `nix run` exits with no window or terminal output | Check `~/.cache/codex-desktop/launcher.log`; the Nix package still requires a user-provided `codex` CLI |
 | `gh auth status` works in terminal but fails inside Codex Desktop | See [GitHub CLI auth in app-launched shells](github-cli-auth.md) |
 | Electron hangs while CLI is outdated | Re-run the launcher and check `~/.cache/codex-desktop/launcher.log` plus `~/.local/state/codex-update-manager/service.log` |
--- a/docs/updater.md
+++ b/docs/updater.md
@@ -20,11 +20,36 @@ installs continue to update through npm, while official standalone installs
 under `~/.codex/packages/standalone` are updated with the official standalone
 installer instead of being replaced through npm.
 
+System-package-managed CLI installs are reused but not mutated through npm or
+the standalone installer flow. On Arch-like hosts, when the resolved CLI lives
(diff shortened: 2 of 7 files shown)


ilysenko:

- This removes the existing `cli_latest_version` field from `codex-update-manager status --json`; `serde(alias = "cli_latest_version")` only reads old state, it does not keep writing the old status field.
- Please either keep `cli_latest_version` as a backward-compatible output alias for the official latest version, or include the required updater versioning/docs for an incompatible CLI/status schema change.

Yo-DDV (commented):
Thanks for taking this one on, the direction makes sense to me.

I found one more issue while stressing the new tests, separate from the `cli_latest_version` compatibility point above. On the current PR head `9cf6944`, `RUST_TEST_THREADS=16 cargo test -p codex-update-manager` failed on run 6 for `codex_cli::tests::preflight_reports_actionable_pacman_update_without_running_npm_install`.

The failing assertion shows an extra `-Q|codex-desktop|` line in the fake pacman log. That command comes from the updater package detection path, not from the new CLI-management path. The new tests put the fake `pacman` on the global `PATH`, so another test can resolve that same fake binary while the pacman preflight test is still running.

A small fix would probably be to keep only the fake `npm` on `PATH`, place the fake `pacman` in a separate temp dir, and pass it only through `CODEX_UPDATE_MANAGER_TEST_PACMAN_PATH`. That should keep these assertions deterministic without serializing unrelated tests.

ilysenko (approved):
Reviewed the current head after the compatibility and test-isolation fixes. Local updater/smoke validation and GitHub CI are green.

PR #707, opened 2026-07-08
Fix frameless titlebar webview menu on Linux

AzureDominus:
## Summary

- update the frameless-titlebar webview patch to target the current main app chrome bundle
- remove Linux from the application-menu safe-area path for the latest upstream bundle shape
- disable the web app application-menu bridge so the File/Edit/View/Help strip is not rendered for frameless Linux windows
- update the feature test to cover the current onboarding bundle shape and exclude the older safe-area bundle path

## Root cause

The latest upstream DMG moved the relevant window chrome logic into the main onboarding bundle. The existing frameless-titlebar descriptor no longer matched that bundle, so the web app still rendered its application menu row even though the main-process titlebar/menu patches applied.

## Validation

- node --test linux-features/frameless-titlebar/test.js
- git diff --check
- local Arch/Hyprland package rebuild with frameless-titlebar enabled
- full-monitor screenshot verification after clearing Electron renderer caches confirmed the File/Edit/View/Help strip was gone
- xhigh subagent code review found no issues and confirmed the diff follows the latest-DMG drift policy


Yo-DDV (commented):
Hi, thanks for tackling this. I tested the patch locally against the current `app-initial~app-main~onboarding-page-CNHnOMz8.js` bundle. The current target does get patched, the transformation is idempotent, and the focused tests pass on my side.

One edge case looks worth tightening before this leaves draft. `hasDisabledApplicationMenuBridge` currently treats any `function x(){return!1}` in the whole bundle as proof that the application-menu bridge is already disabled. The current bundle already contains unrelated functions with that exact shape, so if the real bridge predicate drifts later, the patch can silently skip the bridge without warning.

A small regression test with an unrelated `function unrelated(){return!1}` plus a drifted bridge name should catch it. I think the already-applied check should be tied to the application-menu bridge shape, or the warning should depend on `foundApplicationMenuBridge` rather than a generic `return!1` function.

ilysenko:
Thanks for the update. One blocker remains before this leaves draft: `hasDisabledApplicationMenuBridge` still treats any `function ...(){return!1}` in the whole bundle as proof that the application-menu bridge was already disabled. A local regression probe shows that an unrelated `function unrelated(){return!1}` suppresses the drift warning even when the real bridge predicate no longer matches.

Please tie the already-applied check to the application-menu bridge shape, or make the warning depend on the actual bridge match, and add a regression test for the unrelated-return-false case.

ilysenko (approved):
Reviewed the current head after the frameless titlebar bridge drift-warning fix. Local feature/smoke validation and GitHub CI are green.

PR #708, opened 2026-07-08
Fix Linux feature symlink staging

ryan-mt:
## Summary
- reject symbolic links inside Linux feature resource sources before staging
- verify the real install target parent remains inside the app install directory
- add focused regression tests for source-link and target-parent-link staging attempts

## Root cause
Declarative Linux feature staging validated lexical relative paths, but it copied resources with fs.cpSync and then recursively chmodded the staged target. A symlinked feature source or preexisting symlinked target parent could redirect copy/chmod behavior outside the install tree.

## Validation
- node --test scripts/lib/linux-features.test.js
- node --check scripts/lib/linux-features.js && node --check scripts/lib/linux-features.test.js
- bash scripts/ci/run-node-checks.sh syntax
- git diff --check -- scripts/lib/linux-features.js scripts/lib/linux-features.test.js

Note: the broader Node suite was not green on Windows because existing tests depend on Linux chmod/symlink semantics and Linux platform probes.

### DIFF
--- a/scripts/lib/linux-features.js
+++ b/scripts/lib/linux-features.js
@@ -616,7 +616,52 @@ function chmodRecursive(target, mode) {
   }
 }
 
-function copyInstallFile(source, target, mode) {
+function pathStaysInside(root, candidate) {
+  const relative = path.relative(root, candidate);
+  return (
+    relative === "" ||
+    (relative !== ".." && !relative.startsWith(`..${path.sep}`) && !path.isAbsolute(relative))
+  );
+}
+
+function assertNoSymbolicLinks(target, label) {
+  const stat = fs.lstatSync(target);
+  if (stat.isSymbolicLink()) {
+    throw new Error(`${label} must not contain symbolic links`);
+  }
+  if (!stat.isDirectory()) {
+    return;
+  }
+  for (const name of fs.readdirSync(target)) {
+    assertNoSymbolicLinks(path.join(target, name), label);
+  }
+}
+
+function assertNoSymbolicLinksIfPresent(target, label) {
+  try {
+    assertNoSymbolicLinks(target, label);
+  } catch (error) {
+    if (error?.code !== "ENOENT") {
+      throw error;
+    }
+  }
+}
+
+function assertInstallParentInside(installDir, target, label) {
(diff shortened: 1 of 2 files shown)


Yo-DDV (commented):
Thanks for tightening this up. I think there is still one symlink escape left before the new copy guard runs.

Both cleanup paths run before `copyInstallFile`: `removePreviouslyStagedArtifacts()` removes entries from `.codex-linux/linux-features-staged.json`, and `removeLegacyDeclarativeRuntimeHooks()` scans/removes old hook dirs. Those removals currently only do lexical path validation or `stat` directory checks, so a symlinked parent under `.codex-linux` can still redirect `fs.rmSync()` outside the app dir.

I reproduced it with a staged manifest target `.codex-linux/features/unsafe-link/payload.txt` and `.codex-linux/features/unsafe-link -> /tmp/outside`: staging throws the new “must stay inside the install directory” error, but `/tmp/outside/payload.txt` is already deleted. Same shape with `.codex-linux/prelaunch.d -> /tmp/outside-hooks` deletes `unsafe-link-old-hook.sh` in the legacy cleanup path.

Could we apply the same realpath/no-symlink containment check to cleanup before any `rmSync`, and add tests for an existing staged manifest plus a symlinked legacy hook dir? That should close the remaining pre-staging gap.

ilysenko:
Thanks for tightening the copy/chmod path here. There is still a blocker before this can merge: both cleanup paths can still delete through symlinked parents before the new copy guard runs.

1. `removePreviouslyStagedArtifacts()` / `removeInstallRelativePath()` can follow a symlinked staged manifest parent and `rmSync` a file outside the app dir before staging later throws.
2. `removeLegacyDeclarativeRuntimeHooks()` uses `stat`-based directory checks, so a symlinked `.codex-linux/prelaunch.d` can redirect cleanup outside the app dir and staging may still succeed.

Could you apply the same realpath/no-symlink containment protection before any cleanup `rmSync`, and add regression tests for a stale manifest target plus a symlinked legacy hook directory?

ilysenko (approved):
Reviewed the current head after the cleanup hardening fix. Local targeted/smoke validation and GitHub CI are green.

PR #711, opened 2026-07-09
Document and include optional Codex CLI deps

### LINKED ISSUE #710: CLI self-managed install silently drops @openai/codex-linux-x64 — broken tool calls with no clear error
**Heads up: `npm install -g @openai/codex` silently drops the platform binary**

**CLI self-managed install silently drops `@openai/codex-linux-x64` — broken tool calls with no clear error**

If you manage the Codex CLI yourself rather than letting the bundled runtime handle it, a global `npm install -g @openai/codex` can complete without error but leave the install incomplete — specifically, the optional dependency `@openai/codex-linux-x64` is not placed in the correct nested location.

The result is either Codex failing to start with:
```
Error: Missing optional dependency @openai/codex-linux-x64.
```
or — more insidiously — starting but producing broken tool calls (namespace doubling, malformed output), with no obvious error pointing to the missing binary.

The fix is to run `npm install --include=optional` from inside the installed package directory after the global install:

```bash
npm install -g @openai/codex@latest
cd ~/.nvm/versions/node/$(node -v)/lib/node_modules/@openai/codex
npm install --include=optional
```

This places `codex-linux-x64` in the correct nested `node_modules/@openai/` path. Verified working on Fedora 44 / nvm / Node v24.14.1, Codex CLI 0.143.0.

---

**A note on the rollback advice circulating on GitHub**

If you've been hitting tool-calling issues and have seen advice to roll back to 0.142.5 — please try the fix above first. The rollback advice appears to be cargo-culted: it gets repeated because someone said it worked once, but if your install was incomplete to begin with, no amount of version-switching will fix it. The missing binary is present in both 0.142.5 and 0.143.0 when the install is done correctly, and the tool-calling breakage disappears once it is.

It may be worth adding a row to `docs/troubleshooting.md` covering this, and/or adding `--include=optional` to the documented manual CLI install path.

Thanks for the port Gary - you did what Mr Altman failed to do, so kudos to you.

ilysenko:
Summary
- Add --include=optional to updater-managed npm CLI installs.
- Update launcher and docs manual CLI install guidance so Linux platform optional deps are installed.
- Add troubleshooting guidance for missing @openai/codex-linux-x64 or malformed tool calls after self-managed npm installs.

Tests/checks
- cargo fmt -p codex-update-manager -- --check
- cargo test -p codex-update-manager codex_cli
- git diff --check

Notes
- Fixes #710.


Yo-DDV:
Thanks, this direction matches what I saw locally for #710.

I checked it with npm 11.16.0 on Linux x64. A normal isolated install did bring the platform package on my side, so I could not reproduce the silent omission with the default npm config here. After deleting `node_modules/@openai/codex-linux-x64`, though, the CLI fails immediately on the missing optional dependency path, and `npm install --include=optional` from the installed package restores it.

I also tested the documented `--include=optional --prefix` install form, plus the updater test suite locally. The PR CI is green too. I do not see a reason to open a competing fix from my side.

PR #714, opened 2026-07-09
Fix managed Node launcher PATH setup

ryan-mt:
## Summary

- tolerate an unset or empty inherited `PATH` when prepending the managed Node runtime
- keep the managed Node directory first even if it already appears later in `PATH`
- add a launcher regression probe for both cases

## Root cause

The generated launcher enables `set -u`, but `prepend_managed_node_runtime_to_path` expanded bare `$PATH`. With no inherited `PATH`, startup aborted with `PATH: unbound variable`. The old containment check also skipped prepending when the managed directory appeared later, allowing an earlier executable to win.

## Validation

- focused launcher regression probe: pass
- manual unset, normal, and later-entry PATH probes: pass
- `bash -n install.sh launcher/start.sh.template tests/scripts_smoke.sh`: pass
- `git diff --check`: pass
- full `tests/scripts_smoke.sh`: blocked on this Windows host by pre-existing POSIX mode semantics (expected 755, observed 777)

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -564,13 +564,14 @@ export_packaged_runtime_env() {
 prepend_managed_node_runtime_to_path() {
     [ -x "$MANAGED_NODE_BIN_DIR/node" ] || return 0
 
+    local current_path="${PATH:-}"
     if [ -z "${CODEX_LINUX_USER_PATH+x}" ]; then
-        export CODEX_LINUX_USER_PATH="${PATH:-}"
+        export CODEX_LINUX_USER_PATH="$current_path"
     fi
 
-    case ":$PATH:" in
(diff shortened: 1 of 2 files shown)


ilysenko:
- The new smoke probe inherits `CODEX_LINUX_USER_PATH` from the parent environment but expects it to be empty. This makes `tests/scripts_smoke.sh` fail when run from Codex Desktop. Please unset or otherwise isolate `CODEX_LINUX_USER_PATH` before the first probe call and rerun the combined smoke suite.

ilysenko (approved):
Reviewed the updated head. The inherited CODEX_LINUX_USER_PATH smoke-test issue is fixed, the focused regression probe passes in the Codex environment, and all required CI checks are green.

PR #715, opened 2026-07-09
Reject overlapping Linux feature targets

ryan-mt:
## Summary

- reject duplicate and normalized-alias install targets across enabled Linux features
- reject ancestor/descendant overlaps between resources and runtime hooks
- reserve the framework-owned staged manifest path
- fail during plan construction before cleanup or copy can mutate the existing app tree

## Root cause

The install plan tracked resources and runtime hooks independently without target ownership. Staging copied entries sequentially with `force: true`, so later entries could overwrite earlier artifacts while the staged manifest still recorded both. Parent/child overlaps could also fail after cleanup and leave a partial install.

## Validation

- `node --test scripts/lib/linux-features.test.js`: 10/10 pass
- actual `linux-features.js --stage-install` duplicate-target CLI repro: exits 1, preserves prior staged state
- valid distinct resource/hook CLI plan: stages successfully
- `bash scripts/ci/run-node-checks.sh syntax`: pass
- `node --check` and `git diff --check`: pass
- broader Linux feature/patch suites still contain pre-existing Windows-host chmod, XDG, and Linux-runtime failures unrelated to this diff

### DIFF
--- a/scripts/lib/linux-features.js
+++ b/scripts/lib/linux-features.js
@@ -560,9 +560,26 @@ function modeString(mode) {
 function enabledLinuxFeatureInstallPlan(options = {}) {
   const resources = [];
   const runtimeHooks = [];
+  const installTargetOwners = new Map([
+    [STAGED_FEATURE_MANIFEST_RELATIVE_PATH, "Linux feature staging framework"],
+  ]);
+  const claimInstallTarget = (target, owner) => {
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head and its merge result against the latest main. The collision checks fail before staging mutations, targeted and full Linux feature tests pass, and CI is green.

PR #720, opened 2026-07-09
Harden latest DMG main-process patches

moxw:
## Summary

- follow up on #716 with current-only main-process matching for Codex 26.707.30751 and Electron 42.1.0
- preserve nullish BrowserWindow option semantics, current tray/icon resolution, avatar overlay input handling, and explicit quit lifecycle behavior
- keep repeated full-bundle patching idempotent after the Electron require wrapper is installed
- remove the obsolete `linux-tray-close-setting` descriptor and its legacy `canHideLastLocalWindowToTray` compatibility path

## User-visible behavior

- primary and overlay windows retain valid Linux options when upstream values are null or no icon asset is available
- the native titlebar, avatar overlay, tray, close-to-tray, and explicit quit patches match the current upstream bundle without guessed aliases
- rebuilding an already patched bundle no longer changes tray bindings on the second pass

## Source of truth

- current bundle matching lives in `scripts/patches/impl/main-process/`, `scripts/patches/impl/avatar-overlay.js`, and `scripts/patches/lib/minified-js.js`
- the obsolete descriptor removal is in `scripts/patches/core/all-linux/main-process/launch-actions/patch.js` and `scripts/patches/impl/launch-actions.js`
- regressions are covered in `scripts/patch-linux-window-ui.test.js` and `tests/scripts_smoke.sh`

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` (335 passed)
- `node --test linux-features/*/test.js` (402 passed)
- `MAX_BUILD_THREADS=4 ./install.sh ./Codex.dmg` against app 26.707.30751 / Electron 42.1.0
- second enforced real-DMG patch pass: all 17 required patches `already-applied`; main bundle SHA-256 remained byte-identical
- `git diff --check`, changed JavaScript syntax checks, and `bash -n tests/scripts_smoke.sh`

## Environment and limits

- validated on Arch Linux x86_64
- the full production package artifacts were not rebuilt after the final review fix; the smoke suite exercised Debian, RPM, pacman, and AppImage packaging paths
- the real-DMG report has no required failures; the existing optional Computer Use install-flow bundle patch remains skipped because the current bundle is absent

## Risk

The change is limited to latest-DMG main-process patch matching and its fixtures. No optional feature is enabled or added.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -184,14 +184,10 @@ const fileManagerBundle =
   "var lu=jl({id:`fileManager`,label:`Finder`,icon:`apps/finder.png`,kind:`fileManager`,darwin:{detect:()=>`open`,args:e=>il(e)},win32:{label:`File Explorer`,icon:`apps/file-explorer.png`,detect:uu,args:e=>il(e),open:async({path:e})=>du(e)}});function uu(){}";
 const terminalEnvBundle =
   "var Q0=`xterm-256color`;var t={t(e){return e}};var Backend=class{isLocalTerminalSession(e){return e?.type===`local`}async getWorktreeShellEnvironmentForCwd(e){return null}async buildTerminalEnv(e,n,r){let i={...process.env};if(n!=null&&(i.CODEX_APP_TITLE=n),this.isLocalTerminalSession(r)){let t=await this.getWorktreeShellEnvironmentForCwd(e);if(t!=null){for(let e of t.exclude)delete i[e];Object.assign(i,t.set)}}return process.platform!==`win32`&&(i.TERM=Q0,delete i.TERMINFO,delete i.TERMINFO_DIRS),t.t(i)}};";
-const alreadyOpaqueBackgroundBundle =
-  "process.platform===`linux`?{backgroundColor:e?t:n,backgroundMaterial:null}:{backgroundColor:r,backgroundMaterial:null}";
-const opaqueBackgroundBundleWithDriftingGw =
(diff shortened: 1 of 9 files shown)


ilysenko:
Thanks for the update. I found two blockers:
- `window.js` still keeps obsolete older-DMG branches (`oldLinuxBgPatch` and the pre-26.623 About icon shapes), then treats the shape used by current 26.707 as a 26.623 fallback. Please keep only the current `opaqueWindowSurfaceEnabled` and About icon shapes, in line with the latest-DMG-only policy.
- The current no-icon About path is not idempotent: a second `applyLinuxAboutDialogPatch(source, null)` pass rewrites existing `x == null || x.isEmpty()` guards into duplicated null guards. Please make the second pass byte-identical and cover it with the current-DMG fixture using `iconAsset = null`.

moxw:
Addressed both blockers in d5dfcc7.

- `window.js` now targets only the current `opaqueWindowSurfaceEnabled` helper and current About icon promise shape. The old `opaqueWindowsEnabled`/`oldLinuxBgPatch` paths and pre-26.623 About branches are removed.
- The no-icon and bundled-icon About paths are byte-identical on a second pass. Preflight and replacement use the same alias-aware shapes, so drift is reported without partial mutation.
- Added regressions for obsolete About input, mismatched aliases, missing opaque-surface predicates, and unrelated Linux background branches.

Validation: 336 patcher tests, 402 feature tests, full script/package smoke suite, real Electron 42.1.0 DMG rebuild, and an enforced second pass on the repacked ASAR with an unchanged main-bundle SHA-256. All required patches reported `already-applied` on that second pass.

Yo-DDV:
Hi, thanks for the quick follow-up.

I rechecked `d5dfcc7` against the two blockers Ilya called out. The current `window.js` path looks narrowed to the current `opaqueWindowSurfaceEnabled` and About icon shapes now. The old `oldLinuxBgPatch` / pre-26.623 About handling no longer appears in the implementation; the old About shape is only kept as a negative fixture, and the no-icon About path stayed byte-identical on a second pass in a small local check.

Local checks are green on my side: `node --test scripts/patch-linux-window-ui.test.js`, `bash tests/scripts_smoke.sh`, `make check`, and `make test`. The live PR checks are green too, including the upstream DMG build and package jobs on `d5dfcc7`.

I do not see another blocker in the paths I checked. The one limit is that I did not run a separate local real-DMG rebuild, so I am relying on the green upstream-build CI for that part.

ilysenko (approved):
Looks good. I rechecked the addressed blockers on the current head and the required checks are green.

PR #721, opened 2026-07-09
Repair missing Codex CLI optional dependency

ryan-mt:
## Summary

- repair an existing npm-global Codex CLI when its Linux platform optional dependency is missing
- verify npm provenance before running the sibling npm in the installed package directory, preserving the exact CLI path and install channel
- fail closed for arbitrary executables, system prefixes, standalone installs, Bun, and pnpm; persist repair/update command failures as `failed`

## Verification

- `cargo test -p codex-update-manager codex_cli::tests::` (29 passed)
- `cargo test -p codex-update-manager status_preserves_cli_reconciliation_failure` (1 passed)
- actual WSL binary QA: exact upstream error repaired without `--allow-install-missing`; original CLI path retained; arbitrary/Bun/pnpm layouts made no npm call; failed repair persisted `cli_status=failed`
- full updater suite: 212 passed; three pre-existing WSL/NTFS `feature_picker` failures remained unchanged

## Scope

Automatic repair requires a verified paired-prefix npm layout with `<prefix>/bin/npm`, covering standard npm and nvm installs. Custom prefix installs without a sibling npm continue to use the documented manual repair command.

Total diff: 494 changed lines.

### DIFF
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -3617,7 +3617,11 @@ mod tests {
         }
 
         assert!(result.is_err());
-        assert_eq!(state.cli_status, CliStatus::Updating);
+        assert_eq!(state.cli_status, CliStatus::Failed);
+        assert!(state
+            .cli_error_message
+            .as_deref()
+            .is_some_and(|message| message.contains("npm")));
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head. The repair is restricted to verified npm-managed CLI layouts, fails closed for other installation channels, persists failures correctly, and all required checks are green.

PR #724, opened 2026-07-09
Add optional Omarchy theme integration

dbech:
## Summary

- add an optional, default-disabled Omarchy theme integration
- install the Codex theme template under the current Omarchy config path
- serve the generated CSS through a bounded loopback stylesheet endpoint
- refresh the renderer stylesheet without restarting the app
- bound automatic `omarchy theme refresh` so a hung CLI cannot block app launch

## Validation

- `node --test linux-features/omarchy-theme/test.js` (5/5)
- `bash -n linux-features/omarchy-theme/install-template.sh`
- `python3 -m py_compile launcher/webview-server.py`
- `git diff --check`
- GitHub CI pending for the updated head

## Notes

The integration remains opt-in. It follows the current Omarchy path at `~/.config/omarchy`; the unsupported config-home override was removed. A direct `CODEX_LINUX_WEBVIEW_USER_STYLESHEET` override remains available for trusted CSS files.

### DIFF
--- a/launcher/webview-server.py
+++ b/launcher/webview-server.py
@@ -10,6 +10,10 @@ import sys
 import urllib.parse
 
 
+USER_STYLESHEET_ENDPOINT = "/__codex_user_stylesheet.css"
+MAX_USER_STYLESHEET_BYTES = 256 * 1024
+
+
 def _install_parent_death_signal():
     # Ensure the kernel terminates this process if the launcher (parent) exits
     # without invoking its cleanup trap (SIGKILL, OOM, crash). Without this,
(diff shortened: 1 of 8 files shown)


Yo-DDV (commented):
Hi Daniel, thanks for sending this. I found one case that still looks broken around the config override.

The PR documents `CODEX_OMARCHY_CONFIG_HOME`, and the webview server now serves CSS from that override-aware path. The refresh step still calls `omarchy theme refresh`, though. On current Omarchy, `omarchy-theme-refresh` and `omarchy-theme-set` write under `$HOME/.config/omarchy`, so with `CODEX_OMARCHY_CONFIG_HOME=/tmp/custom` I end up with CSS generated in the default Omarchy dir while Codex keeps serving `/tmp/custom/current/theme/codex-desktop.css` as empty.

The local tests are green, but the fake `omarchy` in `test.js` writes to `CODEX_OMARCHY_CONFIG_HOME`, so it misses this shape. I think either the override should be dropped from the documented behavior, or the prelaunch script and test should make the refresh path match what gets served.

dbech:
> Hi Daniel, thanks for sending this. I found one case that still looks broken around the config override.
> 
> The PR documents `CODEX_OMARCHY_CONFIG_HOME`, and the webview server now serves CSS from that override-aware path. The refresh step still calls `omarchy theme refresh`, though. On current Omarchy, `omarchy-theme-refresh` and `omarchy-theme-set` write under `$HOME/.config/omarchy`, so with `CODEX_OMARCHY_CONFIG_HOME=/tmp/custom` I end up with CSS generated in the default Omarchy dir while Codex keeps serving `/tmp/custom/current/theme/codex-desktop.css` as empty.
> 
> The local tests are green, but the fake `omarchy` in `test.js` writes to `CODEX_OMARCHY_CONFIG_HOME`, so it misses this shape. I think either the override should be dropped from the documented behavior, or the prelaunch script and test should make the refresh path match what gets served.

Yea I think we drop the override. I'll adjust shortly.

Thanks for reviewing!

dbech:
Thanks!

Issue #713, opened 2026-07-09
ChatGPT Work


mnaser:
@ilysenko

Are you planning to do some work/efforts to bring ChatGPT work into Linux like you have done here?  Just wondering where you're at on this and if you have the time to get through with this.

cowboycodr:
Crazy OpenAI's "AGI" can't adapt their electron app to Linux

EricKrouss:
@mnaser I'm working on refactoring the code for ChatGPT's DMG right now. Keep an eye on my fork :)

mnaser:
I got it to run successfully, I'm working on cleaning up the patches and should have a PR soon, hopefully.

mnaser:
i'll pause my efforts, I see #716 is up! :)

ilysenko:
It’s working locally right now so it should work. We’ll still need to fix all the Linux features and related things later, but overall it should be okay.

ilysenko:
I am already using the latest Codex, and for my routine programming tasks, everything is working fine.

shawnshaddock:
Will the codex-update-manager auto update this when ready, or is this going to be a completely different app/pacakge?

ilysenko:
To be honest, I don't remember it anymore.
It would be best to update the repository locally and run the native install, or whatever installation method you used.

Yo-DDV:
Thanks for keeping the discussion grounded. I’ll add my view here, since this touches the boundary of the project.

I agree with the direction kept so far in this repo. The current repo already does the right thing: it helps Linux users run the official OpenAI desktop app from their own authorized copy, with local Linux compatibility glue around it. In practice, that works well as-is, and it keeps the line clean.

OpenAI is still OpenAI. The app code, the product, the name, the marks, the binaries, and the assets are not ours. The MIT license here covers this wrapper and the Linux packaging work, not OpenAI’s own software.

So I would be very careful with anything described as a new “port” of ChatGPT or Codex. If that means maintaining compatibility with the current upstream app bundle, fine. If it means publishing, cloning, redistributing, or rebranding OpenAI-owned parts as if they were community-owned code, that is a different thing entirely, and I do not think this repo should go anywhere near it.

People can take their own risks elsewhere, but I would not expect public repositories that cross that line to stay up for long. More importantly, that is not what this project has been useful for. The useful part is much narrower: keep the Linux wrapper working, respect OpenAI’s ownership, and avoid pretending this is an independent ChatGPT/Codex desktop project.

[...]

PR #727, opened 2026-07-09
Fix Linux native app usage query drift

Yo-DDV:
## Summary

- related to #713
- keep Linux native desktop app usage queries enabled when the Computer Use labels are split out of the webview chunk
- detect the native-apps hook by its returned `{ nativeApps, isLoading }` shape instead of relying only on localized label strings
- scope the platform-gate rewrite so adjacent macOS/windows gates in the same chunk stay untouched
- add regression tests for the unlabeled chunk shape and adjacent-gate case while keeping the unrelated-query guard in place

## Validation

- `timeout 300 bash tests/scripts_smoke.sh`
- `git diff --check`

## Notes

This is intentionally scoped to the existing Computer Use native-apps hook. It does not change the Linux native apps backend, tray behavior, rate-limit footer, packaging, or feature opt-in rules.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -6524,6 +6524,22 @@ test("enables native app usage queries on Linux when Computer Use labels move ou
   assert.doesNotMatch(patched, /a=n&&\(r===`macOS`\|\|r===`windows`\)/);
 });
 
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Looks good. I rechecked the current head: the native-apps hook detection is scoped, adjacent platform gates stay untouched, and required checks are green.

Issue #726, opened 2026-07-09
remote-mobile-control requires ChatGPT auth after successful Desktop re-login


SOSANA:
## Summary

Codex Mobile setup cannot enable remote control on the Linux wrapper after signing out and back into the same ChatGPT account. The app-server rejects the enable request with `remote control requires ChatGPT authentication`.

This resembles the remote-mobile patch drift tracked in #537, but occurs on the currently installed working package and prevents enrollment entirely.

## Environment

- Debian 13, GNOME, X11
- Wrapper package: `2026.06.18.014837+31d8e266`
- Upstream Desktop app version: `26.611.62324`
- Codex CLI: `0.144.0`
- `remote-mobile-control` and `remote-control-ui` enabled
- Desktop app server launched with `--remote-control`

## Reproduction

1. Open Codex Desktop and choose Set up Codex Mobile.
2. Choose Allow devices to control this computer.
3. When it fails, sign out of Codex Desktop, restart it, and sign back in with the same ChatGPT account used on mobile.
4. Repeat the Allow action.

## Actual behavior

The UI reports `Couldn't enable remote control. Try again`. No environment is registered and mobile discovery remains unavailable.

Relevant launcher output after the re-login (tokens omitted):

```text
method=remoteControl/enable errorCode=-32600
error={"code":-32600,"message":"remote control requires ChatGPT authentication"}
Failed to sync local remote control enablement
connectionCount=0
```

At the same time, the app server runs with `CODEX_HOME=/home/sosana/.codex`, and `codex login status` reports `Logged in using ChatGPT`.

## Update state

A newer wrapper candidate (`2026.07.09.192548+4ee90314`) is available but cannot rebuild. Its installer reports skipped remote-control enablement, hydration, and status-read patches, then aborts on critical patch failures.

## Expected behavior

After Desktop re-login, `remoteControl/enable` should register the local environment and advance to mobile pairing rather than rejecting the running app server as unauthenticated.

## Notes

No credentials, cookies, account identifiers, or raw logs are attached.

SOSANA:
## Follow-up: feature-documentation diagnostic

I checked the `linux-features/remote-mobile-control/README.md` after filing this report. The documented managed standalone remote-control runtime is absent on this installation:

```text
~/.codex/packages/standalone/current/codex: missing
codex-remote-control.service: inactive
```

[...]

PR #728, opened 2026-07-09
Fix remote mobile patch targets for current chunks

Yo-DDV:
## Summary

The `26.707.31428` unified app moved the conversation manager signals, completed-item path, remote-control load gate, and status handling into the `b76hmflu` chunk. This updates the affected core and opt-in `remote-mobile-control` selectors while keeping the previous `worktree-init-v2` and `glxlkd48` forms supported. The older `thread-app-shell` fixture remains intentionally unsupported.

The PR also adds a feature-enabled upstream-DMG check. It now fails unless the relevant core patches and remote-mobile descriptors, including the current status-wait patch, are present and successful. That workflow runs only when the remote-mobile feature or its shared patch/report infrastructure changes.

`remote-mobile-control` remains disabled by default. This is still a partial repair for #726; the standalone runtime handoff and ChatGPT-auth path are outside this change.

## Validation

- 814 Node tests
- script smoke suite
- `cargo check`
- 217 Rust tests
- feature-enabled inspect against the `26.707.31428` DMG with exact patch-report validation
- isolated full app build with the same feature config
- syntax check of the patched `b76hmflu` bundle
- raw-DMG selector check: one matching asset per affected descriptor, with no conversation/projectless overlap
- second patch pass: unchanged tree hash, with every affected descriptor reporting `already-applied` without warnings

### DIFF
--- a/.github/workflows/upstream-build-app.yml
+++ b/.github/workflows/upstream-build-app.yml
@@ -10,6 +10,7 @@ on:
       - scripts/patch-linux-window-ui.js
       - scripts/patch-linux-window-ui.test.js
       - scripts/patches/**
+      - linux-features/remote-mobile-control/**
       - scripts/ci/validate-patch-report.js
       - .github/workflows/upstream-build-app.yml
   push:
@@ -21,6 +22,7 @@ on:
       - scripts/patch-linux-window-ui.js
       - scripts/patch-linux-window-ui.test.js
       - scripts/patches/**
+      - linux-features/remote-mobile-control/**
       - scripts/ci/validate-patch-report.js
       - .github/workflows/upstream-build-app.yml
   workflow_dispatch:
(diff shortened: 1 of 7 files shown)


SOSANA (commented):
Thanks for narrowing this to the current chunk drift. The selector changes and negative fixtures look consistent, but I need one verification change before treating it as a proven repair for the remote-mobile feature.

The green upstream-build workflow runs `make build-app` with optional features disabled, so it does not exercise the changed `remote-mobile-control` descriptors against the downloaded DMG. The unit tests use synthetic asset names. Please add a PR-visible real-DMG check with `remote-mobile-control` explicitly enabled and assert that the relevant patch-report entries apply or are already applied, rather than merely being skipped.

Separately, this should remain a partial repair for #726. The currently reproduced blocker is `remoteControl/enable` returning `remote control requires ChatGPT authentication`; the app-server args and standalone daemon/auth handoff are intentionally out of scope here. Please do not close #726 from this PR unless that path is verified separately.

Yo-DDV:
Thanks for the review. This was a real gap: the previous green upstream build did not exercise the opt-in remote-mobile descriptors.

I pushed cbb6c58 with a CI-visible upstream DMG inspect that enables `remote-mobile-control` and fails unless the relevant core and feature patch-report entries are `applied` or `already-applied`. That check is now green on the PR.

I also kept this scoped as a partial #726 repair. The app-server args, standalone daemon handoff, and auth path remain out of scope here.

SOSANA (approved):
Approved: current head cbb6c58 adds a real upstream-DMG check with remote-mobile-control enabled and requires the retargeted patch-report entries to succeed. This remains a partial repair only; #726 stays open for the separate ChatGPT-auth and standalone-runtime handoff.

ilysenko:

Yo-DDV:
Thanks for the nudge. I rebased this onto the current main branch and did another full pass against the `26.707.31428` app.

That pass caught two integration details worth fixing before review: the new status-wait descriptor was not part of the feature-enabled CI assertion yet, and the unified test fixture needed to model `b76hmflu` as one shared bundle instead of writing it twice. I also narrowed the workflow trigger so unrelated Linux features do not start this DMG build.

The full local suite, current-DMG inspect, isolated app build, syntax check, and second-pass idempotence check are green. I have marked the PR ready for review while the refreshed CI runs on this head.

PR #731, opened 2026-07-09
Fix Linux settings page component resolution

colonelpanic8:
## Summary

- validate settings-row candidates by component behavior before importing them
- reject icon-only toggle assets and infer the real toggle from settings call sites
- fall back to generated Linux components when upstream candidates are incompatible
- add a regression fixture for the current DMG decoy asset names

## Root cause

The current DMG includes `settings-row-disclosure-*` and `toggle-left-*` assets. Filename-only discovery selected the disclosure initializer as `SettingsRow` and a Lucide icon as `Toggle`, causing the Linux settings route to fail when rendered.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` (328/328 passing)
- focused current-DMG decoy regression test
- `node --check scripts/patches/impl/keybinds-settings.js`
- `git diff --check`
- verified dependency resolution against the installed current-DMG assets


ilysenko (approved):
Looks good. I rechecked the current head: component candidate validation avoids the current decoy assets, fallback behavior is covered, and required checks are green.

Issue #709, opened 2026-07-09
Proposal: optional Linux pet overlay feature


Gabriel-Kahen:
## Summary

I have a working local Linux feature that makes the Codex pet/avatar behave like a persistent desktop overlay. The current Linux avatar overlay patches cover hit-testing and some window behavior, but on Hyprland/Wayland the pet still needed compositor-aware handling to feel like a real desktop companion.

Working prototype: https://github.com/Gabriel-Kahen/codex-pet-linux-patch

## What the prototype fixes

- removes the visible pet window border on Hyprland
- keeps the pet above other windows and visible across workspaces where supported
- preserves manual pet position instead of snapping back on clicks/tab switches
- keeps the overlay transparent instead of painting black
- anchors the mascot itself to the corner while placing the tray/text above-left
- keeps the existing pet selector usable

## Suggested upstream shape

This probably should not become default behavior for every Linux user. It seems like a good fit for a disabled-by-default Linux feature, for example `linux-features/pet-overlay`, with compositor-specific behavior guarded by settings.

The prototype includes a personal custom pet/default-pet setup, but that part is not the upstream request. The reusable part is the optional Linux overlay behavior.

## Validation

The prototype is working locally after relaunch/reboot testing on a Hyprland/Wayland setup, and its test suite is passing in GitHub Actions.

Yo-DDV:
Thanks for putting this together. I pulled the prototype locally and tested it against current main and a copied installed app bundle.

It does work as a local feature. The loader picks it up, the env hook is staged, and the patch descriptors apply on the copied app bundle without breaking syntax. Your CI is green as well.

I would still avoid upstreaming the repo in its current shape. It mixes overlay behavior, selector changes, a custom Los pet, a default pet change, and an installer that patches an already built app.asar. For this project I think the cleaner path would be a much smaller disabled by default linux-features/pet-overlay feature, without the installer, without the custom pet, and without changing the default pet.

The Hyprland approach is worth exploring. Using execFile is the right direction. I would tighten the window matching before merging anything though, especially the class and title fallback when several Codex windows are floating. I would also add tests around missing hyprctl, invalid client JSON, multiple matching windows, no match, and settings that turn compositor handling off.

So I am in favor of the idea, but I would split it down before a PR. Happy to help shape the narrower version.

ilysenko:
[...]

PR #732, opened 2026-07-09
feat: add optional Linux pet overlay feature

Yo-DDV:
Adds a disabled-by-default `linux-features/pet-overlay` feature for the avatar overlay behavior discussed in #709.

The PR keeps this out of the core Linux patch and does not include the prototype installer, custom pet, default-pet change, or selector changes. The feature adds an optional main-bundle patch plus a launcher default that keeps GPU compositing enabled without overriding an explicit user choice.

The Hyprland path matches the exact overlay title, current process, and window geometry before applying compositor actions. It also rejects stale windows and malformed client data, preserves manual placement during drag, supports compositor-aware locked placement, and keeps the legacy dispatcher fallback bounded. The main-bundle patch applies atomically: if a required current-bundle match drifts, it leaves the bundle unchanged instead of retaining partial changes.

@Gabriel-Kahen tested the feature in a separate Hyprland GUI instance and contributed 79d5c3c. They subsequently confirmed physical mouse dragging and position stability across Codex windows on the updated branch. The patch was also validated against the current upstream DMG. The normal upstream workflow remains feature-off so this optional feature cannot block unrelated core changes.

Refs #709

Tests:
- `node --test linux-features/pet-overlay/test.js` (33/33)
- `cargo fmt --check`
- Semgrep JavaScript rules (0 findings)
- GitHub CI package builds for Debian, RPM, pacman, and Nix
- current upstream DMG patch application on the pre-hardening-equivalent main-bundle patch


Yo-DDV:
For reviewers, this PR intentionally stops at the optional overlay behavior.

I did not bring over the prototype installer, the Los pet, the default-pet change, or selector changes. The part I expect the most review on is the Hyprland matching logic, so the tests focus on ambiguous and failure cases instead of only the happy path.

Gabriel-Kahen:
I tested this draft on Hyprland in a separate Codex instance and put the follow-up hardening in [Gabriel-Kahen/codex-desktop-linux@79d5c3c](https://github.com/Gabriel-Kahen/codex-desktop-linux/commit/79d5c3c).

It keeps the scope agreed in #709: disabled by default, contained in `linux-features/pet-overlay`, and no installer, custom pet, default-pet, selector, or core Linux patch changes.

The follow-up tightens same-process window matching and stale callback handling, makes locked placement compositor-aware, respects disabled workspace/top-order settings, expands drag/layout and failure coverage, and adds the feature to the upstream-build workflow.

Verified:
- full Node suite: 798/798 passing
- pet-overlay suite: 28/28 passing
- real build against the current Codex DMG with the feature patch applied
- separate GUI instance: pet switching and selection persistence across relaunch, borderless overlay, pinning, all-workspace visibility, and position stability through clicks/settings/workspace changes

One remaining manual check: synthetic input could not trigger Electron's native Wayland drag, so physical mouse-drag persistence should be confirmed before taking the PR out of draft.

Happy for this commit to be cherry-picked, or I can open a small PR against `Yo-DDV:feature/pet-overlay`.

Yo-DDV:
Hi @Gabriel-Kahen, thanks for testing this on Hyprland and for putting the follow-up together.

I pulled 79d5c3c into the branch with its original attribution. I also tightened two fallback cases while going through it. Legacy `movewindowpixel` and `alterzorder` arguments now stay grouped as Hyprland expects, and a timed-out Lua dispatch no longer falls through to a toggling legacy `pin`. The locked-drag test now exercises a live window instead of returning through an unrelated guard.

The pet-overlay suite is 30/30 locally, the full Node suite is 800/800, and the packaging smoke suite passes. The upstream-DMG check is now running from the updated branch.

I'll keep the PR in draft for the physical mouse-drag persistence check you called out. Thanks again for doing the real compositor pass.

Gabriel-Kahen:
@Yo-DDV, I tested the updated build on Hyprland. Everything works as intended, including physical pet dragging and position stability across Codex windows. Thanks for the follow-up fixes.

ilysenko:

PR #734, opened 2026-07-09
Require webview entrypoint before launch

### LINKED ISSUE #729: ArchLinux - UI does not load at all (freshly built)
The UI does not work at all, it just shows "Directory listing for /"

Image

Yo-DDV:
## Summary

- fail the webview extraction step when upstream no longer provides `webview/index.html`
- reject package layouts that would ship without `content/webview/index.html`, including the RPM builder path
- stop cold launch before Electron when the bundled local webview is incomplete, while keeping the explicit `ELECTRON_RENDERER_URL` override path working

Fixes #729.

## Tests

- `bash -n launcher/start.sh.template scripts/lib/webview-install.sh scripts/lib/package-common.sh scripts/build-rpm.sh tests/scripts_smoke.sh`
- `shellcheck -s bash -S error launcher/start.sh.template scripts/build-rpm.sh scripts/lib/package-common.sh scripts/lib/webview-install.sh tests/scripts_smoke.sh`
- `semgrep scan --metrics=off --config p/secrets --error launcher/start.sh.template scripts/build-rpm.sh scripts/lib/package-common.sh scripts/lib/webview-install.sh tests/scripts_smoke.sh`

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -2219,12 +2219,35 @@ adopt_discovered_webview_server() {
     return 1
 }
 
+renderer_url_override_is_active() {
+    if truthy_env_value "${CODEX_LINUX_ALLOW_RENDERER_URL_OVERRIDE:-}" &&
+       [ -n "${ELECTRON_RENDERER_URL:-}" ] &&
+       [ "$ELECTRON_RENDERER_URL" != "$WEBVIEW_ORIGIN/" ]; then
+        return 0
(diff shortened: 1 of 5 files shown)


ilysenko (approved):
Looks good. I rechecked the current head: the install, package, and launcher paths now fail clearly when the bundled webview entrypoint is missing, the renderer override escape hatch remains intact, and required checks are green.

PR #735, opened 2026-07-10
Fix safe-area patch matching for the current app shell

Yo-DDV:
## Summary

- Update the Linux window-controls safe-area patch descriptor so it matches the current upstream app shell chunk.
- Add a regression test that covers the current bundle shape and rejects the older stale shape.

## Validation

- `make check`
- `make test`
- `make inspect-upstream` against the current upstream DMG
- `make inspect-upstream-intel` against the current upstream DMG
- `make build-app-fresh` against the current upstream DMG
- `codex-app/start.sh --diagnose-scaling`

The baseline build skips `linux-window-controls-safe-area` on the current DMG because the shell chunk name moved. With this patch, that optional patch is applied again while the other skipped optional patch state stays unchanged.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -979,6 +979,32 @@ test("app-server feature enablement descriptor matches current app-main chunks",
   assert.equal(descriptor.pattern.test("experimental-feature-visibility-Bvp90zWX.js"), false);
 });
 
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Looks good. I rechecked the current head: the stale app-shell pattern is replaced rather than retained as a fallback, regression coverage rejects the old shape, and required checks including the current upstream DMG build are green.

PR #736, opened 2026-07-10
Fix current Read Aloud assistant bundle

joshyorko:
## Summary

- target the current primary assistant-thread bundle for Read Aloud
- reject the removed index, local-conversation, generic app-main, and previous editor-diff compatibility shapes
- apply the assistant button and webview runtime atomically so optional drift cannot leave a partial patch or report a false success

## Root cause

In the 26.707 DMG, the assistant render contract lives in `app-initial~app-main~onboarding-page-*.js`. The previous descriptor retained broad legacy bundle shapes, which could no longer prove that the runtime patch was applied to the current assistant message implementation.

## Validation

- `node --test linux-features/read-aloud/test.js` (49/49)
- exact 26.707 DMG parity preflight reports `feature:read-aloud:assistant-runtime` applied
- regression coverage verifies drift leaves the bundle unchanged and reports `skipped-optional`


ilysenko:

PR #739, opened 2026-07-10
Fix automation update eager patch matching for current app shell

Yo-DDV:
## Summary

- Update the `automation-update-eager-tool` descriptor to match the current upstream app shell chunk.
- Add a regression test that accepts the current dynamic tools bundle and rejects the older stale chunk.

## Validation

- `make check`
- `make test`
- `make inspect-upstream` against the current upstream DMG
- `make inspect-upstream-intel` with the candidate patch report
- `make build-app-fresh` with the current upstream DMG

On `origin/main`, the current DMG skips `automation-update-eager-tool`. With this patch, that optional patch applies again. The remaining `linux-window-controls-safe-area` skip is handled separately in #735.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -979,6 +979,32 @@ test("app-server feature enablement descriptor matches current app-main chunks",
   assert.equal(descriptor.pattern.test("experimental-feature-visibility-Bvp90zWX.js"), false);
 });
 
+test("automation_update eager descriptor matches the current dynamic tools chunk", () => {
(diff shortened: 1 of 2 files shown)


Yo-DDV:
Hi, small update before this gets reviewed.

I am closing this one after rechecking it against the newer `26.707.31428` DMG.

The dynamic tools code moved back to the `k0ede4gb` chunk, which `origin/main` already targets. Keeping this PR open would now replace the correct route with the stale `g4rafana` route.

I reran a fresh inspect from `60f875a`, and `automation-update-eager-tool` applies again. No wrapper change is needed for this DMG, so this PR is obsolete now.

PR #742, opened 2026-07-10
Fix current AppShots chunk matching

Yo-DDV:
## Summary

- Match only the latest shared app-shell chunk for AppShots availability and the current settings chunk.
- Patch the current AppShots hotkey class atomically for Linux, including its default, support state, setter, reconcile guard, Wayland state, and bare-modifier helpers.
- Declare the Linux settings helper at module scope before the final source-map trailer.
- Require every current hotkey and settings call site so partial drift leaves the original bundle unchanged with a warning.
- Remove the superseded availability, controller, and settings bundle shapes while keeping capture and IPC behavior unchanged.

## Validation

- Semgrep against `origin/main`
- Direct two-pass application of all four descriptors to the raw `26.707.31428` DMG
- Current hotkey-class marker, strict ESM, source-map placement, and syntax checks
- Isolated build and packed-ASAR inspection with `appshots` enabled


ilysenko:

Yo-DDV:
Thanks for the reminder. I rebased this onto the current main branch and reran the AppShots checks against the latest `26.707.31428` app.

The raw-bundle pass found one more current-app drift: the hotkey descriptor could report that it had applied after changing shared modifier helpers while the actual AppShots class was still macOS-only. The branch now updates every required class and helper shape atomically, and leaves the original bundle untouched if any part is missing. The older bundle forms have also been removed.

All four descriptors now apply once without warnings, the patched bundles are valid and idempotent, and the full local validation is green. I have marked this ready for review while the fresh CI matrix runs on the updated head.

PR #745, opened 2026-07-10
Fix API-key service-tier drift warnings

Yo-DDV:
## Summary

- Scope API-key service-tier drift warnings to the ordered target function shapes instead of file-wide token combinations.
- Keep warnings for recognizable but unpatchable service-tier gates and model mappings.
- Detect the already-patched gate directly so repeated patch passes stay silent even when the chunk contains earlier unrelated auth bindings.
- Add collision, retained-warning, and warning-idempotence regressions.

## Validation

- `node --test linux-features/api-key-service-tier/test.js`
- Semgrep on the changed JavaScript files
- Two-pass scan of all 65 matching chunks from the `26.707.31428` packed ASAR
- Isolated build and packed-ASAR inspection with `api-key-service-tier` enabled

### DIFF
--- a/linux-features/api-key-service-tier/patch.js
+++ b/linux-features/api-key-service-tier/patch.js
@@ -137,16 +137,46 @@ function applyApiKeyServiceTierPatch(source) {
   return applyFallbackFastTierPatch(applyApiKeyModelMarkerPatch(applyApiKeyServiceTierGatePatch(source)));
 }
 
+function applyCurrentGateAndModelPatch(source) {
+  if (!PATCHED_SERVICE_TIER_GATE.test(source) && !hasApiKeyServiceTierGateShape(source)) {
+    warn("Could not identify current service tier auth gate", "API key service tier gate patch");
+  }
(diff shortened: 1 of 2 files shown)


ilysenko:

Yo-DDV:
Thanks for the reminder. I rebased this onto the current main branch and checked the warning logic against every matching chunk in the `26.707.31428` app.

The final pass found two cases worth tightening. File-wide keyword checks could still combine unrelated functions into a false drift signal, and an earlier unrelated auth binding could trigger a warning on the second patch pass. The branch now recognizes the target function structure directly and keeps the real drift warnings covered.

All 65 matching chunks are silent on both passes, the two intended assets remain byte-idempotent, and the isolated feature build is clean. The full local validation is green, so I have marked this ready for review while the fresh CI matrix runs on the updated head.

ilysenko (approved):
Reviewed the final split-descriptor implementation. Partial and missing current-target drift is independently reported, exact current assets patch warning-free and idempotently, and all checks are green.

PR #746, opened 2026-07-10
Fix current remote mobile status drift

joshyorko:
## Summary

- retarget the remote-control status guard to the current 26.707.31428 webview bundle and remove the superseded bundle selector
- extend the Linux initial-status wait using minifier-independent atom helper matching
- cover descriptor selection, modern bundle aliases, and patch-report application

## Verification

### DIFF
--- a/linux-features/remote-mobile-control/patch.js
+++ b/linux-features/remote-mobile-control/patch.js
@@ -1393,8 +1393,19 @@ function applyLinuxRemoteControlStatusWaitPatch(source) {
     return source;
   }
 
-  const statusWaitRegex =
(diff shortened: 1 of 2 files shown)


Yo-DDV:
Hi, thanks for working on this.

Small coordination note: the latest `26.707.31428` DMG moved the conversation hydration, load-gate, and status code into the same `b76hmflu` chunk. I refreshed #728 because its real-DMG check started failing. That refresh necessarily retargets the existing status-read guard, so Git now reports conflicts with this draft in `patch.js` and `test.js`.

Your 30-second Linux status-wait change is still separate and useful. I tested it on top of the refreshed #728 against the current bundle; both changes apply without warnings, remain idempotent, and the resulting module parses correctly. Once #728 lands, this should be a straightforward rebase: keep the status-wait patch and its tests, then drop the duplicate status-read retarget if Git asks.

ilysenko:
Please always keep your pull requests out of draft status.

ilysenko (approved):
Reviewed the current head after the status-wait target fix and regression coverage. All required checks pass.

PR #748, opened 2026-07-10
Fix authenticated proxy fetch routing for current app

Yo-DDV:
## Summary

- Match the latest main-bundle desktop fetch and ClientRequest shapes, and remove the superseded bundle needles.
- Route requests through ClientRequest only when authenticated proxy credentials are configured; keep the normal `net.fetch` path unchanged otherwise.
- Return credentials only for the configured proxy host and port, and leave incomplete helper state untouched with a drift warning.
- Add current-bundle runtime, negative, fail-soft, drift, and idempotence regressions.

## Validation

- `node --test linux-features/authenticated-proxy/test.js`
- Semgrep on the changed feature patch and tests
- Direct patch application and syntax inspection against the raw `26.707.31428` main bundle
- Exact helper, route, challenge-filter, legacy-needle, and idempotence checks

### DIFF
--- a/linux-features/authenticated-proxy/patch.js
+++ b/linux-features/authenticated-proxy/patch.js
@@ -4,6 +4,8 @@ const {
   inferModuleAlias,
 } = require("../../scripts/patches/lib/minified-js.js");
 
+const JS_IDENT = "[A-Za-z_$][\\w$]*";
+
 function applyAuthenticatedProxyPatch(currentSource) {
   const electronVar = inferModuleAlias(currentSource, "electron");
   if (electronVar == null) {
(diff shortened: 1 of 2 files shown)


ilysenko:

Yo-DDV:
Thanks for the reminder. I rebased this onto the current main branch and reran the validation against the latest `26.707.31428` DMG. The desktop fetch path now uses ClientRequest only when proxy credentials are configured, while the normal no-credentials path still uses `net.fetch`.

The final pass also removed the superseded bundle needles and tightened proxy challenge matching. Credentials are returned only for the configured proxy host and port, and incomplete helper state is left untouched with a drift warning instead of producing a partial patch.

The current bundle applies the feature once without warnings, the resulting main bundle is valid and idempotent, and the full local validation is green. I have marked this ready for review while the fresh CI matrix runs on the updated head.

ilysenko (approved):
Reviewed the final current-DMG-only proxy path. Proxy credentials are scoped to matching proxy challenges, normal net.fetch behavior is preserved without credentials, focused tests pass, and all checks are green.

PR #749, opened 2026-07-10
Fix Copilot reasoning effort chunks for current app

Yo-DDV:
## Summary

- Match only the current quick-chat settings/model-list bundle and composer UI bundle used by the `26.707.31428` app.
- Enable the current Copilot effort dropdown, keyboard actions, and `/reasoning` command while preserving adjacent permission gates.
- Remove older-DMG regexes, descriptor targets, fixtures, and documentation so the feature follows the repository current-DMG-only policy.

## Validation

- `node --test linux-features/copilot-reasoning-effort/test.js`
- `node --test linux-features/*/test.js` — 414 tests passed
- `git diff --check`
- Exact `26.707.31428` settings/model/UI application with warning-free idempotence
- `node --check` on both patched current bundles
- Legacy target/fixture marker scan

### DIFF
--- a/linux-features/copilot-reasoning-effort/README.md
+++ b/linux-features/copilot-reasoning-effort/README.md
@@ -25,12 +25,12 @@ feature to the generated app.
 
 ## What It Patches
 
-- `use-model-settings-*.js` reads and writes
-  `copilot-default-reasoning-effort` next to `copilot-default-model`.
-- `font-settings-*.js` keeps the model's full `supportedReasoningEfforts` list
-  for Copilot auth instead of forcing only `medium`.
-- `index-*.js` keeps reasoning effort dropdown entries and the `/reasoning`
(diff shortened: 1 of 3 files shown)


ilysenko:

Yo-DDV:
Thanks for the reminder. I rebased this onto the current main branch and expanded the final validation against the `26.707.31428` DMG. That pass found that the current composer UI had moved as well, so the branch now covers the effort dropdown, keyboard actions, and `/reasoning` command in addition to persistence and model filtering.

All three feature descriptors apply to the packed app, the resulting bundles remain valid and idempotent, and the adjacent permission gate is unchanged. The local validation and the fresh five-job CI matrix are green on `e4da992`, so I have marked this ready for review.

ilysenko (approved):
Reviewed the current head after removing older-DMG compatibility paths. Exact current-DMG validation and all required checks pass.

PR #750, opened 2026-07-10
Fix Linux desktop settings hooks

GTanger:
## What changed

- Convert the generated Linux Desktop settings toggle and build-info panel from React hooks to class components.
- Keep the same settings persistence, build-info refresh, copy commit, and open commit behavior.
- Add regression assertions that the generated Linux Desktop settings asset no longer contains `React.useState`, `React.useEffect`, `React.useCallback`, or the old `useLinuxSetting` helper.

## Why

In the current upstream ChatGPT bundle, the generated `linux-desktop-settings-linux.js` page can be loaded under a React dispatcher that is not active for the synthetic chunk. Opening Settings -> Linux desktop then trips the React error boundary with a `Cannot read properties of null (reading 'useState')` stack rooted at `LinuxToggle`.

The generated page does not need hooks for this state. Class components avoid the dispatcher path while preserving the existing UI and async setting/build-info behavior.

## Validation


ilysenko:

ilysenko (approved):
Reviewed the current head: focused patch tests and all required checks pass; no merge blockers found.

PR #753, opened 2026-07-10
Fix current ui-tweaks sidebar bundle

Yo-DDV:
## Summary

- Retarget the opt-in sidebar project-name style to the current projects-index and remote-conversation bundle.
- Stop matching the retired generic app page chunk, which no longer contains the sidebar markers.
- Add exact positive and stale-target regression coverage without changing the existing style runtime or configuration behavior.

## Validation

- `node --test linux-features/ui-tweaks/test.js`
- `node --test scripts/patch-linux-window-ui.test.js linux-features/*/test.js`
- `bash tests/scripts_smoke.sh`
- `cargo check --workspace`
- `cargo test --workspace`
- ShellCheck and Semgrep on the changed JavaScript files
- Isolated build and packed-ASAR inspection against the `26.707.31428` DMG
- Local merge with #756, including the combined `ui-tweaks` suite and real-DMG feature application

PR #756, opened 2026-07-10
[codex] Add optional GPT-5.6 model picker tweak

C-Joey:
## Summary

- add an optional `ui-tweaks` model picker setting
- open the advanced picker by default and render model choices inline
- include non-hidden `gpt-5.6-*` models when the upstream `available_models` whitelist is stale

## Why

The local app-server already returns GPT-5.6 Sol, Terra, and Luna with
`hidden: false`, but the Desktop frontend can still remove them through its
feature-config model whitelist. This leaves the CLI able to select the models
while the Desktop model picker does not show them.

The patch preserves the existing whitelist behavior for other model families
and only admits visible GPT-5.6 models. It is fail-soft, idempotent, and lives
inside the disabled-by-default `ui-tweaks` feature.

## User impact

Users who enable `ui-tweaks` can see and select GPT-5.6 Sol, Terra, and Luna
directly from the composer model picker. Baseline installations are unchanged.

## Validation

- `git diff --check`
- verified in the real Electron window on Fedora 44 with the RPM build; the
  picker displayed Sol, Terra, and Luna alongside the existing model families

### DIFF
--- a/linux-features/ui-tweaks/README.md
+++ b/linux-features/ui-tweaks/README.md
@@ -17,6 +17,7 @@ Enable it in the local, gitignored feature config:
 
 | Tweak | Patch module | What it does | Settings |
 | --- | --- | --- | --- |
+| `modelPicker.showModelsByDefault` | `patches/model-picker-model-list.js` | Opens the advanced picker by default and shows model choices inline instead of hiding them behind the compact Power slider and a nested Model submenu. | `tweaks.modelPicker.showModelsByDefault.enabled` |
 | `sidebar.projectName` | `patches/sidebar-project-name.js` | Styles project names in the left sidebar project list. It does not style `Projects` / `Chats` section headings and does not style chat rows. | `tweaks.sidebar.projectName.enabled`, `tweaks.sidebar.projectName.style` |
 
 ## Settings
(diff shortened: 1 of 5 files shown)


Yo-DDV:
Hi, thanks for putting this together.

I checked the current `26.707.31428` DMG. Each of the three model-picker descriptors matches exactly one asset, applies without warnings, remains idempotent, and leaves a valid module. I also merged this locally with #753; the merge is clean, the `ui-tweaks` tests pass, and all four combined descriptors apply on the same DMG.

There are two proof gaps I would close before taking this out of draft. The current green jobs never enable `ui-tweaks`, so they would stay green if all three model-picker selectors drifted. Once the feature-aware validator from #728 is available, an upstream-DMG inspect should enable `ui-tweaks` and require these entries to succeed:

- `feature:ui-tweaks:model-picker-default-advanced-view`
- `feature:ui-tweaks:model-picker-include-gpt-5-6`
- `feature:ui-tweaks:model-picker-inline-model-list`

I would also add a behavioral allowlist test, not only a replacement-string assertion: a visible GPT-5.6 model should be admitted, a hidden GPT-5.6 model should stay rejected, and a visible non-5.6 model should remain gated by `available_models`. The implementation has those semantics today; the extra test would keep them from drifting silently.

ilysenko:

ilysenko (approved):
Reviewed the current head and exact current-DMG behavior. Focused and feature-wide tests pass; the optional model visibility behavior is accepted as intentional.

PR #757, opened 2026-07-10
[codex] Keep reasoning effort labels in English

C-Joey:
## Summary

- add an optional Simplified Chinese locale tweak for reasoning effort values
- keep `None`, `Minimal`, `Low`, `Medium`, `High`, `XHigh`, `Max`, and `Ultra`
  in English
- leave the surrounding picker title and usage warning translated

## Why

The current Simplified Chinese locale translates both `XHigh` and `Ultra` as
`极高`, even though they are separate upstream effort values. Keeping the enum
values in English avoids ambiguous duplicate labels and preserves the exact
terminology used by the CLI and configuration files.

The patch is fail-soft, idempotent, scoped only to the `zh-CN` locale bundle,
and lives inside the disabled-by-default `ui-tweaks` feature.

## User impact

Users who enable `ui-tweaks` and use Simplified Chinese retain a translated
interface while seeing unambiguous reasoning effort values. Baseline
installations and other locales are unchanged.

## Validation

- `git diff --check`
- verified the installed `zh-CN` asset contains all eight English labels

### DIFF
--- a/linux-features/ui-tweaks/README.md
+++ b/linux-features/ui-tweaks/README.md
@@ -17,6 +17,7 @@ Enable it in the local, gitignored feature config:
 
 | Tweak | Patch module | What it does | Settings |
 | --- | --- | --- | --- |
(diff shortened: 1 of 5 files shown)


ilysenko (approved):
Looks good. I rechecked the current head: the tweak remains disabled by default, is scoped to the zh-CN locale, fails soft, preserves surrounding translations, and required checks are green.

PR #758, opened 2026-07-10
Add opt-in API key model visibility feature

y40881110-lang:
## Summary

- add a default-disabled `api-key-model-visibility` Linux feature
- let API-key-authenticated hosts use the non-hidden model catalog returned by Codex CLI instead of the desktop Statsig `available_models` allowlist
- preserve the existing ChatGPT, Copilot, Amazon Bedrock, and CLI-hidden model behavior
- document the feature and add focused drift, idempotency, composition, and behavior tests

## Motivation

An OpenAI-compatible API-key provider can return visible models from `model/list`, but Codex Desktop may hide them afterward when its dynamic `available_models` allowlist has not caught up. This makes the CLI and desktop model pickers disagree even though both use the same provider.

The patch does not hardcode model ids. Future models returned as non-hidden by the CLI are picked up automatically for API-key hosts. It does not grant provider access or force reasoning-effort UI gates such as Ultra.

## Scope

This remains opt-in under `linux-features/` and is disabled by default. The patch only changes the API-key authentication branch of the existing model-catalog filter and fails soft if the upstream minified expression changes.

## Validation

- `node --test linux-features/api-key-model-visibility/test.js linux-features/api-key-service-tier/test.js` (`14/14`)
- `node --test linux-features/*/test.js` (`410/410`)
- `git diff --check`
- rebuilt and inspected upstream app `26.707.31428` / Electron `42.1.0`; the feature descriptor applied exactly once to the expected webview asset

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,7 @@ workarounds.
 |---|---|---|---|
 | Record and Replay (alpha) | Opt-in alpha | `record-and-replay` | [Docs](linux-features/record-and-replay/README.md) |
 | Agent Workspaces | Opt-in | `agent-workspace` | [Docs](linux-features/agent-workspace/README.md) |
+| API key model visibility | Opt-in | `api-key-model-visibility` | [Docs](linux-features/api-key-model-visibility/README.md) |
 | API key service tier | Opt-in | `api-key-service-tier` | [Docs](linux-features/api-key-service-tier/README.md) |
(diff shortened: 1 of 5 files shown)


ilysenko (approved):
Looks good. I rechecked the current head: the feature remains disabled by default, scopes the allowlist bypass to API-key hosts, preserves hidden-model and other-auth behavior, fails soft on drift, and required checks are green.

PR #759, opened 2026-07-10
Adapt Browser Use trusted hash patch to direct builder

yazanabuashour:
## Summary

- Adapt Browser Use trusted-hash injection to the current DMG’s direct `node_repl` runtime builder.
- Remove recognition of the retired async `resourcesPath` shape and add regression coverage for the current builder.

## Why

DMG `26.707.31428` no longer uses the runtime-factory shape covered by #692. The existing patch therefore misses the trusted-hash insertion point, preventing hashes for the Linux-patched bundled Browser and Chrome clients from being added to the trusted set.

## Validation

- DMG SHA-256: `6f67af7e2f934093ab8afebcec11374d40c8db8f9100fb6620f24155401d8319`
- Confirmed `browser-use-node-repl-approval` was applied
- Confirmed the patched JavaScript chunk passes `node --check`

## Scope

Targets the latest DMG shape only. The change is limited to the Browser Use patch implementation.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -1129,6 +1129,13 @@ function currentChromeNativeHostRuntimeBundleFixture() {
   ].join("");
 }
 
+function currentBrowserUseTrustedHashesRuntimeBuilderFixture() {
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Looks good. I rechecked the current head: the patch targets only the current direct node_repl builder, explicitly drops the retired async resourcesPath shape, remains fail-soft, and required checks including the upstream DMG build are green.

PR #761, opened 2026-07-10
Fix open target discovery for current app flow

Yo-DDV:
## Summary

- follow the current main-process open-target registry and detection context
- keep worker-backed detection unchanged on macOS and Windows while resolving Linux targets locally
- restore current availability, bridge detection, and directory-mode matching
- contain failed Linux target probes so one broken desktop entry cannot break the full menu

The feature remains opt-in and no tracked feature configuration changes.

## Validation

- `cargo check --workspace`
- `cargo test --workspace`
- full rebuild from the current `26.707.31428` DMG
- packed ASAR syntax and marker checks
- fresh five-job GitHub Actions matrix on the published head

### DIFF
--- a/linux-features/open-target-discovery/patch.js
+++ b/linux-features/open-target-discovery/patch.js
@@ -124,18 +124,22 @@ function findPropertyBlock(source, propertyName) {
   };
 }
 
-function findAsyncFunctionBlockContaining(source, marker) {
+function findAsyncFunctionBlockContaining(source, marker, predicate = null) {
   let markerIndex = source.indexOf(marker);
   while (markerIndex !== -1) {
     const functionStart = source.lastIndexOf("async function ", markerIndex);
-    const blockStart = functionStart === -1 ? -1 : source.indexOf("{", functionStart);
+    const signatureEnd = functionStart === -1 ? -1 : source.indexOf("){", functionStart);
+    const blockStart = signatureEnd === -1 ? -1 : signatureEnd + 1;
     const block = findBalancedBlock(source, blockStart);
     if (block != null && block.end > markerIndex) {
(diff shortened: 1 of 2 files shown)


ilysenko:
Please remove the pull request from draft status once it is ready.

Yo-DDV:
Thanks for the reminder. I took another full pass after rebasing this onto the current main branch, including a rebuild from the `26.707.31428` DMG. The feature now follows the current main-process target registry and detection context while keeping the worker path unchanged on macOS and Windows. I also added coverage for failed target probes and the current availability, bridge, and directory-mode shapes.

The local suites and the fresh five-job CI matrix are green on `8049e2b`, so I have marked this ready for review.

ilysenko:
- This current-DMG drift fix still retains older open-target bundle needles, fallback branches, and legacy-shape fixtures. The repository supports only the latest DMG, so please remove those compatibility paths and keep the implementation and tests scoped to the current `26.707.31428` shape.

ilysenko:
- This current-DMG drift fix still preserves older compatibility paths in the registry detection, `JN`/`requestOpenInWorker` command needles, `this.options` arrow bridge, and obsolete `ZN`/`iP` execution patch and tests. On 26.707.31428 only the `QN`/`FN` and current class/availability/directory shapes are used; please remove the legacy branches and fixtures, and bind the current registry/context exactly instead of falling back to an unrelated nearby `detect()` or `void 0`.

ilysenko (approved):
Re-reviewed the final head after the current-DMG registry cleanup and exact b76hmflu descriptor fix. The exact bundle patch is warning-free and idempotent, focused tests pass, and all checks are green.

PR #762, opened 2026-07-10
Preserve portable upstream bundled plugins

Yo-DDV:
## Summary

- preserve the upstream Sites, Deep Research, and Visualize plugins in Linux builds
- keep marketplace metadata aligned with the exact portable entries that were staged
- make build staging and launcher cache replacement preserve the previous payload on failure
- reject invalid manifests, symlinks, privileged entries, special files, native executables, and platform-specific bundles from the portable path
- document which bundled plugins use this path and which still require Linux-specific replacements

The allowlist is intentionally explicit. The macOS-native LaTeX and Record and Replay payloads are not copied by this change.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js linux-features/*/test.js` — 755 tests passed
- `cargo check --workspace`
- `cargo test --workspace`
- `shellcheck -s bash --severity=error -x launcher/start.sh.template scripts/lib/bundled-plugins.sh tests/scripts_smoke.sh`
- Semgrep default rules on the modified runtime scripts — no findings
- full isolated build from the current `26.707.31428` DMG
- Sites design picker initialize and tool-list smoke test
- Visualize render smoke test
- isolated launcher cache hydration for all three plugins

The fresh package matrix for this head is green, including Nix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,7 @@ workarounds.
 | GUI install prompts | If installed | Uses `kdialog` / `zenity`, then terminal fallback | [Native setup](docs/native-setup.md) |
 | Linux file manager integration | Always | Built into core Linux patches | [Architecture](docs/architecture.md) |
 | Chrome plugin native host | Always | Installed with bundled plugins | [Architecture](docs/architecture.md) |
+| Portable upstream plugins | When shipped upstream | Sites, Deep Research, and Visualize are staged automatically; upstream rollout still applies | [Architecture](docs/architecture.md#bundled-plugins) |
 | Browser annotations | Always | Built into the patched webview | [Architecture](docs/architecture.md) |
 | Tray and warm-start handoff | Always | Normal app launch | [Architecture](docs/architecture.md) |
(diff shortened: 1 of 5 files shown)


ilysenko:

Yo-DDV:
Hi @ilysenko, thanks for the nudge.

I took one more pass before finalizing this and found a few failure paths worth fixing in the staging and launcher cache flow. The updated head now preserves the previous payload on replacement failures, tightens the manifest and filesystem checks, and adds regression coverage for those cases.

The current DMG build and all five CI jobs, including Nix, are green on `0a8c6af`. I have marked the PR ready for review.

ilysenko:
- `sync_extra_bundled_plugin_cache` runs before the launcher lock is acquired and uses plain `mv` for directory replacement. Two concurrent cold starts can recreate the destination between moves; GNU `mv` then nests the temporary directory inside the existing cache, reports success, and can discard the backup while leaving a mixed payload. Please serialize cache replacement before it begins, or use collision-safe `mv -T`/rollback semantics, and add a concurrent-destination regression test.

ilysenko (approved):
Re-reviewed the final head after the collision-safe cache replacement fix. The deterministic race regression, rollback behavior, portable plugin validation, full smoke suite, and all CI checks pass.

PR #764, opened 2026-07-10
Fix Linux patch drift for latest Codex DMG

OmerFarukOruc:
## Summary

The latest Codex Desktop DMG reshaped several main-process and webview bundles, so a set of anchors in the Linux patchers no longer match. Five `required-upstream` patches fail their anchor lookup, which trips the critical-patch gate and aborts `make install-native`. This realigns the patchers to the current bundle shapes.

## Critical (build-blocking) fixes

- **linux-window-options** — the primary window now guards `focusable` as `...m===void 0?{}:{focusable:m}` instead of a bare `focusable:m,`. Added a void-spread variant so the Linux-primary `focusable:!0` override still applies.
- **linux-native-titlebar** — the `case\`primary\`` block reshaped (per-branch `...appearance===\`quickChat\`?{…}` spreads, `||linux` folded into the win32 branch). Match just the win32/linux `titleBarOverlay` branch and split Linux onto its own overlay helper, and preserve the widened `installApplicationMenuTitleBarOverlaySync` guard tail (`&&t!==\`quickChat\``).
- **linux-tray** — the icon fallback now marks the resolved icon as a template image and derives the running icon via a helper (`setTemplateImage(!0)` + `q9(...)`), and splits the `getFileIcon` fallback onto its own statement. Match that shape and insert the Linux packaged-icon fallback ahead of it while leaving the upstream loop/fallback bodies intact.
- **linux-avatar-overlay-mouse-passthrough** — `applyPointerInteractivityPolicy` is byte-identical but the method that follows it changed, so the needle's trailing anchor drifted. Re-anchor on the method's own closing braces, and realign the four downstream sub-patches whose shapes moved: `endDrag` (reclamp detached from `dragState=null`), `setElementSize` (mascot/tray no longer adjacent in the destructure), the i3 tray-size fallback (`});this.anchor=` moved), and `applyLayout` (`setWindowBounds`/`sendLayoutToRenderer` no longer adjacent).

## Optional fixes

- **linux-terminal-user-path** — `buildTerminalEnv` renamed its env-freeze return method; relaxed that one token.
- **linux-window-controls-safe-area** and **opaque-window-default-resolved-theme** — Vite re-chunked the webview bundles, so the asset filename patterns now point at the current `app-initial~app-main~*` chunks.

## Verification

- `patch-linux-window-ui.js --enforce-critical` exits 0; all five critical patches report `applied`.
- Patched main bundle passes `node --check`; patches are idempotent (re-runs report `already-applied`).
- `node --test scripts/patch-linux-window-ui.test.js` — 343/343 pass (existing upstream-shape fixtures unaffected).
- Installed and confirmed working end-to-end.

## Out of scope

Six other optional patches drifted structurally in the same DMG (app-server conversation hydration, completed-item recovery, profile settings menu, local-environment action modal, keybinds settings, chrome plugin auto-install). They do not block the build and are left for a separate change.


Yo-DDV:
Hi @OmerFarukOruc, thanks for looking into this.

I reran `557b818` and the current `main` against the same `26.707.31428` DMG (`sha256: 6f67af7e2f934093ab8afebcec11374d40c8db8f9100fb6620f24155401d8319`). The PR head does clear the required-patch gate on its original base, but it is now conflicting and the same artifact already passes on `main`: all 17 required patches apply or report already applied, with no optional skip or warning outside the expected disabled and target-specific entries.

On the PR head, that run still reports six optional skips and one applied-with-warnings entry. The new `opaque-window-default-resolved-theme` pattern is one example: it targets `app-initial~app-main~page-CtX5-cLy.js`, but the patch cannot find its insertion point. The mainline fixes from #720 and #735 cover these same current-DMG surfaces with dedicated regression coverage.

Given that, I do not see a residual change here that is still needed on top of current `main`. I think closing this as superseded is safer than resolving all five conflicts and reintroducing the older patch shapes. If you have a specific current-main failure that I missed, I can reproduce that exact case.

ilysenko:
Thanks again for tracking down the original drift. Current `main` has since incorporated current-DMG-only replacements through #716, #720, and #735. I revalidated the current 26.707.31428 artifact: `main` passes all 17 required patches and the 342-test patch suite, while this head now conflicts in all five touched files and fails the required native-titlebar patch. The remaining hunks target intermediate/older bundle shapes or broaden asset matching, so carrying them forward would reintroduce compatibility fallbacks without fixing a current-main failure. Closing this as superseded.

PR #768, opened 2026-07-10
Fix Chrome sidepanel app-server startup on Linux

Yo-DDV:
The current ChatGPT Chrome sidepanel reaches the Linux native host, then stops at `codexRuntime/hello`. That request is still forwarded to the Browser Use client, which does not own the sidepanel runtime protocol, so the extension reports that no handler is registered.

This adds the protocol v2 runtime path to the Linux native host. It selects the matching current install, starts one app-server per sidepanel, and bridges it through a loopback WebSocket protected by the extension Origin and a random token. Restart, tab-context assets and restricted local-file opening are covered as well. Existing Browser Use request and response routing is left unchanged.

The process and file lifecycle is bounded. Pre-auth handshakes use a separate evictable pool, while authenticated upgrades obtain both a connection permit and a per-client process lease before the WebSocket is accepted. App servers are reclaimed after the last connection, `ensure` calls that never connect expire, and active clients are not evicted at capacity. Process-group cleanup keeps the leader unreaped until its descendants are gone, and runtime paths now validate every canonical ancestor with a narrow allowance for root-owned sticky directories.

Tests run locally

- `cargo test -p codex-computer-use-linux` — 317 passed
- focused runtime tests in debug and release — 30 passed in each profile
- `bash scripts/ci/run-node-checks.sh` — 857 passed
- `make check` and `make test` — 217 updater tests passed
- `bash tests/scripts_smoke.sh` — Debian, RPM, pacman and AppImage smoke tests passed

I also exercised the release binary against the current extension manifest and CLI. Two sidepanels completed `hello`, `ensure` and app-server `initialize`; same-client reuse and restart worked, invalid token and Origin attempts returned 403, and every observed app-server process exited with the native host.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,9 +78,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
+version = "1.0.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
 
 [[package]]
 name = "assert-json-diff"
@@ -333,6 +333,15 @@ dependencies = [
 
 [[package]]
 name = "block-buffer"
+version = "0.10.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
+dependencies = [
+ "generic-array",
+]
+
+[[package]]
+name = "block-buffer"
 version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cdd35008169921d80bc60d3d0ab416eecb028c4cd653352907921d95084790be"
@@ -490,6 +499,7 @@ dependencies = [
  "cosmic-protocols",
  "evdev",
  "futures-util",
+ "getrandom 0.4.2",
  "image",
  "libc",
  "mimalloc",
@@ -497,7 +507,9 @@ dependencies = [
  "schemars",
  "serde",
  "serde_json",
+ "sha2",
  "tokio",
+ "tokio-tungstenite",
  "wayland-client",
  "wayland-protocols",
  "xkeysym",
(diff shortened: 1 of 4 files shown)


ilysenko:
Thanks for working on this. I found several blockers that need to be addressed before this can be merged:

- Proxy permits are acquired before Origin and token validation. A local process can hold all permits with stalled TCP handshakes and prevent the legitimate extension from connecting. Pre-auth handshakes need a bounded, evictable pool, separate from authenticated connection limits, with a regression test that fills the entire pool.
- App-server processes are not released when their sidepanel WebSocket closes. After 16 client IDs, later sidepanels can permanently hit the process limit. The lifecycle needs race-free per-client leases, cleanup after the last connection, and bounded cleanup for `ensure` calls that never connect.
- Process cleanup stops following the process group once the direct child exits. Descendants can survive, while signalling a stored PGID after reaping the leader creates a PID-reuse risk. The leader should be observed without reaping, the complete group terminated, and the child reaped only afterward.
- Trusted-path validation returns after checking the first parent, and directory paths do not validate their ancestor chain. Every canonical ancestor must be checked, with only a narrow exception for root-owned sticky directories such as `/tmp`.

Please add focused regression coverage for handshake exhaustion, reconnect-versus-cleanup races, unconnected process reaping, exited leaders with live descendants, and unsafe grandparent directories.

Yo-DDV:
Thanks, you were right on all four. I reproduced each failure on the previous head before touching the implementation.

The proxy now keeps stalled pre-auth handshakes in a bounded, evictable pool. Once Origin and token are valid, the authenticated slot and process lease are taken before the `101` goes back. Processes live until their last lease is released, unconnected `ensure` calls expire, active clients are not evicted at capacity, and an old reconnect cleanup cannot touch a replacement process.

I also keep the leader unreaped until the whole process group is gone, and path checks now walk every canonical ancestor, with only root-owned sticky directories such as `/tmp` allowed through. I added the cases you asked for plus active-capacity, last-connection, stale-lease and shutdown/start coverage. The rebased head passes the full Rust workspace checks, 857 Node tests, 217 updater tests, the cross-format smoke suite, and the release run against the current extension manifest and CLI. All five GitHub checks are green as well.

ilysenko:
Thanks for the update. The previous blockers look fixed, but I found one remaining security blocker:

- Trusted runtime paths currently accept group-writable files and directories when their gid matches the process egid. That does not prove the group is private; another group member could replace the trusted CLI, Node/module content, or an ancestor directory. Please reject group-write unconditionally for trusted entries and ancestors, retain only the explicit root-owned sticky-directory exception, and update the `0770` regression test.

Yo-DDV:
Thanks, agreed. I tightened that path now.

The runtime trust check rejects group-write for trusted files and every checked ancestor, regardless of gid. The only remaining exception is still the root-owned sticky-directory case for `/tmp`-style parents.

I also updated the old `0770` regression and added a parent-directory case, then reran the Rust workspace and the usual patcher/feature tests locally.

ilysenko:
Thanks for the thorough follow-up — the requested group-write hardening looks fixed. One non-blocking follow-up for future hardening: path validation canonicalizes the runtime paths, while process launch, current-directory setup, and environment variables still use the original manifest paths. That leaves a narrow local symlink-swap/TOCTOU window when an original ancestor is group-writable. This does not block this PR; retaining and using the validated canonical paths, with a focused regression test, would close the gap.

ilysenko (approved):
The requested lifecycle and trusted-path fixes are addressed on the current head, focused regression coverage is present, and all required checks pass. The remaining canonical-path TOCTOU observation is non-blocking follow-up hardening.

PR #769, opened 2026-07-10
Fix Read Aloud settings runtime

PinguuSS:
## Summary

- migrate the Read Aloud Linux desktop settings control to the current class-based settings runtime
- keep enabled-state and speech-pace reads and writes on the shared global-state bridge
- fail soft when the current generated settings key boundary drifts, and update the regression fixture to match the latest bundle shape

## Root cause

The generated Linux desktop settings controls moved from `useLinuxSetting` to class components in #750. The optional Read Aloud feature still injected a function component that called the removed hook, so opening **Settings > Linux desktop** with the feature enabled reached the renderer error boundary with `ReferenceError: useLinuxSetting is not defined`.

#736 updates the assistant-response bundle patch and is complementary; it does not update this generated Linux desktop settings component.

## User impact

The Linux desktop settings page now opens normally when Read Aloud is enabled. Its toggle, voice setup actions, and speech-pace slider load and persist through the same runtime contract used by the current generated controls.

## Source of truth

- `CHANGELOG.md`

## Validation

- `node --test linux-features/read-aloud/test.js` (49/49)
- `node --test linux-features/*/test.js` (430/430)
- `node --test scripts/patch-linux-window-ui.test.js` (342/342)
- rebuilt from the official ChatGPT `26.707.31428` DMG with Electron `42.1.0` on Ubuntu 25.10, KDE Wayland
- patch report: all 3 Read Aloud patches applied; 0 post-patch integrity findings
- Electron runtime check: opened **Settings > Linux desktop**, toggled Read Aloud `false -> true -> false`, verified **Choose folder**, **Download voice**, and **Speech pace**, changed/restored pace `1.05 -> 1.10 -> 1.05`, and observed no page errors or relevant console errors

## Scope and limitations

This is limited to the opt-in Read Aloud feature. No packaging or updater behavior changes. The source build and real Electron runtime were validated on the environment above; native package artifacts were not rebuilt because their payload contract is unchanged.


Yo-DDV (commented):
Hi, thanks for fixing the missing `useLinuxSetting` crash. I reproduced it against 26.707.31428, and the class-based path itself works on current `main`.

I did find two async failure cases in the new component. `load()` uses `Promise.all`, so one rejected settings read discards the other successful value. With Read Aloud stored as `true` and only the speed read rejected, the component finishes loading with `enabled: false`.

There is also a stale rollback on the pace slider. I sent `1.15`, then `1.30`, let the second write succeed, and rejected the older first write afterward. The catch restored `1.40`, overwriting the newer successful choice in the UI.

Could you keep the two reads independent and ignore rollback or error callbacks from superseded writes? A small runtime test that instantiates the generated class and settles the bridge requests out of order would cover both cases.

PR #770, opened 2026-07-10
Fix pet overlay patcher reruns

Yo-DDV:
## Summary

- recognize the pet overlay transparency prefix when the core opaque-background patch runs again
- keep the generated Hyprland string escaping readable by the shared minified-bundle scanner
- cover core/feature composition, repeat runs, and malformed drift with regression tests

## Why

Running the patcher a second time on a bundle with `pet-overlay` enabled could report required core patches as missing even though the first pass had applied them. The transparency prefix changed the expected background helper shape, while a quote-matching regex in the generated overlay methods confused the lightweight class scanner on the next pass.

The matcher now accepts only the exact composed transparency prefix and still rejects malformed background helpers. The generated escape helper keeps the same runtime output without leaving an ambiguous regex literal in the class body.

## Validation

- `bash scripts/ci/run-node-checks.sh` (857 tests)
- `node --test linux-features/pet-overlay/test.js` (34 tests)
- `node --test scripts/patch-linux-window-ui.test.js` (346 tests)
- local Semgrep JavaScript scan (68 rules, no findings)
- local current-DMG patch run twice with `pet-overlay`; the second required-patch pass succeeded and the resulting file tree was unchanged

Follow-up to #732.


ilysenko (approved):
Reviewed the current head. The patcher rerun behavior is narrowly scoped, current-DMG coverage is green, and the regression tests cover composition, idempotency, and malformed drift.

PR #774, opened 2026-07-10
Keep Read Aloud settings state consistent

Yo-DDV:
## Summary

- preserve a successful Read Aloud setting when the other settings read fails
- serialize toggle and pace writes, rolling a failed latest change back to the last confirmed value
- keep repeated patch runs stable for existing Read Aloud settings bundles and the shared Electron binding

This follows up on #769.

## Validation

- `node --test linux-features/read-aloud/test.js` (58 tests)
- `bash scripts/ci/run-node-checks.sh` (866 tests)
- Semgrep JavaScript and secrets rules (104 rules, no findings)
- official app version 26.707.31428 inspected with Read Aloud enabled; all three feature patches applied and required-patch validation passed


ilysenko:

- `updateEnabled` and `updateSpeed` roll the newest failed write back to the previous optimistic value rather than the last confirmed value. Starting from `false`, toggling `true → false`, then rejecting both writes leaves the UI at `true` even though neither write succeeded. Please serialize writes or track confirmed state, and add a regression test for consecutive failures.

Yo-DDV:
Thanks, you're right. The generation counters ignored stale failures, but the rollback target was still the previous optimistic value.

I've changed the toggle and pace writes to run serially per setting and track the last confirmed value. The tests now cover both writes failing, an older failure followed by a newer success, and a successful first write followed by a failing second write for both controls.

Locally, the Read Aloud suite is 58/58, the full Node checks are 866/866, the script smoke tests pass, and Semgrep reports no findings.

ilysenko (approved):
Reviewed the updated write serialization and confirmed-state rollback handling. The previous rapid double-failure issue is fixed, the focused regression suite passes, and all CI checks are green.

PR #783, opened 2026-07-10
Pin Linux launcher to its bundled webview

colonelpanic8:
## Summary

- ignore inherited `ELECTRON_RENDERER_URL` values unless the explicit debug override is enabled
- keep each launcher instance on its own verified local webview bundle
- remove the now-redundant Agent Workspaces renderer-pin hook and update its docs/tests

## Root cause

A fresh build launched from an existing ChatGPT Desktop task inherited that app's `ELECTRON_RENDERER_URL` (typically `:5175`). The new process therefore loaded stale assets from the already-running installation instead of the webview bundle it had just built. In the reported case, that stale bundle still contained the hook-based synthetic Linux settings page, so Settings → Linux desktop failed with `Cannot read properties of null (reading 'useState')` even though current `main` generated the corrected class-based asset.

## Validation

- `node --test linux-features/agent-workspace/test.js`
- `node --test scripts/patch-linux-window-ui.test.js` (346 passing)
- focused launcher smoke checks for missing webview handling, template markers, and side-by-side identity
- rebuilt the latest upstream DMG, launched with an intentionally stale inherited `ELECTRON_RENDERER_URL=:5175`, and verified the renderer used its own `:53439` origin
- opened Settings → Linux desktop in that instance and verified all controls rendered without an error boundary or console errors


Yo-DDV (commented):
Thanks for the patch. I checked current head `1dde8728`.

I ran:

- `node --test linux-features/agent-workspace/test.js`

CI is green on the PR. I did not find a blocker or anything odd in the diff. Final review and merge stay with @ilysenko.

ilysenko (approved):
Reviewed current head 1dde8728. The launcher fix is focused, tests and package checks are green, and I found no blockers.

PR #784, opened 2026-07-10
Hide unavailable settings search results

PinguuSS:
## Summary

- hide Linux settings search results for macOS-only Dock icon controls
- hide Suggested prompts search hits when they only point at a non-rendered settings row
- add a core webview patch descriptor and regression coverage for minified alias collisions

## Root cause

The upstream settings search documents include message IDs for controls that do not always render in the generated Linux settings page. Dock icon entries are macOS-only, and Suggested prompts can be indexed under general-settings even when the real row is gated away by account eligibility. The first implementation also exposed a minifier TDZ risk by calling an imported feature-gate alias inside a function that declares the same local name; this version routes that call through a top-level helper.

## Validation

- node --test --test-name-pattern="Linux settings search" scripts/patch-linux-window-ui.test.js
- node --test scripts/patch-linux-window-ui.test.js (348/348)
- bash tests/scripts_smoke.sh
- node --test linux-features/*/test.js (479/479)
- git diff --check
- rebuilt the generated Linux app from the current official DMG with the opt-in feature set enabled
- launched the generated app on Kubuntu/Wayland and verified Settings search via CDP screenshots:
  - prompt keeps real Prompt/Git hits and no longer shows Suggested prompts
  - dock icon returns no Linux result

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- Linux settings search no longer shows unavailable macOS Dock icon controls or
+  Suggested prompts results that do not render in the generated Linux settings
+  page.
 - Read Aloud no longer crashes the generated Linux desktop settings page after
   its shared controls moved from React hooks to class components. The feature's
   enabled state, voice setup actions, and speech pace now use the same
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -167,6 +167,7 @@ const {
   applyLinuxI18nGatePatch,
   applyLinuxOpaqueWindowsDefaultPatch,
   applyLinuxSafeMonospaceFontStackPatch,
+  applyLinuxSettingsSearchVisibilityPatch,
   applyLinuxSkillsListDedupePatch,
   applyLinuxThreadSidePanelNativeTooltipPatch,
   applyLinuxTooltipWindowControlsCollisionPatch,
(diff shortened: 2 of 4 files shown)


Yo-DDV (changes requested):
Thanks for the detailed fix. I found one blocker in the Suggested prompts case.

The current bundle puts the `settings.agent.ambientSuggestions.*` messages in the `general-settings` search section. The new helper removes them from that section unconditionally, even when rollout gate `2425897452` is enabled. The same settings bundle renders the Suggested prompts row when that gate is enabled and the account eligibility check passes, so an eligible user can see the setting but cannot find it through Settings search.

The regression test at `scripts/patch-linux-window-ui.test.js:472` currently expects the result to stay hidden with the gate enabled, which locks in that false negative. Could you align search visibility with the row's effective visibility and cover disabled, enabled-and-eligible, and enabled-but-ineligible cases? Simply changing the condition to `!n` may still expose the result to ineligible accounts.

Yo-DDV (commented):
Thanks for the update here. I rechecked current head `695d32b6` against the settings-search concern I raised earlier.

I ran:

- `node --test --test-name-pattern='Linux settings search' scripts/patch-linux-window-ui.test.js`

The previous false-negative path looks fixed now, CI is green, and I do not see a remaining blocker from my side. Final review and merge stay with @ilysenko.

ilysenko (approved):
Reviewed current head 695d32b6. The prior visibility blocker is fixed, the patch remains optional and fail-soft, and all checks are green.

avifenesh:
Heads-up: the DMG at codex-app-prod/ChatGPT.dmg rolled back to the earlier 26.707.41301 build (see #817 for the full story — the July 10 rebuilt DMG's chunk names are gone again). On the rolled-back build, `applyLinuxSettingsSearchVisibilityPatch` skips because the hardcoded import specifiers (`E$ as …`, and the `T2`/`j7`/`qk` additions) pin the rebuilt build's export letters — the current build exports the statsig gate as `aG`, the account-info hook as `oJ`, and the eligibility predicate as `y1` (from the `…chatg~gwqc41kz` chunk). Since the letters flip per published build, resolving them semantically from the bundle (e.g. locating the gate by a `X(`<gate-id>`)` call and mapping back through the import list, like the function-shape discovery you already do for the search function) would survive both. Happy to help if useful; #817 keeps this patch optional-skip so it doesn't block CI.

PR #785, opened 2026-07-10
test: isolate GNOME rendering probe from WSL

ryan-mt:
## What changed

- preserve the launcher's real WSL detection behind a probe-only wrapper
- let the simulated GNOME Wayland multi-monitor case explicitly opt out of host WSL detection

## Why

`env -i` clears WSL environment variables but cannot hide the WSL kernel marker in `/proc/sys/kernel/osrelease`. On WSLg hosts, the GNOME rendering-profile smoke case was therefore classified as WSLg and failed before exercising the behavior it intended to test.

This change affects only the generated smoke-test probe; production launcher behavior is unchanged.

## Validation

- `bash tests/scripts_smoke.sh` (all script smoke tests passed on WSLg)

### DIFF
--- a/tests/scripts_smoke.sh
+++ b/tests/scripts_smoke.sh
@@ -4566,8 +4566,18 @@ source_path, output_path = sys.argv[1:3]
 source = open(source_path, encoding="utf-8").read()
 start = source.index("is_wsl_environment() {")
 end = source.index("configure_side_by_side_app_env() {")
-probe = "#!/usr/bin/env bash\n" + source[start:end] + r'''
+helpers = source[start:end].replace(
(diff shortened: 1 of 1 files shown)


Yo-DDV (commented):
Thanks for the targeted test fix. I rechecked current head `28bbd99a`.

I ran:

The full smoke suite passes, CI is green, and I did not find a blocker in the GNOME/WSL probe isolation. Final review and merge stay with @ilysenko.

ilysenko (approved):
Reviewed current head 28bbd99a. This is a focused test-only WSL isolation fix; the full smoke suite and all checks pass.

PR #786, opened 2026-07-10
test: normalize skill fixture mode on DrvFS

ryan-mt:
## What changed

- copy the instruction-only skill fixture into the test temp directory
- set the fixture `SKILL.md` mode to `0644` before inspecting it

## Why

WSL DrvFS can report every tracked file as executable. When the repository is checked out under `/mnt/c`, the Record & Replay fixture test therefore classifies an instruction-only skill as conditional instead of supported.

Materializing the fixture with its intended mode keeps the capability test deterministic without changing importer behavior.

## Validation

- `cargo test -p codex-record-replay-linux`
- `cargo clippy -p codex-record-replay-linux --all-targets -- -D warnings`

### DIFF
--- a/record-replay-linux/tests/skills_import_fixtures.rs
+++ b/record-replay-linux/tests/skills_import_fixtures.rs
@@ -39,7 +39,13 @@ fn import_options(source: &Path, target_dir: &Path) -> SkillImportOptions {
 #[test]
 fn instruction_only_defaults_to_supported_importable_skill() {
     let temp = tempfile::tempdir().unwrap();
-    let skill = fixture_skill("instruction-only");
+    let fixture = fixture_skill("instruction-only");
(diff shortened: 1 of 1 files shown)


Yo-DDV (commented):
Thanks for tightening this test fixture. I rechecked current head `c6352421`.

I ran:

- `cargo fmt --check`
- `cargo test -p codex-record-replay-linux --test skills_import_fixtures`

The fixture-mode coverage passes, CI is green, and I did not find a blocker or odd permission change in the diff. Final review and merge stay with @ilysenko.

ilysenko (approved):
Reviewed current head c6352421. This is a focused test-only DrvFS fixture normalization; focused Rust tests and all checks pass.

PR #787, opened 2026-07-10
fix(updater): harden DMG downloads

ryan-mt:
## Summary

- download upstream DMGs into a unique same-directory temporary file and atomically replace the cache only after flush and sync
- preserve an existing valid DMG when HTTP status, streaming, or filesystem operations fail
- bound upstream connect and stalled-read operations without imposing a total large-download deadline
- hash cached DMGs incrementally instead of loading the full image into memory

## Root cause

The updater created `Codex.dmg` before the GET request succeeded. That truncated a valid cache and allowed failed transfers to leave partial data at the recorded path. The wrapper-update path also read the entire DMG into memory to derive its package version.

## Validation

- `cargo fmt --check`
- `cargo test -p codex-update-manager upstream` — 7 passed
- `cargo test -p codex-update-manager wrapper_apply::tests::derives_package_version_with_streamed_dmg_hash` — 1 passed
- updater suite excluding three independently reproduced `upstream/main` feature-picker race failures — 217 passed
- clean `upstream/main` baseline reproduced the same three feature-picker failures (214 passed, 3 failed)
- `git diff --check`

### DIFF
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -11,7 +11,6 @@ use crate::{
 };
 use anyhow::{Context, Result};
 use chrono::{Duration as ChronoDuration, Utc};
-use reqwest::Client;
 use serde::Deserialize;
 use std::{
     ffi::OsString,
@@ -933,7 +932,7 @@ async fn run_check_cycle(
         return Ok(());
     };
 
-    let client = Client::builder().build()?;
+    let client = upstream::http_client()?;
 
     sync_runtime_state(config, state);
     state.status = UpdateStatus::CheckingUpstream;
--- a/updater/src/upstream.rs
+++ b/updater/src/upstream.rs
@@ -5,8 +5,35 @@ use chrono::{DateTime, Utc};
 use futures_util::StreamExt;
 use reqwest::{header, Client};
 use sha2::{Digest, Sha256};
-use std::path::{Path, PathBuf};
-use tokio::{fs::File, io::AsyncWriteExt};
+use std::time::Duration;
+use std::{
+    path::{Path, PathBuf},
+    sync::atomic::{AtomicU64, Ordering},
+};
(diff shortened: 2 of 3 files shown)


Yo-DDV (commented):
Thanks for hardening the updater path. I checked current head `c1af3bf7`.

I ran:

- `cargo fmt --check`
- `cargo test -p codex-update-manager upstream`
- `cargo test -p codex-update-manager wrapper_apply::tests::derives_package_version_with_streamed_dmg_hash`

CI is green. The atomic download/cache behavior and streamed hash path look aligned from my side. Final review and merge stay with @ilysenko.

Leay15:
@ryan-mt Thank you for these hardening changes!  
I reviewed the runtime behavior, and everything looks good.  
Could you please bump the codex-update-manager crate to the next patch version, 0.9.1?

Leay15 (approved):

ilysenko:
Thanks for the update. I found two additional blockers:

- `download_dmg()` publishes every download to the shared `downloads/Codex.dmg`, but the wrapper-apply path does not use the daemon check lock. Concurrent downloads can overwrite each other and leave a caller using SHA/version metadata that no longer matches the file at the returned path. Please serialize publication/use across both paths or publish immutable content-addressed files.
- The temporary file is synced before `rename`, but the parent directory is not synced afterward. Please make the completed publication crash-durable before reporting success and persisting updater state.

avifenesh (dismissed):

ilysenko:
Thank you for identifying and implementing the updater download-hardening direction. We incorporated that work into #834 with the outstanding review concerns addressed: downloads are published to immutable SHA-256-addressed paths so daemon and wrapper flows cannot replace each other's input, the completed file and parent directory are synced, and hashing is streamed in both the updater and acceptance engine. I am closing this PR as superseded by #834.

PR #788, opened 2026-07-10
fix(chrome): bound native bridge resources

ryan-mt:
## Summary

- reject inbound native-messaging frames above 64 MiB before allocating their declared body
- expire pending request correlations after 10 minutes
- cap unanswered correlations at 1024 per bridge direction
- return correlated JSON-RPC errors when a pending map is at capacity

## Root cause

The bridge trusted the native 32-bit length prefix and allocated a matching buffer without a protocol ceiling. Unanswered Chrome/client request correlations also remained in memory until a response or disconnect, allowing long-lived processes to grow state without bound.

## Validation

- `cargo fmt --check`
- four focused frame/prune/cap/correlation regression tests
- `cargo test -p codex-computer-use-linux --bin codex-chrome-extension-host -- --skip tests::runtime_ensure_keeps_child_alive_after_request_returns` — 51 passed
- clean `upstream/main` baseline independently reproduced the skipped fixture failure (`required_path_missing`), 47 passed / 1 failed
- `git diff --check`

Chrome documents 64 MiB as the maximum message sent to a native messaging host.

### DIFF
--- a/computer-use-linux/src/bin/codex-chrome-extension-host.rs
+++ b/computer-use-linux/src/bin/codex-chrome-extension-host.rs
@@ -36,7 +36,10 @@ const MAX_ACCEPTED_FRAME_BYTES: usize = 64 * 1024 * 1024;
 const PENDING_REQUEST_TTL: Duration = Duration::from_secs(10 * 60);
 /// Bounds unanswered correlations independently in each bridge direction.
 const MAX_PENDING_REQUESTS_PER_DIRECTION: usize = 1024;
+/// Bounds retained string IDs to about 4 MiB per direction at the entry cap.
+const MAX_PENDING_REQUEST_ID_STRING_BYTES: usize = 4 * 1024;
 const PENDING_REQUEST_LIMIT_ERROR_CODE: i64 = -32001;
+const INVALID_REQUEST_ERROR_CODE: i64 = -32600;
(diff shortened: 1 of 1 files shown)


Yo-DDV (commented):
Thanks, this one looks useful. I checked current head `6cd3a1d1`.

I ran:

- `cargo fmt --check`
- `cargo test -p codex-computer-use-linux --bin codex-chrome-extension-host -- --skip tests::runtime_ensure_keeps_child_alive_after_request_returns`

The focused resource-bound tests pass, CI is green, and I did not see a suspicious path or blocker in the bridge diff. Final review and merge stay with @ilysenko.

ilysenko:
Thanks for the update. I found one blocker:

- The 1,024-entry cap does not bound retained memory because each pending entry clones the original JSON-RPC `id`, while an accepted frame may be up to 64 MiB. A peer can therefore queue large unanswered IDs and retain tens of GiB despite the cap. Please constrain the accepted ID type/serialized size or enforce a total byte budget for pending correlations, with a correlated error where possible.

avifenesh (dismissed):

ilysenko (approved):
Reviewed current head: the retained-ID memory blocker is fixed, resource bounds are sound, and validation passes.

PR #790, opened 2026-07-10
fix(updater): validate check intervals

ryan-mt:
## Summary

- reject a zero updater check interval before daemon timers start
- reject hour values that overflow seconds, Chrono duration, or the platform monotonic timer range
- centralize checked Tokio and Chrono interval construction
- preserve existing defaults and TOML format while returning config-path-aware validation errors

## Root cause

`check_interval_hours` was deserialized as an unrestricted `u64` and multiplied directly by 3600 before calling `tokio::time::interval`. Zero panics in Tokio and large values could overflow or exceed Chrono/platform timer representation.

## Validation

- `cargo fmt --check`
- config-focused tests — 17 passed after one independently reproduced baseline fixture-race skip
- updater suite — 219 passed after exactly three independently reproduced `upstream/main` feature-picker race skips
- clean `upstream/main` baseline reproduced the same three failures
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -548,7 +548,7 @@ dependencies = [
 
 [[package]]
 name = "codex-update-manager"
-version = "0.9.0"
+version = "0.9.1"
 dependencies = [
  "anyhow",
  "chrono",
--- a/updater/Cargo.toml
+++ b/updater/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "codex-update-manager"
-version = "0.9.0"
+version = "0.9.1"
 edition = "2021"
 
 [dependencies]
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -402,7 +402,7 @@ async fn run_daemon(
     }
     info!("daemon initialized");
 
-    time::sleep(Duration::from_secs(config.initial_check_delay_seconds)).await;
+    time::sleep(config.initial_check_delay_duration()).await;
     if let Err(error) = run_check_cycle_from_disk(config, state, paths).await {
         error!(?error, "initial check failed");
     }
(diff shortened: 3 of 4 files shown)


Yo-DDV (commented):
Thanks for this. I checked current head `00b38289`.

I ran:

- `cargo fmt --check`
- `cargo test -p codex-update-manager config::tests`

The zero/overflow interval cases are covered and CI is green. Nothing blocking from my side. Final review and merge stay with @ilysenko.

Leay15 (changes requested):
@ryan-mt Thank you for the fix!  

Just one comment before I approve:
`check_interval_hours = 0` should be treated as a recoverable configuration error, not a fatal one.  
The updater should automatically fall back to the default value of 6 hours and emit a warning indicating that the configured value is invalid and that the default is being used.  
Values that cannot be safely represented due to `Duration`, `Chrono`, or platform timer overflows should remain fatal errors.  
Please also add a test covering both the fallback behavior and the warning. :)

Leay15:
One more comment: please remember to bump the crate version to the next patch release once the remaining PRs land :)

avifenesh (approved):
- updater/src/config.rs:176-187 — Zero interval fallback is correct: warns with context, substitutes default, proceeds.
- updater/src/config.rs:221-247 — Validation covers all overflow paths: seconds multiply, Chrono duration, platform timer checked_add.
- updater/src/config.rs:645-653 — Test confirms warning emission and fallback to 6h default.
- updater/src/config.rs:656-692 — Overflow cases correctly reject with path-aware errors.

Verified the change is correct and adequately tested.

ilysenko (approved):
The requested recoverable zero-interval behavior is now implemented with a warning and a 6-hour fallback; overflow cases remain fatal. Updater version is bumped to 0.9.1. The full updater suite passes locally (222 tests), and all GitHub checks pass.

PR #791, opened 2026-07-10
fix(linux): skip stale bundled-plugin reconcile workers

### LINKED ISSUE #789: Linux bundled marketplace becomes incomplete during cold-start reconcile
## Summary

On Linux, an already-installed bundled Browser plugin can be temporarily absent
from an early cold-start feature snapshot. An older asynchronous reconcile
worker can then treat that stale snapshot as current, remove the installed
plugin, and rewrite the runtime marketplace without its assets. A later queued
worker sees the current snapshot and installs Browser again.

This creates a first-launch interval in which Browser settings report the
plugin as unavailable. The renderer can also observe missing marketplace assets
while the incomplete snapshot is active.

## Environment

- NixOS 25.11
- Niri / Wayland
- Nix package built from this repository
- upstream payload: `26.707.31428`

## Runtime evidence

Before the fix, a cold launch recorded:

1. a runtime marketplace with three plugins and no Browser;
2. `bundled_plugin_uninstall_requested` for `pluginName=browser` with
   `reason=not_in_bundled_marketplace_plugin_names`;
3. renderer reads against Browser assets while that marketplace was active;
4. a later four-plugin marketplace containing Browser;
5. `bundled_plugin_install_requested` for Browser with `reason=missing`.

Reinstalling Browser from Settings repaired that launch, but the same sequence
could recur after another cold start.

After the fix, two consecutive rebuilt NixOS cold starts recorded:

1. the stale worker completing without marketplace side effects;
2. the current worker writing the complete four-plugin marketplace:
   `browser`, `chrome`, `computer-use`, and `visualize`;
3. Browser uninstall requests: `0`;
4. Browser install requests: `0`;
5. Browser availability resolving to `available` before its Settings assets
   were used.

## Confirmed root cause

The bundled-plugin queue captures a semantic feature snapshot before its async
preflight. During that preflight, a newer startup snapshot can be stored and a
second reconcile queued. The older worker still reaches the destructive
marketplace boundary and treats absence from its stale snapshot as an
authoritative disable.

The first local mitigation retained an already-installed `browser` plugin.
That approach was rejected because it could not distinguish transient startup
state from an explicit disable and was Browser-specific.

## Implemented invariant in #791

The final patch carries the reconciler's existing semantic snapshot hash into
the async worker. Immediately before destructive marketplace mutation, the
worker compares its captured hash with the latest scheduled hash:

- stale workers exit through the existing `try/finally` path;
- only the newest semantic snapshot may mutate the marketplace;
- an authoritative explicit disable still reaches the normal uninstall path;
- no Browser allowlist, cache-retention rule, marketplace reuse, or
  availability bypass is introduced;
- Browser config, WSL, pane, and server-side gates remain unchanged.

This is plugin-agnostic: the guard is based on the complete semantic snapshot,
not a plugin name.

## Covered edge cases

The committed regressions cover:

- stale `false -> true` feature transitions while the first worker is blocked
  in preflight;
- authoritative disabled snapshots reaching the newest worker normally;
- simultaneous `externalBrowserUse` and `inAppBrowserUse` changes, proving the
  comparison uses the latest complete multi-field snapshot;
- valid minified identifiers containing `$`, escaped through the existing
  shared `escapeRegExp` helper;
- ambiguous minified worker shapes failing closed;
- insertion-order drift and exact patch idempotence;
- the current `26.707.31428` bundle applying once and remaining idempotent.

## Validation

- focused patcher tests passed;
- current-DMG apply/idempotence check passed;
- current GitHub CI is green;
- NixOS rebuild and switch passed;
- two consecutive cold-start traces retained the complete marketplace with no
  Browser uninstall/reinstall;
- collaborator re-review confirmed the `$` identifier and multi-field race
  coverage and reported no further blocker in the checked paths.

## Acceptance criteria

- [x] Browser remains installed and its Settings assets are readable throughout
      two consecutive Linux cold starts.
- [x] An absent Browser is not synthesized or installed by the guard.
- [x] An explicitly disabled Browser reaches the normal authoritative
      reconcile path.
- [x] Other bundled plugins retain the existing newest-snapshot reconcile
      behavior; the guard contains no plugin-name special case.
- [x] Browser config, WSL, pane, and server-side gates remain unchanged.
- [x] The current-DMG patch, focused tests, and GitHub CI pass.
- [x] The rebuilt package passes the reported first-screen and two-cold-start
      runtime validation.

## Architecture note

A source-level implementation could instead coalesce queued updates or pass a
typed generation token into the worker. The required invariant is the same:
stale semantic input must not mutate the installed marketplace.

## Related work

- #164 handled a read-only launcher cache and permission monitor.
- #688 made launcher-side bundled-plugin cache synchronization concurrent, but
  that work completes before Electron starts.
- #762 stages portable bundled plugins for packaged Linux builds.

This issue occurs later, inside the Electron main-process reconcile queue.

Vantrongs:
Fixes #789.

## Summary

- carry the bundled-plugin reconciler's existing semantic snapshot hash into
  its async worker;
- skip a stale worker immediately before destructive marketplace mutation;
- keep the descriptor fail-soft as `optional` drift;
- cover stale `false -> true`, authoritative disable, ambiguous minified worker,
  insertion-order drift, dollar-prefixed minified identifiers, multi-field
  availability snapshots, and idempotence.

## Root cause

During Linux cold start, a newer feature snapshot can be scheduled while an
older bundled-plugin worker is still in async preflight. The older worker then
materializes an incomplete marketplace, uninstalls Browser, and the queued
worker installs it again. The renderer can remain in the transient unavailable
state.

The guard uses the semantic hash upstream already computes for reconcile
equivalence. It contains no Browser allowlist, cache retention, marketplace
reuse, or availability bypass. An explicit disable produces a new hash, so old
work becomes stale and the newest worker retains the normal uninstall behavior.

## User impact

Browser no longer disappears when this cold-start race occurs or requires a
manual reinstall from Settings after an affected launch.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js scripts/patches/impl/main-process/browser.test.js` — passed;
- current `26.707.31428` main bundle: patch applies once and is idempotent;
- GitHub CI for `eaf07db` is green;
- two consecutive rebuilt NixOS cold starts: runtime marketplace contained
  `browser`, `chrome`, `computer-use`, and `visualize`; Browser uninstall and
  install requests were both zero;

## Architecture note

A typed generation token or source-level queue coalescing could enforce the
same invariant at another boundary. The current patch keeps that invariant
inside the existing Linux bundle patch without adding a Browser-specific rule.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -51,6 +51,7 @@ const {
 const {
   applyBrowserUseNodeReplApprovalPatch,
   applyBrowserUseNodeReplApprovalAssets,
+  applyLinuxBundledPluginReconcileStaleSnapshotPatch,
   applyLinuxBrowserUseRouteLivenessPatch,
   applyLinuxChromeExtensionStatusPatch,
   applyLinuxExternalOpenEnvPatch,
@@ -864,6 +865,7 @@ test("default core patch descriptors are grouped and unique", () => {
     "linux-chrome-plugin-auto-install",
     "linux-chrome-native-host-runtime",
     "browser-use-node-repl-approval",
+    "linux-bundled-plugin-reconcile-stale-snapshot",
     "linux-browser-use-route-liveness",
     "linux-chrome-extension-status",
     "linux-local-app-server-feature-enablement-handler",
@@ -924,6 +926,12 @@ test("default core patch descriptors are grouped and unique", () => {
   );
   assert.equal(
     descriptors.find((descriptor) => descriptor.id === "local-environment-action-modal-draft")?.ciPolicy,
+    "optional",
+  );
+  assert.equal(
+    descriptors.find(
+      (descriptor) => descriptor.id === "linux-bundled-plugin-reconcile-stale-snapshot",
+    )?.ciPolicy,
     "optional",
   );
   assert.equal(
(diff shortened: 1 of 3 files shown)


Yo-DDV (commented):
Hi, thanks for putting this together. I found one case that seems worth fixing before this leaves draft.

The new patch captures minified symbols and then interpolates them directly into `new RegExp(...)`. That is fragile for valid JS identifiers containing `$`, and this repository already has fixtures with that style of minified name.

I reproduced it with the reconciler shape by using `$` as the captured feature snapshot variable. The patch returns the source unchanged and emits `WARN: Could not find bundled plugin reconcile snapshot`. Since this descriptor is optional drift, CI can stay green while this specific fix is not applied.

The small fix should be to run the captured identifiers through `escapeRegExp` from `scripts/patches/lib/minified-js.js` before building those regexes, plus a regression fixture with `$` in the snapshot, hash, or worker identifiers.

One secondary coverage gap I noticed is that the new tests only exercise `inAppBrowserUse`, while #789 mentions other bundled assets during the transition. I would not block on naming every plugin, but it would be good to prove the guard against a multi-field availability snapshot too.

Vantrongs:
@Yo-DDV Thanks — reproduced and fixed in `eaf07db`.

- imported the existing `escapeRegExp` helper and applied it to every captured
  identifier interpolated into a dynamic regex (feature snapshot, force arg,
  captured snapshot, and worker name);
- added an all-`$` minified-identifier regression that exercises the complete
  stale `false -> true` reconcile path and exact idempotence;
- expanded the race fixture to serialize both `externalBrowserUse` and
  `inAppBrowserUse`, and verified that only the latest full snapshot reaches
  destructive reconciliation;
- preserved the authoritative-disabled regression.

Validation:

- focused patcher tests passed;
- current `26.707.31428` bundle apply/idempotence check passed;
- refreshed GitHub CI is green.

Yo-DDV:
Thanks for the quick update.

I rechecked the two points from my review on `eaf07db`. The `$` identifier case is covered now, and the race fixture does exercise both `externalBrowserUse` and `inAppBrowserUse`.

The focused patcher test passes locally for me, and the PR checks are green now too. I do not see another blocker from my side in the paths I checked.

ilysenko (approved):
Reviewed current head eaf07db2. The stale-worker guard is optional and fail-soft, prior feedback is fixed, and all checks pass.

PR #792, opened 2026-07-10
fix(updater): block rolled-back DMG hashes

ryan-mt:
## Summary

- persist the full SHA-256 of a rolled-back upstream DMG
- block exact hash matches even when a later download receives a new timestamped package version
- retain legacy version fallback only for state written before the hash field existed
- clear both rollback block identifiers on normal install and successful install recovery

## Root cause

Rollback blocking used generated package versions. Those versions include download time, while comparison discards `+hash` metadata, so upstream header drift could make the same bad DMG look like a different candidate.

## Compatibility

The existing `rollback_blocked_candidate_version` field remains readable. The new SHA-256 field is optional with a serde default, so existing state files continue to load.

## Validation

- `cargo fmt --check`
- rollback-focused suite — 17 passed
- updater suite — 220 passed after exactly three independently reproduced `upstream/main` feature-picker race skips
- clean baseline reproduced the same three failures
- `git diff --check`


Yo-DDV (commented):
Thanks for the rollback hardening. I checked current head `409661e9`.

I ran:

- `cargo fmt --check`
- `cargo test -p codex-update-manager rollback`
- `cargo test -p codex-update-manager state::tests`

The legacy-state fallback and new SHA blocking behavior are covered, CI is green, and I did not find a blocker. Final review and merge stay with @ilysenko.

Leay15 (approved):
@ryan-mt Thank you for separating the two changes and hardening the updater path!  
Everything looks good to me. Could you please bump the crate version to the next patch release once the next PR lands? :)

avifenesh (commented):
Hash-based rollback blocking is correct, but one edge case to consider:

**updater/src/app.rs:1488** — `rollback_blocks_candidate` correctly prioritizes hash over version. However, what happens if OpenAI legitimately re-releases the same version with a fixed DMG (same version string, different hash)? The current logic would still block it. Consider whether rollback should be clearable via a flag or timeout, or document that users must manually clear state to retry a same-version re-release.

Tests and backward compat look good.

PR #794, opened 2026-07-10
Suppress tray icons for multi-launch instances

colonelpanic8:
## Summary

- keep `--new-instance` / multi-launch processes out of the Linux system tray
- preserve the primary app tray and its existing user preference
- document that secondary instances are trayless
- add patcher and runtime smoke coverage for the multi-launch guard

## Root cause

The StatusNotifier watcher was not duplicating a single item. Each visible ChatGPT icon had a distinct D-Bus owner PID. Three of the extra items belonged to explicit `--new-instance` dev/test launches. Those launches use isolated profiles, so each profile inherited the default-enabled tray setting and registered another identical SNI item.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` (346 passing)
- focused `test_linux_tray_patch_smoke` and `test_linux_single_instance_patch_smoke` from `tests/scripts_smoke.sh`

The full smoke script is blocked earlier on this NixOS host by the existing Fedora Atomic test invoking `bash` under the hardcoded `PATH=/usr/bin:/bin`; the affected smoke functions pass when run directly.

### DIFF
--- a/docs/build-and-packaging.md
+++ b/docs/build-and-packaging.md
@@ -135,6 +135,14 @@ CODEX_MULTI_LAUNCH_PORT_RANGE=5175-5199 ./codex-app/start.sh --new-instance
 CODEX_MULTI_LAUNCH=1 CODEX_MULTI_LAUNCH_PORT_RANGE=5175-5199 ./codex-app/start.sh
 ```
 
+Secondary instances do not create a system-tray icon by default, and closing
+their last window exits normally. To keep an individual secondary instance in
+the tray and retain close-to-tray behavior, set `CODEX_MULTI_LAUNCH_TRAY=1`:
+
+```bash
+CODEX_MULTI_LAUNCH_TRAY=1 ./codex-app/start.sh --new-instance
+```
+
 ## Package Formats
 
 After `make build-app` or `make build-app-fresh`, build a package from
--- a/docs/build-and-packaging.md
+++ b/docs/build-and-packaging.md
@@ -124,6 +124,10 @@ Open an independent app process:
 ./codex-app/start.sh --new-instance
 ```
 
+Secondary instances do not register system tray items. This keeps temporary
+development and test instances from adding indistinguishable icons alongside
+the primary app's tray item.
+
 Configure the port range or make every launch use multi-instance mode:
 
 ```bash
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -219,7 +219,7 @@ Launches the $CODEX_LINUX_APP_DISPLAY_NAME app.
 
 Options:
   -h, --help                  Show this help message and exit
-  --new-instance              Start a separate app instance on the first free port in the multi-launch range
+  --new-instance              Start a separate trayless app instance on the first free port in the multi-launch range
   --new-chat                  Open the main window on a new chat
   --quick-chat                Open a projectless quick chat
   --prompt-chat               Show the compact prompt for a new chat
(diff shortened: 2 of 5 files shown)


Yo-DDV (commented):
Thanks for handling the multi-launch tray case. I checked current head `102271fb`.

I ran:

CI is green. The env gate stays scoped to multi-launch, and I did not find a blocker from my side. Final review and merge stay with @ilysenko.

colonelpanic8:
@ilysenko this pretty desperately needs a merge. its very annoying

ilysenko:
Thanks for addressing the duplicate tray items. I personally use multiple tray icons to manage separate running instances, so I’d like to preserve that workflow as an explicit option rather than remove it entirely.

I found one blocker and suggest making the behavior configurable:

- Multi-launch currently suppresses tray startup, but the existing Linux close-to-tray handler can still hide the last window. This leaves a secondary process running with no tray item or visible window to restore it.
- Please add an explicit override such as `CODEX_MULTI_LAUNCH_TRAY=0|1`, defaulting to `0` for `--new-instance`. Use the same resolved condition for both tray startup and close-to-tray: when disabled, closing the last secondary window should exit normally; when enabled, preserve the tray and existing hide-to-tray behavior.
- Please document the override and test both modes, including inherited-environment handling so a stale internal marker cannot affect ordinary primary launches.

The default trayless behavior still makes sense for temporary dev/test instances; this would just keep multiple-tray workflows available for users who intentionally want them.

avifenesh (commented):

ilysenko:
Thanks for addressing the requested multi-launch tray behavior. The implementation now looks correct. One stale smoke assertion still blocks CI: please update the exact launcher export check to include `CODEX_LINUX_MULTI_LAUNCH_TRAY`, matching the new export line.

Before pushing updates, please run the full relevant local test suite (`bash tests/scripts_smoke.sh`) in an environment where it can complete, or at minimum ensure the complete launcher smoke section runs. This failure would have been caught locally.

Also, please have an agent independently review each PR before requesting maintainer review. The agent should inspect the full diff and run the relevant tests, not only the newly added focused cases.

Vantrongs (commented):
Reviewed the full current diff at `d149ba8` after the latest launcher-smoke fix. I did not find a blocker.

The internal multi-launch markers are cleared before argument parsing, the same resolved condition controls both tray startup and close-to-tray, and the two supported modes remain paired: trayless secondary instances exit on last-window close, while `CODEX_MULTI_LAUNCH_TRAY=1` keeps the tray and hide-to-tray behavior. Primary/single-instance routing is unchanged.

Verified on the current head:

- `node --test scripts/patch-linux-window-ui.test.js` — 361/361
- `nix develop -c bash tests/scripts_smoke.sh` — passed

GitHub CI is also 6/6 green. The inherited-marker regression is structural rather than a live Electron launch, but the launcher unsets both internal markers directly before parsing and I found no logic gap in that path.

ilysenko:
Thanks for the work on this and for addressing the earlier feedback. After reconsidering the scope, we are not planning to add multi-launch-specific tray behavior at this time. This is not a current priority for the default Linux build, but we may revisit it later if the need becomes broader. Thank you again for the implementation and testing.

ilysenko:
Thanks again for the final implementation. Since this PR comes from a fork with maintainer edits disabled and `main` has moved forward, its reviewed commits have been carried onto the current base in #859, preserving the original authorship. #859 now supersedes this closed PR.

PR #795, opened 2026-07-10
Fix updater builder DMG intel staging

kdegeek:
## Summary

- stage `scripts/lib/upstream-dmg-intel.js` in native packages' update-builder bundle
- assert the helper is present in the Debian packaging smoke test

## Root cause

`scripts/patch-linux-window-ui.js` imports `./lib/upstream-dmg-intel.js`, but `stage_update_builder_bundle` did not copy that dependency into `/opt/codex-desktop/update-builder`. Local update rebuilds therefore failed at the Linux patch step with `MODULE_NOT_FOUND` after successfully extracting the upstream DMG and rebuilding native modules.

## Impact

Packaged installations can rebuild newer ChatGPT/Codex DMGs instead of repeatedly failing before package generation. The assertion protects the update-builder dependency from being omitted again.

## Validation

- verified the new smoke assertion fails before the staging change
- `bash -n scripts/lib/package-common.sh`
- performed a live `codex-update-manager check-now` rebuild from the affected DMG
- inspected the resulting `.deb` and confirmed both the helper file and its generated manifest entry are present

### DIFF
--- a/scripts/lib/package-common.sh
+++ b/scripts/lib/package-common.sh
@@ -773,6 +773,7 @@ stage_update_builder_bundle() {
     cp "$REPO_DIR/scripts/lib/linux-features.js" "$update_builder_root/scripts/lib/linux-features.js"
     cp "$REPO_DIR/scripts/lib/linux-features.sh" "$update_builder_root/scripts/lib/linux-features.sh"
     cp "$REPO_DIR/scripts/lib/linux-target-context.js" "$update_builder_root/scripts/lib/linux-target-context.js"
+    cp "$REPO_DIR/scripts/lib/upstream-dmg-intel.js" "$update_builder_root/scripts/lib/upstream-dmg-intel.js"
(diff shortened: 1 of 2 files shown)


Yo-DDV (commented):
Thanks for the fix. I did a draft pass on current head `f81f7228`.

I ran:

- `bash -n scripts/lib/package-common.sh tests/scripts_smoke.sh`
- Debian staged update-builder smoke path, including `patch-linux-window-ui.js --help`

CI is green. I did not find a blocker in the staged update-builder coverage, but since the PR is still draft I am leaving timing to you. Final review and merge stay with @ilysenko once it is ready.

ilysenko:
Closing as superseded. The same update-builder staging fix and Debian smoke assertion are already present on `main` via commit `1d73851c` from #773, while this draft branch is now substantially behind the current base. Thanks for identifying and validating the missing dependency.

PR #796, opened 2026-07-10
fix(settings): serialize shared writes

ryan-mt:
## Summary
- coordinate updater, Electron, and Read Aloud writes through one bounded cross-process lock protocol
- merge under the lock and atomically replace settings.json from a same-directory temp file
- recover only stale locks owned by dead processes, with ordered recovery claims to prevent lock theft
- await Electron/Read Aloud persistence and cover concurrency, stale recovery, timeout, and cleanup

## Validation
- cargo fmt --check
- focused updater settings tests: 6 passed
- full updater suite: 221/224; the exact 3 failures reproduce on upstream/main (214/217)
- focused core settings tests: 6 passed
- focused Read Aloud writer test: 1 passed
- full core Node suite: 341/349; the exact 8 Windows/environment failures reproduce on upstream/main (338/346)
- full Read Aloud suite: 54/59; the exact 5 Windows/Python/path failures reproduce on upstream/main (53/58)
- node --check on both patch implementations
- git diff --check


Yo-DDV (commented):
Thanks, I rechecked current head `6c53dee1` after the clippy fix.

I ran:

- `cargo fmt --check`
- `cargo test -p codex-update-manager config::tests`
- `node --test --test-name-pattern='settings|Settings|read aloud|Read Aloud' scripts/patch-linux-window-ui.test.js linux-features/read-aloud/test.js`

CI is green now. I did not find a remaining blocker in the shared settings write path. Final review and merge stay with @ilysenko.

ilysenko:
Thanks for tackling the shared settings race. I found three blockers in the recovery protocol:

- Stale recovery performs a final read and then unlinks by pathname. Another recoverer can replace the stale lock between those operations, causing the first process to delete the new owner’s lock and break mutual exclusion. Please use an atomic ownership/inode-safe reclamation design and add a deterministic replacement-lock race test. The Rust path must also recover an empty stale lock left by a crash between `create_new` and owner initialization.
- Recovery follows `settings.json.lock` symlinks and opens them for append, which can modify an unrelated target file. Please reject symlinks/non-regular lock sidecars without following them in both Rust and Node implementations.
- Please make settings publication crash-durable by syncing the parent directory after rename, and bump `codex-update-manager` to the required next patch version.

Recovery claims should also avoid permanent election poisoning when a claimant remains alive but abandons recovery.

avifenesh (commented):

ilysenko:
Thanks for the substantial work on this. The underlying settings-write race is a valid concern, but this PR has grown too broad to review and maintain safely as one change.

It introduces a shared locking and recovery protocol across three separately owned surfaces:

- the Rust update manager;
- the core Electron settings writer;
- the optional Read Aloud feature.

Each surface has different lifecycle, failure, compatibility, and maintenance requirements. Combining them has produced a large cross-language protocol with significant recovery and durability complexity, while the branch is already conflicting with the current `main`.

We are going to close this PR rather than continue expanding it. If you would like to pursue the problem further, please start with a focused issue or design proposal defining the required cross-process contract, then submit a small, ordered series of PRs for the updater, core Electron path, and Read Aloud feature. Each PR should remain independently understandable and keep feature-specific behavior inside its feature directory.

Before opening replacements, please verify the proposed scope against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant tests locally, perform an independent full-diff review with a local LLM, and wait for all required CI checks to complete before requesting maintainer review.

PR #797, opened 2026-07-10
fix(launcher): stage marketplace metadata atomically

ryan-mt:
## Summary
- stage bundled marketplace metadata through a unique destination-adjacent temp file
- chmod the temp and atomically rename it over the live target
- preserve the existing cache and clean temp files when copy, chmod, or rename fails
- add focused smoke coverage and reject delete-before-copy staging

## Validation
- focused atomic staging smoke: pass for cp/chmod/mv failures and successful replacement
- static launcher staging assertions: pass
- git diff --check

The broad WSL smoke entrypoint stops at its existing environment prerequisite because Node is not installed in that WSL environment.

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -713,7 +713,7 @@ stage_bundled_marketplace_metadata() {
     }
     if cp "$source_marketplace" "$marketplace_temp" && \
        chmod u+w "$marketplace_temp" && \
-       mv -f -- "$marketplace_temp" "$marketplace_target"; then
+       mv -fT -- "$marketplace_temp" "$marketplace_target"; then
         return 0
     fi
     rm -f -- "$marketplace_temp"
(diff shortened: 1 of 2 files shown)


Yo-DDV (commented):
Thanks for this cleanup. I checked current head `74ddab74`.

I ran:

The smoke suite covers the atomic marketplace staging and failure cases, CI is green, and I did not find a blocker. Final review and merge stay with @ilysenko.

ilysenko:
Thanks for making the marketplace metadata update atomic and fail-soft. I found one edge case that should be addressed before merge.

`mv -f -- "$marketplace_temp" "$marketplace_target"` treats an existing `marketplace_target` symlink to a directory as a destination directory. In that case the temporary file is moved inside the linked directory, potentially outside `CODEX_HOME`, the symlink remains in place, and the function reports success without replacing the live marketplace metadata.

Please use no-target-directory replacement semantics, such as `mv -fT -- ...` on this Linux-only launcher, or an equivalent safe approach, and add a regression test with `marketplace.json` symlinked to a directory.

The remaining failure handling and destination-adjacent staging look good.

avifenesh (commented):

ilysenko (approved):
Reviewed current head. The atomic destination-adjacent staging is focused, fail-soft, and safely handles destination symlinks.

PR #800, opened 2026-07-10
test: resolve feature fixture tools from PATH

Vantrongs:
## Summary

- resolve Bash from executable regular files in absolute directories from the
  active `PATH`;
- use that absolute path for spawned shell scripts and generated fixture
  shebangs;
- preserve the caller's host tool directories when prepending the temporary
  Omarchy fixture directory;
- ignore empty and relative `PATH` entries so tests cannot resolve tools from
  the current working directory.

## Why

The two optional-feature suites assumed FHS paths such as `/bin/bash` or
`/usr/bin`, even though the implementation being tested does not require those
locations. On NixOS and other non-FHS systems, the fixtures could fail before
exercising feature behavior.

This is test-only. It does not change either feature's runtime implementation
or CI policy.

## Scope

- `linux-features/node-repl-reaper/test.js`
- `linux-features/omarchy-theme/test.js`

The similar open-target and MCP helper fixture failures are intentionally not
included; they belong to separate functional changes and reproduce unchanged
on clean `origin/main`.

## Validation

- `node --check linux-features/node-repl-reaper/test.js`
- `node --check linux-features/omarchy-theme/test.js`
- `node --test linux-features/node-repl-reaper/test.js linux-features/omarchy-theme/test.js` — 10/10
- `env PATH=":.:$PATH" node --test linux-features/node-repl-reaper/test.js linux-features/omarchy-theme/test.js` — 10/10
- GitHub Rust and Smoke Tests job, including the full Node runner
- `git diff --check`

This remains a focused fixture portability change; a shared test-tool resolver
may make sense only if more suites converge on the same contract.

## Checklist

- [x] The change is test-only and limited to the two sister fixtures.
- [x] The focused tests pass with normal and adversarial `PATH` values.
- [x] GitHub CI is green, including the full Node runner.

### DIFF
--- a/linux-features/node-repl-reaper/test.js
+++ b/linux-features/node-repl-reaper/test.js
@@ -11,6 +11,21 @@ const test = require("node:test");
 const REAPER = path.join(__dirname, "reaper.sh");
 const LONG_RUNNING_NODE_ARGS = ["-e", "setInterval(() => {}, 1000)"];
 
+function commandPath(name) {
+  for (const directory of (process.env.PATH || "").split(path.delimiter)) {
+    if (!directory || !path.isAbsolute(directory)) continue;
+    const candidate = path.join(directory, name);
+    try {
+      fs.accessSync(candidate, fs.constants.X_OK);
+      if (!fs.statSync(candidate).isFile()) continue;
+      return candidate;
+    } catch {}
+  }
+  throw new Error(`could not resolve executable from PATH: ${name}`);
+}
+
+const BASH = commandPath("bash");
+
 function makeFakeApp() {
   const appDir = fs.mkdtempSync(path.join(os.tmpdir(), "codex-node-repl-reaper-test-"));
   fs.mkdirSync(path.join(appDir, "resources"));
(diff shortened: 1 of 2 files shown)


Yo-DDV:
Thanks for putting this one together. I checked the two fixture changes and I do not see a code blocker in the paths touched here.

The checks I ran locally line up with the intent of the PR. The two focused tests pass with the normal PATH, with empty/relative PATH entries in front, and with Bash resolved from a temporary absolute directory. The full Node runner also passes for me now at 866/866, and `nix flake check` passes locally too.

One small cleanup before normal review. The PR body still looks like draft text in two places. The 864/866 note seems stale on this head, and I would either remove or trim the provenance sentence at the end since the validation list is enough.

Once you are comfortable with the body, can you mark it ready for review? That should make it easier for @ilysenko to do the final pass.

Vantrongs:
Thanks for checking this. I removed the stale local aggregate note and trimmed the draft provenance text, then marked the PR ready for review.

ilysenko (approved):
Reviewed current head f7c94ed9. This is a focused test-only portability change: Bash is resolved from absolute executable PATH entries, relative/empty entries are ignored, and runtime behavior is untouched. CI is green.

PR #802, opened 2026-07-10
Add Linux global dictation hotkeys

### LINKED ISSUE #798: Add Linux global dictation hotkeys
## What is missing

The current desktop bundle already contains the global dictation controller, its overlay, history, hold-to-talk mode, toggle mode, and command keybindings. Linux is excluded by the platform capability check, and the release watcher used by hold-to-talk has no Linux implementation.

This leaves the global dictation settings unavailable even though the rest of the feature is present in the app.

## Proposed Linux path

I would like to add this as an opt-in Linux feature, alongside the existing conversation-mode work.

On X11, Electron can keep registering the shortcut while a small release watcher stops hold-to-talk when the modifier is released. On Wayland, the XDG GlobalShortcuts portal exposes both activation and deactivation events, so the Linux integration can preserve hold and toggle behavior without reading raw input devices.

The implementation should remain capability-based. If the active desktop does not provide the required portal, registration must fail with a useful error instead of reporting support that is not available.

## Acceptance criteria

- Linux no longer reports global dictation as universally unsupported when the feature is enabled.
- Hold and toggle shortcuts work outside the app on X11.
- Wayland uses the GlobalShortcuts portal and handles both activation and deactivation.
- Failed registration, denied portal access, helper exit, duplicate bindings, and app shutdown clean up without leaving a stale shortcut or recording session.
- The feature stays disabled by default and does not change macOS or Windows behavior.
- Source builds and packaged installs stage the same helper and pass the existing Linux feature checks.

I can work on the first implementation. Feedback on the backend split or desktop environments that need extra coverage would be useful before the patch is finalized.

Yo-DDV:
Closes #798.

## What changed

- add an opt-in `global-dictation` Linux feature without changing the default package
- preserve hold and toggle shortcuts on X11 with an event-driven release monitor
- use the XDG GlobalShortcuts portal on Wayland, including activation and deactivation events
- route paste through X11 tools or a keyboard-only RemoteDesktop portal session
- stage the helper consistently in source builds, updater builds, packaged installs, and Nix

Registration fails cleanly when the active desktop cannot provide the required portal. Helper exits, denied requests, duplicate events, concurrent paste requests, app shutdown, and monitor loss are covered by the regression tests.

## Validation

- full Node test suite: 867 passed
- Rust workspace tests, formatting, and Clippy: passed
- package smoke matrix: passed
- Nix evaluation on x86_64 and aarch64, plus a real x86_64 build: passed
- real DMG install and extracted ASAR verification: passed

The Wayland path was exercised against an isolated D-Bus portal implementation because this workstation does not expose the GlobalShortcuts portal. The X11 and package paths were validated on the current Linux environment.


ilysenko:
Thanks for the thorough implementation and validation. I found three blockers before merge:

1. The X11 hold release monitor tracks only modifier groups. For an accelerator such as `Ctrl+Space`, releasing `Space` while continuing to hold `Ctrl` leaves `all_groups_down()` true, so hold-to-talk can continue until Ctrl is released or the 1800-second watchdog expires. Please resolve and track the non-modifier key as part of the chord, and add a regression test that releases the primary key while the modifier remains down.

2. `applyLinuxGlobalDictationMainProcessPatch()` silently returns the original source when the upstream release-watcher sentinel is absent. Since the patch marker is also absent, the descriptor runner records this as `already-applied` rather than a missing optional patch. Please emit a fail-soft warning and ensure the patch report records an optional skip; add a regression test for the missing sentinel.

3. The PR changes the updater bundle contract by adding `global-dictation-linux` to `REQUIRED_BUNDLE_FILES`, but the updater remains at `0.9.0`. Under the repository versioning rules, this backward-compatible feature addition needs a minor bump to `0.10.0`, including `Cargo.lock` and related version references.

The feature remains correctly opt-in/default-disabled, and the portal security, lifecycle handling, staging, packaging, and Nix wiring otherwise look good.

Yo-DDV:
@ilysenko thanks, I pushed the update.

The three blockers should be covered now: X11 tracks the primary key in the chord too, missing release-watcher sentinels now show up as optional skips in the patch report, and the updater is bumped to `0.10.0`.

Local checks are green on my side, including the global-dictation tests, all linux-features tests, patcher tests, updater Rust tests, script smoke tests, and Nix. Let me know if you still see anything missing.

avifenesh (commented):

ilysenko:
Thanks for the update. The three previous blockers are fixed. I found two remaining concerns:

- The X11 primary-key resolver does not match the Electron/portal accelerator vocabulary. For example, `Ctrl+Plus` falls through to a case-sensitive `Plus` keysym lookup, while `xmodmap` reports `plus`, so the release monitor exits with `unsupported accelerator key`. Please align the X11 mapping with `portal_key()` and add regression coverage for named punctuation plus representative media/numpad aliases.
- Because `global-dictation` is an optional, default-disabled feature, its Rust helper should not become part of the default workspace build or be copied into the default updater bundle when the feature is disabled. Please keep the helper as a standalone crate built through `--manifest-path` only when the feature is enabled, and make updater staging conditional as well. The default package/update path should not compile or carry this feature’s helper sources.

Yo-DDV:
@ilysenko thanks, I pushed the follow-up for those two points.

The X11 release watcher now normalizes the Electron/portal accelerator aliases before resolving xmodmap keysyms, with coverage for `Ctrl+Plus`, media keys, and numpad aliases.

I also kept the Rust helper out of the default workspace and default update-builder bundle. It is built through its own manifest, and the helper source is staged only when `global-dictation` is enabled.

Local checks are green on my side, and GitHub CI is running on `6847ef39` now.

ilysenko (approved):
Looks good after the maintainer follow-up. The optional helper now stays out of the default workspace/default updater payload, update-builder stages only source inputs when the feature is enabled, and CI is green.

PR #803, opened 2026-07-10
Fix Appshots availability and X11 monitor lifecycle

moxw:
## Summary

- retarget the Appshots availability descriptor to the current upstream app page bundle and remove the obsolete bundle shape
- replace the per-device X11 monitor/FIFO fanout with one root XInput2 event stream
- stop the XInput2 listener when its Electron parent exits and terminate it during unregister cleanup
- add executable XInput2 event coverage and document the runtime lifecycle

## Root cause

The current `26.707.41301` DMG moved the Appshots availability atom into `app-initial~app-main~page-*.js`. The optional descriptor still targeted the previous long-form chunk name, so the installed patch report marked availability as skipped while the main-process and settings patches applied.

The X11 bare-modifier helper also started one shell pipeline and `xinput test` process for every slave keyboard device, including virtual devices. Failed or replaced Electron instances could leave those grandchildren adopted by init.

## User impact

Linux Appshots becomes available again on the current DMG. On X11, `Alt + Alt` and `Shift + Shift` now use one XInput2 listener instead of a process tree per keyboard device, and the listener exits with its owning Electron process.

## Source of truth

- `linux-features/appshots/patch.js`
- `linux-features/appshots/bin/bare-modifier-monitor`
- `linux-features/appshots/test.js`
- `linux-features/appshots/README.md`

Generated `codex-app/` and package output are not included.

## Validation

- `bash -n linux-features/appshots/bin/bare-modifier-monitor`
- `node --test linux-features/appshots/test.js` — 17 passed
- `node --test linux-features/*/test.js` — 481 passed
- `git diff --check`
- rebuilt from the current official `Codex.dmg`; installed patch report showed all four Appshots descriptors as `applied`
- built and installed the pacman package on Arch Linux/X11
- verified the installed helper exactly matched the source helper

## Scope and limitations

This remains inside the opt-in `appshots` Linux feature. Wayland behavior is unchanged and continues to use the existing non-bare shortcut/GlobalShortcuts portal path.

### DIFF
--- a/linux-features/appshots/README.md
+++ b/linux-features/appshots/README.md
@@ -40,8 +40,9 @@ Privacy and correctness constraints:
 - `Alt + Alt` and `Shift + Shift` are backed by a feature-local
   `bare-modifier-monitor` helper staged into `resources/native/`. It requires
   the left and right modifier keycodes, so tapping only one physical modifier
-  twice does not trigger AppShots. It uses `xinput` and `xmodmap`, so it is
-  expected to work on X11 sessions and fail closed elsewhere.
+  twice does not trigger AppShots. It reads one root XInput2 event stream and
+  stops that listener when its Electron parent exits. It uses `xinput` and
+  `xmodmap`, so it is expected to work on X11 sessions and fail closed elsewhere.
 - On Wayland, the feature stages an Electron args hook that enables
   `GlobalShortcutsPortal`, and the settings dropdown hides the X11-only bare
   modifier shortcuts.
--- a/linux-features/appshots/bin/bare-modifier-monitor
+++ b/linux-features/appshots/bin/bare-modifier-monitor
@@ -165,6 +165,17 @@ else
 fi
 monitor_pid="$APPSHOT_XINPUT_PID"
 event_fd="${APPSHOT_XINPUT[0]}"
+
+sleep 0.1
+if ! kill -0 "$monitor_pid" >/dev/null 2>&1; then
+    monitor_status=0
+    wait "$monitor_pid" || monitor_status=$?
+    monitor_pid=""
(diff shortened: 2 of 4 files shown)


Yo-DDV (commented):
Thanks for picking this up.

I hit one fail-closed case that looks worth fixing. In `linux-features/appshots/bin/bare-modifier-monitor`, the helper now prints `ready` immediately after starting `xinput test-xi2 --root`. If `xinput` exits straight away, the helper still says it is ready and exits 0, so the app can believe the bare-modifier monitor is available on a broken X11 setup.

Repro I used locally was a stubbed `xinput` that accepts `test-xi2 --root` and then exits 2. On this branch the helper returns:

```text
status=0
stdout=ready
```

On `origin/main`, the same kind of startup failure does not emit `ready`. I think the single-stream version needs a small readiness gate before printing `ready`, plus a regression test for this startup-failure path.

moxw:
Addressed in `52cadf5`.

The helper now gives `xinput test-xi2 --root` a startup grace period and checks that the monitor is still alive before emitting `ready`. If it exits during startup, the helper emits `permission-denied`, preserves a nonzero result, and never reports readiness.

I added the reported stub case as a regression test. Validation:

- `node --test linux-features/appshots/test.js` — 17 passed
- `node --test linux-features/*/test.js` — 481 passed
- current-DMG regeneration — all four Appshots descriptors applied
- rebuilt and installed pacman package; packaged helper hash matches the source

ilysenko (approved):
Reviewed current head 52cadf51. The previous fail-closed startup blocker is fixed and regression-tested; the optional Appshots descriptor now targets only the current DMG bundle shape, and the single XInput2 monitor lifecycle looks correct. CI is green.

PR #806, opened 2026-07-11
fix: refresh pet overlay after selection

Gabriel-Kahen:
## Summary

- refresh the open pet overlay when `selected-avatar-id` changes
- reload only the pet overlay, preserving the main Codex windows and overlay placement
- fail closed with a clear warning if the upstream settings handler changes

## Why

Pet selections were persisted correctly, but the existing overlay renderer kept its initial state. The selected pet therefore appeared only after restarting Codex. This hooks the existing setting update and refreshes the overlay immediately.

## Validation

- `node --test linux-features/pet-overlay/test.js` (36 passed)
- `node --test scripts/patch-linux-window-ui.test.js` (346 passed)
- `node --test linux-features/*/test.js` (481 passed)
- side-by-side app build and packaged-bundle marker verification


ilysenko:
Thanks for the update. I found a blocker before this is marked ready:

- `webContents.reload()` replaces the renderer document, so the transparent CSS inserted with `insertCSS()` and the DOM styles applied with `executeJavaScript()` need to be installed again. The current `__codexPetOverlayTransparentRendererInstalled` flag lives on the `BrowserWindow` and remains set across the reload, causing `codexPetOverlayInstallTransparentRenderer()` to skip reinjection. The selected pet may update while the overlay becomes opaque.

Please reapply the transparent renderer setup after the new document loads (or reset the guard safely before reload) and add a regression test that models stylesheet/DOM state loss across navigation while preserving the window position.

ilysenko (approved):
Reviewed current head 314f1e51. The reload path now reapplies transparent renderer styles on did-finish-load, the regression is covered, the optional feature remains fail-soft/default-disabled, and all required checks pass.

PR #807, opened 2026-07-11
test(updater): resolve fixture tools from PATH

Vantrongs:
## Summary

- resolve fixture executables from regular executable files in absolute
  directories from the active `PATH`;
- generate fake updater build/install scripts with the resolved absolute Bash
  shebang;
- stop assuming fixture tools such as `cat`, `mkdir`, and `chmod` live in
  `/bin` or `/usr/bin`;
- serialize the process-global `PATH` read with the updater test suite's
  existing environment lock while running async build work through a local
  Tokio runtime.

## Why

The updater implementation does not require an FHS filesystem layout, but two
test helpers did. On NixOS, the fake update bundle failed to spawn because it
used `#!/bin/bash`, and standalone CLI tests failed before their assertions
because their tool fixture searched only `/bin` and `/usr/bin`.

This changes test fixtures only. Production updater lookup, bundle contents,
and command execution are unchanged.

## Validation

- `cargo test -p codex-update-manager builder::tests::builds_update_with_fake_bundle -- --exact`
- `cargo test -p codex-update-manager codex_cli::tests::standalone_cli_symlink_updates_with_standalone_installer -- --exact`
- full `cargo test -p codex-update-manager` reaches and passes every affected
  builder/CLI case on NixOS; its only remaining failure is the separate
  Desktop-settings isolation case
- `git diff --check`

## Scope

Only test modules in `updater/src/builder.rs` and `updater/src/codex_cli.rs`
change. PR #802 also touches `builder.rs`, but its global-dictation bundle hunks
do not overlap these fixture helpers.

## Checklist

- [x] The change is test-only and limited to updater fixtures.
- [x] Focused tests, formatting, and clippy pass on current `origin/main`.

### DIFF
--- a/updater/src/builder.rs
+++ b/updater/src/builder.rs
@@ -525,6 +525,7 @@ mod tests {
     use super::*;
     use crate::config::RuntimePaths;
     use anyhow::Result;
+    use std::os::unix::fs::PermissionsExt;
     use tempfile::tempdir;
 
     enum FakePackageOutput {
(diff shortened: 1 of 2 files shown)


avifenesh (approved):
LGTM. The changes correctly fix NixOS compatibility while maintaining test determinism.

**Key points:**
- `host_tool()` searches PATH but filters to absolute directories and verifies executables exist, avoiding relative-path pitfall
- `host_bash_script()` resolves bash shebang from PATH instead of hardcoding `/bin/bash`
- Test still holds `env_lock()` for entire duration, serializing PATH reads across parallel test runs
- `builds_update_with_fake_bundle` correctly switches from `#[tokio::test]` to local runtime since it now needs the env lock before any async work
- `link_test_system_tool` in codex_cli.rs reuses existing `is_executable()` helper instead of raw `.is_file()` check
- CI runs on standard Ubuntu where /bin, /usr/bin exist — this change is NixOS-specific and doesn't break existing CI

**No hermeticity regression:** tests still control PATH via `std::env::set_var("PATH", ...)` after acquiring the lock; the only difference is fixture generation now respects the host's actual tool locations instead of assuming FHS paths exist.

PR #808, opened 2026-07-11
fix(pacman): pin PKGEXT so Debian-hosted makepkg output is collected

avifenesh:
## Summary

- Pin `PKGEXT=.pkg.tar.zst` in makepkg environment to ensure Debian/Ubuntu makepkg produces packages the collector can find

## Root cause

On Debian/Ubuntu hosts, makepkg (from the `pacman-package-manager` package) defaults `PKGEXT` to `.pkg.tar.gz`, but `build-pacman.sh`'s collector at line 167 only searches for `.pkg.tar.zst` and `.pkg.tar.xz`. This causes `make pacman` to build successfully but then fail with "makepkg did not produce a package".

The latest-symlink at line 172 also hardcodes `.pkg.tar.zst`.

## Validation

Reproduced on Ubuntu 26.04 with `pacman-package-manager`'s makepkg.

Commands run:
```bash
bash -n scripts/build-pacman.sh  # syntax valid
grep -A2 'pkg.tar.zst\|pkg.tar.xz' scripts/build-pacman.sh  # collector confirmed
```

Smoke test `tests/scripts_smoke.sh` expects `.pkg.tar.zst` output, which this fix ensures.

This exact fix is proven in the sibling `chatgpt-desktop-linux` fork.

### DIFF
--- a/scripts/build-pacman.sh
+++ b/scripts/build-pacman.sh
@@ -110,7 +110,8 @@ main() {
 	trap "rm -rf '$build_root'" EXIT
 
 	local staging_root="$build_root/staging"
-	local -a makepkg_env=("PKGDEST=$DIST_DIR")
+	# Pin PKGEXT so Debian/Ubuntu makepkg (defaults to .pkg.tar.gz) produces .zst for the collector
+	local -a makepkg_env=("PKGDEST=$DIST_DIR" "PKGEXT=.pkg.tar.zst")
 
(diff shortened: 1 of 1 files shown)


ilysenko (approved):
This is a focused pacman-only fix: it pins the output format already required by the collector and latest symlink. All GitHub checks pass, and the full local script smoke suite passes.

PR #809, opened 2026-07-11
test(patches): reject obsolete terminal sanitizer shape

avifenesh:
## Summary

Keep the Linux terminal PATH patch targeted at the current upstream DMG sanitizer shape (`.t(...)`) and add a regression test proving that the obsolete 26.623 `.$r(...)` shape is not accepted as a fallback.

This preserves the useful fixture from the original proposal while following the repository current-DMG-only policy and keeping ordinary matcher drift fail-soft.

## Validation

- Verified the current `Codex.dmg` 26.707.41301 emits `r.t(i)` in `buildTerminalEnv`
- `node --test scripts/patch-linux-window-ui.test.js` (359 tests passed)
- `git diff --check`


avifenesh:

ilysenko (approved):
Verified against the current upstream DMG. The final diff keeps the production matcher current-only, adds focused regression coverage, and all required checks pass.

PR #811, opened 2026-07-11
test(patches): cover wrapped require alias resolution

avifenesh:
## Summary

Adds focused unit coverage for the current module-alias resolver behavior used by Linux patches.

## Scope

The original proposal generalized wrapped `require()` matching to every wrapper identifier. After review, that behavior was intentionally narrowed back to the current supported contract: only the repository-owned `codexLinuxPatchExternalOpen(require("electron"))` shape is treated as a wrapped Electron alias. Arbitrary wrappers and wrapped non-Electron modules remain rejected so patch discovery cannot silently accept transformed values as raw module aliases.

## Validation

- `node --test scripts/patches/lib/minified-js.test.js scripts/patch-linux-window-ui.test.js` (368 tests passed)
- `git diff --check`

The branch is rebased onto the current `main`.

### DIFF
--- a/scripts/patches/lib/minified-js.test.js
+++ b/scripts/patches/lib/minified-js.test.js
@@ -1,0 +1,54 @@
+const assert = require("node:assert/strict");
+const test = require("node:test");
+
+const { requireName, inferModuleAlias } = require("./minified-js.js");
+
+test("requireName finds direct require assignment", () => {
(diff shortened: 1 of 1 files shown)


avifenesh:

ilysenko (approved):
Reviewed the current rebased head. The broad wrapper generalization was removed; the remaining focused tests cover the supported current-DMG contract and all required checks pass.

PR #812, opened 2026-07-11
fix(ui-tweaks): match both sidebar chunk orderings and class variants

avifenesh:
## Summary

The `sidebar-project-name` patch now handles both Vite chunk segment orderings and both CSS class variants used across releases:
- Chunk patterns accept both `projects-index-page~remote-conversation-page` and `remote-conversation-page~projects-index-page` (Vite reorders merged-chunk segments)
- Content markers recognize both `min-w-0 truncate pr-1` (26.623) and `text-fade-truncate pr-1` (26.707)
- CSS selector targets both class variants under `.group/folder-row`

## Root cause

The original pattern matched only one hardcoded chunk segment ordering and one class name, breaking when:
1. Vite's deterministic ordering changed between builds
2. The app renamed classes between releases (26.623 → 26.707)

## Validation

- All existing tests pass
- New test fixture added for 26.707 class variant
- Pattern now explicitly tests both orderings
- Verified against real 26.707 chunk file (`app-initial~app-main~projects-index-page~remote-conversation-page-THP8fcuf.js`)
- Idempotence confirmed: patch applies once and only once

### DIFF
--- a/linux-features/ui-tweaks/patches/sidebar-project-name.js
+++ b/linux-features/ui-tweaks/patches/sidebar-project-name.js
@@ -2,15 +2,21 @@
 
 const DEFAULT_PROJECT_NAME_STYLE = "font-weight: 700 !important; padding-top: 0.25rem;";
 const PROJECTS_SIDEBAR_ASSET_PATTERN =
-  /^app-initial~app-main~projects-index-page~remote-conversation-page-[^.]+\.js$/;
-const PROJECT_NAME_SELECTOR = ".group\\/folder-row .text-fade-truncate.pr-1";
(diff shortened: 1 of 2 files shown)


ilysenko:

- This repository supports only the current upstream DMG, but the patch deliberately retains the legacy 26.623 `min-w-0 truncate pr-1` markup alongside the current 26.707 `text-fade-truncate pr-1` shape. Please remove the older class fallback and keep the selector, markers, and tests focused on the current 26.707 bundle only.

avifenesh:
Closing — with the current-DMG-only policy applied, everything this PR adds is 26.623 back-compat (legacy chunk ordering + the old `min-w-0 truncate pr-1` class). Removing those leaves no meaningful delta over main, which already targets the current 26.707 shape. Thanks for the review.

PR #813, opened 2026-07-11
feat(computer-use): prefer system install before vendored build

avifenesh:
## Summary

Add an explicit opt-in path for staging the published/system-installed `computer-use-linux` backend instead of building the vendored backend.

The vendored source remains the unchanged default and the existing prebuilt source overrides keep highest priority. External discovery and `cargo install` are used only when `CODEX_LINUX_COMPUTER_USE_SYSTEM_INSTALL=1` is set.

## Resolution order

1. Explicit `CODEX_LINUX_COMPUTER_USE_BACKEND_SOURCE` and `CODEX_LINUX_COMPUTER_USE_COSMIC_SOURCE` overrides
2. Vendored build by default
3. With `CODEX_LINUX_COMPUTER_USE_SYSTEM_INSTALL=1` only:
   - existing `computer-use-linux` and `computer-use-linux-cosmic` binaries
   - `cargo install --locked computer-use-linux`
   - vendored build fallback if the external path is unavailable

## Compatibility fix

The published backend resolves its COSMIC helper through a sibling named `computer-use-linux-cosmic`. Staging now preserves that sibling name for the external backend while retaining the repository's `codex-computer-use-*` staged entrypoints. Vendored and existing prebuilt staging do not receive the external alias.

## Validation

- `git diff --check`
- Regression coverage verifies that the opt-in staged backend can find and execute its COSMIC helper, while the default/prebuilt path remains unchanged

### DIFF
--- a/scripts/lib/bundled-plugins.sh
+++ b/scripts/lib/bundled-plugins.sh
@@ -239,12 +239,37 @@ find_cargo_for_linux_computer_use() {
     return 1
 }
 
+find_system_computer_use_binary() {
+    local name="$1"
+    local candidate
+
+    for candidate in \
+        "$HOME/.cargo/bin/$name" \
+        "$HOME/.local/bin/$name"; do
+        if [ -x "$candidate" ]; then
+            printf '%s\n' "$candidate"
+            return 0
+        fi
+    done
+
+    candidate="$(command -v "$name" 2>/dev/null || true)"
+    if [ -n "$candidate" ] && [ -x "$candidate" ]; then
+        printf '%s\n' "$candidate"
+        return 0
+    fi
+
+    return 1
+}
+
 build_linux_computer_use_backend() {
     local crate_dir="$SCRIPT_DIR/computer-use-linux"
     local backend_binary="$SCRIPT_DIR/target/release/codex-computer-use-linux"
     local cosmic_helper_binary="$SCRIPT_DIR/target/release/codex-computer-use-cosmic"
     local cargo_cmd=""
+    local system_backend=""
+    local system_cosmic=""
 
+    # Step 1: Environment override
     if [ -n "${CODEX_LINUX_COMPUTER_USE_BACKEND_SOURCE:-}" ] || [ -n "${CODEX_LINUX_COMPUTER_USE_COSMIC_SOURCE:-}" ]; then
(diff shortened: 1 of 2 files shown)


ilysenko:
Thanks! I think the default should be the other way around: use the bundled Computer Use build, and allow an external installation only through an explicit environment variable. We can only take responsibility for the code shipped in this repository, so automatically recommending or installing an external version would not feel right.

avifenesh:
Flipped as requested: the vendored build is the default again, unchanged from main's behavior. System-binary discovery and the crates.io install now run only with explicit `CODEX_LINUX_COMPUTER_USE_SYSTEM_INSTALL=1` (removed the `PREFER_VENDORED` inversion). Existing `CODEX_LINUX_COMPUTER_USE_BACKEND_SOURCE`/`COSMIC_SOURCE` prebuilt overrides untouched. Verified with an isolated-HOME harness: default path never touches system binaries; opt-in resolves them; `bash -n` clean.

avifenesh:
@ilysenko Since it's MCP I'm maintaining (the goal is the specific one, not any generic) i found it much more convenient to have one source of truth that gets updated immediately when I add improvements and changes and is trustworthy, at least from my side (I trust myself :) ). But ill follow your direction on this if you think differently.

ilysenko:
Thanks, and just to clarify: this is not about distrusting you or your work. The concern is purely about the supply-chain risk.

Today, this repository is our single source of truth: everything we ship is visible and reviewed here. Recommending or installing an external package by default adds another trust boundary. Even a trusted maintainer can have their account or publishing credentials compromised, allowing spyware, a cryptominer, or other malicious code to reach users without any change being reviewed here.

Keeping the external version as an explicit opt-in preserves the convenience without increasing the default trust surface.

avifenesh:
@ilysenko sure, i understand

ilysenko:
Could you clarify why the system/crates.io binaries are discovered as `computer-use-linux` and `computer-use-linux-cosmic`, but staging renames them to `codex-computer-use-linux` and `codex-computer-use-cosmic`?

The published `computer-use-linux` backend resolves its COSMIC helper by looking for a sibling named `computer-use-linux-cosmic`, or via `COMPUTER_USE_LINUX_COSMIC_HELPER`. After staging, that sibling name appears to be missing. Is there another mechanism that sets the helper path, or should the upstream helper name be preserved in the staged payload?

ilysenko (approved):
External Computer Use resolution is explicitly opt-in, the bundled path remains the default, and the COSMIC helper compatibility is covered by regression tests.

PR #814, opened 2026-07-11
fix(copilot-reasoning-effort): stop pinning webview chunk hashes

avifenesh:
## Summary

Replaced hardcoded content-hash segments in the three webview asset patterns with wildcards so they survive upstream chunk structure changes.

## Root cause

The chunk patterns pinned 8-character Vite content-hash segments (e.g., `k0ede4gb`, `iufn7mg3`):

```javascript
pattern: /^app-initial~app-main~onboarding-page~hotkey-window-thread-page~quick-chat-window-page~chatg~k0ede4gb-[^.]+\.js$/
```

These segments reshuffle with every upstream release when Vite rebuilds, causing the patterns to miss their target chunks.

## Changes

Broadened all three patterns to:

```javascript
pattern: /^app-initial~app-main~.*\.js$/
```

This is safe because:
- Each patch function already includes content-scan fallback that checks for marker strings (e.g., `copilot-default-model`) before applying
- The broader pattern matches the stable prefix (`app-initial~app-main~`) while allowing route order and hash segments to vary
- Tests verify that unrelated bundles don't get patched

## Validation

- ✅ Test suite passes (all 6 tests green)
- ✅ Patterns match old fixture names: `app-initial~app-main~onboarding-page~hotkey-window-thread-page~quick-chat-window-page~chatg~k0ede4gb-current.js`
- ✅ Patterns match new 26.707 chunks: `app-initial~app-main~new-thread-panel-page~onboarding-page~projects-index-page~appgen-libra~ggy53w99-CqMu8hJo.js`
- ✅ Double-apply remains idempotent (already-patched chunks detected via marker strings)

### DIFF
--- a/linux-features/copilot-reasoning-effort/patch.js
+++ b/linux-features/copilot-reasoning-effort/patch.js
@@ -154,7 +154,8 @@ module.exports = {
       id: "settings",
       name: "copilot-reasoning-effort-settings",
       phase: "webview-asset",
-      pattern: /^app-initial~app-main~onboarding-page~hotkey-window-thread-page~quick-chat-window-page~chatg~k0ede4gb-[^.]+\.js$/,
+      // Match app-initial chunks; content-scan fallback ensures correct target
+      pattern: /^app-initial~app-main~.*\.js$/,
       missingDescription: "model settings bundle",
       skipDescription: "Copilot reasoning effort settings patch",
       apply: applyCopilotReasoningEffortSettingsPatch,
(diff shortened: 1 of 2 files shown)


ilysenko:

- Please replace the legacy `k0ede4gb` / `iufn7mg3` chunk fixtures with the current 26.707 chunk shape and verify the broadened descriptors against that shape. This repository supports only the current upstream DMG, so old chunk layouts should not remain as compatibility fixtures.

avifenesh:
Done — replaced the `k0ede4gb`/`iufn7mg3` fixture names with the current 26.707 chunk shapes (`…appgen-libra~ggy53w99…` for settings/model-list, `app-initial~app-main~page-…` for the composer UI gate, matching where those needles actually live in the served bundle). Verified all three descriptors against the real current chunks: each applies and is idempotent. Feature suite 6/6.

ilysenko (approved):
Reviewed the current head. The legacy fixtures are removed, the descriptors target the current upstream shape, and CI is green.

PR #816, opened 2026-07-11
test: remove FHS assumptions from shell smoke fixtures

Vantrongs:
## Summary

- resolve Bash and the no-op executable from absolute directories in the active test `PATH`;
- use those host paths for executable smoke fixtures instead of `/bin/bash`, `/usr/bin/bash`, and `/bin/true`;
- preserve deliberately restricted fixture `PATH` values while keeping their interpreter available;
- normalize the captured host-tool path so empty, relative, and duplicate entries cannot influence the tests.

## Why

`tests/scripts_smoke.sh` failed on non-FHS hosts before reaching its assertions because several generated fixtures and direct invocations assumed executables existed below `/bin` or `/usr/bin`. The production launcher and package behavior were not involved.

This change is test-only. FHS paths that are asserted as package payload data remain unchanged, while executable fixtures now use tools resolved from the host environment.

## Validation

- `git diff origin/main...HEAD --check`
- hostile-PATH normalization check with empty, relative, and duplicate entries
- `nix develop -c bash tests/scripts_smoke.sh`

All script smoke tests passed on the rebased branch.

## Checklist

- [x] The change is limited to test fixtures.
- [x] Sister occurrences of `/bin/bash`, `/usr/bin/bash`, and `/bin/true` were audited.
- [x] The final diff received an independent review and all findings were addressed.

### DIFF
--- a/tests/scripts_smoke.sh
+++ b/tests/scripts_smoke.sh
@@ -3,6 +3,41 @@ set -Eeuo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 REPO_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
+
+normalize_host_tool_path() {
+    local entry
+    local normalized=""
+    local -a entries=()
+
+    IFS=: read -r -a entries <<< "${PATH-}"
+    for entry in "${entries[@]}"; do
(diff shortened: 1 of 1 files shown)


ilysenko (approved):
Focused test-only portability fix. The fixtures now resolve host tools without changing production paths, and all GitHub checks pass.

PR #817, opened 2026-07-11
fix(patches): match both published 26.707 webview chunk layouts

avifenesh:
## Summary
- Un-breaks the `Build App Against Upstream DMG` job, red on `main` since the 2026-07-11 05:21 UTC run.
- Root cause: OpenAI briefly published a rebuilt `ChatGPT.dmg`, then rolled it back — the DMG served at `codex-app-prod/ChatGPT.dmg` today is byte-identical to the earlier 26.707.41301 build again. The July 10 pattern updates pinned the rebuilt build's chunk names (`cha~b76hmflu`, `chatg~k0ede4gb`), so when the CI header cache refreshed to the rolled-back build, these patches stopped matching:
  - `linux-app-server-conversation-hydration`, `linux-completed-item-recovery` (required by the `upstream-build` profile)
  - `feature:remote-mobile-control` load-gate, conversation-hydration, status-read-guard, status-wait (required by the remote-mobile inspect step)
  - `automation-update-eager-tool`, `linux-safe-monospace-font-stack`, `linux-tooltip-window-controls-collision`, `linux-computer-use-install-flow` (optional)
- Fix: widen the asset patterns to accept **both** published layouts instead of re-pinning — keep the rebuilt-DMG segment shape with its volatile hash wildcarded, add the rolled-back build's chunk shape. No impl changes; every impl already applies cleanly to the rolled-back build's chunks (verified individually).

## Still skipping (out of scope)
`linux-settings-search-visibility` remains an optional skip on the rolled-back build: its import specifiers pin the rebuilt DMG's export letters (`E$`, `T2`, `j7`, `qk`). Resolving those aliases semantically needs a deeper change — flagging to #784's author.

## Validation
- Downloaded the currently served DMG (sha256 matches CI's cache headers), extracted `app.asar`, ran the full patcher with `remote-mobile-control` enabled and `CODEX_LINUX_ENABLE_COMPUTER_USE_UI=1`:
  - `node scripts/ci/validate-patch-report.js --profile upstream-build --require-enabled-feature remote-mobile-control --require-applied ...` (the exact CI invocation) **passes**
  - second patch run: all 91 relevant patches report `already-applied` (idempotent)
- `node --test 'scripts/**/*.test.js'`: 394/394 pass (two tests updated — they asserted the rebuilt-DMG-only worldview, e.g. `assert.match(pattern, /b76hmflu/)` and rejecting the rolled-back chunk name as legacy)
- All linux-features suites pass except `appshots` XInput2 stream test, which fails identically on a clean checkout (environment-dependent, pre-existing)

### DIFF
--- a/linux-features/remote-mobile-control/patch.js
+++ b/linux-features/remote-mobile-control/patch.js
@@ -55,7 +55,7 @@ const REMOTE_MOBILE_APP_SERVER_REMOTE_CONTROL_MARKER = "codexLinuxRemoteMobileAp
 const REMOTE_MOBILE_APP_SERVER_ARGS_NEEDLE =
   "[`-c`,`features.code_mode_host=true`,`app-server`,`--analytics-default-enabled`]";
 const REMOTE_MOBILE_CONVERSATION_ASSET_PATTERN =
-  /^app-initial~app-main~pull-request-code-review~onboarding-page~hotkey-window-thread-page~cha~b76hmflu-[^.]+\.js$/u;
+  /^app-initial~app-main~(?:pull-request-code-review~onboarding-page~hotkey-window-thread-page~cha~[^.]+|onboarding-page~hotkey-window-thread-page~quick-chat-window-page~chatg~[^.]+)\.js$/u;
 const REMOTE_CONTROL_SELECTED_TAB_NEEDLE =
   "function rr({selectedConnectionsTab:e,showControlThisMacTab:t,showRemoteControlConnectionsSection:n,showTabbedSshPage:r}){return n?e===`control-this-mac`&&!t||e===`ssh`&&!r?`access-other-devices`:e:`ssh`}";
 const REMOTE_CONTROL_SELECTED_TAB_REPLACEMENT =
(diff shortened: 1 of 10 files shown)


ilysenko:

- This repository supports only the latest upstream DMG. Please remove the alternate rebuilt-DMG layout and keep the patterns and tests focused only on the layout currently served upstream.

avifenesh:
Done — dropped the rebuilt-DMG alternates entirely; patterns and test fixtures now target only the currently served layout. Re-validated: the exact CI invocation (`--profile upstream-build` + the remote-mobile requirements) passes against a fresh extract of the current DMG, second run fully `already-applied`, `node --test 'scripts/**/*.test.js'` 394/394.

Yo-DDV:
Thanks for the update. I reproduced the remaining `Rust and Smoke Tests` failure locally.

The failing path seems to be just the shell smoke fixture, not the patch logic itself. In `tests/scripts_smoke.sh`, `test_linux_computer_use_ui_opt_in_smoke` still writes the install-flow asset under the old `...pull-request-code-review...b76hmflu...` name, while this PR moves the Computer Use UI descriptor to `...quick-chat-window-page~chatg~...`.

I tested the minimal change locally by switching that fixture path to:
`app-initial~app-main~onboarding-page~hotkey-window-thread-page~quick-chat-window-page~chatg~gwqc41kz-test.js`

After that, `bash tests/scripts_smoke.sh` passes end to end on top of `2a5dcf1c`.

Nice catch, @ilysenko. I was finishing the exact same fix when your commit landed, and my local diff was byte-for-byte identical. Thanks for getting it in first.

ilysenko (approved):
Reviewed the current head. The descriptors now target only the current upstream DMG, the smoke fixture is aligned, and all required checks pass.

PR #818, opened 2026-07-11
Fix Chrome sidepanel startup with npm Codex CLI

### LINKED ISSUE #805: Chrome sidepanel app-server exits when codexCliPath points to isolated ESM launcher
## Summary

With the current upstream DMG and current `main`, the ChatGPT Chrome sidepanel can fail with:

> Unable to start ChatGPT  
> Codex app-server exited before becoming ready

The new Unix-socket WebSocket proxy is working, but the runtime manifest can point `codexCliPath` at an isolated copy of the npm ESM launcher that cannot execute from its new location.

## Environment

- Distro: CachyOS / Arch (`.pkg.tar.zst`)
- Kernel: `7.1.3-2-cachyos`
- Desktop: KDE Plasma; app launched through X11
- `codex-desktop`: `2026.07.11.015245-1`
- Upstream app: `26.707.41301`
- Wrapper: `0.9.0`, commit `60fbcf0e9288c570ecb4742b557c04f4e3c8bab3`
- ChatGPT Chrome Extension: `1.2.27203.26575`
- User CLI: `codex-cli 0.144.1`
- Node selected by the isolated launcher shebang: `v26.4.0`

The package was built from a clean checkout. `pacman -Qkk codex-desktop` reported `18270 total files, 0 altered files`, and all 48 `codex-chrome-extension-host` tests passed.

## Reproduction

1. Build and install the current DMG from current `main`.
2. Start ChatGPT Desktop and open the ChatGPT Chrome sidepanel.
3. The sidepanel calls `codexRuntime/ensure`.
4. The native host returns `Codex app-server exited before becoming ready`.

The generated runtime manifest contains:

```json
{
  "paths": {
    "codexCliPath": "$HOME/.codex/plugins/.plugin-appserver/codex"
  }
}
```

Running that path directly reproduces the failure:

```text
Warning: Failed to load the ES module: .../.plugin-appserver/codex
SyntaxError: Cannot use import statement outside a module
```

## Root cause

`$HOME/.codex/plugins/.plugin-appserver/codex` is a copied npm ESM entrypoint. It starts with imports such as:

```js
import { spawn } from "node:child_process";
```

However, the isolated `.plugin-appserver` directory has neither the original package metadata (`"type": "module"`) nor the npm package's `node_modules`. Therefore it cannot be used as a standalone executable.

Adding only a local `package.json` is not sufficient: after ESM parsing is enabled, the copied launcher still cannot resolve `@openai/codex-linux-x64` from the isolated directory.

`computer-use-linux/src/chrome_runtime.rs` trusts the manifest path and launches it directly with `app-server --listen unix://...`, so the process exits before creating the socket.

## Verified workaround

Replacing the isolated launcher with a small wrapper that delegates to the real maintained CLI while preserving the plugin-host feature flag fixes the problem:

```bash
#!/usr/bin/env bash
set -euo pipefail

readonly cli="$HOME/.npm-global/bin/codex"
exec env CODEX_CLI_PATH="$cli" \
  "$cli" -c features.code_mode_host=true "$@"
```

After this change, a full native-messaging integration probe produced:

```json
{
  "connected": true,
  "appVersion": "26.707.41301",
  "cliVersion": "26.707.41301",
  "websocket": "HTTP/1.1 101 Switching Protocols"
}
```

The real Chrome sidepanel then started successfully.

## Suggested fix

Ensure the Linux runtime manifest receives a real executable CLI path instead of the relocated npm JavaScript entrypoint. The `features.code_mode_host=true` override can be preserved separately when spawning the app-server.

Alternatively, make the Linux native host validate/preflight the manifest CLI and select a validated real CLI candidate when that path exits before creating the Unix socket. The fallback should keep the existing path ownership and permission checks.

It would also be useful to add an integration fixture where `codexCliPath` is an isolated npm ESM launcher, since the current fake app-server tests do not reproduce this packaging/runtime boundary.

Yo-DDV:
## Summary

Fixes #805.

On Linux, keep the Chrome sidepanel app-server on the resolved Codex CLI instead of copying an npm ESM launcher into `.plugin-appserver`. The copy loses its package metadata and dependencies, so it cannot start outside the npm package tree.

The native host now retains the validated canonical CLI path and passes `features.code_mode_host=true` explicitly when it starts `app-server`. Chrome plugin isolation remains unchanged on other platforms.

## Validation

- Patched the current extracted DMG bundle twice and confirmed the second pass was byte-identical

The regression fixture executes an npm-style ESM launcher from a path containing spaces, checks that its package-relative import still resolves, and verifies that no isolated Linux copy is created. Separate tests cover non-Linux isolation, canonical symlink resolution, and the app-server feature arguments.

## Checklist

- [x] This targets the current upstream bundle shape and does not add a legacy fallback.
- [x] I added relevant regression coverage and ran the validation listed above.
- [x] I reviewed the final diff twice, addressed the findings, and reran the affected gates.


ilysenko (approved):
Reviewed the current head after synchronization with main. The Linux CLI path remains canonical and executable, non-Linux isolation is preserved, and all required checks pass.

PR #821, opened 2026-07-11
test(updater): isolate Desktop settings fixture

Vantrongs:
## Summary

- isolate the daemon reconciliation test from the developer's real Codex Desktop settings file;
- restore both process environment variables with the shared panic-safe test guard;
- keep the existing global environment lock for the full test lifetime.

## Why

`daemon_reconcile_reloads_waiting_state_written_by_another_process` configured auto-install in its fixture, but still allowed `CODEX_LINUX_SETTINGS_FILE` to resolve to the user's real Desktop settings. A local `autoInstallUpdates` value could override the fixture and make the test miss the expected no-polkit-agent transition.

The test now points the settings lookup at a nonexistent file inside its temporary directory. This preserves the intended default behavior and prevents workstation state from changing the assertion. Production updater behavior is unchanged.

## Validation

- focused regression test: 1 passed
- `cargo test -p codex-update-manager`: 228 passed
- `git diff --check`

The final one-file diff also received an independent review with no findings.

### DIFF
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -2622,11 +2622,18 @@ mod tests {
     #[test]
     fn daemon_reconcile_reloads_waiting_state_written_by_another_process() -> Result<()> {
         let _env_guard = crate::test_util::env_lock();
+        let _restore_env = crate::test_util::EnvRestoreGuard::capture(&[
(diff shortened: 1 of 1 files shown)


ilysenko (approved):
Reviewed the current head. The updater test is properly isolated from user settings, environment restoration is panic-safe, and all required checks pass.

PR #823, opened 2026-07-11
test(global-dictation): use host executables in fixtures

Vantrongs:
## Summary

- resolve Bash from executable regular files in the host `PATH` for generated global-dictation fixtures
- keep fixture command overrides ahead of the real host tool path instead of replacing it with `/usr/bin:/bin`
- use `process.execPath` as the portable executable payload instead of `/bin/true`

## Why

The tests added with global dictation fail before exercising the production scripts on non-FHS systems. On NixOS, six watcher cases cannot find Bash after the fixture replaces `PATH`, while both stage-hook cases assume `/bin/true` or an FHS shell path.

This is test-only: production scripts and feature behavior are unchanged.

## Validation

- baseline global-dictation file on NixOS: 13 passed, 8 failed
- first watcher fixture after the minimal fix: 1/1 passed
- `node --test linux-features/global-dictation/test.js`: 21/21 passed
- `bash scripts/ci/run-node-checks.sh`: 900/900 passed
- `node --check linux-features/global-dictation/test.js`

### DIFF
--- a/linux-features/global-dictation/test.js
+++ b/linux-features/global-dictation/test.js
@@ -21,6 +21,29 @@ const {
   descriptors,
 } = require("./patch.js");
 
+const hostPathEntries = (process.env.PATH ?? "")
+  .split(path.delimiter)
+  .filter((entry) => path.isAbsolute(entry));
+const hostPath = [...new Set(hostPathEntries)].join(path.delimiter);
+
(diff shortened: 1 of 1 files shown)


ilysenko (approved):
Reviewed the current head; the fixture changes are focused, policy-compliant, and all required checks pass.

PR #825, opened 2026-07-11
feat(appimage): support an optional bundled Codex CLI

### LINKED ISSUE #824: AppImage: launch fails invisibly when @openai/codex is not installed — consider bundling the CLI
**What happens**

With an AppImage built via `make build-app && make appimage`, launching on a host without `@openai/codex` installed exits before any window is created. The error only lands in `~/.cache/codex-desktop/launcher.log`, plus a `notify-send` toast that never appeared in my case (COSMIC/Wayland — no visible notification). Launched from a desktop entry, the app appears to simply not start.

**Why the existing flows don't cover this**

The README documents that the Codex CLI is required at runtime and that the first launch can install it — but that flow runs through `codex-update-manager`, which AppImage self-builds intentionally exclude ("no bundled updater"), and `run_gui_cli_prompt` requires an interactive terminal. So for the AppImage specifically there is no working install path and no visible error: the one packaging format users expect to be self-contained is the only one that can fail invisibly on a missing host dependency.

**Suggestion**

Let the AppImage build optionally bundle the CLI, e.g.:

```bash
make appimage CODEX_CLI_BUNDLE_SOURCE=/path/to/node_modules/@openai/codex
```

staging the package tree under `resources/codex-cli` with a small wrapper that runs it through the already-bundled managed Node runtime, and having the launcher prefer the bundled copy. This makes the AppImage fully self-contained (matching AppImage expectations) with no behavior change for native packages, which keep the update-manager install flow.

I have this working in a fork and would be happy to send a PR if you're open to it.

**Environment:** CachyOS (Arch), x86_64, COSMIC desktop (Wayland session), AppImage self-build from current main.

Yo-DDV:
Thanks for the detailed AppImage report in #824.

This adds an opt-in AppImage-only path for bundling the Codex CLI. `make appimage` can now take `CODEX_CLI_BUNDLE_SOURCE` pointing at an installed `node_modules/@openai/codex` package. The builder validates the matching Linux optional package, stages both package trees under `resources/codex-cli`, and lets AppRun use the bundled wrapper when `CODEX_CLI_PATH` is not already set.

I kept the build explicit. It does not fetch npm packages by itself, and native packages keep their current updater/install flow.

Fixes #824.

Checks run

- `bash -n scripts/build-appimage.sh packaging/appimage/AppRun packaging/appimage/codex-cli-wrapper.sh tests/scripts_smoke.sh`
- `git diff --check`
- `make check`
- `make test`
- real npm install dry-run for `@openai/codex@0.144.1`, then AppImage build dry-run with the captured AppDir

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ UPSTREAM_INTEL_PATCH_REPORT ?= $(REBUILD_REPORT_DIR)/patch-report.json
 UPSTREAM_INTEL_IMAGE ?= codex-desktop-linux-devcontainer:local
 PACKAGE_NAME := codex-desktop
 PACKAGE_WITH_UPDATER ?= 1
+CODEX_CLI_BUNDLE_SOURCE ?=
 MAX_BUILD_THREADS ?= 0
 MAX_BUILD_THREADS_VALUE := $(strip $(MAX_BUILD_THREADS))
 MAX_BUILD_THREADS_ENABLED := $(filter-out 0,$(MAX_BUILD_THREADS_VALUE))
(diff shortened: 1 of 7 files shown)


ilysenko (approved):
Reviewed the current head; the AppImage-only implementation is well-scoped, policy-compliant, and all required checks pass.

PR #828, opened 2026-07-11
test(chrome): resolve runtime fixtures from PATH

Vantrongs:
## Summary

- resolve Python and `true` fixture executables from absolute host `PATH` entries
- use `current_exe()` for required executable manifest fields instead of `/bin/true` or a `/proc/self/exe` spelling
- keep the change inside Chrome runtime test modules; production behavior is unchanged

## Why

The Chrome runtime suite currently assumes FHS interpreter and executable paths. On NixOS, the fake app-server cannot start and required manifest paths are missing, which causes ten Chrome host tests to fail before their runtime assertions execute.

This also makes `runtime_ensure_keeps_child_alive_after_request_returns` runnable without the skip used while reviewing #788. The changes are hunk-disjoint from #788's pending-correlation resource accounting.

## Validation

- reproduced the current-main failures with `cargo test -p codex-computer-use-linux`
- repaired one app-server lifecycle case first and verified it independently

### DIFF
--- a/computer-use-linux/src/bin/codex-chrome-extension-host.rs
+++ b/computer-use-linux/src/bin/codex-chrome-extension-host.rs
@@ -1090,9 +1090,13 @@ mod tests {
         fs::set_permissions(&resources, fs::Permissions::from_mode(0o700)).unwrap();
 
         let fake_cli = root.join("fake-app-server");
+        let python = test_executable("python3");
         fs::write(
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head. The test-only portability fix is focused, safe, and all required checks pass.

PR #830, opened 2026-07-11
fix(nix): include libnotify for desktop notifications

### LINKED ISSUE #827: Desktop notifications are unavailable on Linux: `Notification API not supported on this host`
# Desktop notifications are unavailable on Linux: `Notification API not supported on this host`

## Summary

Codex Desktop detects both permission requests and completed turns, but it does
not deliver a desktop notification on Linux.

The notification pipeline reaches the app's `desktop-notifications` service and
then stops with `Notification API not supported on this host`, before any
notification is sent to the Linux desktop session.

This is reproducible while the Codex window is unfocused.

## Environment

- Distribution: NixOS 26.05
- Installation: Nix flake package through the Home Manager module
- `codex-desktop-linux`: `0.9.0`, commit `7a0b79e9cf5ad341ff51b2447ff29fc7a27984fe`
- Upstream Codex Desktop: `26.707.41301`
- Electron: `42.1.0`
- Desktop: Hyprland `0.55.4`
- Session: Wayland
- Notification daemon: Quickshell owns `org.freedesktop.Notifications`
- Do Not Disturb: disabled

## Steps to reproduce

1. Launch Codex Desktop.
2. In Settings -> Notifications, enable permission notifications.
3. Move focus to another application.
4. Start a Codex task that requests a command approval.
5. Wait for the approval request.

The same failure can also be observed for a completed turn.

## Actual behavior

No desktop notification is shown.

The following lines were collected during reproduction from the Linux wrapper
log at `~/.cache/codex-desktop/launcher.log`:

```text
[desktop-notifications] show approval ... kind=commandExecution
[desktop-notifications] forward show kind=permission ...
[desktop-notifications] Notification API not supported on this host
```

For completed turns, the same path ends with the same error:

```text
[desktop-notifications] show turn-complete ...
[desktop-notifications] forward show kind=turn-complete ...
[desktop-notifications] Notification API not supported on this host
```

The failure occurs before the Linux notification daemon receives anything.

## Expected behavior

When Codex needs an approval or finishes a turn while unfocused, it should send
a native Linux desktop notification through the active session notification
backend.

## Scope

Confirmed for:

- permission / approval notifications;
- turn-complete notifications.

## Possible implementation direction

The Linux wrapper could provide a native notification adapter when the Web
Notification API is unavailable, preserving the existing application payload
and click behavior while dispatching through Electron's native notification
path or `org.freedesktop.Notifications`.

If useful, I can test a focused fix on NixOS/Hyprland or prepare a PR.

ilysenko:
## Summary

- add `libnotify` to the Nix Electron runtime library set
- make Electron's dynamically loaded Linux notification backend available through the existing RPATH and `LD_LIBRARY_PATH` wiring

## Root cause

Electron loads `libnotify.so.*` with `dlopen()` before creating its Linux notification presenter. Because `libnotify` was absent from `electronLibs`, NixOS could not resolve the library and `Notification.isSupported()` returned `false` before the desktop notification daemon was contacted.

## Impact

Native permission and turn-complete notifications can reach the active `org.freedesktop.Notifications` service on NixOS. Other package formats are unchanged.

Fixes #827

## Checks

- `git diff --check`
- Nix evaluation/build not run locally because the `nix` CLI is unavailable in the agent environment; the draft PR's `Nix Package Builds` check will validate the package

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -294,6 +294,7 @@
           libxi
           libxtst
           libxscrnsaver
+          libnotify
           libglvnd
           systemd
           wayland


Yo-DDV (commented):
Thanks for picking this up so quickly.

I reproduced the missing runtime dependency before testing this head: the current Nix package closure and generated launcher both lacked `libnotify`. On `6b42d941`, I ran `nix flake check --no-update-lock-file --max-jobs 2`, `bash tests/scripts_smoke.sh`, and a full `nix build --no-update-lock-file --no-link .#codex-desktop`; all passed.

The resulting package closure contains `libnotify-0.8.8`, and both generated launcher surfaces include its library path. That covers the exact `dlopen()` failure described in #827. The code change looks right to me.

The only small follow-up I would suggest is an Unreleased changelog entry, since this restores a user-visible Nix feature.

PR #836, opened 2026-07-11
Fix stale streaming state after completed thread resume

### LINKED ISSUE #835: Renderer freezes when sending in old completed thread: resume marks completed turn as streaming
Codex Desktop Linux freezes when I try to send a new message in one existing chat thread. The freeze is reproducible across app restarts. I have to manually kill the Codex/Electron processes from the CLI.

This appears to be thread-specific. Other/new threads continue to work.

## Environment

- App release from local Sentry event: `codex@26.611.62324`
- Electron: `42.1.0`
- OS/kernel from event: CachyOS/Arch-family, kernel `7.1.2-3-cachyos`
- Desktop: KDE Plasma
- Problem thread id: `019f42a7-db06-79d3-9344-6ad485bfad2a`
- Problem thread title: `Review documents and next steps`

## Symptoms

1. Open the existing thread.
2. Try to send a new message.
3. Codex Desktop freezes.
4. Closing/reopening the app does not fix it.
5. Recovery requires manually killing Codex/Electron processes.

## Local diagnostic evidence

The thread history itself appears complete. The rollout file ends with:

```text
type: event_msg
payload.type: task_complete
turn_id: 019f4d87-31f7-78a2-9eb6-2f3f4fa1f653
```

However, when Desktop resumes the thread, the Electron/Sentry breadcrumbs show:
```text
maybe_resume_success conversationId=019f42a7-db06-79d3-9344-6ad485bfad2a latestTurnId=019f4d87-31f7-78a2-9eb6-2f3f4fa1f653 latestTurnStatus=completed markedStreaming=true turnCount=5
```

Immediately after that, the renderer repeatedly calls:

```text
method=thread/turns/list conversationId=019f42a7-db06-79d3-9344-6ad485bfad2a
```

The app state snapshot before the hang includes:
```text
thread_count_active=1
thread_count_streaming_owner=1
thread_count_streaming_without_active_runtime=1
turn_count_total_loaded=110
item_count_total_loaded=266
max_turns_in_single_thread=107
```
Then Electron records:
```text
Error: Renderer unresponsive
tags.errorType=renderer-unresponsive
```

The latest turn is completed, and the rollout ends with task_complete, but the resume path marks the thread as streaming anyway (`latestTurnStatus=completed markedStreaming=true`).  That seems to leave the UI with a streaming owner but no active runtime, after which the renderer gets into a repeated turn-list/render path and becomes unresponsive.

Expected behavior : A completed thread should resume as non-streaming, accept a new message, or show a recoverable error if the runtime state is stale.

Actual behavior: The renderer freezes before the new message is processed.

Suggested fix: On thread resume, if `latestTurnStatus=completed` and there is no active runtime/inflight turn, clear streaming ownership rather than marking the thread streaming. Also consider guarding thread/turns/list refresh loops when a completed legacy thread is resumed with stale streaming state.

Yo-DDV:
## Summary

- clear stale streaming ownership when a completed thread resumes without an active app-server runtime
- keep the existing ownership behavior for active runtimes and non-completed turns
- cover the resume states with a regression test and document the fix

## Why

The resume path currently marks every restored thread as streaming. For a completed thread with an idle runtime, that leaves an owner without an active turn and can trap the renderer in repeated history refreshes.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` (362 passed)
- `git diff --check`

Fixes #835

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- Resuming a completed thread no longer leaves it registered as streaming when
+  the app server reports no active runtime. This clears stale stream ownership
+  before the next message instead of leaving the renderer in a repeated thread
+  history refresh path.
 - Linux settings search no longer shows unavailable macOS Dock icon controls or
   Suggested prompts results that do not render in the generated Linux settings
   page.
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -5815,6 +5815,67 @@ test("discovers current app-server conversation core Linux webview patches", () 
   }
 });
 
+test("does not retain streaming ownership when a completed thread resumes without an active runtime", () => {
+  const source = [
+    "async function resume(e,t,L,ue,b){",
+    "e.updateConversationState(t,t=>{t.threadRuntimeStatus=L.thread.status});",
+    "let de=ue.at(-1)??null;e.markConversationStreaming(t),e.setConversationStreamRole(t,{role:`owner`}),b&&e.releaseResumeNotificationBuffer(t);",
+    "let fe=e.broadcastConversationSnapshot(t);",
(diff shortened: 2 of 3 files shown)


ilysenko (approved):
Reviewed the completed-thread resume recovery against the current upstream DMG path. The stale streaming ownership is cleared only for completed idle resumes, focused and broad tests pass, and all required checks are green.

PR #837, opened 2026-07-11
Fix persistent status panel on current DMG

PinguuSS:
## Summary

- retarget `persistent-status-panel` to the current unified app status bundle, `app-initial~app-main~page-*.js`
- keep the patch fail-soft when the current status bundle marker is absent, instead of letting unrelated matched assets look successful
- add a current-DMG descriptor test that verifies the patch actually changes the current bundle shape

## Reproduction

With only `persistent-status-panel` enabled, the old descriptor no longer reached the current unified app status component. The `/status` panel still opened, but its open state stayed component-local, so leaving the current task and returning remounted the composer without restoring the panel.

I also verified there is no open issue or PR covering this same `persistent-status-panel` / `/status` / composer status panel problem.

## Visible behavior validated

Current DMG runtime validation used a side-by-side app id and user data directory:

- app version: `26.707.41301`
- Sparkle build: `5103`
- DMG SHA256: `467a25381af2943f2c9b8794adae28b36400bd6da2753daf4889a5fd550c4d65`
- Electron: `42.1.0`
- base repo SHA tested from: `2d2bdf6dea79179e376958d94fb6f1b5ce03be53`
- environment: Ubuntu 25.10, KDE, Wayland session with X11 rendering override for the side-by-side test app

Runtime checks performed:

- patch report records `feature:persistent-status-panel:composer-status-state` as `applied`
- modified asset is `/content/webview/assets/app-initial~app-main~page-hSvsQcNf.js`
- opening `/status` stores `codex-linux-persistent-status-panel-open=1`
- switching between Codex tasks keeps the status panel visible
- Quick Chat overlay does not break the Codex status panel state
- `Sites` and `Pull requests` remove the panel from non-task surfaces, while returning to `New task` restores it
- normal app restart with the key set restores the panel
- explicit `Close` removes the key and, after graceful app close/restart, the panel stays closed
- side-by-side identities stayed isolated during runtime validation

## Validation

- `node --check /tmp/codex-persistent-status-app-fixed/content/webview/assets/app-initial~app-main~page-hSvsQcNf.js`
- `node scripts/ci/validate-patch-report.js dist-next/rebuild/patch-report.json`
- isolated build with the current DMG and only `persistent-status-panel` enabled
- runtime CDP validation with screenshots captured for baseline, open, task switch, Quick Chat, non-task routes, restart, close, and closed restart states
- `git diff --check`

## Limits

This is intentionally scoped to the current upstream DMG shape and only the opt-in `persistent-status-panel` feature. It does not touch generated `codex-app/` output or any unrelated feature.

### DIFF
--- a/linux-features/persistent-status-panel/patch.js
+++ b/linux-features/persistent-status-panel/patch.js
@@ -37,6 +37,7 @@ function applyPersistentStatusPanelPatch(source) {
   }
 
   if (!source.includes("composer.statusSlashCommand.description")) {
+    console.warn("WARN: Could not find Codex status panel bundle marker - skipping persistent status panel patch");
     return source;
   }
 
@@ -76,7 +77,7 @@ const patches = [
     phase: "webview-asset",
     order: 20_800,
     ciPolicy: "optional",
-    pattern: /^composer-.*\.js$/,
+    pattern: /^app-initial~app-main~page-.*\.js$/,
     missingDescription: "composer status panel bundle",
     skipDescription: "persistent status panel patch",
     apply: applyPersistentStatusPanelPatch,
--- a/linux-features/persistent-status-panel/test.js
+++ b/linux-features/persistent-status-panel/test.js
@@ -10,13 +10,15 @@ const test = require("node:test");
 const {
   loadLinuxFeaturePatchDescriptors,
 } = require("../../scripts/lib/linux-features.js");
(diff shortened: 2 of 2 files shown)


ilysenko (approved):
Reviewed against the current 26.707.41301 DMG asset. The descriptor now targets the current unified status bundle, preserves fail-soft behavior, and all required checks pass.

Issue #838, opened 2026-07-11
Add Project locks the GNOME/X11 desktop and floods system logs


ChanJoon:
## Summary

Clicking **Add Project** locked the GNOME/X11 desktop. The cursor still moved, but mouse clicks, keyboard input, Dock previews, and new terminal windows stopped working.

GNOME Shell then emitted continuous JS/Clutter errors until `syslog` grew to about 45 GB and filled the root filesystem.

## Environment

- Ubuntu 22.04, kernel 6.8.0-111-generic
- GNOME Shell 42.9, X11
- NVIDIA RTX 5080, driver 570.172.08
- App 26.707.41301, Electron 42.1.0
- Wrapper commit 8da34a3
- Launched through `codex-app/start.sh`

## Reproduction

1. Launch the generated app on GNOME/X11.
2. Click **Add Project**.
3. The folder picker does not become usable and desktop input locks up.

I did not repeat the test because the first occurrence exhausted the root filesystem.

## Evidence

GNOME Shell repeatedly logged:

```text
2026-07-12T00:19:48+09:00 gnome-shell: == Stack trace for context ... ==
2026-07-12T00:19:53+09:00 gnome-shell: Attempting to run a JS callback during garbage collection. The callback was blocked.
2026-07-12T00:20:18+09:00 gnome-shell: clutter_actor_event: runtime check failed: (retval == CLUTTER_EVENT_PROPAGATE)
2026-07-12T00:20:23+09:00 gnome-shell: The offending signal was grab-op-begin on MetaDisplay.
```

Counts during the first 80 seconds:

- 49,998 stack-trace markers
- 31,073 blocked GC callbacks
- 31,066 blocked `SourceFunc()` callbacks
- 114 blocked `grab-op-begin` signals

No NVIDIA Xid, DRM, or kernel GPU error was recorded during the same window.

A CUDA job was using about 90% GPU compute at the time, but the driver remained responsive.

## Expected behavior

Add Project should open a usable folder picker without locking desktop input, breaking GNOME Shell window actors, or flooding system logs.

## Related issues

- #496 — X11 desktop input lockup from the Codex title-bar menu
- #576 — GUI freeze with broken dragging, minimizing, and keyboard input

ChanJoon:
### Additional source-level finding

I traced the **Add Project** action through the packaged `26.707.41301` main process.

The relevant code is in the generated application bundle:

- `resources/app.asar/.vite/build/main-BOrkpaTV.js`
[...]

PR #839, opened 2026-07-11
Avoid parented project picker on Linux X11

Yo-DDV:
## Summary

- open the Add Project folder picker without a parent window on Linux X11
- preserve the parented dialog on Wayland, macOS, and Windows
- cover explicit and fallback X11 detection against the current main bundle

## Why

The current picker attaches the native directory dialog to the primary `BrowserWindow`. The evidence in #838 points to that modal path entering a GNOME Shell input grab that never recovers. Using the existing unparented Electron overload on X11 avoids that relationship without changing project selection or persistence.

Thanks to @ChanJoon for tracing the packaged handler and narrowing the A/B test.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` (363 passed)
- current `main-BOrkpaTV.js` patch and second-run idempotence check
- `git diff --check`

Addresses #838

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- The Add Project folder picker is no longer parented to the Codex window on
+  Linux X11. This avoids a GNOME Shell modal input grab that could lock desktop
+  input and flood system logs, while preserving parented dialogs on Wayland,
+  macOS, and Windows.
 - Resuming a completed thread no longer leaves it registered as streaming when
   the app server reports no active runtime. This clears stale stream ownership
   before the next message instead of leaving the renderer in a repeated thread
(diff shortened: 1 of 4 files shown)


ilysenko (changes requested):
Thanks for the focused X11 mitigation. Please address two blockers before merge:

- Mark `linux-x11-project-picker` as optional. Drift in this session-specific workaround must warn and continue rather than abort every Linux build as `required-upstream`.
- Do not let the DISPLAY fallback override an explicit `XDG_SESSION_TYPE=wayland`. Apply fallback X11 detection only when the session type is absent/unknown, and add coverage for Wayland with `DISPLAY` set but no `WAYLAND_DISPLAY`.

Yo-DDV:
Thanks, both points are valid.

I changed the picker patch to optional and tightened the fallback so an explicit `XDG_SESSION_TYPE=wayland` always keeps the parented dialog, even when `DISPLAY` is set and `WAYLAND_DISPLAY` is absent. The DISPLAY fallback now applies only when the session type is absent or not recognized.

I also added regression coverage for that Wayland case and for absent or unknown session types with conflicting display signals. The focused tests, full patcher suite, and smoke suite pass locally on the updated head.

ilysenko (approved):
Re-reviewed the updated head. The X11 picker patch is now fail-soft, explicit Wayland sessions no longer fall through to DISPLAY-based X11 detection, regression coverage is present, and all required checks pass.

PR #843, opened 2026-07-11
fix(browser): isolate IAB socket discovery

### LINKED ISSUE #841: IAB discovery disconnects an active Chrome extension task
## What happens

Starting an in-app Browser session in one task can disconnect an active Chrome extension session in another task. Chrome and its tabs stay open, but the original task starts returning `Browser is not available`.

This was reproduced with Desktop payload `26.707.41301`:

1. Task A connects to Chrome through the extension backend and reads a tab.
2. Task B starts the bundled Browser client and selects IAB.
3. Task A tries another read and loses its browser binding.

Expected behavior: discovering IAB must not touch or evict an active Chrome extension client.

## Evidence

In the recorded run, IAB setup began at `16:22:04.814Z`; the Chrome task failed at `16:22:05.145Z`. Chrome did not restart.

Both clients scan the same `/tmp/codex-browser-use` directory. The IAB client connects to candidate sockets before it filters the discovered backends. When it probes an `extension-*` socket, the Chrome native host accepts the new connection and replaces its current Browser client. The IAB task then ignores that Chrome backend, but the original Chrome task has already been disconnected.

The Chrome client must continue discovering `extension-*` sockets; only IAB discovery should ignore them before connecting.

## Current work

Merged PR #843 filters Chrome extension sockets only in the bundled IAB/browser staging path and adds regression, package-staging, and live non-eviction coverage.

This is separate from #788, which bounds native-bridge resources, and #840, which fixes writable bundled-plugin staging.

Vantrongs:
## Summary

- keep IAB discovery away from live Chrome extension sockets on Linux
- patch only the bundled `browser` plugin; leave the `chrome` plugin unchanged
- include the helper in packaged update-builder bundles for deb, rpm, and pacman rebuilds

## Why

The bundled Browser and Chrome clients share `/tmp/codex-browser-use`. Before selecting its IAB backend, the Browser client probes every Unix socket in that directory. Probing a live `extension-*` socket makes the Chrome native host accept a second client, evict the active Chrome client, and clear its pending requests. The IAB task then discards that backend, while the Chrome task fails with `Browser is not available` even though Chrome is still running.

The failure was reproduced with two concurrent tasks: IAB setup began at `16:22:04.814Z`, and the active Chrome task lost its binding at `16:22:05.145Z`.

This change filters `extension-*` entries before the IAB client opens candidate sockets. Unknown or ambiguous bundle shapes are left unchanged with a warning. The native host's single-client behavior and Chrome discovery are not changed.

Fixes #841.

## Validation

- focused socket-scope tests: 3/3
- complete Node test set: 384/384
- `cargo test -p codex-computer-use-linux`: all suites pass, including Chrome native host 57/57
- `nix develop -c bash tests/scripts_smoke.sh`: pass
- `nix flake check --no-build`: pass with existing output warnings
- exact application to Browser `26.707.41301`: one target and idempotent; Chrome staging remains unmodified
- live check: the same Chrome binding returned an identical ordered snapshot of 155 tabs before and after a second client ran IAB setup/discovery
- update-builder staging smoke: helper present
- real deb and rpm builds: helper present in both package payloads
- deb, rpm, and pacman CI payload assertions added; local pacman packaging reached `makepkg` but could not complete in the NixOS host environment, so the Arch CI job is the package-format gate
- independent review of the complete final diff: no code blockers; wording and packaging validation findings addressed

### DIFF
--- a/scripts/ci/container-entrypoint.sh
+++ b/scripts/ci/container-entrypoint.sh
@@ -275,6 +275,7 @@ run_deb_job() {
     assert_contains_file /tmp/deb-contents.txt './opt/codex-desktop/update-builder/install.sh'
     assert_contains_file /tmp/deb-contents.txt './opt/codex-desktop/update-builder/launcher/webview-server.py'
     assert_contains_file /tmp/deb-contents.txt './opt/codex-desktop/update-builder/scripts/lib/upstream-dmg-intel.js'
+    assert_contains_file /tmp/deb-contents.txt './opt/codex-desktop/update-builder/scripts/lib/patch-browser-client-iab-socket-scope.js'
     assert_contains_file /tmp/deb-contents.txt './opt/codex-desktop/.codex-linux/codex-packaged-runtime.sh'
 
     rm -rf dist
@@ -331,6 +332,7 @@ run_rpm_job() {
     assert_contains_file /tmp/rpm-contents.txt '/opt/codex-desktop/update-builder/install.sh'
     assert_contains_file /tmp/rpm-contents.txt '/opt/codex-desktop/update-builder/launcher/webview-server.py'
     assert_contains_file /tmp/rpm-contents.txt '/opt/codex-desktop/update-builder/scripts/lib/upstream-dmg-intel.js'
+    assert_contains_file /tmp/rpm-contents.txt '/opt/codex-desktop/update-builder/scripts/lib/patch-browser-client-iab-socket-scope.js'
     assert_contains_file /tmp/rpm-contents.txt '/opt/codex-desktop/.codex-linux/codex-packaged-runtime.sh'
 
     rm -rf dist
@@ -381,6 +383,7 @@ run_pacman_job() {
     assert_contains_file /tmp/pacman-contents.txt 'opt/codex-desktop/update-builder/install.sh'
     assert_contains_file /tmp/pacman-contents.txt 'opt/codex-desktop/update-builder/launcher/webview-server.py'
     assert_contains_file /tmp/pacman-contents.txt 'opt/codex-desktop/update-builder/scripts/lib/upstream-dmg-intel.js'
+    assert_contains_file /tmp/pacman-contents.txt 'opt/codex-desktop/update-builder/scripts/lib/patch-browser-client-iab-socket-scope.js'
     assert_contains_file /tmp/pacman-contents.txt 'opt/codex-desktop/.codex-linux/codex-packaged-runtime.sh'
 
     rm -rf dist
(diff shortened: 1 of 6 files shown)


Yo-DDV:
Thanks for putting this up. I ran an independent pass on this head since I had a branch in progress for the same issue.

The shape looks right to me. The helper patches the installed 26.707.41301 browser-client bundle cleanly, a second run stays idempotent, and the Chrome staging path does not get the IAB marker. I also reran `node --test scripts/lib/patch-browser-client-iab-socket-scope.test.js`, `node --test scripts/*.test.js scripts/lib/*.test.js`, Semgrep on the touched files, and `bash tests/scripts_smoke.sh`; all green here.

The only thing I did not reproduce locally is the live two-task Chrome/IAB smoke from the checklist. I would keep that as the final runtime proof before taking this out of draft. I do not see a separate PR from my side adding value now.

Vantrongs:
Thanks for the independent pass. I completed the remaining runtime proof on the installed NixOS build: one client held the same live Chrome extension binding while a second client executed the patched IAB setup/discovery path. The Chrome client returned 155 tabs both before and after; the ordered `{id,title,url}` snapshots were byte-for-byte identical, and no browser actions were performed. The second CLI client could not select the Desktop-thread-owned IAB binding (`list=[]`), so I have worded the checklist narrowly as Chrome non-eviction during cross-kernel IAB discovery rather than claiming a full IAB tab smoke. I updated the PR body and am taking it out of draft.

Vantrongs:
Rebased the PR onto current `main` at `a0cfb22` after #788 merged. #788 bounds native bridge frames/correlations but leaves `replace_with_client()` and its `clients.drain()` single-client eviction behavior unchanged, so it does not supersede this fix. Post-rebase validation: focused 3/3, Node 384/384, all `cargo test -p codex-computer-use-linux` suites pass (including Chrome native host 57/57), full Nix smoke passes, and `nix flake check --no-build` passes.

ilysenko (changes requested):
Thanks for the focused socket-discovery fix. One packaging blocker remains:

- Include `scripts/lib/patch-browser-client-iab-socket-scope.js` in `stage_update_builder_bundle()` and add a package payload smoke assertion. Native `.deb`, `.rpm`, and pacman updater rebuilds currently receive the updated `bundled-plugins.sh` without this helper, so the fix is silently skipped during future rebuilds.

Before requesting another maintainer review, please verify the update against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant tests locally, use your local coding model to perform an independent code review of the complete final diff, address its findings, and wait until all required CI checks have completed successfully.

ilysenko (approved):
Reviewed the updated head: the update-builder payload blocker is fixed and all required checks pass.

PR #845, opened 2026-07-11
Add Niri support for pet overlay placement

### LINKED ISSUE #842: Pet overlay cannot float, move freely, or preserve focus on Niri
## What happens

With the optional `pet-overlay` feature on Niri, the pet opens as a normal tiled window. It can only be moved between Niri's tiled insertion positions, cannot be placed freely near screen edges, and may take focus from the active window.

Expected behavior:

- open floating without stealing focus;
- allow free placement within the current output's working area;
- keep manual placement when unlocked and configured corner gravity when locked;
- acquire keyboard focus only while the inline reply is being used, then release it.

## Why this needs Niri integration

The four visible placement choices are Niri tile insertion targets, not the pet's gravity corners. Electron cannot position a native Wayland toplevel reliably on its own, so the optional feature needs compositor cooperation.

The existing pet overlay already has a Hyprland-specific path. Niri provides window-id-targeted commands for moving a window to floating layout and changing its floating position. Any Niri implementation must match only the current process' exact pet window and reject malformed, foreign, stale, or ambiguous window data.

This belongs in the disabled-by-default `pet-overlay` feature. It is separate from the Computer Use Niri backend and should not enable Niri support globally.

## Current work

Merged PR #845 implements the Niri path and Nix feature selection, including output-local coordinate coverage for secondary/negative-origin displays and strict native safe-integer checks for Niri ids and pids. The maintainer accepted the remaining coverage gap for this optional feature: the current test proves non-focusable initial show followed by restored focusability, but not focus acquisition while replying and release afterward.

Yo-DDV:
## Summary

- Adds Niri handling to the optional pet overlay path, using `niri msg` only in detected Niri sessions.
- Matches only the current process' exact `Codex Pet Overlay` window and rejects foreign, malformed, stale, numeric-string, or ambiguous candidates before moving anything.
- Converts desired Electron overlay bounds to output-local working-area coordinates before `move-floating-window`, with secondary-output and negative-origin coverage.
- Keeps the first interactive presentation non-focusable through `showInactive()`, then restores focusability for inline replies.
- Allows `pet-overlay` in the Nix Linux feature list and documents the Niri behavior.

Fixes #842.

## Tests

- `node --test linux-features/pet-overlay/test.js`
- `git diff --check`
- GitHub CI on `93df568a`: Rust and Smoke Tests, Nix Package Builds, Debian/RPM/Pacman package builds


ilysenko (changes requested):
Thanks for the work on the Niri integration. Three blockers remain:

- Convert Electron global overlay coordinates to Niri output-local working-area coordinates before calling `move-floating-window`. Please add coverage for secondary and negative-origin outputs.
- Complete or prove the focus lifecycle required by #842. The current implementation relies on a user-provided `open-focused false` rule and does not test inline-reply focus acquisition and release.
- Validate Niri `id` and `pid` as native safe integers without numeric-string coercion so malformed window data remains fail-closed.

Before requesting another maintainer review, please verify the update against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant tests locally, use your local coding model to perform an independent code review of the complete final diff, address its findings, and wait until all required CI checks have completed successfully.

Yo-DDV:
Thanks @ilysenko. I pushed `93df568a` and addressed the three blockers:

- `move-floating-window` now uses the remembered working-area-local coordinates, with tests for secondary and negative-origin outputs.
- The initial interactive show now keeps the window non-focusable through `showInactive()` and restores focusability on the next tick for inline replies; the Niri lifecycle is covered.
- Niri `id` and `pid` now have to be native positive safe integers, so numeric strings fail closed.

I also rechecked `CONTRIBUTING.md`, reran the relevant local checks listed in the PR body, reviewed the final diff again, and waited for the required CI. All five GitHub checks are green on `93df568a`.

ilysenko (approved):
Reviewed the updated head: the functional Niri blockers are fixed, the remaining focus-lifecycle coverage gap is acceptable for this optional feature, and all required checks pass.

PR #847, opened 2026-07-11
fix(updater): preserve cached DMG on failure

ryan-mt:
## Summary
- stage DMG downloads in a unique sibling file
- atomically replace the cache only after streaming, hashing, and syncing succeed
- preserve the prior cache on HTTP and interrupted-download failures

## Tests
- `cargo test -p codex-update-manager upstream::tests` (5 passed)
- `cargo test -p codex-update-manager` (227 passed; 3 unrelated feature-picker tests failed because Node.js is unavailable in the WSL test environment)

### DIFF
--- a/updater/Cargo.toml
+++ b/updater/Cargo.toml
@@ -16,11 +16,11 @@ serde = { version = "1.0.228", features = ["derive"] }
 serde_json = "1.0.150"
 semver = "1.0.28"
 sha2 = "0.11.0"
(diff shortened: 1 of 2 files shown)


ilysenko:
This work overlaps with the ongoing updater changes in [#834](https://github.com/ilysenko/codex-desktop-linux/pull/834). Please check the open pull requests before opening a new one and coordinate with existing work when there is overlap. Thank you.

Issue #849, opened 2026-07-11
Chrome extension side panel fails on Linux: no handler for codexRuntime/hello


diegomarzaa:
## Environment

- Ubuntu 24.04.3 LTS, x86_64
- GNOME on X11
- `codex-desktop` package: `2026.07.09.225102`
- Brave: `144.1.86.146`
- ChatGPT Chrome extension: `1.2.27203.26575`
- Installed with `make bootstrap-native`

## What happens

Opening the ChatGPT extension side panel in Brave shows:

```text
Unable to start ChatGPT
No handler registered for method: codexRuntime/hello
```

Reinstalling the Chrome plugin from Codex Desktop does not fix it.

## Steps to reproduce

1. Build and install the current Linux package with `make bootstrap-native`.
2. Install or reinstall the bundled Chrome plugin from Codex Desktop.
3. Open Brave using the profile where the extension is installed.
4. Open the ChatGPT extension side panel.

## Expected behavior

The side panel connects to Codex Desktop and starts the local ChatGPT app server.

## Actual behavior

The side panel cannot start because `codexRuntime/hello` has no registered handler.

Browser Use itself works. I verified that Codex can list my existing Brave tabs, claim a tab, inspect its DOM, click controls and type into a form. The problem appears limited to the visible ChatGPT side panel and its app-server startup protocol.

## What I found

The installed extension requests:

```text
codexRuntime/hello
codexRuntime/ensure
codexRuntime/restart
```

It requires native-host protocol version 2.

The Linux host in `computer-use-linux/src/bin/codex-chrome-extension-host.rs` routes Chrome requests to the connected Browser Use client. That client does not implement `codexRuntime/hello`, so it returns the error above.

The current host seems to cover the Browser Use relay but not the v2 side-panel runtime methods. A complete fix may need the Linux host to implement the `codexRuntime/*` handshake and manage the matching `codex app-server`, while preserving the existing Browser Use relay.

I could not find an existing issue for this exact error.

Yo-DDV:
Small follow-up. I opened #852 with the Linux-side case I could still reproduce here.

On current `main`, `codexRuntime/hello` is already handled. The remaining failure I hit happens one step later, when `codexRuntime/ensure` rejects the app-server manifest because the Chrome plugin cache under `~/.codex` was created group-writable by the launcher. The PR keeps the runtime checks strict and hardens that cache before writing the native-host manifests.

Local tests are green, and the full PR CI is green as well now.

[...]

PR #852, opened 2026-07-11
Fix Chrome runtime cache permissions

Yo-DDV:
This follows the side-panel runtime work from #768 and the report in #849.

On my local install the native host already answers `codexRuntime/hello`, but `codexRuntime/ensure` can still fail when the user plugin cache was created with a group-writable umask. The runtime trust checks then reject `~/.codex` or the Chrome plugin cache before the app-server starts.

This keeps those trust checks strict and makes the launcher harden the Chrome plugin runtime cache before writing the browser native-host manifests. Existing cache hits and first-time cache copies both get group/other write removed.

Tests run:
- `cargo fmt --check --all`
- `bash -n launcher/start.sh.template && bash -n tests/scripts_smoke.sh && git diff --check`

### DIFF
--- a/computer-use-linux/src/bin/codex-chrome-extension-host.rs
+++ b/computer-use-linux/src/bin/codex-chrome-extension-host.rs
@@ -1239,7 +1239,9 @@ while True:
         fs::set_permissions(&fake_node, fs::Permissions::from_mode(0o700)).unwrap();
 
         let extension_id = "abcdefghijklmnopabcdefghijklmnop";
-        let extension_host_path = trusted_current_exe_link(&root);
+        let extension_host_path = root.join("extension-host");
+        fs::copy(env::current_exe().unwrap(), &extension_host_path).unwrap();
+        fs::set_permissions(&extension_host_path, fs::Permissions::from_mode(0o700)).unwrap();
         let manifest_path = root.join("chrome-native-hosts-v2.json");
         fs::write(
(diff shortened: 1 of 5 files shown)


ilysenko (changes requested):

- Please rebuild the Chrome plugin cache from bundled resources when it or a relevant ancestor is group/world-writable. Recursively removing write bits from the existing cache can bless previously modified executable files that the partial sync does not verify.
- Please avoid requiring a cross-filesystem hard link in `trusted_current_exe_link`; the focused Rust test fails with `EXDEV` when the workspace and temporary directory are on different filesystems.

Yo-DDV:
Thanks, that was the right call.

I pushed an update for both blockers. The Chrome source bundle gets tightened before it can be used, the cached copy is still tightened after sync, and the runtime test now uses a private copied host binary instead of a hardlink.

The first update tripped the Rust clippy job, so I fixed that too and rebased on current `main`. Local checks are green on my side now, including clippy, the Chrome runtime crate, the patcher tests, and the launcher smoke suite. I’ll watch the new CI from here.

ilysenko (approved):
Re-reviewed the current head. The cache trust blocker is resolved: unsafe caches are detected before chmod and rebuilt from staged trusted resources, with regression coverage for an unchecked tampered module. The cross-filesystem test issue is also fixed, and all required checks are green.

PR #853, opened 2026-07-11
Fix new turns in unowned resumed threads

### LINKED ISSUE #850: Completed CLI thread cannot start a new turn after Desktop resume
## What happens

A completed thread can be opened normally after Desktop resumes it, but sending
the next message fails with:

> Please continue this conversation on the window where it was started.

In the reproduced case the thread was created in Codex CLI and there was only
one Desktop window, so there was no original Desktop window to return to.

## Reproduction

1. Complete a turn in Codex CLI.
2. Open the same thread in Desktop after its in-memory ownership has been reset.
3. Wait for `thread/resume` to finish.
4. Submit a follow-up message.

Desktop displays the thread and its completed turn, but rejects the follow-up
before `turn/start` is sent.

## Evidence

The affected rollout is complete and ends with `task_complete`. Desktop logs
show this sequence:

- `thread/resume` succeeds with `latestTurnStatus=completed`;
- the runtime is idle;
- `markedStreaming=false` and `assignedStreamRole=null`;
- submit immediately fails with the message above;
- repeated resumes recreate the same state and no `turn/start` is routed.

This is a follow-up to #835/#836 rather than the same stale-streaming freeze.
#836 correctly keeps a completed idle thread non-streaming, but the new-turn
path still requires an owner role and can treat the resulting null role as a
follower.

## Expected behavior

Opening a completed thread should remain non-streaming. When the sole Desktop
window submits a new turn, it should safely claim ownership if no real owner is
present. A real follower with an `ownerClientId` must continue forwarding to
that owner.

The error should not direct users to a non-existent Desktop window when the
previous client was CLI.

ilysenko:
## What changed

- claim local stream ownership when a resumed thread has no owner and the user starts a new turn
- keep real follower windows forwarding to their existing `ownerClientId`
- add regression coverage for both the unowned and follower paths

## Why

The completed-thread resume fix correctly leaves an idle completed thread non-streaming, but that also leaves its stream role unset. The new-turn path treated the null role like a follower and rejected submission with “Please continue this conversation on the window where it was started,” even when the thread originated in Codex CLI and no Desktop owner existed.

Fixes #850.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` (365 passed)
- current staged webview asset matches the new patch needle
- applying the patch twice to the current asset is idempotent
- `git diff --check`

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -5957,6 +5957,49 @@ test("does not retain streaming ownership when a completed thread resumes withou
   });
 });
 
+test("claims an unowned completed thread when starting a new turn", async () => {
+  const source = [
+    "async function start(e,t,c){",
+    "let p=await MF({conversationId:t,getStreamRole:t=>e.getStreamRole(t),sendRequest:r=>e.sendThreadFollowerRequest(r,`thread-follower-start-turn`,{conversationId:t})});",
+    "if(p)return UP.abort(c,`follower_window_forwarded`),p.result;",
+    "if(e.getStreamRole(t)?.role!==`owner`)throw Error(Kw);",
+    "if(!e.isConversationStreaming(t))throw K.error(`Conversation is not being streamed.`,{safe:{conversationId:t},sensitive:{}}),Error(`Conversation ${t} is not being streamed.`);",
+    "return e.sendRequest(`turn/start`,{threadId:t})",
+    "}",
+  ].join("");
+
+  const patched = applyPatchTwice(applyLinuxAppServerConversationHydrationPatch, source);
+
+  assert.match(patched, /codexLinuxClaimUnownedTurn/);
+  const context = {};
+  vm.runInNewContext(
+    [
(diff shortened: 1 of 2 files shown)


Vantrongs:
Thanks for picking this up. I reproduced the happy path and follower forwarding against this head. Two edge cases look worth deciding before merge:

- The claim currently happens before `waitForPendingThreadSettingsUpdate`, workspace/permissions resolution, and `turn/start`. If any of those throws, the existing catch restores runtime/permissions but not the stream role, so the completed thread can remain streaming + owner without a new turn (the state #835 was meant to clear). A regression test with `sendRequest("turn/start")` rejecting exposes that boundary.
- `getStreamRole() == null` followed by local `markConversationStreaming` / `setConversationStreamRole` is not atomic across renderers. Two null-role windows submitting together can both pass before either snapshot is observed.

Would you prefer rollback for a claim created by this submit plus an IPC-mediated reservation, or keep the narrow fix and track the two-window race separately? I can adapt the focused fixture either way.

PR #854, opened 2026-07-11
fix(webview): coalesce duplicate turn/start submits

### LINKED ISSUE #851: Single Enter can enqueue the same message twice during thread transitions
# Single Enter can enqueue the same message twice during thread transitions

## Summary

A single Enter press can enqueue the same user message twice in Codex Desktop.

Both copies appear in the UI immediately as separate queued messages. In the
rollout log, they are later persisted as two distinct `user_message` events with
different `client_id` values, even though the message text and `turn_id` are the
same.

This occurred twice in one day. One occurrence happened while the submitted
turn triggered context compaction; the other happened during an ordinary turn
start without a `context_compacted` event.

I have not identified the root cause or a reliable reproduction path. Context
compaction may be relevant to one occurrence, but the second occurrence shows
that it is not required. I am reporting this because the same user-visible
failure happened twice in one day and both occurrences are present in the
rollout logs as separate submissions.

## Environment

- Distribution: NixOS 26.05
- Installation: Nix flake package through the Home Manager module
- `codex-desktop-linux` commit: `2d2bdf6dea79179e376958d94fb6f1b5ce03be53`
- Upstream Codex Desktop: `26.707.41301`
- Codex CLI / app-server: `0.144.1`
- Electron: `42.1.0`
- Desktop: Hyprland `0.55.4`
- Session: Wayland

## Steps to reproduce

The issue is intermittent, but one observed path was:

1. Continue using a long-running Codex thread until the next submitted turn
   triggers context compaction.
2. Enter a new message.
3. Press Enter once.
4. Observe the queued user messages while compaction and turn startup complete.

The same behavior was also observed once during an ordinary turn start without
context compaction.

## Actual behavior

Two identical queued messages appear immediately after one Enter press.

The rollout records each copy as an independent message. For the compaction
occurrence:

```text
2026-07-11T17:56:28.550Z context_compacted

2026-07-11T17:56:28.632Z user_message
  turn_id:   019f5252-0b30-7cc0-bcb4-72a83cee1c21
  client_id: 90a68f68-7605-4bd3-80a9-ebf63e7bb434
  message:   <message B>

2026-07-11T17:56:49.468Z user_message
  turn_id:   019f5252-0b30-7cc0-bcb4-72a83cee1c21
  client_id: a0659bd7-8fd2-44c5-91d9-65a980e3b6a8
  message:   <message B, byte-identical>
```

Although the backend timestamps are about 21 seconds apart, both copies were
already visible in the UI immediately. The backend timestamps therefore do not
represent the time at which the two copies became visible in the UI.

The Linux wrapper log also shows two successful `turn/start` requests next to
each other:

```text
[AppServerConnection] response_routed ... method=turn/start ... requestId=f14cdc0f-823e-4c8a-8406-ddcbd858c71f ...
[AppServerConnection] response_routed ... method=turn/start ... requestId=004da7ec-77d5-42a2-90bf-2a68a0b80c1d ...
```

This is followed by repeated active-turn mismatch errors:

```text
Request failed ... method=turn/interrupt ...
expected active turn id 019f5252-0b70-7593-83ac-c6cc992f46e9
but found 019f5252-0b30-7cc0-bcb4-72a83cee1c21
```

The earlier occurrence shows the same duplicate-message shape without
compaction:

```text
2026-07-11T08:46:48.659Z user_message
  turn_id:   019f505b-9ad1-70c2-9568-bb3204f8c550
  client_id: aa0d3fb6-c5af-4850-ae69-a8de47076f96
  message:   <message A>

2026-07-11T08:46:56.177Z user_message
  turn_id:   019f505b-9ad1-70c2-9568-bb3204f8c550
  client_id: 8d4da0e9-2fee-4d5d-aa23-edbcd2b86abd
  message:   <message A, byte-identical>
```

## Expected behavior

One Enter press should create exactly one queued message and one turn-start
request, including while the thread is compacting or otherwise transitioning
between turn states.

Yo-DDV:
## Summary

Fixes #851.

- Add an optional Linux webview patch that reuses an in-flight `turn/start` request when the same turn and message body are submitted twice before the first request settles.
- Target the current queued request-client path before request creation.
- Register the patch for the current `hotkey-window-thread-page` webview chunks while keeping it fail-soft if upstream drifts again.

## Validation

- `node --test --test-name-pattern "coalesces duplicate queued turn starts|turn/start duplicate guard descriptor|remote conversation hydration out of core|default core patch descriptors" scripts/patch-linux-window-ui.test.js` (`4` tests)
- `node --test scripts/patch-linux-window-ui.test.js` (`362` tests)
- `bash scripts/ci/run-node-checks.sh` (`945` tests)
- `cargo fmt --check`
- `cargo test -p codex-update-manager` (`235` tests)

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -157,6 +157,7 @@ const {
   applyLocalEnvironmentActionModalDraftPatch,
   applyPersistentRateLimitFooterPatch,
   applyLinuxAppServerBackfillWaitPatch,
+  applyLinuxTurnStartDuplicateGuardPatch,
   applyLinuxCompletedItemRecoveryPatch,
   applyLinuxRemoteTerminalStatusRecoveryPatch,
   applyLinuxAppServerFeatureEnablementPatch,
@@ -1046,6 +1047,7 @@ test("default core patch descriptors are grouped and unique", () => {
     "linux-app-sunset-gate",
     "linux-app-server-feature-enablement",
     "linux-app-server-backfill-wait",
+    "linux-turn-start-duplicate-guard",
     "linux-completed-item-recovery",
     "linux-remote-terminal-status-recovery",
     "linux-skills-list-dedupe",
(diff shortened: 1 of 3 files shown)


ilysenko:

- Please remove the older direct `sendRequest -> createRequest` patch path and its tests. This repository supports only the current upstream DMG, so the guard should target only the current queued request-client shape.

Yo-DDV:
Thanks @ilysenko, that makes sense.

I removed the direct `sendRequest -> createRequest` patch path and its test. The guard now only targets the current queued request-client shape, and the warning fallback keys on that shape too.

I also rebased on current `main` after #864/#865, updated the PR body, and reran the focused patcher tests, full patcher suite, node CI checks, smoke script, and Rust checks listed there.

ilysenko:
Thanks for working on #851. We have since merged [#864](https://github.com/ilysenko/codex-desktop-linux/pull/864), which removes the Linux-specific completed-resume ownership reset and submit-time ownership claim and restores the upstream conversation ownership lifecycle.

We are not claiming that #864 fixes #851, since the duplicate submission was observed before those ownership patches were introduced. However, it gives us a clean upstream baseline again. Adding another Linux-only guard in the shared request client now would make it harder to determine whether the problem still exists and where it originates.

The current guard also depends on a `turnId` that is not present in the current outgoing `turn/start` payload, so it would not coalesce the reported production requests as written.

Let’s close this PR for now and continue monitoring #851 on builds containing #864. If the issue reproduces again, the next fix should target the original duplicate submit before optimistic UI state is created, rather than coalescing requests afterward.

Yo-DDV:
Thanks @ilysenko, understood.

I’m closing this PR now and leaving #851 to monitoring on builds that include #864. If it reproduces again, I agree the next fix should target the original duplicate submit before optimistic UI state is created, not request-client coalescing afterward.

PR #858, opened 2026-07-11
fix: destroy tray icon before process exit on Linux quit

Ikkisovi:
## Summary
- keep a reference to the Linux `Tray` instance created by the upstream tray factory
- explicitly call `Tray.destroy()` when the existing quit lifecycle starts
- also destroy it from `before-quit` as a fallback
- keep tray creation and multi-instance behavior unchanged

## Why
The icon could survive application exit because the tray object was left for process teardown. Destroying the existing tray before exit unregisters it through Electron while the main process is still alive.

## Tests
- `node --test scripts/patch-linux-window-ui.test.js` (362/362)
- `git diff --check`
- Node syntax checks for the changed patch implementations

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -2214,9 +2214,9 @@ test("adds the Linux quit guard to the current comma-declared Electron prelude",
 
   const patched = applyPatchTwice(applyLinuxQuitGuardPatch, source);
 
-  assert.match(patched, /let codexLinuxQuitInProgress=!1/);
+  assert.match(patched, /codexLinuxQuitInProgress=!1/);
   assert.match(patched, /codexLinuxExplicitQuitApproved=!1/);
-  assert.match(patched, /codexLinuxMarkQuitInProgress=\(\)=>\{codexLinuxQuitInProgress=!0\}/);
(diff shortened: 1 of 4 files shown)


ilysenko:
Thanks for investigating this and preparing a tested fix. We are not planning to add another quit-lifecycle patch for this issue at the moment. The default Linux build already carries many upstream-sensitive patches, and adding timed delays to multiple quit paths would further increase maintenance complexity for an issue that is not currently a high priority. We may revisit it later if the impact becomes broader or a simpler upstream-aligned solution becomes available. Thank you again for the detailed analysis and work.

ilysenko:
reopened :)

ilysenko:
Thanks for investigating this. Runtime inspection shows that the reported icons belong to still-running secondary `--new-instance` Electron processes, so delaying tray destruction after an explicit Quit does not address the observed behavior. We are fixing the ownership mismatch instead: secondary multi-launch instances are trayless by default and exit when their last window closes. Superseded by #859.

ilysenko (approved):
Verified on the current DMG and across the full CI matrix. The tray instance is explicitly destroyed during the existing quit lifecycle without changing tray creation behavior.

PR #859, opened 2026-07-11
Suppress tray icons for multi-launch instances

ilysenko:
## Summary
- suppress tray creation for secondary `--new-instance` launches by default
- keep close-to-tray behavior aligned with whether that instance owns a tray icon
- allow opting back in with `CODEX_MULTI_LAUNCH_TRAY=1`

## Tests

## Notes
This carries the final reviewed fix from #794 onto the current `main`, because the original fork branch does not allow maintainer edits.


ilysenko:
Closing this replacement because it addressed the wrong behavior. The reported bug is specifically that an existing tray icon is not removed during explicit application exit; tray creation and multi-instance policy are out of scope. Sorry for the detour.

PR #862, opened 2026-07-11
Fix Nix LD_LIBRARY_PATH leaking into Linux child processes

jamie-steele:
## Summary

Fix the packaged Nix `LD_LIBRARY_PATH` leaking from the Electron launcher into host child processes.

The launcher now captures the user’s original `LD_LIBRARY_PATH` before adding packaged runtime libraries, preserving the distinction between:

- unset
- set to an empty string
- set to a non-empty user-defined value

Packaged application processes such as Electron continue receiving the Nix runtime paths they require. Host commands, terminals, login shells, Codex CLI processes, and environment-discovery helpers instead receive the original user environment.

This prevents host programs from loading incompatible Nix libraries. On Ubuntu 22.04-based systems, the leak caused commands such as `flatpak` to load Nix libraries requiring newer GLIBC versions.

While tracing the complete Environments UI path, two additional failure cases were confirmed and addressed:

- Projects without a discovered environment now receive a default environment configuration path.
- `ENOTDIR` is treated as “environment not configured” when `.codex` is a regular compatibility file rather than a directory.

No Flatpak-, distribution-, or Environments-screen-specific process exception was added. The library-path fix is applied at the general host child-process boundary, without changing sandboxing or command restrictions.

## Validation

Automated validation:

- `git diff --check` — passed
- `node --test scripts/patch-linux-window-ui.test.js` — 367/367 tests passed
- `nix build .#codex-desktop --no-link --print-out-paths` — passed

Focused regression coverage includes:

- no original `LD_LIBRARY_PATH`
- an explicitly empty original `LD_LIBRARY_PATH`
- a non-empty user-defined value
- application-added Nix paths not reaching host children
- packaged Electron processes retaining their runtime paths
- local terminal environment restoration
- empty local-environment discovery
- `ENOTDIR` while reading a candidate environment configuration

Manual validation on Pop!_OS 22.04, x86_64, host GLIBC 2.35, Flatpak 1.14.6:

- Settings → Environments loads for a local project
- login shells no longer emit Flatpak/GLIBC loader errors
- `flatpak --version` works from an app-spawned shell
- normal Codex tasks and terminal commands continue working
- Electron starts successfully with its packaged runtime libraries

Known limitation: a project where `.codex` is a regular file cannot physically store `.codex/environments/environment.toml`. Such a path is now correctly reported as not configured instead of making the Environments page unavailable.

CI has not yet run on the pull request.

## Checklist

- [ ] Required pull-request CI checks pass.
- [x] I reviewed the final diff with my coding agent, addressed the findings, and reran the relevant tests.


ilysenko (changes requested):
Thanks for working on the Nix `LD_LIBRARY_PATH` leak. There are several blockers that need to be addressed before this can merge:

1. The global `node:child_process` wrapper overwrites explicitly supplied child `LD_LIBRARY_PATH` values and injects the user value into intentionally minimal environments. It also strips the Nix runtime paths from every bundled/runtime child process. Please restore only inherited application-added paths and preserve caller-provided `env` unchanged.
2. Replacing `exec` and `execFile` globally drops properties from the original Node functions, including `util.promisify.custom`. Please avoid the module-wide monkeypatch or preserve the complete Node API contract through a narrower host-process boundary.
3. The empty-project webview patch recomputes the default configuration path while bypassing the upstream eligibility guards. Please reuse the already guarded candidate and add tests covering the disabled/false eligibility states.
4. The three new descriptors should not be `required-upstream`. These fixes are not required for the core app to remain usable, so future upstream drift must warn and continue instead of rejecting builds and updater candidates for every package format.

Before pushing the next revision:

- Read and verify the complete change against the full repository [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), including the current-DMG-only, fail-soft drift, source-of-truth, testing, and cross-format requirements.
- Run the relevant local test suites before pushing the update.
- Run an independent review of the complete diff with a local LLM using high reasoning effort. Ask it to look specifically for runtime regressions, Node API contract changes, explicit child environment handling, packaged native helpers, updater/build effects, and compatibility across Nix, AppImage, deb, rpm, pacman, and user-local installs. Address its findings before requesting another maintainer review.
- After updating the PR, wait until every required CI check has completed successfully before requesting review or merge.

ilysenko (changes requested):
Thanks for addressing the earlier feedback. The global `child_process` wrapper is gone, explicit child environments are no longer modified globally, and the new descriptors are fail-soft.

One environment-preservation blocker remains. The current implementation can overwrite existing user-defined `LD_LIBRARY_PATH` values:

- A value configured by the user in login-shell startup files such as `~/.profile` is removed from the discovered shell environment. The CLI and terminals then receive the older launcher snapshot instead.
- A value explicitly supplied by a local/worktree environment configuration is overwritten because terminal restoration runs after those project-specific overrides have been applied.

These cases may not be common, but they are valid existing configurations. A fix intended to prevent packaged Nix paths from leaking should remove only the inherited application-added value, without replacing values explicitly supplied by the user or project.

A smaller approach would be to sanitize the inherited base environment before applying explicit overrides, and then preserve any value returned by login-shell discovery. Please add regression coverage for both cases.

To keep this PR easier to review and maintain, please also consider moving the separate `ENOTDIR` and empty-project environment-discovery fixes into a follow-up PR. This is not required for the `LD_LIBRARY_PATH` blocker itself.

Before pushing the next update, please verify the complete change against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant tests locally, and perform an independent full-diff review with a local LLM. After updating the PR, wait for every required CI check to complete successfully before requesting maintainer review or merge.

ilysenko:
Thanks for the work on this. We merged a smaller, `LD_LIBRARY_PATH`-only fix in #877: https://github.com/ilysenko/codex-desktop-linux/pull/877

The merged implementation preserves login-shell and worktree overrides, avoids global `child_process` patching, sanitizes both launcher and Electron updater boundaries, and keeps the ASAR descriptors fail-soft. Closing this pull request as superseded by #877.

jamie-steele:
Hi there,

I noticed that PR #877 retains substantial logic and code that I previously submitted in PR #862, including the three-state LD_LIBRARY_PATH handling, environment markers, host-process boundary logic, and related regression tests.

I intentionally kept “Allow edits from maintainers” enabled so that necessary refinements could be made directly on my branch while preserving contribution history.

Since #862 was closed and the resulting implementation was committed separately, my contribution is absent from the Git history and contributor statistics. Could you please acknowledge #862 and add appropriate co-author attribution in the project history, changelog, or documentation?

Thank you.

PR #863, opened 2026-07-11
Repair swallowed wrapper updater runtime

PinguuSS:
## Summary

- Make the wrapper updater webview patch repair a runtime that was already appended on the same line as a trailing `//# sourceMappingURL=...` comment.
- Add a regression test for the previously-swallowed runtime shape.

## Context

Follow-up to #860. That PR fixed current-DMG clean builds by appending the runtime on a fresh line. This keeps the patch idempotent while also self-healing the exact bad generated shape if the patch is reapplied to an asset that already contains the swallowed runtime.

No open PR overlap was found for wrapper updater sourcemap/runtime repair.

## Validation

- `node --test linux-features/codex-wrapper-updater/test.js`
- `node --check linux-features/codex-wrapper-updater/patch.js`
- `node --check linux-features/codex-wrapper-updater/test.js`
- `git diff --check`

## Out of scope

- No updater Rust changes.
- No settings contract changes.
- No other Linux features.

### DIFF
--- a/linux-features/codex-wrapper-updater/patch.js
+++ b/linux-features/codex-wrapper-updater/patch.js
@@ -99,6 +99,15 @@ function wrapperRuntimeSource() {
 
 function applyWebviewRuntimePatch(source) {
   if (source.includes(`codexLinuxWrapperUpdaterVersion=`)) {
+    const runtimeNeedle = `;(()=>{const VERSION=${JSON.stringify(RUNTIME_VERSION)};`;
+    const runtimeIndex = source.indexOf(runtimeNeedle);
+    if (runtimeIndex > 0) {
+      const lineStart = source.lastIndexOf("\n", runtimeIndex - 1) + 1;
+      const sameLinePrefix = source.slice(lineStart, runtimeIndex);
+      if (sameLinePrefix.startsWith("//# sourceMappingURL=")) {
+        return `${source.slice(0, runtimeIndex)}\n${source.slice(runtimeIndex)}`;
+      }
+    }
     return source;
   }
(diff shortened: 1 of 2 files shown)


ilysenko:
I think I just merged a pull request related to this topic.

ilysenko (changes requested):
Thanks for following up on the wrapper updater runtime issue. The feature itself can remain in the repository, but this PR adds compatibility for stale generated output that was produced before #860.

#860 already fixes the current source path by appending the runtime on a fresh line. New installs and updater rebuilds create a fresh candidate from the current DMG, so they should not need to detect and repair the previously broken generated asset shape.

Please remove the legacy self-repair branch and its test. Keep the implementation focused on the current DMG and the clean-build path. Supporting old generated shapes adds another permanent compatibility path and makes future upstream drift harder to diagnose.

Before updating the PR:

- Read the complete repository [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), especially the current-DMG-only rule, generated-output ownership, source-of-truth discipline, and requirement to remove obsolete compatibility paths.
- Review the complete diff and remove anything that is not needed for a fresh build from the current upstream DMG.
- Run an independent self-review of the complete diff with your local LLM using high reasoning effort. Ask it specifically to identify legacy compatibility branches, unnecessary fallback behavior, generated-artifact repair logic, duplicated paths, and future maintenance cost. Address its findings before requesting another maintainer review.
- Run the relevant local tests before pushing the revision.
- After updating the PR, wait for every required CI check to complete successfully before requesting review or merge.

PR #864, opened 2026-07-11
Restore upstream conversation ownership lifecycle

### LINKED ISSUE #857: Failed turn start can leave a resumed thread falsely owned
## Problem

Follow-up to #850 and merged PR #853.

#853 fixes the normal path for starting a new turn in an unowned resumed thread, but ownership is claimed before the remaining submit pipeline runs. If settings/workspace/permission resolution or `turn/start` fails, no new turn exists while the completed thread can remain marked streaming and owned by the submitting window.

A focused reproduction with `turn/start` rejecting leaves:

```json
{"role":{"role":"owner"},"markedStreaming":true}
```

Expected: ownership created by this submit is rolled back when the submit fails, while pre-existing ownership is preserved.

## Root Cause Analysis

The null-role check and ownership transition happen immediately after follower forwarding. Several fallible operations still run before the new turn is accepted. Existing error recovery restores conversation runtime and permissions, but it does not distinguish or undo ownership created by this submit.

The transition is also renderer-local check-then-set. Two null-role windows can pass it concurrently because there is no reservation or compare-and-set confirmation before either sends `turn/start`.

## TDD Fix Plan

1. **RED**: Reject `turn/start` after a null-role submit claims ownership and assert the completed thread returns to non-streaming/null-role state.
   **GREEN**: Track whether this submit created the claim and roll back only that claim in the submit error path.

2. **RED**: Reject a submit from an already-owned thread and assert its pre-existing ownership is not cleared.
   **GREEN**: Scope rollback to the exact claim generation created by the failing submit.

3. **RED**: Submit concurrently from two null-role windows and assert only one receives permission to send `turn/start`; the other forwards or retries against the winner.
   **GREEN**: Add IPC-mediated ownership reservation/confirmation before dispatch.

## Acceptance Criteria

- [ ] A failed new-turn submit cannot leave a completed thread falsely streaming/owned.
- [ ] Existing owner and real follower behavior remains unchanged.
- [ ] Concurrent null-role windows cannot both send `turn/start`.
- [ ] Regression tests cover rollback and concurrent ownership acquisition.
- [ ] Existing #835/#836/#850 tests remain green.

ilysenko:
## Summary

- remove the Linux-only completed-resume ownership reset
- remove the submit-time unowned-turn ownership claim
- restore the current upstream conversation ownership lifecycle
- keep remote notification hydration scoped to the opt-in remote mobile feature

## Why

`conversationStreaming` tracks which Desktop window owns the conversation event stream; active versus idle turn state is tracked separately. Clearing ownership for completed idle threads diverged from upstream and caused #850. Reclaiming ownership later in the submit pipeline then introduced a non-atomic check-and-set and left false ownership behind when turn setup failed, as reported in #857.

Removing both compensating patches restores the upstream resume and submit behavior instead of maintaining a Linux-only ownership state machine.

Fixes #857.

## Checks

- `node --test scripts/patch-linux-window-ui.test.js` (360 passed)
- `node --check scripts/patches/impl/webview/index.js`
- `git diff --check`

## Notes

- This intentionally supersedes the Linux-specific fixes from #836 and #853.
- No generated app output or package-specific runtime code is changed.


ilysenko:
Follow-up: #851 touches the same general thread-transition area, but it was reproduced on `2d2bdf6d`, before the ownership patches removed by this PR were introduced. It should therefore remain open and should not be considered fixed by #864.

We will continue monitoring for duplicate queued messages and duplicate `turn/start` requests on builds containing this merge.

PR #866, opened 2026-07-11
Test changed Linux Features against upstream DMG

PinguuSS:
## Summary

- run the upstream DMG workflow when a pull request changes a tracked Linux Feature
- derive changed feature ids from the complete PR diff, expand transitive `requires`, and reject conflicts through the existing manifest loader
- generate an uncommitted feature config for one current-DMG build and include it in the diagnostic artifact
- exclude `linux-features/local`, preserve default feature behavior for scheduled and push runs, and keep normal installs/updater rebuilds unchanged

## Why

The shared acceptance gate rejects drift only for features enabled in the candidate patch report. The regular upstream build used the empty feature config, so an opt-in descriptor could stop matching the published DMG while synthetic feature tests and the default build remained green.

This change makes PR-time acceptance aware of the repository features changed by that PR without adding an all-features matrix or changing user configuration.

## Source of truth

- `.github/workflows/upstream-build-app.yml`
- `scripts/ci/changed-linux-features.js`
- `scripts/lib/linux-features.js`
- `docs/upstream-dmg-acceptance.md`

## Current-DMG validation

- repository base: `2752ba912cfa3be48caa081f2fa4b225d71a05f4`
- app: `26.707.41301`
- Electron: `42.1.0`
- DMG SHA-256: `467a25381af2943f2c9b8794adae28b36400bd6da2753daf4889a5fd550c4d65`
- ETag: `0x8DEDED0CF5BF674`
- size: `560409764` bytes
- real isolated build with `ui-tweaks`: verdict `accepted`, all 6 feature patches `applied`, no blockers or warnings
- synthetic drift injected into a copy of that real patch report: verdict `rejected` with `enabled-feature-drift`

## Tests

- `node --test scripts/ci/*.test.js scripts/lib/linux-features.test.js` (42 passed)
- `node --test linux-features/*/test.js` (519 passed)
- `node --test scripts/patch-linux-window-ui.test.js` (360 passed)
- Node syntax checks for the new helper and tests
- YAML parse for `upstream-build-app.yml`
- generator byte-stability check
- `scripts/ci/validate-patch-report.js` against the real build report
- `git diff --check`

## Out of scope

- scheduled canary rotation across all opt-ins
- changes to local installs, updater rebuilds, patch discovery, or the patch engine
- removing GitHub's documented 300-file path-filter limit; that would require a separate lightweight detector job and broader workflow architecture

### DIFF
--- a/.github/workflows/upstream-build-app.yml
+++ b/.github/workflows/upstream-build-app.yml
@@ -65,7 +65,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: ${{ github.event_name == 'pull_request' && 0 || 1 }}
+          fetch-depth: ${{ github.event_name == 'pull_request' && '0' || '1' }}
 
       - name: Set up Node.js
         uses: actions/setup-node@v4
--- a/.github/workflows/upstream-build-app.yml
+++ b/.github/workflows/upstream-build-app.yml
@@ -5,15 +5,19 @@ on:
     - cron: '30 6 * * *'
   pull_request:
     paths:
+      - linux-features/*/**
+      - '!linux-features/local/**'
       - install.sh
       - Makefile
       - scripts/patch-linux-window-ui.js
       - scripts/patch-linux-window-ui.test.js
       - scripts/patches/**
       - scripts/ci/validate-patch-report.js
+      - scripts/ci/changed-linux-features*.js
       - scripts/ci/upstream-dmg-*.js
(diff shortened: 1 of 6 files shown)


ilysenko:
Thanks for the contribution. Please note that the installer already handles this case: if an enabled feature fails or drifts, the new candidate is rejected and the existing working app remains installed. Optional features are a low priority, and our CI time and Actions capacity are limited, so please close this PR for now.

PR #867, opened 2026-07-11
refactor(remote-mobile): move completed-item recovery out of core

Yo-DDV:
## Summary

- move completed-item recovery from the default core patch set into the disabled-by-default `remote-mobile-control` feature
- keep the existing recovery behavior and document its new ownership
- cover feature discovery, recovery behavior, report provenance, and patcher idempotence

## Validation

- `node --test linux-features/remote-mobile-control/test.js`
- `node --test scripts/patch-linux-window-ui.test.js`
- `bash scripts/ci/run-node-checks.sh`
- `bash tests/scripts_smoke.sh`
- `nix flake check --no-build --no-write-lock-file`
- `nix build .#codex-desktop-remote-mobile-control --no-link`
- local patch reports against the upstream `26.707.41301` bundle with the feature disabled, enabled, and applied a second time

Refs #639

PR #868, opened 2026-07-11
refactor(remote-mobile): scope terminal status recovery to feature

Yo-DDV:
## Summary

- Move Linux remote terminal-status recovery out of the default core webview patch set and into the disabled-by-default `remote-mobile-control` feature.
- Keep the existing stale `waitingOnUserInput` recovery behavior covered in feature tests, including older already-patched bundles.
- Make the Node CI syntax check tolerate tracked JavaScript files that are intentionally deleted in a PR, with smoke coverage for that case.

This is a follow-up to #867 and covers the next remote/mobile-specific recovery path from #639. It does not try to close the whole remote/mobile tracking issue.

## Tests

- `node --test --test-name-pattern "remote conversation hydration out of core|remote mobile control feature exposes|remote terminal status" scripts/patch-linux-window-ui.test.js linux-features/remote-mobile-control/test.js`
- `semgrep scan --config p/javascript linux-features/remote-mobile-control/patch.js linux-features/remote-mobile-control/test.js scripts/patches/impl/webview/index.js scripts/patch-linux-window-ui.test.js`
- `nix flake check --no-build --no-write-lock-file`


Yo-DDV:
Thanks for your help @ilysenko

ilysenko (approved):
Verified against the current upstream DMG. The feature-scoped patch now applies atomically, ignores unrelated matching chunks, escapes current minified aliases, and rejects partial drift without mutating the asset. All required checks are green.

PR #869, opened 2026-07-11
refactor(remote-mobile): keep config preservation core-owned

Yo-DDV:
Summary
- remove the duplicate `remote-mobile-control` config-preservation descriptor
- keep the generic Linux `remote_control` config preservation owned by the core patch set
- update the feature patch-report coverage and changelog

Tests
- `node --test linux-features/remote-mobile-control/test.js`
- `node --test scripts/patch-linux-window-ui.test.js`
- `bash scripts/ci/run-node-checks.sh`
- `bash tests/scripts_smoke.sh`
- `semgrep scan --config p/javascript linux-features/remote-mobile-control/patch.js linux-features/remote-mobile-control/test.js scripts/patches/impl/main-process/misc.js scripts/patch-linux-window-ui.test.js`

PR #871, opened 2026-07-11
ci(remote-mobile): validate opt-in feature against upstream DMG

Yo-DDV:
## Summary

- add a CI-only current-DMG compatibility probe for the opt-in `remote-mobile-control` feature
- verify a deterministic second patch pass with stable tree, feature, and descriptor sets
- preserve the feature result in workflow artifacts and include it in scheduled drift reconciliation

## Why

The shared acceptance gate correctly evaluates only the Linux features enabled by a user. The normal CI build has no optional features enabled, so changes under `remote-mobile-control` could otherwise merge without current-DMG coverage.

The probe uses a temporary feature config and `make inspect-upstream`; it does not change normal builds or user installations. Its static descriptor contract covers all 20 feature descriptors expected after the config-preservation cleanup in #869. Generic `remote_control` config preservation remains core-owned.

## Validation

- `bash scripts/ci/run-node-checks.sh` (953 tests)
- `nix flake check --no-build --no-write-lock-file`
- Semgrep JavaScript scan (0 findings)
- workflow YAML, shell syntax, Node syntax, and `git diff --check`
- current `26.707.41301` DMG inspection with `remote-mobile-control` enabled

The current-DMG probe produced an accepted decision with no blockers or warnings. Both patch passes exposed the same 96 descriptors and the extracted tree SHA-256 remained unchanged.

Related to #639.

### DIFF
--- a/.github/workflows/upstream-build-app.yml
+++ b/.github/workflows/upstream-build-app.yml
@@ -10,6 +10,9 @@ on:
       - scripts/patch-linux-window-ui.js
       - scripts/patch-linux-window-ui.test.js
       - scripts/patches/**
+      - linux-features/remote-mobile-control/**
+      - scripts/ci/patch-rerun-check.js
+      - scripts/ci/patch-rerun-check.test.js
       - scripts/ci/validate-patch-report.js
       - scripts/ci/upstream-dmg-*.js
       - scripts/validate-upstream-dmg.js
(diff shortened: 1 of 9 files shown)


ilysenko:

- This reintroduces the optional-feature DMG probing declined in #866. Enabled-feature drift is already rejected before installation, while optional features are low priority and our CI time and Actions capacity are limited. Please remove this probe and close the PR.

Yo-DDV:
Thanks for clarifying. I missed that this falls under the same CI policy you set in #866. I'll close this PR and keep these current-DMG feature checks local instead of adding them to Actions.

PR #872, opened 2026-07-12
fix(remote-mobile): follow current webview chunks

Yo-DDV:
## Summary

- target the current upstream webview chunks for remote-control feature sync, settings visibility, host enablement, and active conversation status
- handle the current enablement bundle ordering where the log marker is declared after the request handler
- keep the bridge search bounded and add coverage for current, legacy, distant-anchor, and second-run cases

## Why

The previous filename patterns still matched older or unrelated chunks, so these descriptors were reported as `already-applied` even though their current target symbols had moved. On the current DMG, that left five `remote-mobile-control` transformations unapplied.

This keeps the fix within the repository's current-DMG policy and does not add a content-scan fallback. The feature remains disabled by default.

## Validation

- current DMG `26.707.41301`: all five descriptors apply on the first pass and report `already-applied` on the second
- extracted app tree hash is unchanged by the second pass
- `node --test $(rg --files -g '*.test.js')` (421 passed)
- `node --test linux-features/*/test.js` (526 passed)

Related to #639.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- Remote mobile control now patches the current upstream webview chunks for
+  feature sync, settings visibility, host enablement, and active conversation
+  status. The enablement bridge also accepts the current bundle ordering where
+  its log marker is declared after the request handler.
 - Automated user-local updates no longer inherit or set developer overrides
   that could replace a running Electron app or bypass DMG acceptance. Manual
   and timer rebuilds now fail safely at promotion, transactional installs retain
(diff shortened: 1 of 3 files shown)


Yo-DDV:
Hi @ilysenko, one useful follow-up from the current-DMG validation: the revoke setup-reset descriptor still matches the retargeted settings chunk, but its callback shape has changed, so it currently returns the source unchanged.

I have a separate fix ready locally. It also accounts for the Desktop's own client identity when deciding whether the last controller was removed. I will send that as a small follow-up after this PR lands, rather than expanding this one. #872 itself remains green and ready for review.

ilysenko (approved):
Verified against the current upstream DMG. The five descriptors now target the current chunks, apply idempotently, and the obsolete bundle and patched-output compatibility paths have been removed. All required checks are green.

PR #874, opened 2026-07-12
test(remote-mobile): cover staged artifact lifecycle

Yo-DDV:
## Summary

- verify that disabling `remote-mobile-control` exposes no staging entrypoint and enabling it registers the exact `stage.sh` entrypoint
- run the feature staging script twice and assert the exact markers, script contents, and executable mode
- preload a stale Desktop ownership marker and verify that a build without the matching app patch removes it

## Why

The existing stage tests covered a single successful pass and a clean missing-marker case. They did not prove manifest-to-staging registration, repeated staging, or stale ownership cleanup, even though all three affect updater and package rebuild confidence outside the patch descriptor report.

This is a test-only change. Runtime behavior is unchanged.

## Validation

- `node --test linux-features/remote-mobile-control/test.js` (83 passed)
- `node --test linux-features/*/test.js` (525 passed)
- built both default and `remote-mobile-control` Nix packages and verified artifact presence, absence, and modes

Related to #639.

### DIFF
--- a/linux-features/remote-mobile-control/test.js
+++ b/linux-features/remote-mobile-control/test.js
@@ -10,6 +10,7 @@ const test = require("node:test");
 const vm = require("node:vm");
 
 const {
+  enabledLinuxFeatureStageHooks,
   loadLinuxFeaturePatchDescriptors,
 } = require("../../scripts/lib/linux-features.js");
 const {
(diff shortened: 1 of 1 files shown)


ilysenko (approved):
Verified the test-only stage lifecycle coverage against the current feature manifest and stage hook. The assertions are focused, the effective merge is clean, and all required checks are green.

PR #875, opened 2026-07-12
Test Linux desktop settings rendering

colonelpanic8:
## Summary

- execute the generated Linux desktop settings and fallback toggle modules in the patcher test harness
- recursively render the generated settings tree and verify all four accessible switches
- exercise a switch click and assert that the updated value reaches the callback

## Why

The Linux desktop settings page recently broke when an inferred upstream toggle depended on private lazy initialization. Existing fixture tests verified generated source strings and imports, but did not prove that the generated page could render or that its switches behaved correctly. This adds a lightweight behavioral regression test for that failure class.

## Validation

- node --test scripts/patch-linux-window-ui.test.js (358 passed)
- current upstream DMG inspection: keybinds-settings patch applied
- required upstream patch-report validation passed
- Linux settings shell smoke assertions passed; the broader smoke run later hit an unrelated launcher watchdog timing threshold

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -1655,6 +1655,38 @@ function createModernNativeKeyboardShortcutsSettingsFixture() {
   return { extractedDir, assetsDir };
 }
 
+function evaluateGeneratedSettingsModule(source, bindings, exportExpression) {
+  const executable = source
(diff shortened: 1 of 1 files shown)


ilysenko (approved):
Focused regression coverage for the generated Linux settings runtime path. The behavioral harness and full patcher suite look good.

PR #879, opened 2026-07-12
refactor(remote-mobile): remove legacy visibility fallback

colonelpanic8:
## Summary

- remove the obsolete exact remote-control visibility needle/replacement fallback
- keep only the current generic visibility bundle matcher
- remove the legacy-shape test and exercise the current bundle shape in the integration fixture

## Context

Follow-up to the maintainer feedback on #876. The original PR merged before the review was submitted, so this isolates the requested current-DMG-only cleanup on top of the latest upstream `main`.

## Validation

- `node --test linux-features/remote-mobile-control/test.js` (81 passed)
- verified the current extracted `26.707.51957` visibility asset is matched, patched, and idempotent
- `nix build .#codex-desktop-remote-mobile-control .#checks.x86_64-linux.nix-linux-features-multi-feature --no-link -L` (both accepted with no blockers)

PR #882, opened 2026-07-12
refactor(remote-mobile): drop native workarounds

Yo-DDV:
## Summary

- remove the account-compatibility workaround now covered by the current upstream enrollment flow
- remove the selected-tab workaround now covered by the current eight-parameter tab resolver
- delete the obsolete fixtures and regression cases while keeping the remaining Linux workarounds

## Why

The current DMG already handles both account-scoped and authenticated user identities during enrollment. Its tab resolver also covers Control this computer, Control other devices, SSH, and WSL without the Linux fallback.

This removes only the duplicate transforms. The feature remains disabled by default, and the other Linux-specific behavior is unchanged.

## Validation

- current DMG `26.707.51957`: account identity and tab resolver cases exercised directly
- first and second feature passes produce the same extracted-app tree hash
- 80 targeted remote-mobile tests and 962 Node tests
- complete script smoke suite
- Nix package build for `codex-desktop-remote-mobile-control`
- Semgrep JavaScript scan with zero findings
- current `main` plus #879 with 79 remote-mobile tests

Related to #639.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Changed
 
+- Remote mobile control now relies on the current upstream account-enrollment
+  compatibility and Connections tab resolver instead of patching duplicate
+  Linux-specific fallbacks into those paths.
 - Remote notification hydration, replay, completed-item recovery, and remote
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Current upstream natively covers both removed compatibility paths. The cleanup follows the current-DMG-only policy and the complete CI matrix passes.

PR #883, opened 2026-07-12
fix(remote-mobile): select one runtime owner

Yo-DDV:
## Summary

- choose one remote-mobile runtime owner at cold start
- prefer an installed systemd user service over the Desktop app-server and standalone fallback
- validate a versioned Desktop owner marker and replace marker symlinks safely during staging
- log the selected owner and cover each precedence path with regression tests

## Why

The cold-start hook previously recognized systemd only while the unit was active. An installed unit that was restarting, disabled, or failed could therefore be bypassed by a second standalone daemon. The Desktop path also trusted the marker by existence alone.

This keeps the existing opt-in fallback behavior while making ownership deterministic and observable.

## Validation

- remote-mobile feature tests: 93 passed
- Node checks: 975 passed
- script smoke suite
- Bash syntax and ShellCheck
- Semgrep JavaScript rules: 0 findings

Part of #639.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- Remote mobile cold starts now select one runtime owner deterministically.
+  Explicit systemd user-service configuration takes precedence over the
+  Desktop app-server and standalone fallback, while a versioned Desktop marker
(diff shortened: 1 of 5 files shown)


ilysenko (approved):
Reviewed the current head; the runtime ownership change is focused, tested, and keeps the behavior inside the opt-in feature.

PR #884, opened 2026-07-12
docs(remote-mobile): define control topology ownership

Yo-DDV:
## Summary

- define the mobile-host, outbound-control, Remote SSH, and shared-boundary responsibilities inside `remote-mobile-control`
- map every current patch descriptor to exactly one primary responsibility
- document feature-owned hooks, the Nix service, nested Remote SSH patches, and their RPC boundaries
- keep the matrix synchronized with the descriptor registry through a regression test

## Why

The feature now spans three distinct control paths that share settings and host-routing code. Without an explicit ownership contract, a compatibility patch for one path can easily enable, query, or reset another path by accident.

The added test makes descriptor additions and ownership changes deliberate while leaving runtime behavior unchanged.

## Validation

- remote-mobile feature tests: 80 passed
- Node checks: 962 passed
- script smoke suite

Part of #639.

### DIFF
--- a/linux-features/remote-mobile-control/README.md
+++ b/linux-features/remote-mobile-control/README.md
@@ -64,6 +64,75 @@ What it changes:
   when it is missing, then starts the managed app-server daemon with
   `remote-control start`.
 
+## Control topology boundaries
+
+This feature touches three different control paths. They must stay independent:
+
+- `mobile-host`: a mobile client controls this Linux installation. This owns the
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head; the topology ownership documentation is accurate, descriptor coverage is enforced by tests, and the change adds no runtime compatibility fallback.

PR #885, opened 2026-07-12
fix(remote-mobile): harden device key store

Yo-DDV:
## Summary

- serialize device-key updates with a per-store `flock` lock
- publish writes through an exclusive temporary file, `fsync`, and atomic rename
- reject unsafe paths, file types, ownership, permissions, malformed records, and oversized stores
- migrate the existing unversioned store on the next successful write without discarding valid keys

## Why

The Linux provider previously used an unlocked read-modify-write sequence and treated malformed JSON as an empty store. Concurrent enrollment or revocation could lose keys, while corruption could be silently replaced.

This keeps the file-backed provider experimental, but makes its persistence behavior bounded and fail-closed.

## Validation

- remote-mobile feature tests: 87 passed
- Node checks: 969 passed
- script smoke suite
- Semgrep changed-files baseline: 0 new findings
- extracted DMG main bundle: patch applied, syntax passed, idempotence passed

Part of #639.


ilysenko (changes requested):
Thanks for hardening the device-key store. I found two blockers:

- Please make the lock helper work on NixOS. The provider currently checks only `/usr/bin/flock` and `/bin/flock` and invokes `/bin/sh`, but pure NixOS exposes these tools through Nix or system-profile paths. Please use a safely resolved/staged runtime path and add a test where `flock` is unavailable under `/usr/bin` and `/bin`.
- Please resolve the current `main` conflicts while preserving the remote-mobile topology documentation and descriptor coverage added by #884.

Yo-DDV:
Thanks, that makes sense.

I pushed another pass for this. The lock helper now resolves both `flock` and `sh` from PATH first, with the Nix/system-profile locations in the search list, instead of assuming `/usr/bin/flock`, `/bin/flock`, or `/bin/sh`.

I also added a regression test that hides those fallbacks and only exposes wrapper helpers through PATH, then verifies the spawned commands use those resolved paths. The branch is rebased on current `main` as well, so the topology docs and descriptor coverage from #884 stay in place.

Local checks are green on my side, including the remote-mobile feature tests, patcher tests, script smoke, and the remote-mobile Nix dry-run. I’ll watch the new CI run now.

ilysenko (approved):
Re-reviewed the current head. The previous NixOS helper-path and main-conflict blockers are addressed, focused tests pass, and all required checks are green.

PR #887, opened 2026-07-12
fix(remote-mobile): classify log identifiers as sensitive

Yo-DDV:
## Summary

- classify remote-mobile conversation, turn, and host identifiers as sensitive log fields
- keep aggregate counters, retry attempts, and boolean state in safe diagnostics
- add regression coverage for the generated hydration and auto-connect payloads

## Why

The Linux compatibility paths currently place opaque identifiers in safe log metadata. Comparable diagnostics already treat this context as sensitive. This keeps the useful operational fields available while preventing conversation, turn, and remote host identifiers from being handled as ordinary diagnostic values.

## Validation

- remote-mobile feature tests: 86 passed
- full Node test suite: 968 passed
- script smoke suite: passed
- `nix build .#codex-desktop-remote-mobile-control --no-link`
- extracted the official 26.707.51957 package ASAR and verified the active auto-connect payload

Part of #639.


ilysenko (approved):
Reviewed the current head: the sensitive-field classification is correct, focused tests pass, and all required checks are green.

PR #888, opened 2026-07-12
fix(remote-mobile): validate daemon pid identity

Yo-DDV:
## Summary

- validate daemon PID records against their owner, start time, and managed Codex executable
- bound PID-file parsing and reject invalid process identifiers
- remove stale records without signaling unrelated processes, while preserving incomplete or unverifiable reservations

## Why

The Desktop-owner cleanup currently relies on `kill -0`. A recycled PID can therefore make stale daemon state look active even when it belongs to another process. This change verifies the record and live process identity before deciding whether the PID file is still authoritative.

## Validation

- remote-mobile feature tests: 92 passed
- full Node test suite: 974 passed
- script smoke suite: passed
- `bash -n` and ShellCheck
- `nix build .#codex-desktop-remote-mobile-control --no-link`
- reproduced the PID record and process identity with Codex 0.144.1 in an isolated `CODEX_HOME`

Part of #639.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   Explicit systemd user-service configuration takes precedence over the
   Desktop app-server and standalone fallback, while a versioned Desktop marker
   prevents stale or forged marker content from suppressing the fallback.
+- Remote mobile cleanup now validates daemon PID records against the process
+  owner, recorded start time, and managed standalone executable before
+  preserving them. Recycled or foreign PIDs no longer make stale state appear
+  live, while incomplete in-flight reservations are left untouched.
 - Remote mobile control now patches the current upstream webview chunks for
   feature sync, settings visibility, host enablement, and active conversation
   status. Revoking the final controller now also clears the current mobile setup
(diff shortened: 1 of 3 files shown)


ilysenko:
I’m not sure—there are too many makeshift patches needed to fix `control`. I would skip continuing the CLI branches altogether. I’d handle everything the way it works in the original version, because these kinds of changes tend to drag in future modifications and so on. I’d keep the `control` fixes as close as possible to how they work in the original version, without adding any enhancements.

ilysenko (changes requested):
Thanks for working on this. The PID reuse concern is valid, but this duplicates the upstream daemon’s PID handling without following its locking protocol.

Two blockers need to be addressed:

- The hook validates a record and later removes the PID file without acquiring the upstream `.pid.lock` or rechecking the record. The daemon can publish a new valid record between those operations, which the hook would then remove.
- Empty or incomplete PID files are always preserved, even when no process holds the corresponding lock. A crash can therefore leave stale state permanently.

More broadly, this project should remain as close to upstream behavior as possible. Adding custom process-management logic for an experimental, lightly used feature creates disproportionate maintenance cost. These workarounds accumulate, interact, and produce further edge cases that require even more patches. This is becoming too complex to support reliably.

Please prefer the official daemon lifecycle and remove this custom cleanup unless it is strictly necessary. If separate Linux handling is unavoidable, it must follow the complete upstream lock-and-recheck protocol and cover concurrent replacement and abandoned reservations with tests.

Before requesting another review, please verify the update against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant tests and an independent local-LLM review of the full diff, and wait for all required CI checks.

Yo-DDV:
Thanks, agreed. I’m closing this one rather than trying to reshape it.

The PID reuse concern was real, but this branch fixes it by adding more custom lifecycle around `control`, which goes against the direction you laid out. Better to stop here and keep the next remote-control work closer to the upstream daemon behavior instead of layering another workaround.

PR #890, opened 2026-07-12
fix(nix): reject store-backed remote control environment files

Yo-DDV:
## Summary

- Treat `remoteControl.environmentFile` as a runtime string in both Nix modules.
- Reject direct and optional store-backed paths.
- Add Home Manager and NixOS evaluation coverage and document the safe syntax.

## Why

The option was typed as a Nix path. A path literal could therefore copy an
environment file into the world-readable Nix store before the generated user
service referenced it.

Quoted runtime paths such as `/run/secrets/codex-remote-control.env` keep the
file outside the store. The module now rejects path literals and strings that
resolve into the store, including systemd's optional `-` form.

## Testing

- `nix-instantiate --parse` for both modules and the evaluation test
- `nix build --no-link .#checks.x86_64-linux.nix-linux-features-evaluation`
- `nix build --no-link .#checks.x86_64-linux.nix-linux-features-multi-feature`

### DIFF
--- a/docs/nix.md
+++ b/docs/nix.md
@@ -144,6 +144,14 @@ in
 }
 ```
 
+Set `remoteControl.environmentFile` to a quoted absolute runtime path such as
+`"/run/secrets/codex-remote-control.env"`. Prefix it with `-` only when systemd
+should ignore a missing file. Empty paths, relative paths, Nix path literals and
+strings that resolve inside the Nix store are rejected. Do not interpolate a
+path containing secrets: Nix can copy it into the store before module validation
+rejects the configuration. The referenced runtime file must be readable by the
+user service and should remain owner-only.
+
 Pinning `github:sadjow/codex-cli-nix` to a release tag or commit is
 recommended for fully reproducible configurations.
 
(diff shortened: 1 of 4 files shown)


ilysenko:

- After changing `remoteControl.environmentFile` to a string, please validate that the value after the optional `-` prefix is an absolute path, as required by systemd. The current assertion rejects Nix-store paths but still accepts values such as `secrets.env` or an empty string, which produces an invalid service at runtime; please cover both Home Manager and NixOS.

Yo-DDV:
Thanks, good catch. I updated both modules to validate the path after the optional `-` prefix and require it to be absolute.

The tests now cover empty, relative, optional-relative, and valid optional absolute paths for Home Manager and NixOS. I also rebased on current `main`; all five CI checks are green on `594e7329`.

ilysenko (approved):
Reviewed the current head after hardening the environment-file validation. The Nix-context and non-canonical store-path bypasses are covered for both Home Manager and NixOS, and all required checks pass.

PR #895, opened 2026-07-12
fix(linux): preserve actions in approval notifications

### LINKED ISSUE #893: Linux approval notifications drop upstream action buttons
## Summary

Approval notifications produced by a **Custom** approval policy configured
through `config.toml` show only `View` on Linux. Responding to an approval still
requires focusing the Codex Desktop window.

Codex Desktop already constructs this action payload:

```js
[
  {
    id: "approve",
    title: "Approve",
    actionType: "approve",
  },
  {
    id: "approve-session",
    title: "Approve for session",
    actionType: "approve-for-session",
  },
  {
    id: "decline",
    title: "Decline",
    actionType: "decline",
  },
]
```

The existing callback maps these actions to `accept`, `acceptForSession`, and
`decline`. Electron receives the payload, but its Linux notification backend
does not expose the buttons or their action events.

I initially considered implementing this as an opt-in Linux feature. Finding
that the payload and callbacks already exist changed the scope to restoring
existing behavior as a Linux compatibility fix.

## Expected behavior

The notification displays the existing actions, allowing the request to be
approved or declined directly from the notification without focusing the Codex
Desktop window.

## Actual behavior

The notification contains only `View`.

## Reproduction

1. Use a **Custom** approval policy configured through `config.toml`.
2. Start a task that requires command or file-change approval.
3. Switch away from the Codex Desktop window.
4. Wait for the approval notification.

The built-in permissions `Ask for approval` flow does not currently emit a
desktop notification and is outside the scope of this issue.

assembledev:
## Summary

Closes #893.

- route action-bearing Linux notifications through the standard freedesktop
  notification interface
- forward the selected action index to the existing Codex notification callback
- retain the current Electron notification as a fail-soft fallback when actions
  are unavailable
- stage the native bridge for source installs, native packages, AppImage,
  updater rebuilds, and Nix builds

The bridge does not execute commands or make approval decisions. It only
returns the selected notification action to the existing upstream handler.

This is default-on Linux compatibility behavior and does not add an opt-in
Linux feature.

## Scope

This change affects approval notifications that already contain upstream
notification actions, such as command and file-change approvals produced by a
**Custom** approval policy configured through `config.toml`.

The built-in permissions `Ask for approval` flow does not currently emit a
desktop notification and is outside the scope of this change.

## Validation

- `cargo test -p codex-notification-actions-linux`
- `cargo clippy -p codex-notification-actions-linux -- -D warnings`
- `node --test scripts/patches/impl/main-process/notifications.test.js`
- `node --test scripts/patch-linux-window-ui.test.js` — 366 passed
- `nix flake check path:$PWD --no-build`
- built the Nix `notification-actions-linux` check
- verified patch idempotence against the current upstream main-process bundle
- verified a real **Custom** approval policy notification with action buttons on
  Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- Approval notifications now preserve the upstream Approve, Approve for
+  session, and Decline actions on Linux. A small freedesktop notification
+  bridge forwards the action and close signals that Electron's Linux
+  notification backend does not expose, with the existing View-only Electron
+  notification retained as a fail-soft fallback.
 - Remote mobile cold starts now select one runtime owner deterministically.
   Explicit systemd user-service configuration takes precedence over the
   Desktop app-server and standalone fallback, while a versioned Desktop marker
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -517,6 +517,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "codex-notification-actions-linux"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "futures-util",
+ "serde",
+ "serde_json",
+ "tokio",
+ "zbus",
+]
+
+[[package]]
 name = "codex-read-aloud-linux"
 version = "0.1.0-linux-alpha1"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,9 @@
 [workspace]
-members = ["computer-use-linux", "read-aloud-linux", "updater", "record-replay-linux"]
+members = [
+  "computer-use-linux",
+  "notification-actions-linux",
+  "read-aloud-linux",
+  "updater",
+  "record-replay-linux",
+]
 resolver = "2"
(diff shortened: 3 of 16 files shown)


ilysenko (changes requested):
Thanks for the well-isolated, fail-soft implementation. There is one lifecycle issue to address before merge: in bridge mode, `close()` sets `closed = true` before the bridge emits its `closed` event, so `emitClose()` suppresses the registered close callback. The wrapped Electron notification preserves that event. Please preserve the close-event contract and add a regression test covering programmatic `close()`.

ilysenko (changes requested):
Thanks for fixing the original close-event suppression and adding the regression test. Two blockers remain:

1. `closeRequested` is not respected by `show()`, `startFallback()`, or the pre-shown failure paths. Closing while the bridge is starting can therefore show the Electron fallback afterward. Please prevent fallback/show after close is requested and add a close-before-shown regression test.
2. The Nix installer does not export the prebuilt `CODEX_NOTIFICATION_ACTIONS_SOURCE` and has no Cargo runtime input, so `nix run .#installer` silently installs without the bridge. Please stage the prebuilt bridge in that path as well.

ilysenko (approved):
Reviewed the current head after resolving the notification lifecycle, close/fallback race, stdout-drain, and Nix installer staging issues. Regression coverage is in place and all required checks pass.

PR #897, opened 2026-07-12
Fix Linux tray recovery after session resume

### LINKED ISSUE #891: Tray Quit can stop responding after prolonged GNOME screen lock on Ubuntu X11
## Summary

After the Ubuntu/GNOME screen has been locked for some time and then unlocked, selecting **Quit** from the Codex tray menu can leave the tray icon and the complete Codex Desktop process tree running. The app then has to be terminated externally.

This has occurred repeatedly on one machine, but the lock-duration trigger still needs independent confirmation.

## Environment

- OS: Ubuntu 24.04.4 LTS
- Desktop: Ubuntu GNOME
- Session: X11
- Kernel: 6.8.0-134-generic
- Package format: deb
- codex-desktop-linux: 0.9.5 at `2e04d928babaf09b321c870f4d8b9befd28921a5`
- Installed package version: `2026.07.12.122632`
- Electron: 42.1.0
- Upstream Codex app: 26.707.51957
- Optional Linux features: none

## Steps observed

1. Start Codex Desktop normally.
2. Lock the GNOME desktop.
3. Leave it locked for an extended period.
4. Unlock and return to the existing desktop session.
5. Open the Codex tray menu and select **Quit**.

## Actual behavior

The tray icon remains and the launcher, Electron main process, app-server, renderer, and webview server remain alive. Repeated Quit selections do not reliably terminate the app.

## Expected behavior

The explicit tray Quit action should terminate Codex Desktop within the bounded quit-drain timeout and remove the tray icon.

## Existing patch status

The installed patch report says all relevant required patches were applied:

- `linux-quit-guard`
- `linux-explicit-quit-prompt-bypass`
- `linux-explicit-quit-drain-timeout`
- `linux-explicit-tray-quit`
- `linux-explicit-ipc-quit`

The focused quit/tray test selection passes locally (24 tests). The existing 3-second timeout only helps after the tray callback reaches the Electron quit lifecycle. One possibility is that the native context-menu callback becomes stale across GNOME lock/unlock.

## Runtime observations

Around affected desktop lifecycle periods, `~/.cache/codex-desktop/launcher.log` contains messages such as:

```text
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 destroyed listeners added to [WebContents]
Error occurred in handler for 'codex_desktop:message-from-view': TypeError: Object has been destroyed
CompositorAnimationObserver is active for too long
```

There are currently no explicit markers for tray Quit click, `before-quit`, or `will-quit`, so the exact stage where the action stops is not observable.

## Possible direction

- Log tray Quit, `before-quit`, and `will-quit` transitions.
- Listen for Electron `powerMonitor` `unlock-screen` / `resume` events.
- Refresh or recreate the Linux tray context menu after session resume.
- Add a GNOME lock/unlock manual acceptance case.

joshyorko:
## Summary

Fixes #891.

- refresh the Linux tray context menu after Electron `unlock-screen` and `resume` events;
- keep one stable recovery handler retargeted to the current tray controller;
- register recovery lazily after the first tray controller exists and remove it on `will-quit`;
- add EventEmitter-backed behavioral coverage for repeated controller creation, a destroyed old tray, session recovery, and quit cleanup.

## Root cause

The Linux tray menu was built only during tray setup and ordinary tray-state updates. After a prolonged GNOME/X11 lock cycle, the native menu could retain stale callbacks. Rebuilding the context menu after session recovery restores current callbacks.

Recovery handlers must not capture a specific controller: tray reconstruction would leave an old handler targeting a destroyed tray. The patch therefore stores only the current controller and installs one stable listener pair.

## Validation

- TDD regression failed before the implementation (`actual: 1`, `expected: 3`) and passed after it.
- Tightened lifecycle regression failed with eager registration (`actual: 1`, `expected: 0`) and passed after lazy registration.
- `node --test --test-name-pattern='tray' scripts/patch-linux-window-ui.test.js`
- `node --check scripts/patches/impl/main-process/tray.js`
- tray-specific `test_linux_tray_patch_smoke`
- `git diff --check`

`bash tests/scripts_smoke.sh` reaches the pre-existing marketplace-metadata symlink replacement failure. The same failure reproduces from a clean detached `upstream/main` worktree at `eb057a00`; the tray-specific smoke test passes.

## Runtime scope

The change is source-tested against the current main-process bundle fixture. A prolonged GNOME/X11 lock/unlock acceptance run remains required for installed-runtime proof.


ilysenko (changes requested):
Thanks for working on this fix. One fail-soft blocker remains: recovery-helper insertion and the constructor call are applied independently. If the tray fields still match but the class declaration shape drifts, the patch can insert `codexLinuxSetTrayController(this)` without defining that helper, causing a Linux runtime `ReferenceError`.

Please gate the constructor rewrite on confirmed helper installation, preserve the existing context-menu setup with a warning when helper insertion fails, and add a regression test for this partial-drift case.

Please submit a pull request as a finished, independently reviewed change rather than an initial implementation idea. Repeatedly discovering issues during maintainer pre-review takes substantial time. Before requesting another review, please:

- verify the complete change against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md);
- run all relevant tests locally;
- have your local LLM perform an independent code review of the complete diff and address its findings;
- wait until every required CI check has completed successfully after the final update.

This will help keep the review process focused on validating a merge-ready change instead of iterating through issues that can be caught before submission.

ilysenko (approved):
Re-reviewed the current head after the fail-soft and current-DMG matcher fixes. The helper/call insertion is atomic, partial drift remains safe, current upstream DMG acceptance passes, and the full cross-format CI matrix is green.

joshyorko:
> Re-reviewed the current head after the fail-soft and current-DMG matcher fixes. The helper/call insertion is atomic, partial drift remains safe, current upstream DMG acceptance passes, and the full cross-format CI matrix is green.

Thanks for cleaning 🧹 this up.  I sometimes can get them dropped and can't follow up for a few hours due to family constraints so appreciate it.

PR #899, opened 2026-07-12
Fix Linux desktop settings React runtime

colonelpanic8:
## What changed

- export the initialized React and JSX runtimes from the native keyboard-shortcuts settings chunk
- make the generated Linux desktop settings page and fallback settings components reuse those initialized runtimes
- add regression coverage for the generated bridge imports and exports

## Why

The generated Linux desktop settings chunk invoked the raw React factory independently from the native settings chunk. With the current upstream bundle, that produced a separate/uninitialized hook dispatcher and crashed the page with `Cannot read properties of null (reading 'useState')`.

Reusing the native chunk's initialized runtime keeps the Linux page in the same React module instance as the rest of Settings.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` (359 passing)
- `git diff --check`
- `nix build .#codex-desktop-computer-use-ui-remote-mobile-control --no-write-lock-file -L`
- rendered the exact generated `linux-desktop-settings-linux.js` from the built package in a live browser runtime; all four Linux settings controls and Build information rendered without a React runtime error

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -4914,7 +4914,7 @@ test("renders the generated Linux desktop settings page with working switches", 
     const React = { Component, Fragment: "fragment" };
     const Toggle = evaluateGeneratedSettingsModule(
       fs.readFileSync(path.join(assetsDir, "linux-settings-toggle-linux.js"), "utf8"),
-      { __jsxFactory: () => jsxRuntime },
+      { $: jsxRuntime, __jsxFactory: () => jsxRuntime },
       "t",
     );
     const SettingsPage = ({ title, subtitle, children }) =>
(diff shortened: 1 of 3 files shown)


ilysenko (changes requested):
Thanks for fixing the current settings runtime crash.

One blocker remains: please remove the direct React/JSX factory fallback used when the initialized native runtime cannot be inferred. That path recreates the separate, uninitialized runtime and conflicts with the repository’s current-DMG-only policy. Missing runtime inference should be treated as optional drift: warn and report it, then skip the Linux settings patch without writing generated assets.

Please read and verify the complete PR against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), especially the requirement to remove obsolete upstream-DMG compatibility paths when fixing drift.

Before pushing the update, run the relevant tests locally and perform an independent review of the complete diff with a local LLM. After updating the PR, wait for every required CI check to pass before requesting another review or merge.

ilysenko (approved):
Reviewed the current head. The runtime fallback blocker is fixed, current-DMG drift handling is fail-soft, and the relevant checks pass.

PR #900, opened 2026-07-12
Fix Niri pet drag move queue

### LINKED ISSUE #856: Niri pet drag queues stale compositor moves
## Problem

Dragging the Pet overlay on Niri feels dramatically over-sensitive and can keep moving in the previous direction after the pointer reverses. Fast direction changes produce overshoot and delayed correction instead of tracking the current pointer target.

Reproduction:

1. Enable `pet-overlay` with Niri support and `lockPosition: false`.
2. Start dragging the Pet horizontally.
3. Move quickly in one direction, then reverse.

Expected: the overlay follows the latest pointer target with bounded latency and the released position remains stable.

Actual: queued compositor moves continue to complete after newer pointer targets, causing jumps, lag, and delayed reversal.

## Root Cause Analysis

The Niri bootstrap positioning path is also used at pointer-event rate during drag. Every desired-bounds update schedules multiple discovery and move subprocesses. Timer cancellation removes callbacks that have not started, but it cannot cancel Niri discovery or action processes already in flight. There is no single-flight guard, drag generation, or latest-target queue, so older absolute moves can complete after newer ones.

The output-local coordinate conversion itself is correct. Upstream orb physics may add intentional elasticity for some pets, but it is separate from this Niri transport regression.

## TDD Fix Plan

1. **RED**: Simulate drag targets `100 → 600 → 120` with asynchronous Niri actions and assert that at most one move is in flight and the next emitted target is `120`, never stale `600`.
   **GREEN**: Split pointer-rate drag movement from bootstrap retries and add a single-flight, latest-only move queue.

2. **RED**: Start drag while the matched Pet is still tiled and assert it becomes floating before its first move.
   **GREEN**: Serialize the initial float transition ahead of the drag move.

3. **RED**: End drag while the final move is pending and assert the confirmed final bounds are persisted once, with no later snap-back.
   **GREEN**: Drain and persist the final target after action completion.

4. **RED**: Replace the Pet window or Niri window id while discovery/action callbacks are pending and assert stale callbacks cannot move another window.
   **GREEN**: Scope validation and cached ids to one drag generation and window identity.

5. **RED**: Fail discovery/action and remove the Niri executable; assert state is cleaned up and recovery is bounded.
   **GREEN**: Make completion callbacks total and add bounded rediscovery without an infinite retry loop.

6. **RED**: Apply the updated patch to an already #845-patched bundle and assert the new drag pipeline is installed exactly once.
   **GREEN**: Version the injected method block and include the new marker in completeness checks.

**REFACTOR**: Keep bootstrap placement retries separate from the live drag transport and leave upstream orb physics unchanged.

## Acceptance Criteria

- [ ] Niri drag has at most one compositor move action in flight.
- [ ] Intermediate pointer targets are replaceable; the latest target always wins.
- [ ] A tiled Pet is floated before the first drag move.
- [ ] The final released position is persisted after the queue drains.
- [ ] Stale window ids and callbacks cannot move a replacement or foreign window.
- [ ] Discovery/action failures recover without a stuck or unbounded queue.
- [ ] Existing #845-patched bundles upgrade to the new pipeline.
- [ ] All new and existing Pet tests pass.

ilysenko:
## Summary

- serialize live Niri Pet drag discovery, float, and move actions
- keep only the latest queued pointer target and persist once after the final move drains
- document the Niri drag behavior and add asynchronous regression coverage

## Root cause

The bootstrap retry path also ran at pointer-event rate. Clearing timers could not cancel already-running `niri msg` processes, so stale absolute moves could finish after newer pointer targets.

## Tests

- `node --test linux-features/pet-overlay/test.js`
- `git diff --check`
- current generated `app.asar` patch/idempotence probe

`./scripts/ci-local.sh pr` completed Rust checks and tests but its Ubuntu 24.04 container stopped at the pre-existing `node --check computer-use-linux/gnome-shell-extension/extension.js` ESM parse failure under distro Node 18. The same full node check passes locally with Node 22.

## Notes

A live Niri smoke test was not available in the current GNOME session.

Fixes #856.


Yo-DDV (dismissed):
Thanks for asking me to take a look. I found one remaining race in the handoff between Niri drag generations.

`codexPetOverlayBeginNiriDrag()` replaces the current drag state, but an `execFile` move already dispatched by the previous drag can still be running. A second drag on the same window can then discover the same Niri id and dispatch its new move before the old process completes. In a completion-order test, the new move to `x=300` completed first, then the old move to `x=100` completed last and became the final compositor position. The generation checks correctly suppress stale callbacks, but they cannot undo the side effect of a command that has already been sent.

The same overlap is possible when a bootstrap positioning command is already in flight as a drag starts. Could you keep the single-flight serialization across drag generations and the bootstrap-to-drag handoff, then add a regression test that completes the newer move before the older one?

I reproduced both paths on `a8d8dddb`. The existing local suites and all five CI jobs are green, so this ordering case is the only blocker I found on this head.

joshyorko (dismissed):
I found one remaining replacement-window state leak on `c1fa46b1`.

After a Niri drag's discovery and move processes have both completed, the drag state correctly remains active until `endDrag`. If the overlay window closes or is replaced during this idle gap, however, there is no pending callback or retry timer left to call `codexPetOverlayNiriDragCurrent()`. The replacement window then reaches `codexPetOverlayScheduleNiriHints()`, which returns because `codexPetOverlayNiriDragState` is still non-null, so its bootstrap Niri hints remain suppressed indefinitely.

I reproduced this sequence deterministically:

1. Begin the drag.
2. Complete discovery and the first move so no Niri process or retry remains.
3. Replace the overlay window without calling `endDrag`.
4. Schedule hints for the replacement.

The resulting timer list is `[]`; the expected bootstrap delays are `[0, 80, 300, 1000]`.

The new callback-cleanup regression covers replacement while an old subprocess is still pending, but not this idle state. Please clear or validate stale drag state during scheduling/window teardown and add coverage for the no-pending-callback case.

PR #901, opened 2026-07-12
Link tray build info to source commit

colonelpanic8:
## Summary

- replace the native tray build-info message box with a small secure modal that renders the source commit hash as a GitHub link
- capture and preserve the source commit subject in build metadata
- show the commit message in both the tray modal and Linux desktop settings

## Why

Electron's native message box cannot make the displayed hash itself clickable. A sandboxed HTML modal allows the hash to be presented as a real link while the main process validates and handles navigation.

## Validation

- focused build-info tests
- `bash -n scripts/lib/package-common.sh tests/scripts_smoke.sh`
- packaging smoke coverage passed for Debian, RPM, pacman, AppImage, and source-metadata propagation; the broader smoke run later stopped at the unrelated sudo-alert sound test because the expected system sound file was unavailable


ilysenko (changes requested):
Thanks for the work. Two blockers remain:

1. This is optional Linux UX and must not remain in core. Further development will be accepted only as an opt-in `linux-features/` module disabled by default. Move the entire custom dialog and everything related to it—commit-link and subject handling, tray/settings integration, hooks, and tests—into the feature. Core may contain only the smallest generic extension point.

2. Nix builds have no `.git` metadata, so the current implementation cannot provide the commit URL or subject. Define reproducible Nix behavior and cover it with a regression test.

Verify the update against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant tests locally, review the complete diff with a local LLM, and wait for all required CI checks before requesting review.

PR #913, opened 2026-07-12
Fix Linux desktop settings React runtime

colonelpanic8:
## Summary

- bridge React and JSX from the active settings route module instead of the separately initialized keyboard-shortcuts chunk
- compose runtime-export and route patches transactionally when they target the same asset
- update fixtures and regression coverage to require the route-owned runtime

## Verification

- `node --test linux-features/*/test.js` (550 passed)
- `nix build .#codex-desktop-computer-use-ui-remote-mobile-control --no-write-lock-file -L`
- launched the built app side-by-side with app id `codex-settings-fixed`, an isolated Electron user-data directory, and webview port 5176
- used Linux Computer Use accessibility actions to open File -> Settings; the Linux desktop settings asset and all support assets returned HTTP 200
- verified the isolated launcher log after the route load contains no `error boundary`, `useState`, or `Cannot read properties` failure

The installed pre-fix build reproducibly logs `Cannot read properties of null (reading 'useState')` at `LinuxToggle`; the isolated fixed build does not.


ilysenko (changes requested):
Thanks for fixing the settings React runtime ownership and asset patch composition. The implementation looks sound, but the required smoke suite now fails. Please update the Linux desktop settings smoke fixture so the active route asset contains the inferred React/JSX runtime, and update the expected generated import from `keyboard-shortcuts-settings-test.js` to the route asset. Then rerun the full smoke suite.

ilysenko (changes requested):
Thanks for fixing the settings React runtime and updating the smoke fixture. The settings fix now looks correct, but this PR still includes an independent Nix source-metadata change: `flakeSourceRemote`, `CODEX_LINUX_SOURCE_REMOTE`, and its smoke assertion.

Please remove that Nix change from this PR. It is independently releasable, unrelated to the settings runtime bug, and overlaps the source-link work tracked in PR #901. Handle it separately and sequentially after the earlier pull request is merged or closed.

Please verify the cleaned diff against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), rerun the relevant local tests, review the complete diff with a local LLM, and wait for all required CI checks to pass before requesting review again.

ilysenko (approved):

PR #904, opened 2026-07-12
Extend frameless-titlebar to Quick Chat

edshamis:
## What changed

- extend the opt-in frameless-titlebar feature to Quick Chat windows
- remove Linux titlebar-overlay creation and refresh calls for both primary and Quick Chat windows while the feature is enabled
- update feature tests, documentation, and changelog

## Why

Quick Chat windows retained Electron-drawn minimize, maximize, and close controls even when frameless-titlebar was enabled, unlike the primary window. This made compositor-managed decoration behavior inconsistent between the two window types.

## Impact

When frameless-titlebar is enabled, primary and Quick Chat windows both rely on compositor-managed controls. The feature remains disabled by default. Windows and macOS behavior is preserved.

## Validation

- node --test linux-features/frameless-titlebar/test.js (6 passed)
- node --test scripts/patch-linux-window-ui.test.js (368 passed)
- independent read-only local LLM review of the complete diff against origin/main
- manual NixOS/Hyprland check: Quick Chat has no Electron controls; conversation scrolling works

## Out of scope

Quick Chat content zoom was removed from this PR and will not be submitted until this PR is merged or closed, per maintainer guidance.

PR #914, opened 2026-07-12
Apply zoom to popped-out Quick Chat

edshamis:
## What changed

- apply the shared window zoom viewport to popped-out Quick Chat roots
- inventory every structurally discovered Quick Chat root and patch the asset atomically
- validate the exact per-root CSS-module binding before inserting `zoomedViewport`
- add idempotence, multi-root, partial-drift, alias-boundary, and contract-drift regression coverage

## Why

Popped-out Quick Chat received Zoom In, Zoom Out, and Actual Size commands, but its content did not scale. The route stored the shared `--codex-window-zoom` value while its root retained fixed `h-dvh w-full` sizing instead of consuming the existing `zoomedViewport` CSS module.

## Impact

Quick Chat content now follows the same zoom value as the rest of the app while preserving its conversation scroll container. The patch is Linux-only, fail-soft on upstream drift, and leaves the asset unchanged if any discovered Quick Chat root or style contract is unrecognized.

## Validation

- `node --test --test-name-pattern='Quick Chat|zoom contract' scripts/patch-linux-window-ui.test.js` (14 passed)
- `node --test scripts/patch-linux-window-ui.test.js` (382 passed)
- `node --test linux-features/frameless-titlebar/test.js` (6 passed)
- `git diff --check`
- applied idempotently with no warnings to the current installed Quick Chat asset
- independent read-only local LLM review of the final implementation: no findings
- manual NixOS/Hyprland validation: Zoom In/Out/Actual Size scale Quick Chat and conversation scrolling remains functional

## Scope

This is the content-zoom follow-up requested after the separate `frameless-titlebar` Quick Chat change merged in #904.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 - The opt-in `frameless-titlebar` feature now removes Electron-drawn window
   controls from Quick Chat as well as the primary window, keeping compositor-
   managed decoration behavior consistent across both window types.
+- Popped-out Quick Chat roots now consume the shared window zoom value, so Zoom
+  In, Zoom Out, and Actual Size scale content while preserving conversation
+  scrolling instead of updating only hidden state and titlebar geometry.
 - Remote mobile cold starts now select one runtime owner deterministically.
   Explicit systemd user-service configuration takes precedence over the
   Desktop app-server and standalone fallback, while a versioned Desktop marker
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -171,6 +171,7 @@ const {
   applyLinuxFastModeModelGuardPatch,
   applyLinuxI18nGatePatch,
   applyLinuxOpaqueWindowsDefaultPatch,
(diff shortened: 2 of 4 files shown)


ilysenko (changes requested):
Thank you for the thoughtful implementation and the extensive test coverage. However, we do not want to add this behavior to the default core patch set.

This changes the standard Quick Chat behavior for every user and introduces another dependency on the structure of the minified upstream webview bundle. Even with fail-soft handling and good tests, it creates ongoing maintenance work whenever the upstream DMG changes. The project already has a large review and maintenance load, and upstream updates regularly break these patch points, so we need to keep core focused on compatibility required for the app to work for most Linux users.

If you would like to keep this behavior for your own installation, the repository provides a dedicated place for private features under `linux-features/local/<feature-id>/`. Local features are gitignored, can use the same patch descriptor framework, and can be enabled through `linux-features/features.json`.

For these reasons, we are not planning to merge this customization into core. Thank you for understanding.

edshamis:
Understood—thank you for reviewing it and for clarifying the project boundary. I will keep this customization out of the default core patch set and move it to my private build path as a local opt-in feature. Closing this PR.

PR #916, opened 2026-07-12
Preserve source remote in Nix builds

colonelpanic8:
## Summary

- pass the canonical repository remote into Nix package builds through `CODEX_LINUX_SOURCE_REMOTE`
- add smoke coverage for the Nix environment wiring

## Why

Nix clean-source builds retain the source revision but do not have Git repository metadata available at build time. Without an explicit remote, build info cannot reliably derive a clickable GitHub commit URL.

## Validation

- `nix flake check --no-build --no-write-lock-file` (passed)
- `bash tests/scripts_smoke.sh` (all package and source-metadata coverage passed; the suite later stopped at the pre-existing environment-dependent sudo sound assertion: `Expected alert before authentication`)
- `git diff --check`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -53,6 +53,7 @@
           ];
         };
         flakeSourceCommit = self.rev or (self.dirtyRev or "");
+        flakeSourceRemote = "https://github.com/ilysenko/codex-desktop-linux.git";
         flakeSourceDateEpoch = toString (self.lastModified or 1);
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head. The focused Nix provenance fix is correct, covered by smoke validation, and all required checks pass.

PR #920, opened 2026-07-13
Fix Remote Mobile patch targeting for current app bundle

Yo-DDV:
## Summary

- retarget the seven Remote Mobile runtime patches to the current `iufn7mg3` webview chunk
- use patch-specific completion markers for visibility and Linux copy transforms
- cover unrelated Linux user-agent checks, incremental upstream copy/gates, and byte-identical repeat application

## Verification

- `node --test linux-features/remote-mobile-control/test.js` (100 passed)
- `node --test linux-features/*/test.js` (557 passed)
- exact 26.707.51957 DMG build: 18 Remote Mobile patches applied, then 18 already applied on a byte-identical second pass
- Semgrep baseline scan: no introduced findings

Addresses the current-DMG patch-report regression reported in #639.

### DIFF
--- a/linux-features/remote-mobile-control/patch.js
+++ b/linux-features/remote-mobile-control/patch.js
@@ -52,7 +52,7 @@ const REMOTE_MOBILE_APP_SERVER_REMOTE_CONTROL_MARKER = "codexLinuxRemoteMobileAp
 const REMOTE_MOBILE_APP_SERVER_ARGS_NEEDLE =
   "[`-c`,`features.code_mode_host=true`,`app-server`,`--analytics-default-enabled`]";
 const REMOTE_MOBILE_RUNTIME_ASSET_PATTERN =
-  /^app-initial~app-main~new-thread-panel-page~appgen-library-page~hotkey-window-thread-page~ho~iufn7mg3-[^.]+\.js$/u;
+  /^app-initial~app-main~new-thread-panel-page~onboarding-page~login-route~appgen-library-page~~gpgl9un5-[^.]+\.js$/u;
 const REMOTE_CONTROL_APP_MAIN_PAGE_ASSET_PATTERN =
   /^app-initial~app-main~page-[^.]+\.js$/u;
 const REMOTE_MOBILE_ACTIVE_STATUS_ASSET_PATTERN =
@@ -73,6 +73,8 @@ const REMOTE_CONTROL_LINUX_COPY_REPLACEMENTS = [
   ["Control Mac apps from your phone", "Control Linux apps from your phone"],
   ["Let Codex control the apps on your Mac.", "Let Codex control apps on this Linux desktop."],
   ["Let Codex control the apps on your Mac", "Let Codex control apps on this Linux desktop"],
+  ["Let ChatGPT control apps on your Mac", "Let ChatGPT control apps on this Linux desktop"],
+  ["connected to ChatGPT on a Mac", "connected to ChatGPT on this Linux desktop"],
   ["Connect a device to this Mac", "Connect a device to this Linux desktop"],
   ["Connect your phone to this Mac", "Connect your phone to this Linux desktop"],
   ["Add device to control this Mac remotely", "Add a device to control this Linux desktop remotely"],
@@ -1378,7 +1380,7 @@ module.exports = [
   {
     id: "linux-remote-control-visibility",
     phase: "webview-asset",
-    pattern: REMOTE_MOBILE_RUNTIME_ASSET_PATTERN,
+    pattern: REMOTE_CONTROL_APP_MAIN_PAGE_ASSET_PATTERN,
     order: 20_120,
     ciPolicy: "optional",
     missingDescription: "remote-control connections visibility bundle",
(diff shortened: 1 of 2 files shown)


PinguuSS (changes requested):
I found one current-DMG blocker in the copy descriptor.\n\nOn the exact 26.707.51957 artifact (), the active  imports . The visible setup copy lives in that dialog asset: it currently contains 10  values with , including , , and .\n\nThe proposed descriptor matches  but does not match . The flow loader itself contains none of those copy needles, so it is left unchanged without a warning. Because  does change, the aggregate descriptor can still report  and then  while the active setup dialog remains unpatched. That is the same false-success class this PR is intended to remove.\n\nPlease target the current dialog asset that owns the copy (and remove the flow target if it owns no current copy), then add an exact current-filename fixture/assertion proving the dialog is changed on pass one and byte-identical on pass two. The focused 100-test suite is otherwise green locally, and the seven  runtime retargets match the current bundle.

PinguuSS:
Formatting correction for my review above: the shell dropped the inline-code tokens from the submitted review.

On the exact 26.707.51957 artifact (`a66a0645fd4de6f4f22e75ca87bedb1585aeb57f3457a3df55509bf8da110879`), the active `codex-mobile-setup-flow-D4gEouvj.js` imports `codex-mobile-setup-dialog-CoNNd4IV.js`. The visible setup copy lives in that dialog asset: it currently contains 10 `defaultMessage` values with `Mac`, including `Keep this Mac awake`, `Use your Mac apps while locked`, and `Connect your phone to this Mac`.

The proposed descriptor matches `codex-mobile-setup-flow-*` but does not match `codex-mobile-setup-dialog-*`. The flow loader itself contains none of those copy needles, so it is left unchanged without a warning. Because `remote-connections-settings-*` does change, the aggregate descriptor can still report `applied` and then `already-applied` while the active setup dialog remains unpatched.

Please target the current dialog asset that owns the copy (and remove the flow target if it owns no current copy), then add an exact current-filename fixture/assertion proving the dialog is changed on pass one and byte-identical on pass two. The focused 100-test suite is otherwise green locally, and the seven `iufn7mg3` runtime retargets match the current bundle.

PinguuSS:
One additional result from applying the PR's own `applyLinuxRemoteControlCopyPatch` directly to the current assets:

- `codex-mobile-setup-dialog-CoNNd4IV.js`: 10 Mac-bearing `defaultMessage` values before, 1 after. The remaining visible message is `Let ChatGPT control apps on your Mac`.
- `remote-connections-settings-CwdoPaGR.js`: 15 before, 2 after. The remaining messages include `iPhone Pro and Samsung Galaxy devices connected to ChatGPT on a Mac` and the setup instruction that now mixes `Control this Linux desktop` with the original Mac/PC wording.
- Both outputs are byte-stable on the second pass with no warning.

So changing only the filename matcher is not sufficient: the current source uses `Let ChatGPT...`, while the new synthetic fixture still uses `Let Codex...`. Please build the regression fixture from the exact current dialog/settings copy shapes and assert that the feature-owned visible Mac messages are gone before treating the marker as complete.

PinguuSS:
I also reproduced the report-level failure on a reflink copy of the exact 26.707.51957 app, using the PR head and `remote-mobile-control`:

- pass 1: `feature:remote-mobile-control:linux-remote-control-copy = applied`
- pass 2: `feature:remote-mobile-control:linux-remote-control-copy = already-applied`
- after pass 2, `codex-mobile-setup-dialog-CoNNd4IV.js` has no `codexLinuxRemoteControlCopy` marker and still has all 10 Mac-bearing `defaultMessage` values
- the marked `remote-connections-settings-CwdoPaGR.js` still has 2 Mac-bearing `defaultMessage` values

This is not only a static matcher concern: the current patch report produces the false success on the real asset tree. The other webview descriptors applied as expected; the test copy intentionally lacked the extracted main-process build, so the unrelated app-server descriptor skipped in this focused reproduction.

PinguuSS:
I validated a minimal current-DMG-only adjustment locally on top of `8b31996`; no competing branch or PR was published.

The working shape is:

- copy pattern: `/^(?:codex-mobile-setup-dialog|remote-connections-settings)-.*\.js$/`
- remove the absent current targets `codex-mobile-setup-flow-*` and `use-codex-mobile-connected-settings-*`
- replace the legacy `Let Codex...` entries with the current `Let ChatGPT control apps on your Mac` contract
- add the current settings copy `connected to ChatGPT on a Mac`
- use an exact `codex-mobile-setup-dialog-test.js` fixture and assert the two removed targets do not match

Results:

- Remote Mobile focused suite: 99/99
- current app tree pass 1: copy descriptor `applied`
- pass 2: `already-applied`
- `codex-mobile-setup-dialog-CoNNd4IV.js` and `remote-connections-settings-CwdoPaGR.js` both contain the completion marker
- neither asset retains a visible Mac-bearing `defaultMessage` after excluding the internal `controlMac` message tag
- syntax and `git diff --check` pass

This should keep the PR current-only while closing the report-level false success.

ilysenko (approved):
Reviewed against the current upstream DMG. The retargeting is current-only, the PR stays scoped to Remote Mobile, and all required checks pass.

PR #922, opened 2026-07-13
Fix updater lookup for user-local build tools

magicJie:
## Summary

- include `~/.local/bin` in the updater build command PATH when the directory exists
- keep packaged managed Node.js first while preferring user-local tools over system binaries
- add a regression test that verifies the user-local directory precedes `/usr/bin`

## Why

`scripts/install-deps.sh` installs a modern `7zz` into `~/.local/bin` by default. A systemd user service commonly starts without that directory in PATH, so `codex-update-manager` could fall back to the legacy `/usr/bin/7z` and reject modern APFS DMGs even though the supported `7zz` was installed.

## Validation

- `cargo fmt --check`
- `cargo test -p codex-update-manager builder::tests:: -- --nocapture --test-threads=1` (12 passed)
- local end-to-end updater rebuild progressed from the legacy 7-Zip rejection through DMG extraction, Electron 42.1.0 native-module rebuild, patching, and DEB generation

## Additional test note

A full updater test run encountered the unrelated environment-sensitive failure `app::tests::prompt_install_cli_does_not_treat_non_executable_file_as_installed`; that failure then left the suite's shared environment lock blocked. The complete builder test module passes serially.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -560,7 +560,7 @@ dependencies = [
 
 [[package]]
 name = "codex-update-manager"
-version = "0.9.5"
+version = "0.9.6"
 dependencies = [
  "anyhow",
  "chrono",
--- a/updater/Cargo.toml
+++ b/updater/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "codex-update-manager"
-version = "0.9.5"
+version = "0.9.6"
 edition = "2021"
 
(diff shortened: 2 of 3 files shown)


ilysenko (approved):
Reviewed the current head after the updater patch-version fix. The focused implementation and full CI matrix look good.

Issue #892, opened 2026-07-12
Settings > Linux desktop can crash with generic error after GNOME lock/unlock


psypeal:
## Summary

After returning from a prolonged GNOME screen lock, opening **Settings > Linux desktop** can render only the generic error boundary:

```text
Oops, an error has occurred
```

This was observed on the same session where tray Quit later stopped responding. The two symptoms may share a damaged Electron/WebContents resume lifecycle, but they are reported separately because that relationship is not yet proven.

## Environment

- OS: Ubuntu 24.04.4 LTS
- Desktop: Ubuntu GNOME
- Session: X11
- Kernel: 6.8.0-134-generic
- Package format: deb
- codex-desktop-linux: 0.9.5 at `2e04d928babaf09b321c870f4d8b9befd28921a5`
- Installed package version: `2026.07.12.122632`
- Electron: 42.1.0
- Upstream Codex app: 26.707.51957
- Optional Linux features: none

## Steps observed

1. Start Codex Desktop normally.
2. Lock the GNOME desktop for an extended period.
3. Unlock and return to the existing app instance.
4. Open Settings.
5. Select **Linux desktop**.

## Actual behavior

The settings route displays only:

```text
Oops, an error has occurred
```

## Expected behavior

The generated Linux Desktop settings page should render its launcher, tray, warm-start, update, and build-information controls after session resume.

## Build and static checks

The installed patch report records `keybinds-settings` as applied. The generated asset:

```text
/opt/codex-desktop/content/webview/assets/linux-desktop-settings-linux.js
```

- passes `node --check`;
- has all referenced module exports present;
- imports successfully in an ESM runtime;
- returns a valid React element when `LinuxDesktopSettings({})` is evaluated.

The existing generated-page unit test also passes, but it renders simplified mocked components and does not exercise the real Electron suspend/resume lifecycle.

## Runtime observations

The launcher log contains repeated lifecycle errors around affected sessions:

```text
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 destroyed listeners added to [WebContents]
Error occurred in handler for 'codex_desktop:message-from-view': TypeError: Object has been destroyed
```

The generic error boundary does not currently log the renderer exception or component stack to `launcher.log`, which prevents identifying the exact crashing component.

## Possible direction

- Log the Linux Desktop route error-boundary exception and component stack.
- Validate or recreate destroyed auxiliary `WebContents` after unlock/resume.
[...]

PR #933, opened 2026-07-13
Preserve renderer error stacks in structured logs

PinguuSS:
## Summary

- preserve the original JavaScript stack as a primitive string before renderer error-boundary logs cross the Electron bridge
- target only the active error-boundary bundle in the current unified app
- keep the patch optional, atomic, and idempotent, with an explicit drift warning

## Why

Issue #892 now has a structured `errorStack`, but current-DMG runtime probes show that nested `Error` objects are reserialized by the renderer logger bridge. Unrelated failures therefore acquire the same `Object.postMessage -> dispatchMessage` transport frames.

A live probe logged both a nested `Error` and a primitive copy of `Error.stack`. The nested `errorStack` pointed at the transport, while the primitive field retained the original `issue892OriginalThrowSite` frame. This change adds that primitive `originalErrorStack` field at the error boundary so the next reproduction can identify the actual caller.

This is observability only. It does not claim to fix the Linux Desktop Settings crash or alter WebContents/React lifecycle behavior.

## Current-DMG validation

- repo base: `bce8d36f72eda4cabfbf32a95054e6fc79737722`
- ChatGPT Desktop: `26.707.62119`
- Electron: `42.1.0`
- DMG SHA-256: `c243c94f8de6a51f5530ffe1f8d0c1588733d890ac692e34aaca06d95ba637ca`
- DMG ETag: `0x8DEE0AB667193CC`
- DMG size: `615738501` bytes
- environment: KDE Plasma / Wayland
- isolated no-feature build: accepted with only the pre-existing optional tooltip collision warning
- patch report: `renderer-error-stack-preservation: applied`
- generated asset: syntax-valid and byte-stable on the second pass

## Tests

- `node --test scripts/patch-linux-window-ui.test.js` (370 passed)
- `node --test linux-features/*/test.js` (560 passed)
- `node --check` on the generated current-DMG asset
- `git diff --check`

Closes no issue; follow-up evidence for #892.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -174,6 +174,7 @@ const {
   applyLinuxSafeMonospaceFontStackPatch,
   applyLinuxSettingsSearchVisibilityPatch,
   applyLinuxSkillsListDedupePatch,
+  applyRendererErrorStackPreservationPatch,
   applyLinuxThreadSidePanelNativeTooltipPatch,
   applyLinuxTooltipWindowControlsCollisionPatch,
   applyLinuxWindowControlsSafeAreaPatch,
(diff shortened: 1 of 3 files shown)


ilysenko (changes requested):
Please keep this PR and its branch available, and leave a comment on #892 linking to them. Ask the reporter to build or install it, reproduce the failure, and provide the resulting `originalErrorStack`. Once that evidence is available, please replace the diagnostic probe in this PR with a focused fix for the actual failure.

I understand that keeping this PR open uses one of your available PR slots and limits your ability to submit other pull requests. Still, if you have time to continue working on it, this will keep the effort focused on #892, which is a particularly valuable task for the project.

PinguuSS:
Understood. All required checks completed successfully. I linked the tested branch from #892, asked the reporter to reproduce with it, and will keep the branch available for collecting the original stack. Closing this diagnostic PR so the next submission can contain the focused behavior fix once that evidence exists.

PR #934, opened 2026-07-13
Require Read Aloud for Conversation Mode

PinguuSS:
## Summary

- declare Read Aloud as a required dependency of Conversation Mode
- document that both opt-in features must be enabled explicitly
- cover the missing-dependency error and valid explicit configuration in focused feature tests

## Behavior

Conversation Mode uses the local Read Aloud backend for assistant speech. Its manifest now declares `"requires": ["read-aloud"]`, so the existing feature validation rejects a configuration that enables Conversation Mode without Read Aloud.

Users continue to enable both features explicitly. This PR does not add automatic dependency resolution or change setup and updater behavior.

## Validation

- `node --test linux-features/conversation-mode/test.js` (56 passed)
- `node --test linux-features/conversation-mode/test.js linux-features/example-feature/test.js`
- `node --test --test-concurrency=1 linux-features/*/test.js scripts/patch-linux-window-ui.test.js` (931 passed)
- `git diff --check`

## Scope

This PR changes only `linux-features/conversation-mode/README.md`, `feature.json`, and its focused test. It does not change Conversation Mode runtime, current-DMG asset matchers, the generic feature loader, setup wizard, or updater.

### DIFF
--- a/linux-features/conversation-mode/README.md
+++ b/linux-features/conversation-mode/README.md
@@ -21,8 +21,10 @@ rebuilding:
 ```
 
 The feature is disabled by default. A session starts only after the user clicks
-the composer voice control. `read-aloud` should be enabled too because
-conversation mode speaks through the local Kokoro Read Aloud handler.
+the composer voice control. Enable `read-aloud` explicitly alongside
+`conversation-mode`; it is required because conversation mode speaks through
+the local Read Aloud backend. The Read Aloud settings toggle can remain off
+because Conversation Mode invokes the backend for its own responses.
 
 ## Scope
 
--- a/linux-features/conversation-mode/feature.json
+++ b/linux-features/conversation-mode/feature.json
@@ -2,6 +2,9 @@
   "id": "conversation-mode",
   "name": "Conversation mode",
   "description": "Disabled-by-default Linux voice conversation loop for normal Codex turns using dictation and Read Aloud.",
+  "requires": [
+    "read-aloud"
+  ],
   "entrypoints": {
     "patchDescriptors": "./patch.js"
   }
--- a/linux-features/conversation-mode/test.js
+++ b/linux-features/conversation-mode/test.js
@@ -436,11 +436,25 @@ test("conversation mode stays disabled until listed in features.json", () => {
   });
 });
 
-test("conversation mode exposes optional patch descriptors when enabled", () => {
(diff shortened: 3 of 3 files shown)


PinguuSS:
#938 merged the current-DMG Conversation Mode retarget that originally lived in this PR. I rebased onto that merge and replaced the now-duplicated two-file diff with the separately validated manifest dependency transition: Conversation Mode now requires Read Aloud, and the setup/updater selection paths resolve that contract consistently. The title, description, validation evidence, and head have been updated accordingly.

ilysenko (changes requested):
Thanks for working on this. The current scope is too large for a disabled-by-default optional feature that is likely used by a relatively small number of users.

The existing `requires` validation is sufficient here. Please keep this PR feature-only:

- Add `"requires": ["read-aloud"]` to the Conversation Mode manifest.
- Update its README and focused feature tests.
- Let users explicitly enable Read Aloud when the existing validation reports that it is required.
- Remove the generic auto-resolution, setup wizard, and updater changes from this PR.

Please avoid spending more time on further Conversation Mode refinements unless there is clear user demand. This is not currently a repository priority, and expanding core and updater behavior for it creates disproportionate maintenance and regression risk.

Please verify the reduced change against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant tests locally before pushing the update, review the complete diff independently with a local LLM, and wait for every required CI check to pass before requesting maintainer review again.

PinguuSS:
Updated as requested. The PR is now feature-only: it adds the `read-aloud` requirement to the Conversation Mode manifest, documents explicit enablement of both features, and updates only the focused Conversation Mode tests. The generic resolver, setup wizard, and updater changes were removed. Local focused tests, the serial 931-test feature/patcher suite, smoke tests, and `git diff --check` pass. All required GitHub checks, including the current upstream DMG build, are now green. Ready for re-review.

ilysenko (approved):
Reviewed the reduced feature-only scope on the current head. The previous requested changes are addressed, the dependency contract is covered by focused tests, and all required checks pass.

ilysenko:
@PinguuSS Thanks for the update.

As a general rule, changes to core or shared paths receive much closer scrutiny because they affect all users and package formats. Opt-in, disabled-by-default Linux features have a smaller blast radius, so there is usually more room for flexibility.

PR #935, opened 2026-07-13
Harden Linux settings runtime inference

psypeal:
## Summary

- Harden the Linux settings React/JSX runtime inference by anchoring it to the lazy route factory that owns `linux-desktop` and `general-settings`.
- Require the wrapper to use one React local for both `lazy` and `useState`, then require the React and JSX factories to be initialized in the same statement before exporting the bridge.
- Add a regression fixture with earlier decoy `useState` and `jsx` calls so unrelated runtime locals are rejected.

The previous inference selected the first `useState` and `jsx` references independently from the whole bundle, so it could pair unrelated runtime locals. This change narrows that inference defensively.

This work was motivated by the investigation of #892, but it does not identify or reproduce the reported failure and does not claim to resolve it. #892 should remain open until the actual failure is identified and a fix is verified.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` — 368 passed
- `node --test linux-features/*/test.js` — 560 passed
- `bash tests/scripts_smoke.sh` — passed, including Debian/RPM/pacman/AppImage packaging smoke coverage
- `./install.sh ./Codex.dmg` with upstream 26.707.62119 / SHA256 `c243c94f8de6a51f5530ffe1f8d0c1588733d890ac692e34aaca06d95ba637ca` — `accepted_with_warnings`; the sole warning is the unrelated pre-existing optional tooltip collision patch drift
- Rebuilt-app sanity check: opened Linux desktop settings and completed 100 General ↔ Linux desktop remount cycles with Build information present and no generic error. This was not a reproduction of #892.
- `node --check` on the generated Linux desktop settings and toggle assets
- `git diff --check`
- Required CI checks — 7 passed; the upstream drift reconciliation job was intentionally skipped after the upstream build passed

## Checklist


ilysenko (changes requested):
Please rename this PR to present the change as runtime-inference hardening rather than a fix, and remove `Fixes #892` and any claim that #892 is resolved from the description. The issue should remain open until the actual failure is identified and the fix is verified.

Before requesting another review, please verify the final scope against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant tests locally, review the complete diff independently with a local LLM, and wait for all required CI checks to pass.

ilysenko (approved):
The requested scope correction is addressed: this is now presented as runtime-inference hardening and does not claim to resolve #892. The unchanged code and current-DMG checks were re-reviewed.

PR #940, opened 2026-07-13
Fix Linux settings asset cache invalidation

psypeal:
## Summary

- content-version the generated Linux Desktop settings route with a SHA-256 query token
- content-version imports of locally generated fallback settings components
- refresh existing plain or versioned route references idempotently during patching
- verify both route and fallback digests in patcher and package smoke tests

## Root cause

The generated Linux Desktop settings modules used stable filenames across rebuilds. In the reproduced failure, Electron executed an older hook-based module even though the installed and served asset was the newer class-based implementation. Clearing the browser cache and reloading with cache bypass immediately restored the route, proving that a stale module graph—not the current bridge or React runtime—caused the generic error boundary.

`Cache-Control: no-store` on the local webview server was not sufficient to prevent this Chromium/Electron module reuse. Content-derived query tokens give each generated source revision a distinct module URL while preserving deterministic, idempotent patch output.

## User impact

After rebuilding or updating the Linux package, **Settings > Linux desktop** loads the generated settings implementation instead of reusing a stale incompatible module.

Related to #892.

## Validation

- `git diff --check`
- `node --test scripts/patch-linux-window-ui.test.js` — 368/368 passed
- `bash tests/scripts_smoke.sh` — all script and package-format smoke tests passed
- rebuilt and installed the Debian package from upstream app `26.707.62119`
- live Electron/CDP check opened **Settings > Linux desktop**, rendered all four controls and build metadata, requested the versioned route URL, and recorded no new renderer error boundary

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -5189,7 +5189,7 @@ test("writes only missing Linux settings fallback components after required chec
     assert.equal(fs.existsSync(path.join(assetsDir, "linux-settings-group-linux.js")), false);
     assert.match(
       fs.readFileSync(path.join(assetsDir, linuxDesktopSettingsAsset), "utf8"),
-      /import\{n as SettingsRow\}from"\.\/linux-settings-row-linux\.js"/,
+      /import\{n as SettingsRow\}from"\.\/linux-settings-row-linux\.js\?v=[a-f0-9]{12}"/,
     );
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed the focused cache-invalidation change and its current-DMG coverage. The content-derived asset tokens are deterministic and idempotent, and all required checks are green.

PR #942, opened 2026-07-13
Keep service tier fallback patch atomic

PinguuSS:
## Summary

- require the API-key fast-tier fallback patch to contain all current helper, resolver, and options markers before treating it as successful
- discard in-memory partial changes and emit a warning when one current insertion point drifts
- cover both the direct patch result and the patch-report `skipped-optional` status

## Behavior

When the current fallback helper and resolver match but the options insertion point does not, the feature now leaves the asset byte-identical and reports drift. Previously it returned a partial patch without a warning, so the descriptor was recorded as `applied` even though the fallback option was incomplete.

## Current DMG

- App: `26.707.62119`
- Electron: `42.1.0`
- SHA-256: `c243c94f8de6a51f5530ffe1f8d0c1588733d890ac692e34aaca06d95ba637ca`
- ETag: `0x8DEE0AB667193CC`
- Size: `615738501` bytes

An isolated build with only `api-key-service-tier` enabled applied all three feature descriptors. The generated current assets passed `node --check`. A second pass reported all three descriptors as `already-applied` and kept the asset tree byte-identical.

## Validation

- `node --test linux-features/api-key-service-tier/test.js` (`16/16`)
- `node --test --test-concurrency=1 linux-features/*/test.js scripts/patch-linux-window-ui.test.js` (`933/933`)
- `node --check` on the changed source and generated current assets
- `scripts/ci/validate-patch-report.js` on the isolated current-DMG report
- `git diff --check`

## Scope

This changes only `linux-features/api-key-service-tier/patch.js` and its focused tests. The separate Copilot settings atomicity finding is intentionally out of scope.

### DIFF
--- a/linux-features/api-key-service-tier/patch.js
+++ b/linux-features/api-key-service-tier/patch.js
@@ -87,7 +87,20 @@ function hasApiKeyModelListMappingShape(source) {
   return MODEL_LIST_MAPPING_SHAPE.test(source);
 }
 
+function hasCompleteFallbackFastTierPatch(source) {
+  return (
+    source.includes(`function ${PATCH_MARKER}(`) &&
+    source.includes(`??${PATCH_MARKER}(`) &&
+    source.includes(`[${PATCH_MARKER}(`) &&
+    source.includes(".filter(Boolean)).map")
+  );
+}
+
 function applyFallbackFastTierPatch(source) {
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head. The atomic fail-soft behavior is focused, well tested, and safe to merge.

PR #943, opened 2026-07-13
Fail closed on candidate promotion setup errors

tjdonley:
## Summary

- stop candidate promotion before atomic exchange when recovery-journal preparation fails
- propagate first-install rename failures instead of reporting a successful promotion
- make failed preparation remove only journal and marker metadata owned by its transaction
- use `Path.lstat()` so journal identity capture works with Python 3.8/3.9
- add focused regression coverage for all fail-closed paths

## Root cause

`install.sh` checks `promote_candidate_install` from an `if !` condition. Bash suppresses `errexit` inside functions used in that context, so unchecked journal preparation and first-install rename failures could be ignored. The replacement path could proceed to atomic exchange without a durable recovery journal; the first-install path could report success without installing the app.

Preparation also needed to account for failure after the journal had already been atomically replaced. The helper now verifies transaction ownership, removes the journal before its matching disposable-candidate marker, and leaves foreign transaction metadata untouched.

## Impact

Promotion now fails before changing the working app when its journal cannot be prepared, failed initial renames propagate correctly, and partial preparation cannot leave metadata that blocks future installs after the candidate is discarded.

## Validation

- `bash -n scripts/lib/candidate-install.sh tests/scripts_smoke.sh`
- `/usr/bin/python3 -m py_compile scripts/lib/candidate-promotion.py` using Python 3.9.6
- focused smoke coverage for failed journal preparation, post-journal cleanup and retry, and failed first-install rename
- reviewed on the fork with the actionable Codex review finding addressed before upstream submission

### DIFF
--- a/scripts/lib/candidate-install.sh
+++ b/scripts/lib/candidate-install.sh
@@ -144,12 +144,17 @@ promote_candidate_install() {
         backup="$(candidate_backup_path "$final_dir")"
         journal_file="$(candidate_promotion_journal_path "$final_dir")"
         transaction_id="$(date -u +%Y%m%dT%H%M%S)-$$-${RANDOM:-0}"
-        python3 "$CANDIDATE_PROMOTION_HELPER" prepare \
+        if ! python3 "$CANDIDATE_PROMOTION_HELPER" prepare \
             --candidate "$candidate_dir" \
             --final "$final_dir" \
             --backup "$backup" \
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed the current head. The promotion failure paths now fail closed, preserve the working app, and have focused regression coverage.

PR #950, opened 2026-07-13
Reject incompatible ydotool fallback commands

### LINKED ISSUE #949: Computer Use reports success with incompatible pre-1.0 ydotool
## Summary

Computer Use treats any installed ydotool binary plus a connectable ydotoold socket as a working fallback. Ubuntu/Kubuntu 25.10 currently ships ydotool 0.1.8, whose CLI is incompatible with the raw-event commands emitted by the backend.

## Reproduction

Environment:
- KDE Plasma Wayland
- ydotool 0.1.8-3build1
- live ydotoold and connectable /tmp/.ydotool_socket
- upstream main bb6791d

The backend emits commands such as:

`ydotool mousemove --absolute -- 1108 863`

ydotool 0.1.8 prints `error: unrecognised option '--absolute'` but exits with status 0. The backend therefore reports that the action was sent even though no coordinate input occurred. Its legacy `key` command also interprets current raw key events as literal input.

The doctor report currently marks ydotool available based only on its executable path and advertises it as an input backend.

## Expected behavior

- Detect and reject the incompatible pre-1.0 named-key CLI.
- Do not advertise it as a usable input backend even when ydotoold and the socket are healthy.
- Do not report success when ydotool writes a CLI error to stderr with status 0.
- Keep portal and direct uinput paths unchanged.

I reproduced this while testing Computer Use against Claude Desktop. Claude's sparse AT-SPI tree is a separate application-side limitation; this issue covers only the false-success ydotool fallback.

PinguuSS:
## Summary

- detect whether the installed `ydotool` exposes the raw-event CLI required by Computer Use
- reject legacy pre-1.0 commands before sending input and catch CLI errors that exit with status 0
- advertise `ydotool` only when the binary, daemon, and socket are all usable
- document the minimum compatible version

## Root cause

Ubuntu/Kubuntu 25.10 ships `ydotool` 0.1.8. Its CLI accepts named keys and relative mouse movement, while the backend emits the raw key events and absolute movement syntax introduced in 1.0. The legacy binary prints errors such as `unrecognised option '--absolute'` but exits successfully, so Computer Use previously reported an action even though no input occurred.

This change fails closed instead of attempting unsafe coordinate emulation. Portal and direct `/dev/uinput` input remain unchanged.

Closes #949.

## Validation

- reproduced with `ydotool 0.1.8-3build1`, a live `ydotoold`, and a connectable `/tmp/.ydotool_socket` on KDE Plasma Wayland
- confirmed the updated doctor report rejects the legacy CLI and removes it from `capabilities.input` while retaining the available portal backend
- called `press_key` through the isolated MCP server with the keyboard fallback forced to the live legacy installation; it returned `ok: false` before sending input
- called `click` through the isolated MCP server with the pointer fallback forced to the live legacy installation; it returned `ok: false` before sending input
- confirmed the default Wayland selector prefers the portal when a connectable socket belongs to an incompatible legacy CLI, while preserving the current-CLI preference
- `cargo test -p codex-computer-use-linux` (357 passed)
- `git diff --check`

## Scope

This does not change AT-SPI extraction or attempt to work around Electron applications that expose incomplete accessibility trees.

### DIFF
--- a/computer-use-linux/src/diagnostics.rs
+++ b/computer-use-linux/src/diagnostics.rs
@@ -2,6 +2,7 @@ use crate::windowing::registry::{
     self, COSMIC_WAYLAND_BACKEND, GNOME_SHELL_EXTENSION_BACKEND, GNOME_SHELL_INTROSPECT_BACKEND,
     HYPRLAND_BACKEND, KWIN_BACKEND, NIRI_BACKEND,
 };
+use crate::ydotool;
 use schemars::JsonSchema;
 use serde::Serialize;
 use std::{
@@ -211,7 +212,7 @@ fn capability_map(
     if portals.remote_desktop.ok {
         input_backends.push("portal".to_string());
     }
-    if input.ydotool_socket.ok {
+    if input.ydotool.ok && input.ydotoold.ok && input.ydotool_socket.ok {
         input_backends.push("ydotool".to_string());
     }
 
@@ -635,7 +636,10 @@ fn check_from_backend_probe(probe: &registry::BackendProbe) -> Check {
 
 fn input_report() -> InputReport {
     InputReport {
-        ydotool: command_path_check("ydotool"),
+        ydotool: match ydotool::ensure_supported() {
+            Ok(detail) => Check::ok(detail),
+            Err(detail) => Check::fail(detail),
+        },
         ydotoold: process_check("ydotoold"),
         ydotool_socket: ydotool_socket_check(),
         uinput: read_write_path_check(Path::new("/dev/uinput")),
@@ -803,10 +807,6 @@ fn user_id() -> Option<String> {
         .success()
         .then(|| String::from_utf8_lossy(&output.stdout).trim().to_string())
         .filter(|value| !value.is_empty())
-}
-
-fn command_path_check(command: &str) -> Check {
-    command_check("sh", &["-c", &format!("command -v {command}")])
 }
 
 fn process_check(process_name: &str) -> Check {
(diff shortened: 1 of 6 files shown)


Yo-DDV (changes requested):
Hi @ilysenko, I found one remaining blocker here. The PR fixes the diagnostics side, but the runtime selector can still let a legacy `ydotoold` socket mask a working portal backend.

The mismatch is in `computer-use-linux/src/server.rs`. `should_prefer_portal_pointer_backend()` still returns portal only when `ydotool_socket().is_none()`, and the keyboard selector has the same default shape for non-KDE Wayland. The action paths for click, scroll and drag only open the portal when that selector says yes, then fall back to `run_ydotool_sequence()`. After this PR, `run_ydotool()` calls `ydotool::ensure_supported()` before sending input, so a legacy CLI with a live socket fails there instead of letting the portal handle the action.

That means the exact class of setup described in the PR is still only half covered. With legacy `ydotool` 0.1.8, a connectable `ydotoold` socket and an available RemoteDesktop portal, `doctor` can correctly remove `ydotool` from `capabilities.input` while keeping `portal`, but the real pointer path still suppresses the portal because the socket exists. The result can be `ok: false` even though the usable fallback was available.

I think the selector should treat ydotool as available only when the socket is connectable and the CLI probe succeeds, then add a regression test around Wayland plus portal plus a connectable legacy ydotool socket. Forced ydotool can still fail closed, but the default path should not let an unusable ydotool install block a working portal.

I also re-ran `cargo fmt --all -- --check`, `cargo clippy -p codex-computer-use-linux --all-targets -- -D warnings`, and `cargo test -p codex-computer-use-linux`. They are green, but they do not cover this selector/capability mismatch yet.

ilysenko (approved):
Reviewed the current head and verified the stale portal-selection blocker is addressed. The required checks are green.

PR #951, opened 2026-07-13
fix(linux): commit quit teardown at will-quit

### LINKED ISSUE #939: Linux: a vetoed quit wedges the app — tray destroyed on cancellable `before-quit`, leaving it unclosable and unopenable
## Summary

On Linux, a quit attempt that gets vetoed leaves the app permanently wedged. The panel keeps painting the tray icon, but the tray item behind it is already destroyed: right-click shows no menu, left-click does not restore the window. The process stays alive holding the single-instance lock, so relaunching the app does nothing either. The app ends up simultaneously **unclosable and unopenable**, and only `kill -TERM` recovers it.

The trigger is a re-entrancy bug in the Linux quit patches: `before-quit` is treated as a point of no return, but in Electron it is a cancellable event that can fire more than once.

## Environment

- Distro: CachyOS (Arch-based)
- Desktop: KDE Plasma 6, Wayland session; Electron runs with `--ozone-platform=x11`. `plasmashell` is the StatusNotifierItem host.
- Package format: pacman, `codex-desktop 2026.07.13.093748-1`, installed at `/opt/codex-desktop`
- Affected component: the packaged runtime (Electron main process / ASAR patches). Not `install.sh`, not the updater, not the generated launcher.

## Actual behavior

After a few hours of normal use, the app entered a state where:

- the tray icon was still visible in the KDE panel, but **right-click produced no menu and left-click did not restore the window**;
- there was no window to close, and no way to quit from the UI;
- **relaunching from the app menu or from `/usr/bin/codex-desktop` did nothing at all** — no window, no error;
- the old process was still alive, idle: state `S`, main thread parked in `ppoll`, having consumed only 1m35s of CPU across 3h36m of wall time. It was not deadlocked or spinning — its entire context had already been torn down and nothing was left to drive it.

`kill -TERM <electron-pid>` terminated it cleanly and immediately, and the app then relaunched normally.

## Expected behavior

A quit either completes and the process exits, or it is cancelled and the app returns to a fully working state — tray alive, window restorable, relaunch working. A cancelled quit must never leave a half-destroyed process behind.

## Evidence

**The tray's D-Bus object was unregistered while the bus name stayed alive.** `busctl tree` still listed `/StatusNotifierItem` and `/com/canonical/dbusmenu` on the app's bus name, but every property read failed instantly:

```
Error org.freedesktop.DBus.Error.UnknownMethod: Method is no longer available
```

Other tray apps on the same session answered their properties normally, so the probe itself was sound. The item was destroyed; Plasma simply kept drawing the last icon it had been given. That is the husk in the panel.

**The warm-start socket was gone.** `$XDG_RUNTIME_DIR/codex-desktop/` was empty — no `launch-action.sock` — so the launcher fell back to the Electron `second-instance` handoff, which the wedged process could no longer service (`Failed to open window on second instance`).

**The launcher log shows a teardown that started and never finished**, then silence:

```
IAB_LIFECYCLE removed browser session window windowId=1
AppServerConnection Stopping app-server transport ... cause=stop_process
<ssh app-server proxy> Killed by signal 15
[IpcRouter] I am the router
```

So: transports stopped, browser-session window removed, socket disposed — but the process never exited.

## Root cause

`scripts/patches/impl/main-process/quit-lifecycle.js:15` (`applyLinuxQuitGuardPatch`) injects, at module top level:

```js
app.on(`before-quit`, () => codexLinuxDestroyTray());
```

`before-quit` is **cancellable and re-entrant**. This listener treats it as a commit point and destroys the tray unconditionally, on every fire.

The sequence that produces the husk:

1. A quit starts. `before-quit` fires → **the tray is destroyed** (SNI object unregistered; the panel keeps drawing the stale icon).
2. The quit proceeds into `will-quit`. The drain runs and `codexLinuxFinalizeQuit()` disposes everything — app-server transports, the browser-session window, the warm-start socket — and then calls `app.quit()` a **second** time.
3. That second `app.quit()` **re-fires `before-quit`**, which runs upstream's confirmation prompt again (`showMessageBoxSync({type:'warning', buttons:['Quit','Cancel']})`). `codexLinuxShouldBypassQuitPrompt()` only returns true for tray-Quit and the `quit-app` IPC, so it does not bypass here — and the quit is **vetoed**.
4. Terminal state: a live process whose entire context has been disposed, still holding the single-instance lock, idle in `ppoll` forever.

The exact user action that started the quit in my session was not captured, but the mechanism above is deterministic from the shipped bundle and every piece of post-mortem evidence matches it.

### Contributing defects

1. **`before-quit` destroys the tray unconditionally** (`quit-lifecycle.js:15`). A cancellable event is used as a commit point.
2. **`codexLinuxQuitInProgress` is a one-way latch.** It is set during `before-quit` and never reset, so a single vetoed quit poisons the process for its whole lifetime: the tray menu refuses to open, the close-to-tray path changes behavior, and the `second-instance` handoff is swallowed.
3. **Re-entrant veto hole.** `codexLinuxFinalizeQuit()` ends in `app.quit()`, which is cancellable — so the *final* step of teardown can itself be vetoed, which is exactly what happens.
4. **Scope split in the quit-state helpers.** `scripts/patches/impl/launch-actions.js:18-19` injects a second, leaner copy of the quit-state helpers inside the minified startup function, where it **shadows** the real module-scope copy. That copy has no tray variables and its flag is never written, so the launch-action quit guards are dead code.
5. **`canHideLastWindowToTray()` reports the tray *setting*, not tray liveness.** With the tray object destroyed, the setting is still `true`, so the window can be hidden into a tray that no longer exists.

## Fix direction

Make `will-quit` the single commit point, and hold this invariant:

- **Before `will-quit`:** a quit is fully recoverable. Nothing is destroyed, no sticky state is set, a veto leaves the app completely functional.
- **After `will-quit`:** teardown is irreversible and process exit is **guaranteed** — the drain gets an unconditional timeout and the final step is a non-cancellable exit, not another `app.quit()`.

Plus: gate hide-to-tray on the tray actually being alive (fail closed), and remove the shadowed duplicate of the quit-state helpers so there is exactly one.

## Workaround

If the app wedges into this state:

```bash
pkill -TERM -f '/opt/codex-desktop/resources/app.asar'
```

It exits cleanly on `SIGTERM` and relaunches normally afterwards.

OmerFarukOruc:
## Summary

Fixes #939.

On Linux, a quit that starts but does not finish leaves the app permanently wedged: the tray icon is gone (so it cannot be closed) and the single-instance lock is still held (so it cannot be reopened). The process sits idle at ~0% CPU with its entire context already disposed.

The root cause is that `before-quit` was treated as a point of no return. It is not — it is cancellable and re-entrant. The patch set registered `app.on('before-quit', () => codexLinuxDestroyTray())` at module top level, so **every** quit attempt destroyed the tray, including ones that were later vetoed. It also latched `codexLinuxQuitInProgress` on with no reset, and ended the will-quit drain in a second, cancellable `app.quit()` — which re-fires `before-quit` and can be vetoed again, stalling the quit forever with the tray already gone.

This change makes **`will-quit` the single commit point for irreversible app state**. An accepted Linux quit first enters a reversible `ATTEMPTING` state. `markAppQuitting()` and quick-chat disposal are deferred until `will-quit`; a later `before-quit` veto clears the attempt before either mutation runs. After `will-quit`, teardown is irreversible and process exit is guaranteed.

- Move tray teardown off `before-quit` and onto `will-quit` — including **upstream's own** `before-quit` tray-destroy listener, which the previous fix attempt missed. Without that, relocating only our listener would not have fixed the bug.
- Replace the sticky `codexLinuxQuitInProgress` latch with explicit `ATTEMPTING` and `COMMITTED` states plus a one-shot explicit-quit ticket. `ATTEMPTING` suppresses close-to-tray and other launch actions without mutating upstream quit state; the commit callback performs `markAppQuitting()` only when `will-quit` fires.
- Patch the real SQLite backfill progress-window close listener. It prevents the outer close and calls `app.quit()` recursively; Electron ignores that nested quit while the first quit is in progress, so this was a reachable pre-commit cancellation path. During an accepted Linux quit attempt the progress window now closes normally instead of cancelling and re-entering quit.
- End the Linux drain finalizer in `app.exit(0)` / `process.exit(0)` instead of a cancellable `app.quit()`, with an unconditional drain timeout. `Browser::Exit()` bypasses both the `before-quit` and `will-quit` veto paths, so a committed quit can no longer be stalled by any listener.
- Arm an 8s `unref()`d hard-exit watchdog at `will-quit`, in the same commit step that runs the deferred upstream mutation and destroys the tray. It backstops teardown once the audited listener graph reaches its commit point.
- Require a **live** tray for close-to-tray. `canHideLastWindowToTray()` reported the tray *setting*, not tray liveness, so the window could be hidden into a destroyed tray. The gate is now fail-closed on Linux: no live tray means a real close.
- Remove the duplicate quit-state helper copy that was being injected at depth 1 inside the minified startup function, where it shadowed the module-scope copy and was never written — making the launch-action quit guards dead code. The rebuilt bundle now contains exactly one copy.
- Remove the dead `appVar`/`directHandler` branch (it always received `null`) and the legacy second-instance migration path, per the "no legacy patch paths" rule.
- Make the prompt-bypass patch report its own drift. Its warning keyed on a `showMessageBoxSync` dialog shape that upstream has since localized, so it could never fire. It now recognizes the confirmation site through `canQuitWithoutPrompt()` or through the `showMessageBoxSync` and `markQuitApproved()` pair, so a rename of either anchor surfaces as drift instead of silently recording this `required-upstream` patch as applied.

## Validation

Review-fix validation, run against the amended tree:

| Check | Result |
| --- | --- |
| `node --test scripts/patch-linux-window-ui.test.js` | 386/386 pass |
| `node --test linux-features/*/test.js` | 560/560 pass |
| `bash tests/scripts_smoke.sh` | All script smoke tests passed |
| `node scripts/ci/validate-patch-report.js codex-app/.codex-linux/patch-report.json` | Required upstream profile passed |

**Real-DMG rebuild.** `MAX_BUILD_THREADS=0 make build-app-fresh` refreshed the upstream DMG and rebuilt end to end against `26.707.72221` (Electron 42.1.0), which is newer than the `26.707.71524` build requested in the review. DMG SHA-256: `40e34814e74e30943c209ebd4da94cd4de3581a52c5bffbe2bcf2e488d6361c6`.

- Acceptance verdict `accepted_with_warnings`, with **zero required blockers**. The two warnings are unrelated optional drift in `automation-update-eager-tool` and `linux-tooltip-window-controls-collision`.
- All six lifecycle descriptors, including the new `linux-sqlite-backfill-quit-close`, report `applied` with `required-upstream` policy.
- The emitted `app.asar` contains the reversible attempt state, deferred `markAppQuitting()` callbacks, the `will-quit` commit callback, and the SQLite close guard. It contains no renderer `will-prevent-unload` or `onbeforeunload` handler; the sole `beforeunload` string is a Computer Use dialog enum.

**Runtime reachability proof.** A clean Electron 42.1.0 reproduction of the audited SQLite listener graph recorded `before-quit -> close-outer -> timeout`, with no `will-quit`. The nested `app.quit()` did not restart the lifecycle and the outer `close` remained prevented. This confirms that cancellation was reachable and that merely narrowing the old test would not fix the bug.

**Tests added.** The synthetic “fully recoverable” test called out in review is gone. Its replacement carries one mutable upstream state object through a later `before-quit` veto and retry, proving the veto leaves `isAppQuitting` false, quick-chat undisposed, the tray alive, and close-to-tray usable. A separate behavioral test matches the actual SQLite close listener and proves it no longer prevents an accepted Linux quit. Existing tests still cover the committed watchdog/force-exit path, ticket expiry, drift reporting, and byte-identical reapplication.

**Known risk / scope.** The proof and patch target the complete cancellation graph present in `26.707.72221`: the close-to-tray listener, the SQLite backfill progress listener, the upstream quit confirmation, and the upstream drain listener. A future upstream window or renderer veto is drift and requires a fresh audit; this PR does not claim recoverability for listener shapes absent from the validated bundle.

## Checklist

- [x] I added or updated the relevant tests, ran the validation listed above, and confirmed the required CI checks pass.
- [x] I reviewed the final diff with a coding agent at maximum reasoning effort, addressed all findings, and reran the relevant tests.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -84,6 +84,7 @@ const {
   applyLinuxExplicitQuitPromptBypassPatch,
   applyLinuxExplicitTrayQuitPatch,
   applyLinuxQuitGuardPatch,
+  applyLinuxSqliteBackfillQuitClosePatch,
   applyLinuxWillQuitDrainTimeoutPatch,
 } = require("./patches/impl/main-process/quit-lifecycle.js");
 const {
@@ -905,6 +906,7 @@ test("default core patch descriptors are grouped and unique", () => {
     "linux-quit-guard",
     "linux-ready-to-show-window-state",
     "linux-explicit-quit-prompt-bypass",
+    "linux-sqlite-backfill-quit-close",
     "linux-explicit-quit-drain-timeout",
     "linux-explicit-tray-quit",
     "linux-explicit-ipc-quit",
@@ -983,6 +985,15 @@ test("default core patch descriptors are grouped and unique", () => {
 
   assert.equal(new Set(ids).size, ids.length);
   assert.deepEqual([...ids].sort(), [...expectedIds].sort());
+  const lifecycleOrder = [
+    "linux-quit-guard",
+    "linux-explicit-quit-prompt-bypass",
+    "linux-explicit-quit-drain-timeout",
+    "linux-tray",
+    "linux-single-instance",
+    "linux-launch-actions",
+  ].map((id) => ids.indexOf(id));
+  assert.ok(lifecycleOrder.every((index, position) => position === 0 || lifecycleOrder[position - 1] < index));
   assert.ok(descriptors.every((descriptor) => descriptor.sourcePath.includes(`${path.sep}core${path.sep}`)));
   assert.equal(
     descriptors.find((descriptor) => descriptor.id === "package-desktop-name")?.phase,
@@ -1179,7 +1190,7 @@ test("subagent nickname metadata descriptor follows upstream metadata bundle nam
 
 function trayBundleFixture() {
   return [
-    "async function Hw(e){return process.platform!==`win32`&&process.platform!==`darwin`?null:(zw=!0,Lw??Rw??(Rw=(async()=>{let r=await Ww(e.buildFlavor,e.appBrand,e.repoRoot),i=new n.Tray(r.defaultIcon);return i})()))}",
+    "var ire=!1;function G9(){zw=!1}async function Hw(e){return process.platform!==`win32`&&process.platform!==`darwin`?null:(zw=!0,Lw??Rw??(ire||=(n.app.on(`before-quit`,()=>{G9()}),!0),Rw=(async()=>{let r=await Ww(e.buildFlavor,e.appBrand,e.repoRoot),i=new n.Tray(r.defaultIcon);return i})()))}",
     "async function Ww(e,t,i){if(process.platform===`darwin`){return null}let r=K9(e,t,i);return r==null?{defaultIcon:await n.app.getFileIcon(process.execPath,{size:`small`}),chronicleRunningIcon:null}:{defaultIcon:r,chronicleRunningIcon:null}}",
     "function K9(e,t,r){let a=[(0,i.join)(r,`electron`,`src`,`icons`,`tray.png`)];for(let e of a){let t=n.nativeImage.createFromPath(e);if(!t.isEmpty())return t}return null}",
     "var pb=class{nativeTrayClickSuppressionReason=null;clearNativeTrayClickSuppressionTimeout=null;chronicleTrayIconRefreshInterval=null;chronicleTrayIconState=`default`;isNativeTrayMenuOpen=!1;trayMenuThreads={runningThreads:[],unreadThreads:[],pinnedThreads:[],recentThreads:[],usageLimits:[]};constructor(){this.tray={on(){},setContextMenu(){},popUpContextMenu(){}};this.onTrayButtonClick=()=>{};this.tray.on(`click`,()=>{this.onTrayButtonClick()}),this.tray.on(`right-click`,()=>{this.openNativeTrayMenu()})}async handleMessage(e){switch(e.type){case`tray-menu-threads-changed`:this.trayMenuThreads=e.trayMenuThreads;return}}openNativeTrayMenu(){this.updateChronicleTrayIcon();let e=n.Menu.buildFromTemplate(this.getNativeTrayMenuItems());e.once(`menu-will-show`,()=>{this.isNativeTrayMenuOpen=!0}),e.once(`menu-will-close`,()=>{this.isNativeTrayMenuOpen=!1,this.handleNativeTrayMenuClosed()}),this.tray.popUpContextMenu(e)}updateChronicleTrayIcon(){}getNativeTrayMenuItems(){return[]}}",
@@ -1210,8 +1221,12 @@ function explicitQuitBundleFixture() {
 
 function beforeQuitConfirmationBundleFixture() {
   return [
-    "n.app.on(`before-quit`,o=>{let s=BI(),c=t.sr().some(e=>e.status===`ACTIVE`);if(e||i.canQuitWithoutPrompt()||r||!s&&!c){g=!0,a.markAppQuitting();return}let l=n.app.getName();if(n.dialog.showMessageBoxSync({type:`warning`,buttons:[`Quit`,`Cancel`],defaultId:0,cancelId:1,noLink:!0,title:`Quit ${l}?`,message:`Quit ${l}?`,detail:vB({hasInProgressLocalConversation:s,hasEnabledAutomations:c})})!==0){o.preventDefault();return}i.markQuitApproved(),g=!0,a.markAppQuitting()});",
+    "n.app.on(`before-quit`,o=>{let s=BI(),c=t.sr().some(e=>e.status===`ACTIVE`);if(e||i.canQuitWithoutPrompt()||r||!s&&!c){g=!0,a.markAppQuitting();return}let l=n.app.getName(),u=h.H(),f=u.formatMessage({messageId:x6,defaultMessage:S6,values:{appName:l}});if(n.dialog.showMessageBoxSync({type:`warning`,buttons:[u.formatMessage({messageId:`desktop.quitConfirmation.quit`,defaultMessage:`Quit`}),u.formatMessage({messageId:`desktop.quitConfirmation.cancel`,defaultMessage:`Cancel`})],defaultId:0,cancelId:1,noLink:!0,title:f,message:f,detail:K6({nativeIntl:u,appName:l,hasInProgressLocalConversation:s,hasEnabledAutomations:c})})!==0){o.preventDefault();return}i.markQuitApproved(),g=!0,a.markAppQuitting()});",
   ].join("");
+}
+
+function sqliteBackfillQuitCloseBundleFixture() {
+  return "s.on(`close`,t=>{if(!(i||o)){t.preventDefault(),o=!0;try{e()}finally{o=!1}}})";
 }
 
 function willQuitDrainBundleFixture() {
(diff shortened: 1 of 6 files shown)


Yo-DDV (changes requested):
Hi @ilysenko, I found one blocking lifecycle regression in the final diff.

In `scripts/patches/impl/main-process/quit-lifecycle.js` (the generated branches at lines 69, 82, and 88), the hard-exit watchdog is armed before Electron has finished dispatching the remaining `before-quit` listeners and before window `beforeunload` handlers have had a chance to cancel. I reproduced that ordering against the generated helper: a later listener called `preventDefault()`, `codexLinuxQuitCommitted` remained false and the tray stayed alive initially, but the watchdog remained armed. Firing it then destroyed the tray and called `app.exit(0)` / `process.exit(0)`.

Electron v42.1.0 explicitly allows `before-quit` and window `beforeunload` to cancel `app.quit()`, while `app.exit()` bypasses those hooks. This turns a canceled pre-commit quit into forced termination eight seconds later. It contradicts the PR's recoverability invariant and can discard user state.

Please keep the watchdog unarmed until the chosen commit point, or add a proven reset/disarm path for pre-commit vetoes, and cover a later-listener or window cancellation after the patched listener has run. I also tested the PR merged onto the current `main`: the 383 patch tests, 565 feature tests, and script smoke suite all pass. I did not find another blocker in the tray, launch-action, patch-drift, cleanup, or current-main integration paths.

OmerFarukOruc:
@Yo-DDV You are right. The pre-commit arms are gone.

**No disarm path existed.** `codexLinuxQuitWatchdogTimer` only ever got assigned, never cleared. The one `clearTimeout` under `scripts/patches/impl/` sits in `browser.js:469` and belongs to xdg-open. Once any of those three branches armed the timer, it fired.

**Why they were there.** An earlier max-reasoning review raised the inverse hole: accept the quit at `before-quit`, let a window veto its own close, `will-quit` never fires, and nothing backstops an app that already marked itself quitting. I hedged by arming at the accepted branch. That same reviewer later walked the rebuilt ASAR and cleared the scenario as unreachable in the current bundle. `markAppQuitting()` disables the close-to-tray veto before any window closes. The SQLite-backfill window reissues `app.quit()` itself. The bootstrap-installer veto is macOS-only. The bundle carries no renderer `beforeunload` cancellation and no `will-prevent-unload` handler. So the arms guarded nothing reachable, and they opened the hazard you reproduced.

**What changed.** I dropped the arm call from the literal `beforeQuitPatch` branch and from the semantic `beforeQuitRegex` replacement. Both sites still inject the bypass expression, so the prompt bypass survives untouched. I deleted `acceptedPromptRegex` and its `.replace()` block: with the arm gone it emitted text byte-identical to its match while still marking a `required-upstream` patch as applied. I deleted the second drift warning too, since it keyed on `markQuitApproved(),typeof codexLinuxArmQuitWatchdog`, a string the patch no longer emits.

**Grep of the rebuilt `app.asar`.** `codexLinuxArmQuitWatchdog` = 2 (the definition, plus the single call inside `codexLinuxCommitQuit`, which runs on `will-quit`). `markQuitApproved(),typeof codexLinuxArmQuitWatchdog` = 0. `codexLinuxQuitCommitted=!1` = 1. The accepted `before-quit` branch is back to a bare commit with the bypass prefix intact:

```js
codexLinuxShouldBypassQuitPrompt())||e||r.canQuitWithoutPrompt()||t||!s&&!l){_=!0,i.markAppQuitting();return}
```

**Why #939 stays closed without a pre-commit watchdog.** Two independent things hold it. The Linux drain finalizer ends in `app.exit(0)` then `process.exit(0)`, and `Browser::Exit()` skips both `HandleBeforeQuit()` and `NotifyAndShutdown()` ([browser.cc#L125-L148](https://github.com/electron/electron/blob/v42.1.0/shell/browser/browser.cc#L125-L148)), so no listener gets a second chance to stall a committed quit. And `codexLinuxCommitQuit` still arms the watchdog at `will-quit`, which covers anything that hangs during teardown.

**The one cost.** A quit accepted at `before-quit` that then stalls before `will-quit` fires now has no backstop. That stall has one reachable shape, and it is an Electron runtime fault: Electron closes the BrowserWindows, a renderer never finishes its `beforeunload`/`unload` close handshake, and `will-quit` never fires. No application code in the rebuilt bundle holds a quit open at that point. An unresponsive renderer still can. I took that trade over forcing termination on a quit the user cancelled.

**Tests.** `scripts/patch-linux-window-ui.test.js` gains "a later before-quit veto keeps an accepted explicit quit fully recoverable", which drives your exact sequence: accept the quit, let a later `before-quit` listener call `preventDefault()`, advance timers past the 8s window, then assert the tray survives and `app.exit` never runs. Restore a pre-commit arm and it fails. Remove it and it passes.

**One more thing from the same pass.** The prompt-bypass drift warning could never fire. It keyed on the ``showMessageBoxSync({type:`warning`,buttons:[`Quit`,`Cancel`]`` literal, and upstream has since localized those buttons, so the AND was always false on a `required-upstream` patch. It now recognizes the confirmation site through `canQuitWithoutPrompt()` or through the `showMessageBoxSync` and `markQuitApproved()` pair, so a rename of either anchor surfaces as drift instead of a silent already-applied.

**Validation.**

| check | result |
|---|---|
| `node --test scripts/patch-linux-window-ui.test.js` | 385 pass |
| `node --test linux-features/*/test.js` | 560 pass |
| `bash tests/scripts_smoke.sh` | pass |
| `bash -n` install.sh, launcher template, `scripts/lib/*.sh`, 4 build scripts | pass |
| `cargo check -p codex-update-manager` | clean |
| `cargo test -p codex-update-manager` | 242 pass |

`./install.sh ./Codex.dmg` exits 0 with verdict `accepted_with_warnings` and `"blockers": []`. The only warning is the pre-existing `linux-tooltip-window-controls-collision`, which drifted before this change and stays untouched by it. No patch in this change reports drift.

Thank you for running it merged onto current `main`. Your 565 feature tests against my 560 tells me `main` gained tests after I branched.

ilysenko (changes requested):
Thanks for fixing the pre-commit watchdog issue. One lifecycle blocker remains.

The new test claims that a later `before-quit` veto leaves the app fully recoverable, but the upstream handler has already called `markAppQuitting()`. In the current 26.707.71524 bundle, that sets `WindowManager.isAppQuitting = true` and disposes the quick-chat controller. The test then checks close-to-tray using a separate mock with `isAppQuitting: false`, masking the real post-veto state.

Please either avoid/recover that irreversible upstream state before the commit point, or prove that this veto is unreachable in the current bundle and narrow the recoverability claim and test accordingly. The regression coverage should carry the actual post-`markAppQuitting()` state into the post-veto behavior check.

joshyorko:
@OmerFarukOruc  Just reviewed , came to same conculsion as Gary @ilysenko .  Lets GTM.  YOLO

OmerFarukOruc:
@ilysenko Addressed in `6c666b4`. Your review was correct: the old test recorded `markAppQuitting()` on one mock, then asserted recoverability through a separate `isAppQuitting: false` mock.

I refreshed the upstream DMG and rebuilt against `26.707.72221` (Electron 42.1.0), newer than the requested `26.707.71524`. The real bundle audit found a reachable cancellation path: the SQLite backfill progress window prevents its outer `close` and calls `app.quit()` recursively. A clean Electron 42.1.0 reproduction recorded `before-quit -> close-outer -> timeout`, with no `will-quit`; the nested quit is ignored while the first quit is in progress.

The patch now uses reversible `ATTEMPTING` state before commit. Both accepted confirmation branches defer `markAppQuitting()` and quick-chat disposal in a commit callback. A later `before-quit` veto clears the attempt on the event microtask. `will-quit` changes the state to `COMMITTED`, runs the callback once, destroys the tray, and arms the watchdog. The SQLite backfill close listener now allows its window to close during an accepted Linux attempt instead of cancelling and re-entering quit.

I removed the synthetic “fully recoverable” test. Its replacement carries one mutable upstream state object through veto and retry, proving the veto leaves `isAppQuitting` false, quick-chat undisposed, the tray alive, and close-to-tray usable. A separate behavioral test matches the actual SQLite close listener.

Validation:

- Fresh DMG acceptance: `accepted_with_warnings`, zero required blockers; all six lifecycle descriptors are `applied` under `required-upstream`.
- `node --test scripts/patch-linux-window-ui.test.js`: 386/386 pass.
- `node --test linux-features/*/test.js`: 560/560 pass.
- `bash tests/scripts_smoke.sh`: pass.
- Required patch-report validation: pass.

I also narrowed the PR description to the listener graph audited in `26.707.72221` and removed the old blanket recoverability claim.

PR #961, opened 2026-07-14
Add reliable browser reload shortcuts on Linux

KyongSik-Yoon:
## Purpose

Allow the browser/app panel inside the Linux desktop build to reload after external content changes, without requiring a full ChatGPT restart.

## Implementation

- Inject a small capture-phase `Ctrl+R` handler into the webview entry bundle. `Ctrl+Shift+R` follows the same reload path.
- Keep the handler independent of the upstream Keyboard Shortcuts settings route, so it still applies when that route is re-chunked or unavailable.
- Preserve the existing Keybinds settings, persisted overrides, and physical-key fallback behavior.

## Effect

Users can refresh the currently open app/web panel with the standard shortcuts instead of restarting the desktop app.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` — 369 passing


ilysenko:

avifenesh (commented):

---

I found one behavior regression in the new reload shortcut handler.

scripts/patches/impl/keybinds-settings.js
```diff
@@ -1068,6 +1079,40 @@ function applyKeybindsSettingsIndexPatch(currentSource) {
   return applyLinuxKeybindOverridesRuntimePatch(patchedSource);
 }
 
+function applyLinuxBrowserReloadShortcutRuntimePatch(currentSource) {
+  const runtimeMarker = ";function codexLinuxBrowserReloadShortcutRuntime()";
+  const runtimePatch = `;function codexLinuxBrowserReloadShortcutRuntime(){try{if(typeof window=="undefined"||window.__codexLinuxReloadShortcutInstalled)return;window.__codexLinuxReloadShortcutInstalled=!0;window.addEventListener("keydown",event=>{if(event.defaultPrevented||event.repeat||event.altKey||event.metaKey||!event.ctrlKey||event.code!=="KeyR")return;event.preventDefault();event.stopImmediatePropagation();window.location.reload()},!0)}catch{}}codexLinuxBrowserReloadShortcutRuntime();`;
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This unconditional capture-phase handler bypasses the existing keybind settings/override path. `reloadBrowserPage` and `hardReloadBrowserPage` are already configurable/persisted shortcuts, but this listener reloads on every `Ctrl+R`/`Ctrl+Shift+R` without checking the stored overrides (or whether those actions were remapped) and then stops propagation. A user who has remapped browser reload away from `Ctrl+R` would still get a reload here. Please gate this on the same persisted shortcut state, or only install the fallback when no override/remap exists.

KyongSik-Yoon: Implemented in `4a98d2b` (building on the earlier reload-routing commits).

I removed the capture-phase, hardcoded `Ctrl+R` / `Ctrl+Shift+R` fallback entirely. The Linux implementation now leaves the upstream native menu accelerator resolution untouched, so `reloadBrowserPage` and `hardReloadBrowserPage` continue to use their persisted/remapped shortcut arrays as the single source of truth.

On Linux, the native command callback is enabled without requiring an active Browser Sidebar page and routes:
- reload -> the app `BrowserWindow.reload()`
- hard reload -> `BrowserWindow.webContents.reloadIgnoringCache()`

The ASAR patch is fail-soft and only applies when it finds exactly one correlated native reload command pair. It supports the current upstream `c.webContents.getFocusedWebContents()` shape while rejecting ambiguous, computed, or call-expression providers.

Validation:
- `node --test scripts/patch-linux-window-ui.test.js` — 380/380 passed
- `PATH=/usr/bin:/usr/sbin:/bin:/sbin:$PATH bash tests/scripts_smoke.sh` — passed
- in-memory patch proof against the current extracted `app.asar` main bundle — applied and inserted the Linux reload marker
- manual Linux verification confirmed both reload actions work with remapped shortcuts

ilysenko (approved):
Current head reviewed. The previous capture-phase shortcut blocker is resolved; native remap-aware accelerators remain authoritative, and the Linux reload patch is optional/fail-soft.

PR #976, opened 2026-07-14
fix(remote): suppress rollout-forced reasoning summaries

Mizoreww:
## Summary

- Prevent Linux Remote turns on the local Desktop host from inheriting rollout-forced reasoning summaries when no per-turn summary is requested.
- Preserve explicit per-turn summaries and non-local/Remote SSH behavior.
- Add an opt-in `remote-mobile-control` descriptor, drift warnings, idempotence coverage, and documentation.

## Why

The closed-source Codex Desktop runtime can enable its reasoning-summary rollout even when the local Codex configuration disables summaries. On Linux Remote, that produced repeated English reasoning-title cards in the mobile transcript instead of the useful progress/detail stream.

The patch is scoped to the Linux wrapper's opt-in Remote feature. It only applies on Linux, for the local host, when the turn did not explicitly request a summary. It does not change the upstream event stream or Remote SSH sessions.

## Validation

- `node --test linux-features/remote-mobile-control/test.js` (100/100)
- `node --test scripts/patch-linux-window-ui.test.js` (368/368)
- `git diff --check`
- Rebuilt Debian package and verified the current 26.707 webview asset reports `linux-remote-mobile-reasoning-summary-none` as `applied` and contains the marker.
- `tests/scripts_smoke.sh` reached an unrelated existing `better-sqlite3-electron-42` assertion; no files in that path are changed by this PR.

### DIFF
--- a/linux-features/remote-mobile-control/README.md
+++ b/linux-features/remote-mobile-control/README.md
@@ -48,6 +48,9 @@ What it changes:
   missing from the local turn state.
 - Recovers stale remote terminal status when `waitingOnUserInput` remains active
   after the matching input request has already cleared.
+- Keeps local Linux Remote turns on `summary = "none"` unless a turn explicitly
+  requests a reasoning summary, preventing Desktop's rollout gate from adding
+  repeated English reasoning titles to the mobile transcript.
 - Keeps Chrome Browser Use available to remote/mobile controlled sessions when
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed the current head. The optional feature patch is scoped, fail-soft, idempotent, and covered by current-DMG and package CI.

PR #979, opened 2026-07-14
Match Linux desktop settings to native styling

PinguuSS:
## Summary

- align the generated Linux settings page, sections, groups, and rows with the current native Settings layout
- replace the custom inline switch styling with the current theme-token and state-class contract
- keep the generated controls self-contained while matching upstream spacing, typography, colors, borders, and interaction cursor

## Root cause

The generated Linux settings fallback components still used an older layout and a custom inline switch palette. After the Settings runtime moved to self-contained generated controls, those values no longer tracked the current native component contract, so Linux desktop looked visually separate from the rest of Settings and enabled switches could render with the wrong theme color.

## Runtime validation

Validated side-by-side on KDE Plasma Wayland with:

- ChatGPT Desktop 26.707.72221
- Electron 42.1.0
- DMG SHA-256: `40e34814e74e30943c209ebd4da94cd4de3581a52c5bffbe2bcf2e488d6361c6`
- DMG ETag: `0x8DEE17F53C306BB`
- DMG size: `615731826` bytes

The Linux desktop page now uses the same panel width, typography, row spacing, borders, active switch color, thumb position, and pointer cursor as General. No error boundary or console errors were observed.

## Tests

- `node --check` on all generated Linux settings assets
- isolated current-DMG rebuild and patch-report acceptance
- `git diff --check`


ilysenko (approved):
Reviewed current head fe78ef50. The Linux Settings fallback styling change is focused, covered by patcher/current-DMG checks, and I found no blockers.

PR #981, opened 2026-07-14
Recover stalled Browser webview attachment on Linux

tannerpolley:
## Summary

- recover an active Linux Browser webview when Electron never emits `did-attach-webview`
- coordinate one remount and both bounded attachment deadlines in the retained-webview store, including reassociation and overlapping React watchers
- keep success, failure, deactivation, tab removal, host replacement, and stale-callback cleanup deterministic without changing compositor behavior

The failure occurs when Browser Use becomes active but the retained renderer host never reaches `will-attach-webview` or `did-attach-webview`. The renderer now gets two bounded five-second attachment windows within the existing main-process timeout. Recovery state is shared across host/component replacement so duplicate or stale watchers cannot restart, terminate, or extend another host's recovery.

This is distinct from the wrong-window rebound loop reported in #485: the route window remains stable, but the `<webview>` guest never attaches. The patch targets only the latest upstream DMG shape and remains optional, atomic, fail-soft, and idempotent.

## Validation

- `node --test --test-name-pattern='Browser webview|Browser recovery' scripts/patch-linux-window-ui.test.js` (14 passed)
- `node --test scripts/patch-linux-window-ui.test.js` (382 passed)
- patched the current `app-initial~app-main~onboarding-page-CIkoyvFz.js` asset twice, verified byte-identical second application, and passed `node --check`
- `git diff --check`
- fresh post-commit review passes for lifecycle/races, current-DMG patching, and test coverage

Not tested in this isolated worktree: a fresh live-app X11 capture with both GPU and software compositing. The change is limited to missing-attachment recovery and does not attempt to address the separate blank/tiled compositor symptom.

## Checklist

- [x] I followed [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), kept the change focused, and edited source files rather than generated output.
- [x] The patch targets only the latest `CODEX.DMG` shape.
- [x] I added and ran relevant local tests; required CI is pending on this head.
- [x] I reviewed the final diff, addressed all findings, and reran the relevant tests.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -7761,38 +7761,40 @@ test("patches the current Browser webview store and host atomically", () => {
     patched,
     /let r=Ef\(e,t\);return this\.webviews\.get\(r\)!==n/,
   );
-  assert.match(patched, /linuxBrowserUseRemountKeys\.has\(r\)/);
+  assert.match(patched, /linuxBrowserUseRemountDeadlines\.has\(r\)/);
+  assert.match(patched, /linuxGetWebviewRemountDeadline\(e,t\)/);
   assert.match(
     patched,
-    /r\|\|this\.linuxBrowserUseRemountKeys\.delete\(Ef\(e,n\)\)/,
+    /r\|\|this\.linuxBrowserUseRemountDeadlines\.delete\(Ef\(e,n\)\)/,
   );
   assert.match(
     patched,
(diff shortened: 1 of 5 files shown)


ilysenko:

ilysenko (approved):
Reviewed the final head after the fix/review cycles. Current-DMG patching, lifecycle races, focused/full tests, packaging smoke tests, and all required CI checks are clean.

PR #984, opened 2026-07-15
Fix Pet overlay dragging on KDE Plasma Wayland

### LINKED ISSUE #947: Current DMG drift rejects enabled Pet Overlay feature
## Summary

The current DMG changed the avatar overlay `endDrag` completion shape. When the optional `pet-overlay` feature is enabled, its Niri transport hook no longer matches, the feature discards its partial changes, and candidate acceptance correctly rejects the Linux build.

## Current upstream input

- App version: `26.707.62119`
- DMG SHA-256: `c243c94f8de6a51f5530ffe1f8d0c1588733d890ac692e34aaca06d95ba637ca`
- Environment reproduced: Bluefin Linux, local upstream-release acceptance profile

## Actual behavior

```text
feature:pet-overlay:pet-overlay-main
status: skipped-optional
WARN: Could not identify current avatar overlay drag completion shape - skipping Niri transport hook
```

Because the feature is enabled, this is reported as `enabled-feature-drift` and the candidate is rejected instead of replacing the working app.

## Local evidence packet

The current main bundle no longer docks directly after persisting/reclamping the window. It now computes presentation geometry and calls `dockPresentation` only when a `shouldDock` helper accepts the target. The existing completion regex expects the previous direct call shape, so the required Niri completion callback is not installed.

A focused current-DMG probe against `main-NFVREb5D.js` after retargeting the completion shape reports:

```json
{"id":"pet-overlay","changed":true,"idempotent":true,"warnings":[]}
```

Focused Pet Overlay tests, the complete Linux feature test set, and the patch-runner suite pass with the current shape. The obsolete explicit legacy avatar-overlay fixture and contract test can be removed under the current-DMG-only policy.

## Expected behavior

Retarget the Niri drag completion hook to the current geometry-gated dock sequence while preserving the complete upstream sequence inside the deferred completion callback and fallback path. The patch must remain fail-closed, atomic, and idempotent.

## Scope

- Current DMG only; no fallback regex for the previous completion shape.
- `linux-features/pet-overlay/` plus its shared current-bundle fixture.
- No unrelated Pets settings, custom pet asset, or bundled-skill changes.

markjouh:
## Summary

On KDE Plasma Wayland, the optional Pet overlay was visible, but KWin did not reliably honor Electron's client-side top-order and position changes. The Pet could fall behind other windows, and renderer-driven dragging jumped or bounced because absolute window moves crossed the app/compositor boundary. KWin's generic interactive move action also warped the cursor to the center of the full transparent overlay instead of preserving the point that was grabbed.

This change is intentionally limited to `linux-features/pet-overlay/`:

- add an opt-out KWin path that applies keep-above, all-desktop, task-switcher, border, and locked-position hints to the single window matching the exact Pet title and current PID
- run live Plasma dragging in one short-lived KWin script, preserving the original cursor-to-window offset and moving on KWin's compositor cursor signal without spawning a process for each pointer update
- start renderer drags only from the existing visible mascot and notification-tray hit regions, leaving the transparent remainder out of the drag target
- retarget the shared drag-completion hook to the current DMG's geometry-gated docking tail so the enabled feature applies atomically again; no obsolete completion-shape fallback is retained

The feature remains disabled by default. The KWin behavior runs only in a detected KDE/Plasma session when `kwin` is enabled, can be disabled with `CODEX_PET_OVERLAY_KWIN=0`, and falls back to the existing Electron path if KWin scripting or qdbus is unavailable. Hyprland and Niri transports are otherwise unchanged. No core, installer, package, Remote, or generated files are modified.

Fixes #947.

## Maintenance and safety

- KWin scripts reject foreign or ambiguous windows before changing geometry or window state.
- Temporary script files use private permissions and collision-resistant names, and normal completion unloads the script and removes the file.
- Pointer-rate movement stays inside KWin; qdbus is used only to load, start, and unload the drag script.
- Both patches are idempotent and fail soft. The renderer patch is scoped to the one handler owning `avatar-overlay-drag-start`; structural upstream drift warns and leaves the asset unchanged instead of patching another pointer handler.
- Minified symbol and hashed asset-name changes remain supported. A real structural upstream change intentionally requires review through the existing enabled-feature drift gate.

## Validation

- `node --test linux-features/pet-overlay/test.js` — 64/64 passed
- `node --test linux-features/pet-overlay/test.js scripts/patch-linux-window-ui.test.js` — 432/432 passed
- `bash tests/scripts_smoke.sh` — passed, including Debian/RPM/pacman/AppImage packaging smoke coverage
- `./scripts/rebuild-candidate.sh` against app `26.707.72221`, DMG SHA-256 `40e34814e74e30943c209ebd4da94cd4de3581a52c5bffbe2bcf2e488d6361c6` — `accepted_with_warnings`; both Pet descriptors applied. The two warnings are unrelated existing optional core drift for the automation-update tool bundle and tooltip bundle.
- `node scripts/ci/validate-patch-report.js dist-next/rebuild/patch-report.json` — required patch validation passed
- Manual Plasma Wayland verification on Ubuntu 26.04, KWin 6.6.5, 125% display scale: Pet remained above normal windows, followed direct mouse movement smoothly, kept the original grab point under the cursor, and ignored drag starts in the transparent overlay area
- `git diff --check` — passed

The repository-wide Node run reached 1081/1082 on this host. The only failure is the unchanged AppShots XInput timing test (`bare modifier monitor emits one transition from one XInput2 stream`), which also fails when `linux-features/appshots/test.js` is run alone; no AppShots files are touched. Nix is not installed on this machine, so the Nix check was left to CI.

## Checklist

- [x] The current-DMG drift fix targets only the latest `CODEX.DMG` and does not retain the obsolete completion shape.
- [x] I added relevant regression tests and ran the validation listed above.
- [x] I reviewed the complete final diff at maximum reasoning effort and addressed the findings.
- [x] Required GitHub CI checks pass.

### DIFF
--- a/linux-features/pet-overlay/README.md
+++ b/linux-features/pet-overlay/README.md
@@ -37,6 +37,7 @@ Feature settings can be overridden in the gitignored `features.json` file:
         "lockPosition": false,
         "mode": "interactive",
         "hyprland": true,
+        "kwin": true,
         "niri": true
       }
     }
@@ -60,6 +61,7 @@ the configured screen corner on every layout pass.
 | `lockPosition` | `true` or `false` | `false` | Pins to `gravity` when true; otherwise keeps a visible manual position. |
 | `mode` | `interactive` or `passive` | `interactive` | `passive` makes the pet window non-focusable. |
 | `hyprland` | `true` or `false` | `true` | Uses `hyprctl` only in detected Hyprland sessions. |
+| `kwin` | `true` or `false` | `true` | Uses KWin scripting only in detected KDE Plasma sessions. |
 | `niri` | `true` or `false` | `true` | Uses `niri msg` only in detected Niri sessions. |
 
 Runtime overrides are also supported after restart:
@@ -70,6 +72,7 @@ CODEX_PET_OVERLAY_GRAVITY=bottom-left
 CODEX_PET_OVERLAY_MODE=passive
 CODEX_PET_OVERLAY_LOCK_POSITION=1
 CODEX_PET_OVERLAY_HYPRLAND=0
+CODEX_PET_OVERLAY_KWIN=0
 CODEX_PET_OVERLAY_NIRI=0
 ```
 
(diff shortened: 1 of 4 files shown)


ilysenko (approved):
Reviewed current head 3ef0df1069a413e42a0cdb0a8c576f2ff3ced60f. The KWin handling is scoped to the disabled-by-default pet-overlay feature, fails soft when qdbus is unavailable, and matches only the current process's exact overlay window.

PR #988, opened 2026-07-15
Add an opt-in shared app-server socket for Desktop and SSH clients

### LINKED ISSUE #987: Feature request: opt-in shared app-server socket for Codex Desktop
## Problem

Codex Desktop and an SSH-based Codex client on the same Linux host currently
start separate app-server authorities. The two processes use the same account
and local Codex home, but they expose different live thread namespaces. An SSH
client therefore cannot reliably discover or continue the threads owned by the
running Desktop instance.

This matters for users who keep Desktop as their primary local interface but
also use the existing SSH connection path from another device or client. The
desired behavior is for those clients to attach to Desktop's authority instead
of starting a second authority.

## Expected behavior

Provide an opt-in way for an SSH-launched Codex client to attach to the
app-server authority already owned by a running Desktop instance. Both clients
should then see the same threads, while existing Desktop and SSH behavior stays
unchanged for users who do not enable the feature.

## Possible implementation direction

A local prototype uses a disabled-by-default Linux feature that:

- starts the selected Codex CLI as `app-server --listen unix://PATH` under the
  Desktop process;
- connects Desktop through the stock `codex app-server proxy --sock PATH`
  transport;
- allows another local or SSH-launched stock proxy to attach to the same
  user-private Unix socket;
- passes the app-server WebSocket byte stream through without interpreting,
  filtering, or translating protocol methods;
- scopes the default socket by Linux application identity so side-by-side
  Desktop profiles do not share an authority accidentally; and
- keeps authority lifetime owned by Desktop.

The SSH client should continue invoking the ordinary two-argument
`codex app-server proxy` command. A user-owned wrapper on the host may redirect
only that command to `proxy --sock PATH`, so no client-side protocol changes are
required.

## Compatibility considerations

- The socket should remain a local, owner-only control endpoint. It should not be
  exposed directly over TCP or forwarded as a network service.
- The feature should remain opt-in and disabled by default.
- Remote/SSH hosts and normal local stdio behavior should remain unchanged when
  the feature is disabled.
- Patching should be idempotent and fail soft when the current upstream bundle
  shape changes.
- Reusing the stock Codex authority and proxy commands avoids introducing a
  second app-server protocol implementation.

markus-li:
Fixes #987

## Summary

- add a disabled-by-default `shared-app-server-socket` Linux feature
- let Desktop own one Codex CLI authority listening on a user-private Unix
  socket
- connect Desktop and external clients through the stock
  `codex app-server proxy --sock` transport
- document feature enablement, stable socket configuration, and an exact SSH
  wrapper that redirects only the ordinary two-argument proxy command
- add focused patch, staging, drift, permission, and real-CLI lifecycle tests

## Why

Desktop and an SSH-based Codex connection otherwise start separate app-server
authorities. Even when they use the same account and Codex home, the SSH client
does not attach to Desktop's live thread namespace.

This feature gives both clients one authority without adding a second protocol
implementation. The Desktop runtime still uses its existing WebSocket
transport. The selected Codex CLI supplies both `app-server --listen` and the
stock `app-server proxy --sock` tunnel, so methods, notifications, and unknown
protocol messages are not interpreted or filtered by this feature.

## Design and scope

The feature is contained in `linux-features/shared-app-server-socket/` and is
disabled by default. It adds no core, installer, updater, package-format, or
generated-file changes.

When enabled for the local host transport:

1. Desktop starts the selected `CODEX_CLI_PATH` as
   `app-server --listen unix://PATH`.
2. Desktop opens its existing WebSocket transport through a stock
   `app-server proxy --sock PATH` child.
3. Additional local proxies may attach to that same authority.
4. Disposing Desktop's transport closes its proxies, stops its authority child,
   and removes the socket.

The default path is scoped by `CODEX_LINUX_APP_ID` under `XDG_RUNTIME_DIR`.
`CODEX_LINUX_APP_SERVER_BRIDGE_SOCKET` can select a stable owner-private path
for SSH use. Remote hosts and the normal local fallback are unchanged.

The main-bundle descriptor is optional, idempotent, and narrowly anchored to
the current SSH WebSocket transport and local transport factory. If that
structure drifts, it warns and leaves the bundle unchanged.

## Security

- the authority listens only on a Unix socket
- the authority creates the socket without group or other access
- the documented SSH wrapper redirects only the exact
  `codex app-server proxy` invocation
- the socket is documented as a local control endpoint that must not be exposed
  over TCP or forwarded as a network service
- the feature does not add authentication, protocol parsing, method filtering,
  or a background service independent of Desktop

## Validation

- `CODEX_CLI_PATH=/usr/local/bin/codex node --test linux-features/shared-app-server-socket/test.js`
  - 8 passed, 0 failed
  - includes a real Codex authority, owner-only Unix socket, stock proxy, and
    WebSocket `/rpc` upgrade
- `CODEX_CLI_PATH=/usr/local/bin/codex node --test linux-features/*/test.js`
  - 582 passed, 1 failed
  - the sole failure is the unchanged AppShots XInput2 timing test,
    `bare modifier monitor emits one transition from one XInput2 stream`; it
    reproduces on upstream without this feature
- isolated rebuild from upstream app `26.707.72221`, DMG SHA-256
  `40e34814e74e30943c209ebd4da94cd4de3581a52c5bffbe2bcf2e488d6361c6`
  - acceptance verdict: `accepted_with_warnings`
  - `shared-app-server-socket` descriptor: `applied`
  - no blockers
  - two unrelated existing optional core warnings:
    `automation-update-eager-tool` and
    `linux-tooltip-window-controls-collision`
- manual side-by-side validation with isolated Desktop profiles and external
  SSH clients confirmed shared thread discovery and turn execution
- `git diff --check` passed

Nix and native package builds were not run locally. The feature changes an
optional ASAR descriptor and declarative launcher hook; package-format coverage
will come from the repository's required CI checks.

## Known risk

The feature depends on the current minified local transport factory and SSH
WebSocket adapter shape. Current-DMG tests and the enabled-feature acceptance
gate cover that dependency. Structural upstream drift intentionally warns and
requires review rather than retaining old bundle fallbacks.

## Checklist

- [x] I followed `CONTRIBUTING.md`, kept the change focused, edited source files
  rather than generated output, and removed unrelated changes.
- [x] This is not an upstream-drift fix. The optional patch targets the latest
  `CODEX.DMG` and does not retain obsolete bundle-shape fallbacks.
- [x] I added or updated relevant tests, ran the validation listed above, and
  confirmed that required CI checks pass. Local validation is complete; this
  item can be checked after the public-fork CI run.
- [x] I reviewed the final diff at maximum reasoning effort, addressed the
  findings, and reran the relevant tests.


ilysenko (changes requested):
Thanks for the focused feature and thorough documentation. I found several lifecycle issues that need to be addressed before merge:

- The documented SSH wrapper redirects the proxy, but upstream SSH setup still starts a separate app-server authority. Please configure/document `CODEX_SSH_SKIP_APP_SERVER_BOOT=true` and test the actual SSH lifecycle.
- Please add exclusive socket ownership and serialized authority startup. Startup/dispose must not unlink a live or non-owned socket.
- Handle authority child `error` events so spawn failures reject the transport instead of crashing Electron main.
- Clean up the WebSocket and proxy child on open errors/timeouts.

Please add executable transport-level tests for these failure and cleanup paths; the current real-CLI test exercises the CLI primitives but not the injected Desktop transport.

markus-li:
Thank you, I really appreciate the quick and detailed feedback. I was probably too single-minded about getting the feature working and did not give enough attention to process ownership and failure handling around it.

I have pushed an update that addresses the points you raised:

- serialized authority startup with exclusive socket ownership
- fail-closed handling of existing socket and lock paths
- inode-aware cleanup so replaced paths are not removed
- synchronous and asynchronous child-process failure handling
- cleanup only after confirmed authority exit
- proxy and WebSocket cleanup on errors and timeouts
- corrected documentation for configuring `CODEX_SSH_SKIP_APP_SERVER_BOOT=true` in the remote login environment
- executable transport tests, including a real Codex authority/socket/proxy/WebSocket integration path

The feature remains opt-in and protocol-transparent. I hope this covers what is needed, and I’ll be happy to review and fix anything else you find lacking.

ilysenko (approved):
Current head reviewed. The previous SSH boot, socket ownership, process error, timeout, and cleanup blockers are resolved; the feature remains opt-in, isolated, and covered by focused lifecycle tests.

PR #991, opened 2026-07-15
Fix updater fallback bundle inputs

danielcadev:
## Summary

- include `notification-actions-linux` in fallback update-builder copies
- include `launcher/cli-launch-path.py` in the same fallback path
- extend the fallback bundle fixtures and assertions to cover both required inputs
- bump `codex-update-manager` to `0.10.1`

## Root cause

The packaged update-builder manifest already captures these files, but the fallback `REQUIRED_BUNDLE_FILES` list used by source checkouts without that manifest was stale.

`notification-actions-linux` became a Cargo workspace member in #895, while `launcher/cli-launch-path.py` became an installer input in #962. A fallback rebuild therefore copied the root workspace manifest without the notification crate and later omitted the CLI launch helper expected by `install.sh`.

This caused rebuilds to fail first while Cargo loaded the workspace and then, after working around that missing crate, while the installer copied the CLI launch helper.

## User impact

Updater rebuilds from a manifest-less builder checkout now receive the complete Cargo workspace and launcher inputs instead of failing partway through local package generation.

## Validation

- `cargo test -p codex-update-manager` (261 passed)
- Fedora 44 end-to-end updater rebuild produced the RPM and reached `waiting_for_app_exit`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -560,7 +560,7 @@ dependencies = [
 
 [[package]]
 name = "codex-update-manager"
-version = "0.10.0"
+version = "0.10.1"
 dependencies = [
  "anyhow",
  "chrono",
--- a/updater/Cargo.toml
+++ b/updater/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "codex-update-manager"
-version = "0.10.0"
+version = "0.10.1"
 edition = "2021"
 
 [dependencies]
--- a/updater/src/builder.rs
+++ b/updater/src/builder.rs
@@ -122,12 +122,7 @@ pub async fn build_update_from(
     state.artifact_paths.workspace_dir = Some(workspace.workspace_dir.clone());
     state.save(&paths.state_file)?;
 
-    let feature_config = crate::config::effective_feature_config_path(config);
-    copy_builder_bundle(
-        bundle_source,
-        &workspace.bundle_dir,
-        feature_config.as_deref(),
-    )?;
+    copy_builder_bundle(bundle_source, &workspace.bundle_dir)?;
 
     state.status = UpdateStatus::PatchingApp;
     state.save(&paths.state_file)?;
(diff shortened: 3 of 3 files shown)


ilysenko (changes requested):
Thanks for fixing the two stale fallback inputs. One blocker remains in the same manifest-less fallback: it copies the `global-dictation` feature but not the `global-dictation-linux` source required by its stage hook. Because updater rebuilds still apply the saved enabled-feature configuration, enabling Global Dictation makes this fallback fail while looking for `global-dictation-linux/Cargo.toml`. Please make the fallback include this source when the feature is enabled, without making a disabled optional feature mandatory, and add a regression test covering the enabled configuration.

danielcadev:
Thanks for catching this and for the detailed feedback.

Addressed the requested fallback behavior in `a08157e`, with the follow-up lint cleanup in `e5a8c50`.

The manifest-less fallback now reads the effective saved Linux feature configuration and copies `global-dictation-linux` only when `global-dictation` is enabled. When the feature is disabled, the optional source remains unnecessary.

I also added regression coverage for the enabled configuration while preserving the existing disabled-case assertion. Clippy, the full updater test suite, and the required CI checks pass.

ilysenko:
Sorry for the churn here. We re-checked the review and the earlier request to handle Global Dictation in this PR was our mistake.

This PR's original scope was the manifest-less updater fallback bundle inputs, and the missing `notification-actions-linux` plus `launcher/cli-launch-path.py` entries are valid core requirements. The Global Dictation fallback issue is a separate pre-existing optional-feature follow-up and should not have blocked this focused fix.

I pushed a small revert commit that removes the Global Dictation-specific changes and keeps the original fallback fix intact.

ilysenko (approved):
Reviewed the final d29e2cf0 diff after the revert. The PR is back to the focused manifest-less updater fallback input fix, CI is green, and there are no remaining blockers.

PR #994, opened 2026-07-15
Fix launcher lock on Python without pidfd APIs

### LINKED ISSUE #993: timed out waiting for launcher.lock
Attempting to run this on a clean OpenSUSE install gives the following message:

"WARN: timed out waiting for launcher.lock after 5s; refusing to start another app process."

No other diagnostic information is produced, so it's unclear what causes this. Notably this occurs after less than 1 second, not 5 seconds.

Yo-DDV:
## Summary

- make normal launcher lock acquire and release work without Python pidfd wrappers
- keep pidfd reserved for identity-verified stale Electron termination
- add regression coverage for Python builds that do not expose the pidfd APIs

Fixes #993.

## Root cause

The lock helper exited before publishing lock status when Python lacked `os.pidfd_open` or `signal.pidfd_send_signal`. The launcher then treated the missing status as a `launcher.lock` timeout even though it was not actually waiting on another launch.

## Changes

Normal launcher serialization still uses the parent-death-bound `flock` helper so launcher children cannot inherit the lock. The release path now verifies the current helper PID, start time, parent, and command line before sending that child a normal `SIGTERM`; it does not add a status-file release protocol or use pidfd escalation.

Stale Electron recovery keeps the existing boundary. Terminating an identity-verified stale Electron process still requires pidfd, and the pidfd-unavailable path remains fail-closed.

The shared app-server socket change was removed from this PR so #994 stays focused on #993.

## Validation

- `bash -n launcher/start.sh.template tests/launcher_warm_start_recovery.sh tests/scripts_smoke.sh`
- `shellcheck -s bash -S error launcher/start.sh.template tests/launcher_warm_start_recovery.sh tests/scripts_smoke.sh`
- `bash tests/launcher_warm_start_recovery.sh`
- `CODEX_TEST_DISABLE_WARM_START=1 bash tests/launcher_warm_start_recovery.sh`
- `CODEX_TEST_KILL_DURING_PRELAUNCH=1 bash tests/launcher_warm_start_recovery.sh`
- `CODEX_TEST_DISABLE_PIDFD=1 CODEX_TEST_NORMAL_LOCK_ONLY=1 bash tests/launcher_warm_start_recovery.sh`
- `CODEX_TEST_DISABLE_PIDFD=1 CODEX_TEST_KILL_DURING_PRELAUNCH=1 bash tests/launcher_warm_start_recovery.sh`
- `timeout 1800 bash tests/scripts_smoke.sh`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- Launcher startup no longer requires Python's pidfd wrappers for normal
+  launcher lock acquire and release. Pidfd remains reserved for the
+  identity-verified stale Electron termination path.
 - Approval notifications now preserve the upstream Approve, Approve for
   session, and Decline actions on Linux. A small freedesktop notification
   bridge forwards the action and close signals that Electron's Linux
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -3873,9 +3873,6 @@ parent_pid = int(sys.argv[1])
 lock_path, wait_text, status_path = sys.argv[2:]
 wait_seconds = int(wait_text)
 
-if not hasattr(os, "pidfd_open") or not hasattr(signal, "pidfd_send_signal"):
-    sys.exit(1)
-
 libc_name = ctypes.util.find_library("c") or "libc.so.6"
 try:
     libc = ctypes.CDLL(libc_name, use_errno=True)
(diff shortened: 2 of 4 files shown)


ilysenko (changes requested):
Thanks for jumping on #993. The root cause is identified correctly: the launcher lock helper started requiring Python pidfd wrappers for the normal acquire/release path, and on older distro Python builds it exited before publishing lock status. That made the launcher report a misleading `launcher.lock` timeout even though no real lock wait happened.

There are two blocking concerns with the current shape of this PR.

1. This fixes the symptom, but it keeps and extends the complexity that caused it.

The important boundary should be:

- normal launcher serialization must not depend on pidfd;
- pidfd should only be needed for the narrow stale-Electron termination path;
- if pidfd is unavailable during stale-Electron recovery, that path should fail closed with an explicit message, but ordinary startup should still work.

This PR removes the early pidfd requirement from `acquire_launcher_lock`, which is good. But it then adds a second control protocol around the parent-death lock helper: a private status file, `locked`/`release` states, ownership/mode/content checks, polling, and pidfd escalation if the file-based release does not complete. That is compatible, but it is not simpler. It makes the launcher lock depend on another ad hoc state channel and gives future maintainers one more lifecycle mechanism to reason about.

Please step back and try to reduce the launcher lock model instead of adding another layer to it. If the parent-death helper really has to stay, the PR should explain why a simpler lock ownership/release model is not sufficient. Otherwise, prefer a smaller change that decouples normal lock acquire/release from pidfd without introducing a new status-file release protocol.

2. The shared app-server socket fix is unrelated to #993.

The `ctimeMs` socket identity change may be valid, but it is a separate race in a disabled Linux feature. It should be split into its own PR. Keeping it here mixes a core launcher regression fix with optional feature socket lifecycle behavior, which makes this harder to review and harder to revert safely.

Suggested direction:

- keep this PR focused on #993 only;
- make normal launcher lock acquire/release work without `os.pidfd_open` or `signal.pidfd_send_signal`;
- keep pidfd required only for identity-bound stale Electron termination, with a clear fail-closed diagnostic when unavailable;
- split the shared app-server socket `ctimeMs` fix and test changes into a separate PR;
- add/keep the regression test that simulates Python without pidfd wrappers.

Please also verify the updated PR against https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md, run the relevant launcher tests locally before pushing updates, run an independent complete-diff review with a local LLM before asking for maintainer review, and wait until every required CI check has completed successfully after each update.

Yo-DDV:
@ilysenko thanks for the review. I pushed a smaller version in 7cc47e53.

This is now only the #993 launcher fix. The shared app-server socket change is out of this PR. I also removed the status-file release path; the helper is released through a verified child signal, so normal acquire and release no longer depend on `os.pidfd_open` or `signal.pidfd_send_signal`.

I kept pidfd on the stale Electron termination path only. If pidfd support is missing there, that path still fails closed.

The local launcher variants and full script smoke suite pass, including the missing-pidfd normal-lock case. The required GitHub checks are green on the updated head.

ilysenko (approved):
Reviewed the current head; the previous launcher-lock blockers are addressed.

PR #999, opened 2026-07-15
fix(browser): silence Aura allowlist fallback output

0xdeafbeef:
## Summary

- Stop the Linux Browser Use `/aura/site_status` allowlist fallback from writing its diagnostic endpoint into user-visible tool output.
- Preserve the existing narrow endpoint/error predicate and strict fail-open `false` result.
- Keep patch idempotency with a non-executing marker and add direct behavior coverage for the current two-argument generated `fetchBlocked(url, label)` shape.

The generated Browser client starts its site-status fetch before consulting the hostname cache. On Linux, each rejected Aura request entered the compatibility fallback, whose `console.warn` was forwarded as plaintext in the Browser tool result. This change removes only that diagnostic side effect; it does not alter cache/request ordering or broaden the fallback.

## Validation

- `bash tests/scripts_smoke.sh` — passed with `All script smoke tests passed`
- Focused generated-client coverage verifies strict `false`, zero warnings, byte-identical second application, one marker, original-error propagation for both predicate-negative cases, HTTP non-OK behavior, same-object JSON rejection, and successful response parsing.
- Independent implementation review, automated test ownership, manual QA, and final validation passed with no findings.

Not tested: a live latest-DMG rebuild or installed Browser/Chrome runtime smoke. Four unrelated interactive setup-native smoke cases were skipped because `script(1)` is unavailable in this environment.

## Checklist

### DIFF
--- a/scripts/lib/bundled-plugins.sh
+++ b/scripts/lib/bundled-plugins.sh
@@ -1090,33 +1090,17 @@ import sys
 
 path = Path(sys.argv[1])
 source = path.read_text(encoding="utf-8")
-patterns = [
-    re.compile(
-        r'async fetchBlocked\((?P<url>[A-Za-z_$][\w$]*)\)\{'
-        r'let (?P<response>[A-Za-z_$][\w$]*)=await (?P<fetch>[A-Za-z_$][\w$]*)'
-        r'\((?P=url)\.endpoint,\{method:"GET"\}\);'
-        r'if\(!(?P=response)\.ok\)throw new Error\((?P<format>[A-Za-z_$][\w$]*)'
-        r'\(`Browser Use cannot determine if \$\{(?P=url)\.displayUrl\} is allowed\. '
-        r'Please try again later or use another source\.`\)\);'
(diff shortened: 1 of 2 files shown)


ilysenko (changes requested):
Thanks for the focused fix. The runtime behavior change looks right, but this PR touches the Browser Use generated-client patcher while leaving the old one-argument `fetchBlocked(url)` shape in the patch patterns. This repository targets only the latest upstream `CODEX.DMG`, and patch/drift changes need to remove obsolete generated shapes in the same change. Please drop the legacy one-argument pattern and update the fixtures/tests so the patcher targets only the current two-argument `fetchBlocked(url, label)` shape.

0xdeafbeef:
tested the build, all works, warning spam vanished

ilysenko (approved):
Reviewed current head af643f5fa03f5dceab67ae294d49e0b5ae9aaa4f: the previous current-DMG cleanup blocker is fixed and the rerun CI checks pass.

PR #1000, opened 2026-07-15
Restore Agent Workspaces in current settings navigation

PinguuSS:
## Summary

- patch the current Settings catalog that owns the visible navigation list
- insert `agent-workspaces` into both current catalog representations
- reject mixed or incomplete catalog state without writing any assets

## Problem

The existing patch creates the Agent Workspaces route, metadata, and settings page, but the current shared Settings catalog still omits `agent-workspaces`. The active renderer filters the route out, so the patch report reports the feature as applied while the navigation entry remains invisible.

## Current upstream validation

- base repository: `c06cd5ab3d658becb1cb57f1728c850b3b4a7dd0`
- app: ChatGPT Desktop `26.707.72221`
- Electron: `42.1.0`
- DMG SHA-256: `40e34814e74e30943c209ebd4da94cd4de3581a52c5bffbe2bcf2e488d6361c6`
- DMG size: `615731826` bytes

An isolated build with only `agent-workspace` enabled reports both feature descriptors as `applied`. In a side-by-side runtime, Agent Workspaces appears between Environments and Worktrees and the settings page renders successfully.

The branch was rebased after the prior administrative closure and revalidated against the current `main`. No open or merged PR supersedes this fix.

## Tests

- `node --test linux-features/agent-workspace/test.js` (32 passed)
- `node --test linux-features/*/test.js` (599 passed)
- `node --test scripts/patch-linux-window-ui.test.js` (396 passed)
- current-DMG isolated build and patch-report acceptance
- `node --check` on generated assets containing Agent Workspaces markers
- `git diff --check`

### DIFF
--- a/linux-features/agent-workspace/patch.js
+++ b/linux-features/agent-workspace/patch.js
@@ -1918,6 +1918,40 @@ function isAgentWorkspaceSettingsNavigationBundleSource(currentSource) {
   );
 }
 
+const CURRENT_SETTINGS_CATALOG_SLUGS = "local-environments.worktrees.environments";
+const PATCHED_SETTINGS_CATALOG_SLUGS = "local-environments.agent-workspaces.worktrees.environments";
+const CURRENT_SETTINGS_CATALOG_ITEMS = "{slug:`local-environments`},{slug:`worktrees`}";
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the diff and relevant local tests; looks good.

PR #1009, opened 2026-07-15
Expose GSettings schemas to the Nix launcher

colonelpanic8:
## Summary

- add the GTK 3 and desktop GSettings schema roots to the Nix launcher environment
- preserve any existing `XDG_DATA_DIRS` entries
- add a flake check that verifies both schema roots are embedded in the built wrapper

This lets `gsettings` calls from the launcher resolve their schemas on NixOS without downstream source patches.

## Validation

- `nix build -L .#checks.x86_64-linux.nix-gsettings-schema-wrapper`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,6 +220,7 @@ jobs:
             .#codex-desktop-computer-use-ui
             .#codex-desktop-remote-mobile-control
             .#codex-desktop-computer-use-ui-remote-mobile-control
+            .#checks.x86_64-linux.nix-gsettings-schema-wrapper
             .#checks.x86_64-linux.watchdog-linux-features
           )
           for output in "${outputs[@]}"; do
--- a/flake.nix
+++ b/flake.nix
@@ -345,6 +345,15 @@
           stdenv.cc.cc.lib
           zlib
         ]);
+        gsettingsSchemaPackages = with pkgs; [
+          gsettings-desktop-schemas
+          gtk3
+        ];
+        gsettingsSchemaRoot = pkg:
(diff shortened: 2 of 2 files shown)


ilysenko (changes requested):
Thanks for the focused Nix fix. One runtime issue needs to be addressed before merge.

`makeWrapper --prefix XDG_DATA_DIRS` turns an unset or empty variable into only the two Nix schema roots. This removes the effective XDG defaults (`/usr/local/share:/usr/share`) and can hide system desktop, icon, and MIME data from the launcher. Please prefix the schema roots onto `${XDG_DATA_DIRS:-/usr/local/share:/usr/share}` while preserving explicitly configured entries.

Please also add an executed runtime check covering unset, empty, and explicitly populated `XDG_DATA_DIRS`, and ensure that check is built by CI rather than only evaluated with `nix flake check --no-build`.

colonelpanic8:
Addressed the requested XDG behavior in d0120b8:

- unset or empty `XDG_DATA_DIRS` now falls back to `/usr/local/share:/usr/share` before the Nix schema roots are prefixed
- explicitly populated values are preserved after the schema roots
- the check executes the actual generated wrapper for unset, empty, and populated cases
- CI now explicitly builds `.#checks.x86_64-linux.nix-gsettings-schema-wrapper`

Validated locally with `nix flake check --no-build` and a successful build of the targeted check. The complete CI rerun is green: Rust/smoke, Debian, RPM, pacman, and Nix package builds all passed. Ready for re-review.

ilysenko (approved):
Reviewed current head d0120b84571697b77ef80a428c1177546c706165. The XDG_DATA_DIRS default-preservation blocker is fixed, the wrapper runtime check now covers unset/empty/populated values, and CI is green.

PR #1010, opened 2026-07-15
Fix inactive Browser Use webview attachment on Linux

tannerpolley:
## Summary

- retarget Browser webview recovery to the current split store, host, and hidden-host renderer assets
- allow an inactive Browser Use conversation to mount its hidden webview while a different conversation owns the visible Browser panel
- keep the hidden host suppressed when the same logical Browser tab is already visible, avoiding duplicate webviews
- report optional drift when any matched recovery asset loses its current production needles
- add patch-report regressions for all three recovery descriptors

## Root cause

The Browser Use store was synchronizing correctly, but `HiddenBrowserUseWebviewHost` returned `null` whenever its conversation was not the active route and any Browser panel was open. Browser automation can target an inactive synthetic conversation while the visible task owns the panel. In that state the hidden host chunk loaded, but no `<webview>` was mounted, so neither `will-attach-webview` nor `did-attach-webview` fired and the 15-second wait timed out.

The current upstream DMG also splits the recovery seams across three renderer assets, so the obsolete combined onboarding-bundle descriptor can no longer restore the store and host halves together.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js`: 397/397 passed
- `bash tests/scripts_smoke.sh`: passed
- current upstream DMG `26.715.31925`: isolated build accepted with unrelated optional warnings; all three Browser recovery descriptors applied
- full-restart check: after terminating and relaunching the isolated app, a new task owned by the restarted instance opened `https://example.com/` through Browser and returned `Example Domain` without an attachment timeout
- generated Browser assets passed `node --check`; `git diff --check` passed

The branch is rebased onto the current `main` and the pending full-restart gate is complete.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -8877,6 +8877,60 @@ test("current split Browser webview assets apply all recovery descriptors withou
   }
 });
 
+test("reports drift when current split Browser recovery assets lose their primary needles", () => {
+  const cases = [
+    {
+      assetName:
+        "app-initial~artifact-tab-content.electron~app-main~appgen-settings-page~page~pull-request-r~mxek7o2y-BH5mkLvE.js",
+      patchName: "linux-browser-use-webview-attach-recovery-store",
+    },
+    {
+      assetName:
+        "app-initial~app-main~pull-request-route~onboarding-page~hotkey-window-thread-page~quick-cha~mo2avlln-Be1pn_Z1.js",
+      patchName: "linux-browser-use-webview-attach-recovery-host",
+    },
+    {
+      assetName: "browser-sidebar-hidden-browser-use-webview-host-DbLBblbO.js",
+      patchName: "linux-browser-use-hidden-host-ownership",
+    },
+  ];
+
+  for (const { assetName, patchName } of cases) {
+    const tempRoot = fs.mkdtempSync(path.join(os.tmpdir(), "codex-browser-recovery-drift-"));
+    try {
+      const buildDir = path.join(tempRoot, ".vite", "build");
+      const assetsDir = path.join(tempRoot, "webview", "assets");
+      fs.mkdirSync(buildDir, { recursive: true });
+      fs.mkdirSync(assetsDir, { recursive: true });
+      fs.writeFileSync(path.join(buildDir, "main.js"), "module.exports={}");
+      fs.writeFileSync(path.join(assetsDir, assetName), "function A(){return null}");
+
+      const report = createPatchReport();
+      const corePatchRoot = path.join(
+        __dirname,
+        "patches",
+        "core",
+        "all-linux",
+        "webview",
+        "browser-use-attach-recovery",
+      );
+      captureWarns(() => patchExtractedApp(tempRoot, { report, corePatchRoot }));
+
+      assert.equal(
+        report.patches.find((patch) => patch.name === patchName)?.status,
+        "skipped-optional",
+        patchName,
+      );
+      assert.ok(
+        optionalDriftFromReport(report).some((patch) => patch.name === patchName),
+        patchName,
+      );
+    } finally {
+      fs.rmSync(tempRoot, { recursive: true, force: true });
+    }
+  }
+});
+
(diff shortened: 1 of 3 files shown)


Yo-DDV (changes requested):
Thanks for the patch. I think this still needs one change before it can be treated as fixing the current upstream drift.

The store descriptor added at `scripts/patches/core/all-linux/webview/browser-use-attach-recovery/patch.js:17` does not match the upstream DMG that this PR's own `Build App Against Upstream DMG` job tested. That job passed overall, but its patch summary says `linux-browser-use-webview-attach-recovery-store` was `skipped-optional`, and the final acceptance verdict was `accepted_with_warnings`. So the host half can apply while the store half is still missing, which means the recovery patch is not actually fully restored on the current DMG.

Could you retarget the store descriptor to the real current store chunk and add a regression/report check that catches this case? A green CI job is a bit misleading here because optional patch drift is allowed to pass, but for this PR the skipped store half is the behavior being fixed.

ilysenko:

ilysenko (changes requested):
Thanks for the update. The ownership change itself matches the reported inactive-conversation failure, and the previous descriptor issue was fixed for DMG 26.707.72221.

The currently supported DMG is now 26.715.21425, however. Against the current DMG, `linux-browser-use-webview-attach-recovery` is `skipped-optional` while only `linux-browser-use-hidden-host-ownership` applies. The descriptor still targets the obsolete combined onboarding chunk, and the new “current assets” regression test reproduces that older layout rather than the current split store/host assets.

Please rebase onto current `main`, retarget the recovery descriptors and implementation exclusively to the current DMG layout, remove the obsolete combined-bundle path as required by the current-DMG-only policy, update the patch-report regression with current production shapes, and rerun the upstream build. Please also complete the pending full-restart live Browser attachment validation and mark the PR ready for review.

PinguuSS:
I reproduced the current-DMG failure independently and prepared the requested retarget on top of current `main` plus this PR:

https://github.com/PinguuSS/codex-desktop-linux/commit/a4ca14d

The current `26.715.21425` renderer has three separate assets:

- store: `app-initial~artifact-tab-content.electron~app-main~appgen-settings-page~page~pull-request-r~mxek7o2y-BH5mkLvE.js`
- host: `app-initial~app-main~pull-request-route~onboarding-page~hotkey-window-thread-page~quick-cha~mo2avlln-Be1pn_Z1.js`
- hidden host: `browser-sidebar-hidden-browser-use-webview-host-deXVCWnB.js`

The commit removes the obsolete combined-bundle descriptor, creates current-only store and host descriptors, updates the host lifecycle fixture to the production shape, and adds a patch-report regression requiring all three descriptors to be `applied`.

Validation completed against DMG SHA-256 `ff459150991612007549270d2d28c5e78cec6bd6ac200a7ada5ed6c031369b87` (`26.715.21425`, Electron `42.3.0`):

- current-DMG isolated build: `accepted_with_warnings`
- all three Browser recovery descriptors: `applied`
- second pass over the three generated assets: byte-stable / already applied
- `node --check` on all three generated assets
- `node --test scripts/patch-linux-window-ui.test.js`: 377/377
- `bash tests/scripts_smoke.sh`: passed
- `git diff --check`: passed

Live side-by-side validation also passed: `tabs.new()` returned tab `1` immediately, the webview reached `will attach -> did attach -> runtime attached -> open wait resolved`, and navigation to `https://example.com/` returned `Example Domain`. The installed unpatched build instead disposes/removes the hidden host immediately and times out after 15 seconds.

I restarted the isolated app and confirmed that it relaunched cleanly, but the external test task binding did not re-register with that restarted side-by-side instance. I am therefore not claiming the pending full-restart Browser attachment check as complete; that last check still needs to be run from a task actively owned by the restarted instance.

After rebasing the PR branch onto current `main`, the commit above should be available to cherry-pick.

ilysenko:

ilysenko (approved):
Reviewed the current head after the split current-DMG Browser recovery retarget, patch-report regression coverage, and full-restart validation update. The previous blockers are resolved.

PR #1016, opened 2026-07-16
fix(browser): scope socket directory per Linux user

### LINKED ISSUE #1015: Scope the Browser Use socket directory per Linux user
## Summary

The Chrome extension host and bundled Browser clients use the global Linux socket directory:

```text
/tmp/codex-browser-use
```

The host correctly requires this directory to be owned by the effective UID and restricts it to mode `0700`. However, because the default path is not scoped by UID, the first user who creates it prevents other users on the same Linux system from starting Browser Use.

## Reproduction

1. User A creates or starts Browser Use using `/tmp/codex-browser-use`.
2. The directory becomes owned by User A and is restricted to mode `0700`.
3. User B starts the Chrome extension host with the default configuration.
4. Startup fails during `prepare_socket_dir` because the directory belongs to another UID.

The ownership check can be reproduced by pointing the host at an existing directory owned by another UID:

```text
Error: unix socket directory is owned by uid 0, expected 1000: /tmp
```

## Expected behavior

Each Linux user should receive an isolated Browser Use socket directory without weakening the existing ownership, permission, socket-mode, or `SO_PEERCRED` checks.

A suitable default could be:

```text
$XDG_RUNTIME_DIR/codex-browser-use
```

with a fallback such as:

```text
/tmp/codex-browser-use-$UID
```

The Browser/IAB discovery path and the Chrome extension host must continue resolving the same directory.

## Related work

Issue #841 and PR #843 addressed interference between IAB discovery and active Chrome extension sockets inside the shared `/tmp/codex-browser-use` directory.

This is a separate multi-user isolation issue: those changes prevent one backend from disrupting another for the same user, while this issue concerns ownership conflicts between different Linux users.

## Impact

A local user—or simply the first legitimate user to start Browser Use—can prevent other users on the same machine from using the feature.

This affects availability only. The existing `0700` directory mode, `0600` socket mode, unpredictable socket name, and `SO_PEERCRED` validation continue protecting confidentiality and authorization.

danielcadev:
## Summary

- resolve Browser Use sockets under `$XDG_RUNTIME_DIR/codex-browser-use` on Linux
- fall back to a UID-scoped `/tmp/codex-browser-use-$UID` directory
- keep the Rust extension host plus bundled Browser and Chrome clients on the same path
- preserve `CODEX_BROWSER_USE_SOCKET_DIR` as the explicit override

## Why

The global `/tmp/codex-browser-use` default is created with owner-only permissions. That is correct for one user, but on a multi-user system the first user to create it prevents every other user from starting Browser Use.

This change scopes the default per Linux user without weakening the existing directory ownership, socket permissions, unpredictable socket names, or peer-credential checks. The Windows named-pipe path is unchanged.

Fixes #1015
Related to #841
Follow-up to #843

## Validation

- `node --test scripts/lib/patch-browser-client-iab-socket-scope.test.js` (5 passed)
- `cargo test -p codex-computer-use-linux --bin codex-chrome-extension-host` (60 passed)
- `bash tests/scripts_smoke.sh` (all passed, including Debian, RPM, pacman, AppImage, Browser, and Chrome staging)
- applied the patch helper to the current bundled Browser and Chrome client shapes

`rustfmt` and `clippy` are not installed in the local environment; the required GitHub CI job will run both.

### DIFF
--- a/.github/workflows/upstream-build-app.yml
+++ b/.github/workflows/upstream-build-app.yml
@@ -14,10 +14,14 @@ on:
       - scripts/ci/upstream-dmg-*.js
       - scripts/validate-upstream-dmg.js
       - scripts/lib/candidate-install.sh
+      - scripts/lib/bundled-plugins.sh
+      - scripts/lib/browser-client-node-repl-runtime.test.js
+      - scripts/lib/patch-browser-client-iab-socket-scope.js
       - scripts/lib/patch-validation.js
       - scripts/lib/upstream-dmg-acceptance.js
       - scripts/lib/upstream-dmg-release-profile.js
       - scripts/lib/linux-features.js
+      - computer-use-linux/src/bin/codex-chrome-extension-host.rs
       - linux-features/**
       - .github/workflows/upstream-build-app.yml
   push:
@@ -33,10 +37,14 @@ on:
       - scripts/ci/upstream-dmg-*.js
       - scripts/validate-upstream-dmg.js
       - scripts/lib/candidate-install.sh
+      - scripts/lib/bundled-plugins.sh
+      - scripts/lib/browser-client-node-repl-runtime.test.js
+      - scripts/lib/patch-browser-client-iab-socket-scope.js
       - scripts/lib/patch-validation.js
       - scripts/lib/upstream-dmg-acceptance.js
       - scripts/lib/upstream-dmg-release-profile.js
       - scripts/lib/linux-features.js
+      - computer-use-linux/src/bin/codex-chrome-extension-host.rs
       - linux-features/**
       - .github/workflows/upstream-build-app.yml
   workflow_dispatch:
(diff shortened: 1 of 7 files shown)


ilysenko (changes requested):
Thanks for the focused multi-user isolation fix. I found three runtime blockers:

- The injected socket resolver reads `process.env` and `process.getuid()` inside the packaged `node_repl`, where `process` is intentionally unavailable. Importing the patched staged Browser client through the real runtime fails with `process is not defined`; the unit test masks this by injecting a fake `process`. Please use a supported bridge and add a real-runtime import test for both staged clients.
- The Browser client and Chrome native host resolve `$XDG_RUNTIME_DIR` independently in different processes. If their environments differ, they select different rendezvous directories. Please use one deterministic cross-process path or explicitly pass a shared resolved path.
- Please guard the complete AF_UNIX socket path length and fall back to a short UID-scoped path when `$XDG_RUNTIME_DIR` would exceed `sun_path`.

The existing directory ownership, `0700`/`0600` modes, and `SO_PEERCRED` checks otherwise remain intact.

danielcadev:
Thanks for catching these runtime issues. I rebased onto current main and addressed the three blockers:

- The injected resolver no longer accesses the unavailable global `process` inside `node_repl`; it now uses the supported `node:os` API.
- The Browser client and Chrome native host use the same deterministic UID-scoped default under `/tmp`, independent of `XDG_RUNTIME_DIR`.
- The native host validates the complete AF_UNIX socket path before binding.

I also added a regression test that imports both staged Browser and Chrome clients through the packaged `node_repl` runtime.

Local rustfmt, Clippy, Rust tests, Node tests, smoke tests, actionlint, and the real-runtime import test pass.

ilysenko (approved):
Reviewed current head 8494ee9. The previous runtime blockers are resolved: the Browser and Chrome paths now share a deterministic UID-scoped socket directory, the node_repl client import no longer depends on process, and the native host validates the full Unix socket path while preserving ownership, mode, and SO_PEERCRED checks.

PR #1022, opened 2026-07-16
Normalize Linux Browser sidebar user agent

psypeal:
## Summary

- Normalize only the Linux Browser sidebar webview user agent.
- Set the webview's `params.useragent` during `will-attach-webview`, before its initial navigation.
- Leave `app.userAgentFallback`, the main renderer, Browser Settings webviews, and all session-level user agents unchanged.

## Root cause

The upstream desktop user agent includes `ChatGPT/<version>` and `Electron/<version>` product tokens. Some sites reject or alter their login flow when those tokens are present. The earlier implementation broadened the workaround to the app fallback and newly created web contents, which could affect unrelated renderers and sessions.

This revision uses the Browser sidebar's dedicated webview-preferences seam and writes only that guest's `useragent` parameter before navigation.

## Scope

This PR is intentionally limited to user-agent normalization. Window-open disposition routing and the obsolete navigation-patch cleanup are being moved to a separate focused PR.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` — 401/401 passing.
- Regression tests verify the main renderer, Browser Settings webview, unrelated session, and Browser session default user agent remain unchanged.
- Fresh upstream DMG 26.707.91948 candidate accepted with `linux-browser-use-user-agent: applied`.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -57,6 +57,7 @@ const {
   applyLinuxBundledPluginCopyPermissionsPatch,
   applyLinuxBundledPluginReconcileStaleSnapshotPatch,
   applyLinuxBrowserUseRouteLivenessPatch,
+  applyLinuxBrowserUseUserAgentPatch,
   applyLinuxChromeExtensionStatusPatch,
   applyLinuxExternalOpenEnvPatch,
 } = require("./patches/impl/main-process/browser.js");
@@ -947,6 +948,7 @@ test("default core patch descriptors are grouped and unique", () => {
     "linux-bundled-plugin-reconcile-stale-snapshot",
     "linux-bundled-plugin-copy-permissions",
     "linux-browser-use-route-liveness",
+    "linux-browser-use-user-agent",
     "linux-chrome-extension-status",
     "linux-notification-actions",
     "linux-local-app-server-feature-enablement-handler",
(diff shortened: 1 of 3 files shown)


ilysenko (changes requested):
Thanks for the careful investigation and live validation. I found two blockers:

1. Please scope the user-agent normalization to the Browser Use session or reliably identified Browser guest. `app.userAgentFallback` is global, and the `webviewTag !== true` check does not identify Browser Use contents, so the current implementation can change the main renderer and unrelated web contents or sessions. Please avoid the app-wide fallback and add regression tests proving that unrelated renderers and sessions keep their original user agents.
2. Please split the user-agent normalization and window-open disposition routing into separate focused PRs. They are independent core fixes with separate descriptors, tests, and commits. The obsolete navigation-patch cleanup can stay with the window-open fix.

psypeal:
> Thanks for the careful investigation and live validation. I found two blockers:
> 
>     1. Please scope the user-agent normalization to the Browser Use session or reliably identified Browser guest. `app.userAgentFallback` is global, and the `webviewTag !== true` check does not identify Browser Use contents, so the current implementation can change the main renderer and unrelated web contents or sessions. Please avoid the app-wide fallback and add regression tests proving that unrelated renderers and sessions keep their original user agents.
> 
>     2. Please split the user-agent normalization and window-open disposition routing into separate focused PRs. They are independent core fixes with separate descriptors, tests, and commits. The obsolete navigation-patch cleanup can stay with the window-open fix.

wait a sec, this version still has some problems. I'm polishing it, and I'll test it locally to make sure it run smoothly

psypeal:
Addressed both blockers from the review:

1. The UA normalization is now scoped to the Browser sidebar's own `will-attach-webview` preferences seam. It sets only that guest's `params.useragent` before navigation. It no longer reads or writes `app.userAgentFallback`, registers a global `web-contents-created` listener, or changes any session-level user agent. Regression tests verify the main renderer, Browser Settings webview, unrelated session, and Browser session default UA all remain unchanged.
2. I force-updated this PR to one UA-only commit and moved all window-open disposition/popup/routing work plus the obsolete navigation-patch cleanup to #1030.

Validation for this PR:

- 401/401 patch tests passing.
- Fresh 26.707.91948 DMG candidate accepted.
- Patch report: `linux-browser-use-user-agent: applied`.
- Packaged asset contains no `userAgentFallback` mutation.

ilysenko:
Thanks for the work on this and for narrowing the implementation.

We are going to close this PR because the project currently prioritizes preserving the standard upstream ChatGPT Desktop behavior rather than changing the Browser sidebar user agent.

We are also deliberately limiting the number of downstream patches we maintain. Upstream DMG changes are creating more patch drift than we can reliably keep up with, so new behavior-changing patches need to be reserved for compatibility fixes required to make the standard app work on Linux.

This is a project-scope decision, not a concern with the quality of the implementation. Thank you for the investigation and testing.

PR #1024, opened 2026-07-16
Restore current Computer Use host platform support

colonelpanic8:
## Summary

- rebase onto the latest `main`
- add a focused opt-in descriptor for the current `nrw3o0ql` Computer Use host-platform bundle
- preserve the newer `jgoqfqy2` plugin-detail install-flow patch merged in #1028
- restore an idempotent host-platform patch that recognizes Linux without bypassing the upstream feature and rollout gates

## Why

The current DMG still ships the host-platform contract in the `nrw3o0ql` asset family. After #1028 moved the install-flow descriptor to the plugin-detail contract, the host-platform gate needed its own descriptor and implementation rather than replacing the newer fix.

## User impact

With the Computer Use UI opt-in enabled, Linux is accepted as a compatible host platform while the existing required-feature and rollout gates remain intact.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` (376 passed)
- `git diff --check`
- `bash tests/scripts_smoke.sh` passed Computer Use, package-format, updater, promotion, feature-staging, and launcher warm-start checks; it later stopped in the unrelated launcher CLI policy test because the scrubbed `/usr/bin:/bin` PATH on this Nix host contains no `bash`.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -1035,11 +1035,12 @@ test("default core patch descriptors are grouped and unique", () => {
   const computerUseInstallFlow = descriptors.find((descriptor) => descriptor.id === "linux-computer-use-install-flow");
   assert.equal(
     computerUseInstallFlow.pattern.test(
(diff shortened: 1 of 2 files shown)


colonelpanic8:
Rebased onto current main and resubmitted as #1050.

Issue #952, opened 2026-07-14
Computer Use installs successfully but UI says unavailable and plugin detail is not found on 26.707.62119


vitolob:
## Summary

On Ubuntu 24.04 / GNOME X11, the bundled Linux `computer-use` plugin is staged, registered, and installed successfully, but the UI still reports **Unavailable in this context** and the plugin detail route reports **Plugin not found**.

The Computer Use settings route loads, but it shows only Google Chrome (browser extension not connected); the expected **Any App** / native-app control is absent.

## Environment

- Ubuntu 24.04 (Noble)
- GNOME on X11
- VMware guest
- Native package installation under `/opt/codex-desktop`
- Bundled upstream/plugin build observed in launcher logs: `26.707.62119`
- Computer Use UI opt-in:

```json
{"codex-linux-computer-use-ui-enabled": true}
```

## UI symptoms

1. **Plugins** lists Computer Use with its toggle enabled, but displays:
   - `Unavailable in this context`
   - `Control Linux desktop apps from Codex`
2. **Settings > Computer use** loads, but contains only:
   - Google Chrome
   - `Browser extension not connected`
   - no `Any App` / native desktop app entry
3. Opening the Computer Use plugin detail page produces:
   - `Plugin not found`
   - `This plugin could not be loaded from its marketplace entry.`

## Patch report

All relevant patches report applied:

```text
linux-computer-use-ui-feature: applied
linux-computer-use-plugin-gate: applied
linux-computer-use-native-desktop-apps: applied
linux-computer-use-ui-availability: applied
linux-computer-use-install-flow: already-applied
```

## Marketplace/install evidence

Installed resources and the runtime marketplace entry exist:

```text
/opt/codex-desktop/resources/plugins/openai-bundled/plugins/computer-use
STATUS: OK

~/.codex/.tmp/bundled-marketplaces/openai-bundled/plugins/computer-use
STATUS: OK
```

The launcher's `latest` cache link is missing after synchronization:

```text
~/.codex/plugins/cache/openai-bundled/computer-use/latest
STATUS: MISSING
```

Relevant launcher log excerpts:

```text
Computer Use plugin cache synced from bundled resources:
~/.codex/plugins/cache/openai-bundled/computer-use/0.1.2-linux-alpha2

[BundledPluginsMarketplace] bundled_plugins_runtime_marketplace_written
pluginNames=["sites","browser","chrome","computer-use","visualize"]

[...]

PR #1028, opened 2026-07-16
Fix current Computer Use UI availability

PinguuSS:
## Summary

- retarget the Computer Use Settings and plugin-detail descriptors to the current upstream chunks
- replace the legacy matcher set with current-DMG-only, marker-backed patches
- make the two Settings insertion points atomic so partial matches remain byte-identical and report drift
- cover explicit and persisted opt-in paths, including the path used by updater rebuilds

## User-visible behavior

With the Computer Use UI opt-in enabled on Linux:

- Computer Use appears in the Plugins page
- its plugin detail opens instead of showing unavailable or not found
- Settings > Computer use shows the Any App control alongside Chrome

## Current upstream validation

Validated from repo SHA `b24fa4f506659aa295c79e9fd7b2937c4a3083f2` against:

- ChatGPT Desktop `26.707.91948`
- Electron `42.1.0`
- DMG SHA-256 `5a2ab9689f4ba38fcb135565246d5ca2f124d539336a0a32afcdb72040d21466`
- DMG ETag `0x8DEE2FD72E9886C`
- DMG size `615797208` bytes
- KDE Plasma Wayland

An isolated current-DMG build and an updater-equivalent rebuild candidate using the persisted Settings flag both reported all five Computer Use patches as `applied`. Reapplying the generated assets was byte-stable. Runtime validation covered Plugins list, plugin detail, Settings > Computer use, and restart persistence in a side-by-side instance.

The candidate was accepted with two unrelated optional warnings for tooltip collision and Browser Use webview attachment recovery; no Computer Use descriptor warned or skipped.

## Tests

- `node --test scripts/patch-linux-window-ui.test.js` (385 passed)
- `node --check scripts/patches/impl/computer-use.js`
- `node scripts/ci/validate-patch-report.js dist-next/computer-use-updater-flow/patch-report.json`
- `git diff --check`

Related to #952.


ilysenko:
You closed it at that stage, when the robot was already about to come to merge.

PinguuSS:
> You closed it at that stage, when the robot was already about to come to merge.

It's because there is another PR doing almost the same as mine, so I didn't want to be hungry about PR's

PR #1034, opened 2026-07-16
Restart updater daemon when a package upgrade replaces its binary

subsy:
## Summary

The updater daemon now detects that a package upgrade replaced its binary on disk and exits with a dedicated nonzero status code, so the already-deployed `Restart=on-failure` unit relaunches it on the new binary.

## Problem

Package upgrades replace `/usr/bin/codex-update-manager` while the daemon keeps running the old, now-deleted image. The post-install/post-upgrade hooks deliberately leave an active service untouched (`codex_start_one_enabled_user_service` returns early when the unit is active — reasonable, since the daemon itself may have triggered the pacman transaction and restarting it mid-install would kill the install). The result is that a running daemon survives **every** upgrade until the next reboot.

Observed on Arch (pacman package): a daemon started at boot predated the `record-replay-linux` workspace member (6ed86d4) and the manifest-based rebuild staging (751e912). On every periodic check it staged a builder workspace using its old hardcoded copy list, while copying the current bundle's `Cargo.toml` — which references crates the old list never copies. Every rebuild then failed with:

```
error: failed to load manifest for workspace member `.../builder/record-replay-linux`
Caused by: No such file or directory (os error 2)
```

producing a "update failed" notification every 6 hours, with `/proc/<pid>/exe → /usr/bin/codex-update-manager (deleted)` confirming the stale image. Manually rebuilding (`make update-native`) installs a fresh binary but never restarts the daemon, so the cycle repeats after the next upstream release. Any future divergence between a long-running daemon and the installed builder bundle can re-trigger the same class of failure.

## Fix

- New `updater/src/restart.rs`: reads `/proc/self/exe`, strips the kernel's ` (deleted)` suffix, and compares device+inode of the running image against the installed path. Returns the replacement path only when a different binary actually exists on disk, so package *removal* still flows through the existing packaged-runtime shutdown path instead of a restart loop.
- `run_daemon` checks this at the top of each loop iteration and exits with new code `12` (`BINARY_REPLACED_RESTART_EXIT_CODE`) when a replacement is found. With the 15s reconcile tick, a replaced binary is picked up within seconds regardless of whether the daemon or a manual package install performed the upgrade.

Exiting nonzero (rather than switching the unit to `Restart=always` with a clean exit) was chosen deliberately: it works retroactively on machines still running the old unit file, and it does not interfere with the daemon's intentional clean-exit paths (packaged runtime removed, shutdown signal). No packaging or unit-file changes are needed.

After a package upgrade, the updater daemon restarts itself onto the new binary within ~25 seconds (systemd `RestartSec=10` + reconcile tick) instead of running the old binary until reboot. The journal shows one intentional `exit-code` service restart per upgrade.

## Validation

- `cargo check -p codex-update-manager` and `cargo clippy -p codex-update-manager --all-targets` clean.
- New unit tests in `restart.rs` (6): ` (deleted)` suffix stripping and pass-through, same-inode → no replacement, different-inode → replacement, missing on-disk binary → no replacement, and a live check that the running test binary is not reported as replaced.
- `cargo test -p codex-update-manager`: 268 passed. 5 pre-existing failures (`codex_cli`/CLI-probe tests) fail identically on a clean `main` checkout on this host — they probe the host's real npm-installed Codex CLI and are unrelated to this change.
- Real-world root cause verified on the affected machine: stale daemon (deleted inode) staged workspaces missing `record-replay-linux`/`notification-actions-linux`; after restarting the service onto the current binary, the next check cycle succeeded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- The updater daemon now detects that a package upgrade replaced its binary
+  on disk and exits with a nonzero status so systemd's `Restart=on-failure`
+  relaunches it on the new binary. Previously a running daemon survived every
+  upgrade and kept staging rebuild workspaces with outdated logic, failing
+  each periodic update until the next reboot.
 - Launcher startup no longer requires Python's pidfd wrappers for normal
   launcher lock acquire and release. Pidfd remains reserved for the
   identity-verified stale Electron termination path.
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -5,7 +5,8 @@ use crate::{
     cli::{Cli, Commands},
     codex_cli,
     config::{RuntimeConfig, RuntimePaths},
-    diagnostics, feature_picker, install, install_rollback, liveness, logging, notify, rollback,
+    diagnostics, feature_picker, install, install_rollback, liveness, logging, notify, restart,
+    rollback,
     state::{CliStatus, PersistedState, UpdateStatus},
     upstream, wrapper, wrapper_apply,
 };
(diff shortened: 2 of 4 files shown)


ilysenko (approved):
Reviewed current head 28f03cf. The updater restart logic is narrowly scoped, preserves package-removal clean shutdown, and avoids systemd restart loops while letting Restart=on-failure pick up the replaced package binary.

PR #1041, opened 2026-07-17
Fix upstream DMG drift

ilysenko:
## Summary
- repair Linux compatibility drift for the current upstream DMG
- remove obsolete compatibility paths replaced by the current DMG shape

## Validation
- exact current-DMG acceptance recorded by watchdog v2
- GitHub CI must pass, including Nix Package Builds

### DIFF
--- a/linux-features/open-target-discovery/patch.js
+++ b/linux-features/open-target-discovery/patch.js
@@ -692,12 +692,12 @@ function applyOpenInTargetsBridgeDetectionPatch(currentSource) {
   }
 
   const currentClassMatch = currentSource.match(
-    /async detectTarget\(\{target:([A-Za-z_$][\w$]*)\}\)\{if\(this\.requestOpenInWorker==null\)throw Error\(`Open in worker unavailable`\);let\{command:([A-Za-z_$][\w$]*)\}=await this\.requestOpenInWorker\(\{method:`get-target-command`,params:([A-Za-z_$][\w$]*)\(this\.settingsStore,\1\)\}\);return\{available:\2!=null\}\}/u,
+    /async detectTarget\(\{target:([A-Za-z_$][\w$]*)\}\)\{let\{command:([A-Za-z_$][\w$]*)\}=await this\.#[A-Za-z_$][\w$]*\(\)\(\{method:`get-target-command`,params:([A-Za-z_$][\w$]*)\(this\.settingsStore,\1\)\}\);return\{available:\2!=null\}\}/u,
   );
   if (currentClassMatch != null) {
     const [needle, targetVar, commandVar, paramsFn] = currentClassMatch;
     const replacement =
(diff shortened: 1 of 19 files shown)


linux-features/open-target-discovery/patch.js
```diff
@@ -692,12 +692,12 @@ function applyOpenInTargetsBridgeDetectionPatch(currentSource) {
...
   );
   if (currentClassMatch != null) {
     const [needle, targetVar, commandVar, paramsFn] = currentClassMatch;
     const replacement =
-      `async detectTarget({target:${targetVar}}){if(process.platform===\`linux\`){let ${commandVar}=await codexLinuxOpenTargetRegistryCommand(this.settingsStore,${targetVar});return{available:${commandVar}!=null}}if(this.requestOpenInWorker==null)throw Error(\`Open in worker unavailable\`);let{command:_codexWorkerCommand}=await this.requestOpenInWorker({method:\`get-target-command\`,params:${paramsFn}(this.settingsStore,${targetVar})});return{available:_codexWorkerCommand!=null}}`;
+      `async detectTarget({target:${targetVar}}){if(process.platform===\`linux\`){let ${commandVar}=await codexLinuxOpenTargetRegistryCommand(this.settingsStore,${targetVar});return{available:${commandVar}!=null}}let{command:_codexWorkerCommand}=await this.#n()({method:\`get-target-command\`,params:${paramsFn}(this.settingsStore,${targetVar})});return{available:_codexWorkerCommand!=null}}`;
```
avifenesh, line 700: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

The matcher accepts any private worker accessor (`this.#[A-Za-z_$][\w$]*()` on the line above), but this replacement hard-codes `this.#n()`. If upstream/minification emits `#a`/`#r` instead, the regex still matches and the patch rewrites the class to call an undeclared private method, which is a syntax error for private fields. Please capture the matched private method name and reuse it in the replacement (or narrow the regex to only match `#n` so drift fails closed).

scripts/patches/impl/main-process/tray.js
```diff
@@ -1,374 +1,46 @@
...
-  if (
-    traySetupForDiagnostics != null &&
-    sourceWithTrayDiagnostics === patchedSource &&
-    !patchedSource.includes("[codex-linux] Failed to set up system tray")
-  ) {
-    console.warn("WARN: Could not find tray setup catch handler — skipping Linux tray diagnostics patch");
   }
-  patchedSource = sourceWithTrayDiagnostics;
 
   return patchedSource;
```
avifenesh, line 45: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This function now returns after only guarding close-to-tray and registering the `Tray`; it no longer rewrites the current startup call to consult `codexLinuxIsTrayEnabled()`. The Linux Desktop settings page still exposes `codex-linux-system-tray-enabled` as the “System tray” toggle, but with the upstream startup shape left as `(E||process.platform===`linux`)&&oe()` the tray is still initialized on every Linux cold start even when that setting is disabled. Please restore the startup gate around the Linux tray initializer.

avifenesh (commented):

---

I found two correctness issues in the patch update: one can generate an invalid bundle when the private method minifier name drifts, and one drops the existing System tray setting gate for Linux tray startup.

PR #1045, opened 2026-07-17
Fix Linux IAB socket discovery after per-user scoping

danielcadev:
## Summary

- align the packaged Electron IAB server with the deterministic UID-scoped Linux socket directory already used by the staged Browser and Chrome clients
- preserve `CODEX_BROWSER_USE_SOCKET_DIR` as the explicit cross-process override
- fail closed for invalid or foreign-owned directories, normalize the directory to `0700`, and normalize the listening socket to `0600`
- add regression coverage that executes the producer and staged-client resolvers independently and requires them to select the same path

## Root cause

The per-user socket work in #1016 moved the staged consumers to `/tmp/codex-browser-use-$UID`, but the packaged Electron main process still created IAB sockets under the legacy global `/tmp/codex-browser-use` directory. The server could be healthy and listening while `agent.browsers.list()` returned no IAB browser because the consumers searched a different directory.

This follow-up patches the actual IAB producer instead of adding a legacy consumer fallback, so the multi-user isolation from #1016 remains intact.

## Validation

- `node --test scripts/lib/patch-browser-client-iab-socket-scope.test.js scripts/patches/impl/main-process/browser.test.js` — 15/15 passed
- `node --test scripts/patch-linux-window-ui.test.js` — 374/374 passed
- rebuilt a transactional candidate from the latest upstream DMG `26.715.21425` (`ff459150991612007549270d2d28c5e78cec6bd6ac200a7ada5ed6c031369b87`); the new descriptor applied to the current main-bundle shape and the candidate was accepted with unrelated optional-patch warnings
- launched the candidate side by side and confirmed the real Electron process created `/tmp/codex-browser-use-$UID` as `0700` with its listening socket as `0600`
- end-to-end validation selected `Codex In-app Browser` through `agent.browsers.list()`, opened an IAB tab, and navigated successfully to `https://example.com`

## Scope and risk

The Windows named-pipe path is unchanged. The change is limited to the Linux main-process producer patch and its regression tests; it does not restore the global socket directory or weaken the existing ownership boundary.

## Checklist

- [x] This is not an upstream-drift compatibility workaround; it was validated against the latest `CODEX.DMG` only.
- [x] I added or updated relevant tests, ran the validation listed above, and confirmed that required CI checks pass. Local validation is complete; GitHub CI is pending.

### DIFF
--- a/scripts/lib/patch-browser-client-iab-socket-scope.test.js
+++ b/scripts/lib/patch-browser-client-iab-socket-scope.test.js
@@ -8,8 +8,66 @@ const os = require("node:os");
 const path = require("node:path");
 const test = require("node:test");
 const { pathToFileURL } = require("node:url");
+const vm = require("node:vm");
+
+const {
+  applyLinuxBrowserUseSocketDirectoryPatch,
+} = require("../patches/impl/main-process/browser.js");
 
 const patcher = path.join(__dirname, "patch-browser-client-iab-socket-scope.js");
+
+test("main-process producer and staged Browser client resolve the same Linux socket directory", async () => {
+  const workspace = fs.mkdtempSync(path.join(os.tmpdir(), "codex-socket-alignment-"));
+  const clientPath = path.join(workspace, "browser-client.mjs");
+  const clientFixture =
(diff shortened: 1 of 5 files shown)


ilysenko (approved):
Reviewed current head a1aaeada. The producer and staged clients now share the UID-scoped Linux socket directory, permissions remain owner-only, current-DMG validation passes, and no merge blockers remain.

PR #1047, opened 2026-07-17
Exit cleanly after Linux bootstrap failure

jheinem1:
## Summary

- bound the Linux native startup-failure dialog wait to 15 seconds
- exit the failed Electron instance after the dialog resolves or times out
- register the bootstrap recovery patch as required for current upstream builds
- cover descriptor registration, patch idempotence, and a never-resolving dialog

## Root cause

When `runMainAppStartup()` fails after the Linux warm-start socket has been created, upstream bootstrap handling destroys every `BrowserWindow` and waits indefinitely for the native failure dialog. If that dialog is hidden or never resolves, the windowless process retains both the single-instance lock and launch-action socket. Every later launcher invocation receives a successful warm-start acknowledgement from an instance that can no longer create a UI.

This was reproduced after an expired OAuth refresh token delayed `codex app-server` initialization beyond Desktop's 30-second bootstrap timeout. The late app-server response could not recover the already-failed Desktop bootstrap, and all later clicks were handed off to the windowless process.

## Behavior

On Linux, the failure dialog remains available for up to 15 seconds. The app then calls `app.exit(1)`, which releases the single-instance lock and removes the warm-start socket through normal process teardown. macOS and Windows keep the upstream dialog behavior unchanged.

## Validation

- focused patch tests: 3/3 passed
- `node --check` on the implementation and descriptor
- `git diff --check`
- applied in memory to the exact affected installed `bootstrap.js`: changed, idempotent, syntax-valid, and contained the bounded exit
- full patch test suite: 375/376 passed; the single failure is an existing macOS-host assumption in the unrelated bootstrap-lock runtime test

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -131,10 +131,14 @@ const {
   featurePatchDescriptors,
 } = require("./patches/runner.js");
 const {
+  applyExtractedAppPatchDescriptors,
   applyMainBundlePatchDescriptors,
   discoverCorePatchDescriptors,
   normalizePatchDescriptors,
 } = require("./patches/engine.js");
+const bootstrapPatchDescriptors = require(
+  "./patches/core/all-linux/extracted-app/bootstrap/patch.js",
+);
 const {
   detectLinuxTargetContext,
   linuxTargetSummary,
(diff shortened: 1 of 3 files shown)


ilysenko (changes requested):
Thanks for tackling this recovery failure. One blocker remains: the file-level patch targets `.vite/build/bootstrap.js`, but the current DMG uses a hashed `.vite/build/bootstrap-*.js` bundle loaded by `early-bootstrap.js`. The wrapper therefore returns `changed: false`, and the descriptor is incorrectly reported as `already-applied`, so the green upstream check is a false positive and the runtime fix is absent. Please resolve the current hashed bootstrap target, remove the obsolete adjacent fixed-path handling per the current-DMG-only policy, fail required status for missing or ambiguous targets, and add a production-layout integration test covering file mutation and patch-report status.

ilysenko (approved):
Reviewed current head 25ab6a0: the production hashed bootstrap target is resolved, obsolete fixed-path handling is removed, failure recovery is fail-soft optional, and the full CI matrix passes.

PR #1049, opened 2026-07-17
Fix packaged wrapper update detection

magicJie:
## Summary

- resolve the wrapper remote from packaged source metadata when the frozen update-builder has no Git checkout
- create or refresh a managed full-history wrapper checkout so ancestry and no-downgrade checks remain authoritative
- use that managed checkout for opt-in packaged wrapper detection while preserving stale-candidate cleanup on offline/fetch failures
- document the managed checkout behavior for updater and feature users

## Validation

- `env -u CODEX_CLI_PATH -u NVM_DIR cargo test -p codex-update-manager wrapper -- --test-threads=1` (24 passed)
- isolated live `check-wrapper --json` against the installed frozen `/opt/codex-desktop/update-builder`; resolved the packaged GitHub remote, created a non-shallow managed checkout, and reported installed wrapper `b24e5ff / 0.10.1` aligned

### DIFF
--- a/docs/updater.md
+++ b/docs/updater.md
@@ -245,6 +245,9 @@ enable_wrapper_updates = true
 
 in `~/.config/codex-update-manager/config.toml`.
 
-This is intended for git-checkout/dev update-builder installs. Frozen
-native-package builders without a `.git` directory report no wrapper candidate
-and receive wrapper changes through normal package upgrades.
+This remains opt-in. Git-checkout/dev builders inspect their configured remote
+without changing the working tree. Frozen native-package builders prepare a
+managed checkout under `~/.cache/codex-update-manager/wrapper-src`, using the
+remote recorded in the package's source metadata, and inspect updates there.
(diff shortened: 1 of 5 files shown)


ilysenko (changes requested):
Thanks for putting this together and for the work on packaged wrapper detection.

We do not want to merge a feature-specific solution here right now. The plan is to add this capability soon through one shared mechanism for all Linux features, not only `codex-wrapper-updater`.

Landing a special-case implementation first would create another path that the general solution would later need to work around or replace. That is how workarounds start stacking on top of each other and increasing the long-term support burden.

This PR is therefore out of scope for now. Please do not spend more time extending it; we will revisit this capability as part of the unified implementation.

For future work, please verify the change against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant tests locally before opening or updating the PR, perform an independent full-diff review with a local LLM, and wait for all required CI checks to pass before requesting maintainer review or merge.

PR #1050, opened 2026-07-17
Restore current Computer Use host platform support

colonelpanic8:
## Summary

- rebase onto the latest `main`
- target the current DMG `26.715.21425` Computer Use assets
- patch the `ekwfx4j1` host-platform contract without bypassing required-feature or rollout gates
- patch the `iaq4jiqv` plugin-detail availability contract
- refresh the Settings contract for `computer-use-settings-DsM_pz8i.js`
- remove the obsolete `nrw3o0ql`, `jgoqfqy2`, and old Settings fixtures

## User impact

With the Computer Use UI opt-in enabled, Linux is accepted as a compatible host platform, the bundled Computer Use card appears in Settings, and plugin details load while upstream feature and rollout gates remain intact.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` (377 passed)
- targeted `test_linux_computer_use_ui_opt_in_smoke`
- exact DMG `26.715.21425` ASAR validation: all three Computer Use patches reported `applied` on the first pass and `already-applied` on the second
- `git diff --check`


ilysenko (changes requested):
Thanks for restoring the Linux host-platform gate without bypassing the upstream rollout or required-feature checks. The descriptor targets are stale for the currently supported DMG 26.715.21425: an opt-in inspection reports the host-platform, Settings availability, and install-flow patches as skipped. Please retarget the host contract to the current `ekwfx4j1` asset, plugin detail to `iaq4jiqv`, refresh the Settings implementation for `computer-use-settings-DsM_pz8i.js`, and remove the obsolete `nrw3o0ql`/`jgoqfqy2` targets and fixtures. The green upstream job does not exercise this path because Computer Use UI is disabled there.

ilysenko (changes requested):
Thanks for the focused fix. I found one current-DMG blocker before this can merge.

The new host-platform descriptor targets the `nrw3o0ql` bundle, and the existing install-flow descriptor/test fixture still targets `jgoqfqy2`, but the current supported DMG `26.715.21425` has moved these Computer Use contracts to newer assets (`ekwfx4j1` / `iaq4jiqv`, with `computer-use-settings-DsM_pz8i.js`). With Computer Use UI opt-in enabled, the proposed host-platform patch will not be applied to the current DMG.

Please retarget the Computer Use UI descriptors and tests to the current DMG assets, update any drifted Settings/install-flow contract as needed, and remove the obsolete `nrw3o0ql` / `jgoqfqy2` paths rather than keeping older-DMG compatibility fixtures.

ilysenko (approved):
Reviewed current head cdcc40eb4dc8e17126459a04e55be0cfc9bd2c74; the prior current-DMG Computer Use blockers are resolved and checks are green.

PR #1056, opened 2026-07-17
Fix/pet overlay drag for COSMIC

Leay15:
## Summary

- Adds a generic Linux avatar-overlay input hook so optional features can opt into full-window input without changing default overlay behavior.
- Updates `pet-overlay` to provide a frameless, transparent mascot window on Linux. In the unlocked mode, transparent space moves the native window while the mascot remains draggable within its view and tray controls remain interactive.
- Preserves session-only positions, `lockPosition`, and existing Hyprland/Niri behavior.
- Makes notification-tray placement adaptive: it stays above the mascot when space permits, moves below near the top edge, and snaps the mascot to the nearest viable vertical zone when neither side has enough room to avoid overlap.
- Adds regression coverage for Linux input shaping, Wayland passthrough, window creation, internal mascot dragging, tray bounds/placement, locking, and patch idempotence.

## Validation

- `node --test linux-features/pet-overlay/test.js` — 54/54 passing.
- `node --test scripts/patch-linux-window-ui.test.js` — 364/364 passing.
- Manual validation in COSMIC remains pending after rebuilding the application with `pet-overlay` enabled.

### DIFF
--- a/linux-features/pet-overlay/README.md
+++ b/linux-features/pet-overlay/README.md
@@ -46,8 +46,13 @@ Feature settings can be overridden in the gitignored `features.json` file:
 ```
 
 `lockPosition: false` preserves manual pet moves when the existing window
-position is visible. Set it to `true` only when you want the mascot pinned to
-the configured screen corner on every layout pass.
+position is visible. In this default unlocked mode the overlay is frameless:
+transparent space moves the native `356×320` window, while the mascot is a
+`no-drag` region so its renderer gesture can reposition the mascot within the
+view. Existing tray controls keep their own interactive regions. Set
+`lockPosition` to `true` only when you want the mascot pinned to the configured
+screen corner on every layout pass; that mode removes the full-surface
+drag/input opt-in.
 
 ## Options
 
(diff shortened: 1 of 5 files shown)


ilysenko (changes requested):
Thanks for the focused COSMIC pet-overlay work. I found one blocker before this can merge.

This PR removes the existing KWin/KDE Plasma implementation and its regression coverage while the feature manifest and README still advertise `kwin: true`, KDE Plasma hints, and Plasma drag validation. The final patch no longer defines the KWin qdbus/script helpers or the `CODEX_PET_OVERLAY_KWIN` runtime override; only swallowed calls to missing KWin helpers remain. That silently regresses the documented Plasma path for pet-overlay users.

Please preserve the existing KWin support and tests while adding the COSMIC/full-window input path, or split any intentional KWin removal into a separate PR with updated feature defaults/docs and a clear rationale.

ilysenko (approved):
Approved after restoring the Niri live-drag transport, preserving KWin coverage, and verifying the current head with local and GitHub checks.

PR #1059, opened 2026-07-17
Add project group Last updated sorting

moxw:
## Summary

- add an opt-in `project-group-last-updated-sort` Linux feature
- make Last updated reorder project groups as well as task rows
- preserve upstream Priority and Manual order behavior
- keep the feature disabled by default

## Root cause

Upstream first computes project-group recency, then reapplies the persisted project order. When that saved order contains every project, it fixes the project headers in place even though task rows are sorted by Last updated.

## User-visible behavior

With `project-group-last-updated-sort` enabled, Last updated sorts unpinned project groups by the newest project/task activity. Priority and Manual order continue to use the upstream saved project order.

## Validation

- `node --test linux-features/project-group-last-updated-sort/test.js` — 8 passed
- `node --test --test-reporter=dot linux-features/*/test.js` — 607 passed, 1 unrelated sandbox-only failure because the Thorium cleanup test could not write `~/.codex/hooks.json`
- current-DMG rebuild with `project-group-last-updated-sort` enabled — feature descriptor applied once
- real bundle patch — first pass changed, second pass idempotent, `node --check` passed

## Limitations

The isolated current-DMG candidate reached the feature patch successfully, then promotion remained inconclusive because a disabled cleanup hook attempted to write `~/.codex/hooks.json` under sandbox restrictions. Existing optional core drift warnings are unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,7 @@ workarounds.
 | Open Target Discovery | Opt-in | `open-target-discovery` | [Docs](linux-features/open-target-discovery/README.md) |
 | Persistent status panel | Opt-in | `persistent-status-panel` | [Docs](linux-features/persistent-status-panel/README.md) |
 | Pet overlay | Opt-in | `pet-overlay` | [Docs](linux-features/pet-overlay/README.md) |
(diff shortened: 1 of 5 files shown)


ilysenko (approved):
Reviewed the current head. The feature is isolated, opt-in, covered by focused tests, and safe to merge.

PR #1062, opened 2026-07-17
Fix shared app-server socket DMG drift

moxw:
## Summary

- update the shared app-server socket detector for the current upstream SSH WebSocket transport class
- match the current local transport factory's WebSocket fallback block
- keep the patch fail-soft, idempotent, and latest-DMG-only
- add regression coverage that rejects the previous bundle shape

## Root cause

The current DMG added fields before the SSH transport's `kind` marker and changed the local transport factory from a ternary fallback to an `if (websocketUrl)` block. The feature descriptor therefore skipped, and enabled-feature acceptance rejected the candidate.

## User impact

Native bootstrap and updater rebuilds can accept the current DMG when `shared-app-server-socket` is enabled. The feature's runtime behavior is unchanged.

## Validation

- `node --test linux-features/shared-app-server-socket/test.js` — 24 passed
- real current-DMG main bundle — first pass changed, second pass byte-identical, one transport injected
- `node --check` passed for the feature sources and patched current bundle
- current-DMG transactional rebuild — `accepted_with_warnings`; `shared-app-server-socket` applied once with no enabled-feature blockers

## Limitations

The rebuild retains unrelated optional core drift warnings already present for the current DMG.

### DIFF
--- a/linux-features/shared-app-server-socket/patch.js
+++ b/linux-features/shared-app-server-socket/patch.js
@@ -3,7 +3,11 @@
 const IDENT = "[A-Za-z_$][\\w$]*";
 
 function findTransportSymbols(source) {
-  const classMatch = source.match(new RegExp(`var (${IDENT})=class\\{kind=\\\`websocket\\\``));
+  const classMatch = source.match(
+    new RegExp(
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head bd127c0016a2b21ed55d5626e90a287a8e607242. No blockers found; focused optional feature drift fix with passing checks.

PR #1063, opened 2026-07-17
feat(linux): add bounded working-tree watcher

### LINKED ISSUE #983: Codex Desktop on Linux consumes ~65k inotify watches and exhausts the system limit
Codex Desktop version

codex-desktop 2026.07.14.164133

The accompanying Codex app-server/CLI reports codex-cli 0.144.4.
Environment

    Debian GNU/Linux 13 (trixie), x86_64
    Kernel 6.12.95+deb13-amd64
    Workspace size: approximately 185,015 filesystem entries
    Kernel default: fs.inotify.max_user_watches = 65536
    fs.inotify.max_user_instances = 1024

What issue are you seeing?

After an extended Codex Desktop session, the main Codex Electron process owned 65,042 inotify watches across three inotify instances. This consumed nearly the entire default per-user Linux limit

Measured process ownership:

Codex Desktop main Electron process: 65,042 watches / 3 instances
Codex app-server process:                 32 watches / 3 instances
Kernel max_user_watches:              65,536

The main process had been running for about 11 hours when measured. Its command began with:

/opt/codex-desktop/electron --no-sandbox --class=codex-desktop ...

Steps to reproduce

    On Linux with the default fs.inotify.max_user_watches=65536, open a moderately large workspace in Codex Desktop (this workspace has about 185k entries).
    Use Codex Desktop normally for an extended session.
    Identify the main Electron PID and count watches:

pid=$(pgrep -f '/opt/codex-desktop/electron --no-sandbox --class=codex-desktop' | head -1)

for info in /proc/$pid/fdinfo/*; do
  grep -c '^inotify ' "$info" 2>/dev/null || true
done | awk '{ total += $1 } END { print total }'

Expected behavior

    Codex Desktop should keep file watching bounded and avoid consuming nearly the entire per-user inotify allowance.
    Large generated/runtime directories and Git internals should be excluded where possible.
    Watches associated with inactive workspaces/tasks should be released.
    If a workspace would exceed a safe budget, Codex should degrade gracefully or show an actionable warning rather than causing desktop-wide ENOSPC failures.

Workaround

Increasing the system limit works around the problem:

fs.inotify.max_user_watches = 524288

This is only a mitigation; the Codex Desktop process still owns about 65k watches.

gadicc:
## Summary

Fixes #983.

The upstream app targets macOS, where recursive filesystem watching can use an efficient tree-level backend. In the current Linux Electron runtime, translating this directly to `fs.watch(path, { recursive: true })` allocates inotify watches for files as well as directories. Large dependency and build trees can therefore exhaust the per-user inotify pool.

This adds `directory-only-working-tree-watch` as a disabled-by-default Linux feature. Its runtime patch, descriptor, documentation, and focused tests live under `linux-features/directory-only-working-tree-watch/`; the accompanying Nix and watchdog registrations expose and continuously validate it.

When enabled, the feature:

- Installs non-recursive watches on relevant directories, not individual files, while still receiving their child file events.
- Within covered paths, reconciles directory creation, deletion, rename, moved-in trees, filename-less events, and same-path inode replacement.
- Prunes only Git-ignored, untracked directories. Tracked and force-added paths remain covered, and ignored files such as `.env.local` still produce events through their watched parent.
- Has no default name-based exclusions. Directories such as `node_modules` are pruned only when Git ignores them or the user explicitly configures their name.
- Excludes `.git` from ordinary working-tree traversal while retaining Codex’s dedicated Git subsystem and adding bounded refresh watches around the Git index and `.git/info/exclude`.
- Shares a process-wide budget across this feature’s working-tree and Git-refresh watches. The default limit is the smaller of 8192 and one eighth of the current kernel `fs.inotify.max_user_watches`; an override is capped at 65536. Roots receive priority, with remaining capacity distributed fairly in bounded reservations.
- Preserves Codex’s focus-recovery path during partial coverage, expands coverage when capacity is released, and retries transient Git, filesystem, and watch-resource failures with bounded backoff.
- Uses asynchronous Git subprocesses and promise-based, serialized topology scans so reconciliation does not synchronously block Electron.
- Emits low-noise launcher-log diagnostics: one warning per partial-budget episode, one message after complete recovery, and one process-wide error for operating-system watch exhaustion.

On my system, this reduced the app’s watch count from approximately 190,000 to approximately 900 with the default settings.

This is intentionally an opt-in Linux feature for initial field validation. If it proves stable, it should be reconsidered for default or core Linux compatibility: bounded watcher resource use is baseline platform behavior rather than an end-user feature. Before that promotion, persistent zero-watch coverage and operating-system resource exhaustion should receive actionable in-app notices, and a native recursive watcher backend should be evaluated as a possible simplification.

## Validation

- `node --test linux-features/directory-only-working-tree-watch/test.js` — 92/92 passed under Node.js 22, 24, and 25.
- `node --test linux-features/*/test.js` — 692/692 passed.
- `node --test scripts/patch-linux-window-ui.test.js` — 374/374 passed.
- `node --test scripts/lib/linux-features.test.js` — 10/10 passed.
- `node --test scripts/ci/upstream-dmg-acceptance.test.js` — 14/14 passed.
- `bash tests/scripts_smoke.sh` — passed.
- `git diff --check upstream/main...HEAD` — passed.
- Current-DMG rebuild from clean commit `831f16fb`:
  - app version `26.715.21425`
  - DMG SHA-256 `ff459150991612007549270d2d28c5e78cec6bd6ac200a7ada5ed6c031369b87`
  - feature patch applied successfully
  - acceptance verdict `accepted_with_warnings`
  - no blockers; all eight warnings were unrelated optional core-patch drift
- Manual runtime measurement: approximately 190,000 watches before and approximately 900 after enabling the feature.
- Not run locally: Nix evaluation because `nix` is unavailable, or full `.deb`, `.rpm`, pacman, and AppImage artifact builds. Nix/feature registration tests and the cross-format script smoke suite passed.

Known limitations:

- Once the configured budget is full, immediate event coverage is deliberately partial until capacity becomes available. Codex’s focus refresh remains active, but does not itself rebuild watch topology.
- Changes to a global Git excludes file require an app restart; `.gitignore`, index, and `.git/info/exclude` changes are refreshed dynamically.
- Budget diagnostics include the affected local working-tree path in the existing local launcher log.
- Persistent zero-watch coverage and operating-system watch exhaustion remain log-only for this opt-in release; actionable UI is explicitly deferred until before any default/core rollout.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,7 @@ workarounds.
 | Wrapper updater button | Opt-in | `codex-wrapper-updater` | [Docs](linux-features/codex-wrapper-updater/README.md) |
 | Conversation mode | Opt-in | `conversation-mode` | [Docs](linux-features/conversation-mode/README.md) |
 | Copilot reasoning effort defaults | Opt-in | `copilot-reasoning-effort` | [Docs](linux-features/copilot-reasoning-effort/README.md) |
+| Directory-only working-tree watch | Opt-in | `directory-only-working-tree-watch` | [Docs](linux-features/directory-only-working-tree-watch/README.md) |
 | Example Linux Feature | Developer example | `example-feature` | [Docs](linux-features/example-feature/README.md) |
 | Frameless titlebar | Opt-in | `frameless-titlebar` | [Docs](linux-features/frameless-titlebar/README.md) |
 | Global Dictation | Opt-in | `global-dictation` | [Docs](linux-features/global-dictation/README.md) |
--- a/docs/nix.md
+++ b/docs/nix.md
@@ -205,6 +205,7 @@ The Home Manager and NixOS modules accept these feature IDs through
 | Feature ID | Purpose |
 | --- | --- |
 | `appshots` | Linux AppShots capture integration |
+| `directory-only-working-tree-watch` | Bounded directory-only working-tree watches |
 | `frameless-titlebar` | Hide app-provided titlebar controls for compositor-managed decorations |
 | `mcp-helper-reaper` | Cleanup for stale configured MCP helper processes |
 | `node-repl-reaper` | Cleanup for leaked Browser Use `node_repl` helpers |
--- a/linux-features/directory-only-working-tree-watch/README.md
+++ b/linux-features/directory-only-working-tree-watch/README.md
@@ -1,0 +1,149 @@
+# Directory-Only Working-Tree Watch
+
+This opt-in Linux feature replaces Codex Desktop's recursive working-tree
+`fs.watch` call with one non-recursive watch per relevant directory. Electron's
+Node runtime currently allocates one inotify watch for every file and directory
+when `recursive: true` is used on Linux, so large dependency and build trees can
+exhaust the per-user inotify limit.
+
+The feature keeps recursive change detection by maintaining directory watches
+as directories are created, renamed, deleted, or replaced. It skips recursive
+traversal of every `.git` directory; Codex retains its dedicated Git metadata
+watches, and this feature adds small directory watches around the Git index and
+`.git/info/exclude` so ignore topology follows index and exclude changes.
+
+Directories are pruned only when Git confirms that the directory itself is
+ignored and untracked. An unignored wrapper directory remains watched even if
+all of its current contents are ignored, and a force-added tracked file keeps
+its containing directory watched. Ignored files are not skipped: their parent
+directories remain watched, so a root-level file such as `.env.local` still
+produces an event.
+
+There are no default name-based exclusions. A tracked directory named `build`
+or `node_modules` remains watched even if its name resembles generated output.
+
+By default, the watcher uses at most 8192 inotify watches per app process across
+all active working trees, or one eighth of the kernel's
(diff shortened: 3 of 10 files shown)


gadicc:
**Failing CI check noted**.  It's late here now but I'll address this first thing in morning.

A few things worthy of discussion:

- Maybe an easier first pass would have been to simply replace `fs.watch()` with [`@parcel/watcher`](https://www.npmjs.com/package/@parcel/watcher).  Alone it doesn't cover as much as this patch, and also introduces a native C++ dependency, which could be a pro or a con -- we'd have to evaluate separately.

- If the inotify budget concept is excessive for a first attempt.  It definitely adds welcome value but perhaps is too ambitious.

- Likewise for all the git handling.

However, that does make this first contribution quite comprehensive, instead of delaying the inevitable.

ilysenko (approved):
Reviewed current head; no blockers found. The opt-in feature is isolated, covered by tests/CI, and the current-DMG feature patch was locally verified against 26.715.31251.

gadicc:
Thanks, @ilysenko!

PR #1070, opened 2026-07-18
Isolate Electron runtime temp files per user

oksure:
## Summary

- default Electron temporary files to a private per-user directory under `XDG_RUNTIME_DIR`
- fall back to the app state directory when no writable runtime directory exists
- preserve an explicitly configured `TMPDIR`
- document the multi-user IPC failure mode

## Problem

On a multi-user Linux host, Electron creates `/tmp/codex-ipc/ipc-<uid>.sock`. The first user to launch Codex owns the shared `/tmp/codex-ipc` parent, so another user can fail with `EACCES` while creating their UID-specific socket. Single-instance routing then fails, duplicate Codex Desktop processes can start, and Remote Control can immediately turn itself off.

## Verification

- reproduced the original cross-user ownership failure and verified Remote Control reconnects after IPC isolation

The full smoke suite passes.


ilysenko (approved):
Reviewed current head 189710faac3fc551dd7de51362ca2c871f410023. No blockers found; CI is green and the launcher TMPDIR isolation is appropriate for core/default behavior.

PR #1080, opened 2026-07-19
Normalize Linux Browser sidebar user agent

psypeal:
## Summary

Linux in-app Browser sidebar guests can expose ChatGPT/Electron product tokens that cause services such as X to reject or stall the login/page bootstrap. This PR applies a browser-only user-agent normalization before the guest's first navigation.

## Scope

- Sets only the Browser sidebar webview's pre-navigation user-agent parameter on Linux.
- Removes only `ChatGPT/...` and `Electron/...` product tokens.
- Does not change `app.userAgentFallback`, any global session default, unrelated renderers, Browser Settings webviews, or non-Linux behavior.
- Idempotent and fail-soft when the current upstream seam drifts.

This is a replacement for closed PR #1022; the branch contains no window-open, attachment-recovery, Chrome-runtime, multi-tab, or Dock-icon changes.

## Validation

- Rebased onto current upstream `main` (`3bc05c5`).
- Full patch suite: 385/385 passed.
- `git diff --check`: passed.
- Diff: exactly three UA-only files.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -58,6 +58,7 @@ const {
   applyLinuxBundledPluginCopyPermissionsPatch,
   applyLinuxBundledPluginReconcileStaleSnapshotPatch,
   applyLinuxBrowserUseRouteLivenessPatch,
+  applyLinuxBrowserUseUserAgentPatch,
   applyLinuxChromeExtensionStatusPatch,
   applyLinuxExternalOpenEnvPatch,
 } = require("./patches/impl/main-process/browser.js");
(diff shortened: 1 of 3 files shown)


Yo-DDV (changes requested):
Hi, thanks for narrowing this to the Browser sidebar and for adding the regression coverage. @ilysenko, I found one project-scope blocker.

This is the same user-agent normalization that was closed in #1022 after the decision to preserve the standard upstream Browser sidebar behavior and avoid another downstream behavior-changing patch unless it is required for Linux compatibility. The implementation here is narrower, its tests pass, and the current-DMG build applies the descriptor, but this PR does not link a new reproducible Linux failure or a later project decision that changes that scope.

I do not think this should merge in its current form. If a standard app flow is now broken specifically on Linux, that needs a linked issue with the exact site or flow, current build details, an upstream baseline, and a reproducible before-and-after result. Without that evidence, the scope decision from #1022 still applies.

PR #1084, opened 2026-07-19
Allow Device Remote Control keys storage

Leay15:
## Summary

- Upgrade Linux Remote Mobile Control device-key storage to schema v2.
- Encrypt private key material with Electron `safeStorage` when GNOME Secret Service/libsecret or KWallet is available; `basic_text` is not accepted as a keychain backend.
- Retain the existing `0600` PEM file fallback when no protected keychain is available, recording the detected backend without exposing key material.
- Migrate legacy v1 PEM records to `safeStorage` on read when possible, while preserving the original record if migration fails.
- Update upstream bundle matching and retain Linux-specific connection-tab selection without changing non-Linux behavior.
- No linked issue.

## Validation

- `node --test linux-features/*/test.js` — 750 tests passed.
- `node --test linux-features/pet-overlay/test.js` — 69 tests passed.
- `bash tests/scripts_smoke.sh` — validated Debian, RPM, pacman, AppImage, Fedora, and Fedora Atomic paths.
- `git diff --check 73a3edd..HEAD`

Known risk: GNOME/libsecret and Plasma/KWallet were not manually tested in live desktop sessions; automated coverage verifies both code paths and the secure fallback.


ilysenko (changes requested):
Thanks for improving Linux remote-control key storage. I found three blockers:

- Please serialize the read-triggered PEM-to-safeStorage migration with the existing flock, or keep reads side-effect-free and perform a locked re-read/write. The current path can overwrite a concurrent create/delete with a stale snapshot. Please also make post-rename failure handling and documentation match the actual commit point.
- Please add tests that inject `electron.safeStorage` and cover protected ciphertext creation/signing/decryption, `basic_text` and unavailable-backend fallback, migration success/failure, absence of plaintext, and migration versus concurrent updates. The current tests do not execute the new protected-storage branches.
- Please remove the obsolete selected-tab constants and multi-shape drift fallbacks added here. The selected-tab workaround was deliberately removed in #882 because the current DMG handles it, and this repository supports only the current upstream DMG.

Thank you!

ilysenko (approved):
Reviewed the current head after the requested storage, locking, test coverage, and current-DMG cleanup updates. The previous blockers are resolved.

PR #1085, opened 2026-07-19
Fix V2 pet cursor feedback for Linux Computer Use

### LINKED ISSUE #1061: Support V2 pet looking directions during Linux Computer Use
## Summary

V2 pets appear to include two additional spritesheet rows for pointer-facing poses. While checking the official built-in Codex pet on Linux, I noticed that these poses can be selected by the renderer when a cursor event is supplied, but Linux Computer Use does not appear to forward its cursor location to the avatar overlay.

I may be overlooking another Linux path, so I am opening this issue mainly to confirm whether this behavior is expected or currently unsupported.

## Evidence

- The official Codex spritesheet is `1536 × 2288`, with all 16 cells in rows 9–10 populated.
- Controlled cursor events make the deployed renderer select those V2 look frames.
- The avatar overlay listens for `avatar-overlay-computer-use-cursor-changed`.
- In the deployed main process, the cursor-location handler registered on Darwin but not on Linux.

## Expected behavior

When Linux Computer Use performs a pointer action, an awake V2 pet should look toward the corresponding pointer position and return to its normal animation afterward.

V1 pets would keep their existing behavior.

## Environment

- Upstream Codex Desktop: `26.707.91948`
- Tested codex-desktop-linux build: `481fc5e`
- Ubuntu 22.04, GNOME X11

## Notes

This appears separate from the optional `pet-overlay` feature and from pet asset loading.
I have confirmed the missing Linux registration in the built bundle, but have not yet run an end-to-end Linux Computer Use task.

Yo-DDV:
## Summary

- Forward successful Linux Computer Use pointer activity to the existing V2 pet cursor path.
- Keep the bridge local, instance-scoped, bounded, and fail-soft.
- Preserve Darwin behavior and leave non-pointer accessibility actions unchanged.

## Why

The current desktop bundle only registers the Computer Use cursor callback on Darwin. Linux click, scroll, and drag actions therefore never reach the avatar overlay, even though the renderer already supports the V2 look frames.

The Linux backend now sends a small private runtime signal after a pointer action succeeds. The main process reads the actual Electron cursor position and forwards it through the existing avatar event, then returns the pet to its idle animation.

Fixes #1061

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` (385 passed)
- `TMPDIR=/tmp cargo test -p codex-computer-use-linux` (151 passed)
- `cargo build --release -p codex-computer-use-linux`
- `make inspect-upstream` against ChatGPT Desktop `26.715.31925`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- V2 pets now look toward the live pointer position after successful Linux
+  Computer Use click, scroll, and drag actions, then return to their normal
+  animation. The bridge is isolated per app instance and fails softly when its
+  private runtime socket is unavailable.
 - The updater daemon now detects that a package upgrade replaced its binary
   on disk and exits with a nonzero status so systemd's `Restart=on-failure`
   relaunches it on the new binary. Previously a running daemon survived every
--- a/computer-use-linux/src/server.rs
+++ b/computer-use-linux/src/server.rs
@@ -31,16 +31,21 @@ use rmcp::{
 };
 use serde::{Deserialize, Serialize};
 use std::{
-    env,
+    env, fs,
     future::Future,
-    os::unix::net::{UnixDatagram, UnixStream},
-    path::PathBuf,
+    os::unix::{
+        ffi::OsStrExt,
+        fs::{FileTypeExt, MetadataExt},
+        net::{UnixDatagram, UnixStream},
+    },
+    path::{Path, PathBuf},
     process::{Command, Output, Stdio},
     sync::{Arc, Mutex},
     time::Duration,
 };
 use tokio::{
     io::{AsyncRead, AsyncReadExt, AsyncWriteExt},
+    net::UnixStream as TokioUnixStream,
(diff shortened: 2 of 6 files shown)


ilysenko (changes requested):
Thanks for the focused fix and thorough bridge coverage. I found two blockers:

- Please change `linux-computer-use-avatar-cursor` from `required-upstream` to `optional` and update the descriptor policy test. Pet cursor feedback is not required for launch or core usability, so its upstream drift must not reject Linux installs or rebuilds.
- Please emit a warning when the upstream cursor-handler marker is completely absent and add a negative test. The current path returns unchanged source without warnings, causing the patch engine to record `already-applied` and silently hide drift.

Thanks!

Yo-DDV:
Thanks for catching both of these. I changed the cursor descriptor to `optional` and added an engine-level negative test for a completely missing upstream marker. That case now leaves the source unchanged but records `skipped-optional` with an explicit warning instead of `already-applied`.

I also rebased the branch onto the current `main` and reran the full patcher test file locally: all 386 tests pass.

ilysenko (approved):
Reviewed the current head. The prior drift-policy concerns are resolved, the relevant checks pass, and no remaining blockers were found.

PR #1086, opened 2026-07-19
Fix X11 Global Dictation runtime helpers

### LINKED ISSUE #1064: Global Dictation external paste and hold release fail on X11
## Summary

While checking the optional Global Dictation feature on Ubuntu X11, I noticed two problems.

When the toggle shortcut is used from an external application, the transcription is saved in the Voice history, but no text is pasted into the active input. The hold shortcut also appears to start and stop the microphone repeatedly instead of maintaining one recording until the shortcut is released.

The corresponding in-app dictation paths work normally.

I may be overlooking another X11 dependency or runtime path, so I am opening this issue mainly to confirm whether these behaviors are expected and whether the current feature is still affected.

## Evidence

- The built-in `Ctrl+Shift+D` dictation shortcut works inside Codex.
- The configured `Ctrl+Space` Global Dictation toggle also works while Codex is focused.
- When `Ctrl+Space` is used from an external application, the transcript appears in Voice history, but the external input receives no text.
- The external paste failure is accompanied by:
  - `Global dictation paste failed`
  - `TypeError: k7 is not a function`
- Holding the configured `Ctrl+Shift+Space` shortcut repeatedly invokes the microphone and does not produce a usable hold-to-dictate session.
- The hold path is accompanied by:
  - `Failed to watch global dictation hotkey release`
  - `TypeError: e.split is not a function or its return value is not iterable`

## Expected behavior

When Global Dictation is started from an external application, the completed transcript should be pasted into the input that was active when dictation began.

The hold shortcut should start one recording and stop it once when the shortcut is released.

## Environment

- Upstream Codex Desktop: `26.707.91948`
- Tested codex-desktop-linux build: `481fc5e`
- Ubuntu 22.04.5, GNOME X11
- `global-dictation` enabled
- `xdotool`, `xinput`, and `xmodmap` installed

## Notes

The current `main` branch is `f3037182`, with upstream Codex Desktop `26.715.21425` pinned.

A read-only check against that pinned DMG suggests that the same X11 patch area may still be affected, but I have not rebuilt and reproduced the behavior end to end on the latest DMG.

I also checked #798, #802, and #912. They appear related to the feature, but I could not find an existing report for these X11 symptoms. In particular, #912 appears specific to the Wayland portal path.

Yo-DDV:
## Summary

- stop relying on minified main-bundle aliases for the X11 release watcher and paste path
- return a disposable Linux watcher directly and run the fixed `xdotool` paste command through a Linux-owned helper
- migrate already-patched `v1` bundles to the corrected `v2` shape without accepting partial legacy state

Fixes #1064

## Validation

- `node --test linux-features/global-dictation/test.js`
- applied the descriptor to the pinned `26.715.31925` main bundle, including an old-`v1` to `v2` migration replay

### DIFF
--- a/linux-features/global-dictation/patch.js
+++ b/linux-features/global-dictation/patch.js
@@ -1,6 +1,5 @@
 "use strict";
 
-const LEGACY_PATCH_MARKER = "codex-linux-global-dictation-v1";
 const PATCH_MARKER = "codex-linux-global-dictation-v2";
 const IDENT = "[A-Za-z_$][\\w$]*";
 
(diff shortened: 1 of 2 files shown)


ilysenko (changes requested):
Thanks for fixing the X11 paste and release-watcher contracts. The fresh current-DMG changes look well scoped, but this repository supports only the latest upstream DMG and must not preserve legacy patched-output shapes. Please remove `LEGACY_PATCH_MARKER`, `migrateLegacyLinuxGlobalDictationPatch`, and the v1 migration fixtures/tests. Keep only fresh current-DMG patching and current-marker idempotency.

ilysenko (approved):
Reviewed the corrected current-DMG-only implementation. The legacy patched-output migration has been removed, the focused Global Dictation tests pass, and all required checks are green on this head.

PR #1090, opened 2026-07-20
[codex] Keep separated Linux header actions right-aligned

### LINKED ISSUE #1074: Terminal and side-panel buttons are displaced after safe-area fix #1071
## Summary

The terminal and side-panel buttons are rendered in the wrong position in the thread conversation header after the Linux window-controls safe-area fix from #1068 / #1071.

Instead of remaining aligned with the thread toolbar immediately before the native Linux window controls, the buttons are displaced into a lower secondary row and shifted left, leaving a large empty region beneath the minimize/maximize/close controls.

This is a layout regression distinct from the mouse-input lockup tracked in #1073.

## Environment

- Ubuntu 24.04.4 LTS
- GNOME Shell 46.0
- X11 session
- Package: `codex-desktop 2026.07.18.111000`
- Upstream app: `26.715.31925`
- Electron: `42.3.0`
- Installed wrapper source: `f7de27215d637a41f615423f436263257da9016b`
- Package format: `.deb`
- Window configuration: default Linux title bar; `frameless-titlebar` is not enabled

## Fix-presence verification

The installed wrapper source contains merge commit `6fff69dd9f232146f03b1bbbf0091520a83a2c6f` from PR #1071. The installed patch report reports the following as applied:

- `linux-native-titlebar`
- `linux-window-controls-safe-area`
- `linux-tooltip-window-controls-collision`
- `linux-thread-side-panel-native-tooltip`

This reproduction therefore occurs after the #1071 safe-area fix, rather than on the older build reported in #1068.

## Reproduction

1. Launch ChatGPT Desktop on Linux using the default title bar.
2. Open a conversation thread.
3. Observe the terminal and side-panel toggle buttons near the top-right of the conversation.

## Actual behavior

- The terminal and side-panel controls appear in a lower, separate row beneath the native title bar.
- The buttons are shifted toward the left instead of being aligned with the right-side thread toolbar.
- A large unused header area remains beneath the native window controls.
- The resulting layout makes the controls look detached from the conversation toolbar.

## Expected behavior

The terminal and side-panel buttons should retain their normal thread-toolbar alignment and remain on the intended header row. The Linux safe-area patch should reserve only the space needed for minimize/maximize/close controls; it should not move or wrap the application actions into a separate row.

## Suspected area

PR #1071 changed the right-side header slot from the normal `pe-2` handling to `pe-(--spacing-token-safe-header-right)` for end-aligned entries. The current result suggests that the safe-area value is affecting the layout of the entire end action slot rather than only keeping that slot clear of the native window controls.

This is an inference from the merged patch and the post-fix screenshot; the exact DOM/CSS layout still needs runtime inspection.

## Related work

- #1068 — original overlap/tooltip report.
- #1071 — safe-area fix that closed #1068 and is present in this build.
- #1073 — separate post-fix GNOME/X11 mouse-input lockup.

## Acceptance criteria

- [ ] Terminal and side-panel actions remain on the intended thread-toolbar row.
- [ ] The controls are right-aligned immediately before the reserved native window-control area.
- [ ] The safe-area does not introduce an empty secondary header row.
- [ ] The layout remains correct in maximized and windowed modes.
- [ ] Narrow window widths and supported zoom/scaling do not overlap or incorrectly wrap the controls.
- [ ] Mouse hit targets match the visible button positions.
- [ ] Live GNOME/X11 validation and a current-asset regression fixture cover the final layout.

SOSDancho:
## Summary

- keep the 138px right safe area when the app header shares the native titlebar row
- restore the normal `pe-2` alignment when the application-menu layout moves the app header below the titlebar
- keep the patch fail-soft on upstream drift and cover both layout modes with runtime class-rule tests

## Root cause

#1071 correctly reserved the native window-controls footprint, but the current app shell positions the main header at `top-toolbar-sm` when the application menu is enabled. The right slot still consumed the titlebar safe area even though it was vertically below the controls.

This does not revert #1071: the application-menu/titlebar row retains the 138px inset, and an app header overlaid on the titlebar continues to use the safe-area padding.

Fixes #1074

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` (395 passed)
- `TMPDIR=/tmp make build-app` against ChatGPT 26.715.31925 (`accepted_with_warnings`; required patch validation passed)
- `./scripts/build-appimage.sh`

### DIFF
--- a/scripts/patches/impl/webview/index.js
+++ b/scripts/patches/impl/webview/index.js
@@ -16,6 +16,7 @@ const LINUX_SAFE_MONOSPACE_FONT_STACK =
   "\"Noto Sans Mono\", \"DejaVu Sans Mono\", \"Liberation Mono\", \"Ubuntu Mono\", ui-monospace, \"SFMono-Regular\", \"SF Mono\", Menlo, Consolas, monospace";
 const LINUX_TOOLTIP_COLLISION_PADDING_TOP = 44;
 const LINUX_WINDOW_CONTROLS_SAFE_AREA_RIGHT = 138;
+const LINUX_WINDOW_CONTROLS_SAFE_AREA_PROP = "codexLinuxUseWindowControlsSafeArea";
 
 function applyLinuxSafeMonospaceFontStackPatch(currentSource) {
   const safeLinuxMonoFontPattern =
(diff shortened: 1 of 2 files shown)


ilysenko:

avifenesh (commented):

scripts/patches/impl/webview/index.js
```diff
@@ -141,35 +142,118 @@ function applyLinuxOpaqueWindowsDefaultPatch(currentSource) {
...
+  const patchedHeaderSource = headerSource.replace(
+    endSlotPattern,
+    `$1,${prop}:!${menuEnabledAlias},side:\`end\``,
+  );
+  const patchedSlotSource = slotSource
+    .replace(
+      slotMatch[0],
+      slotMatch[0].replace("}){", `,${prop}}){`),
+    )
+    .replace(
```
avifenesh, line 211: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This is patching minified ASAR bundle content, so it needs to satisfy the repo invariant that all eligible occurrences are handled (or uniqueness is asserted). Here `paddingPattern` is only matched once and then passed to a non-global `replace`; if the slot helper contains a second identical `"pe-2":side===\`start\`&&…||side===\`end\`` rule, the first one gets the new prop but the second stays with the old unconditional end padding. Because `alreadyPatched` only checks for the new markers, a second pass would then treat that partially-patched bundle as complete. Please either collect/count the padding matches and fail-soft unless it is unique, or replace all matches by index/range.

ilysenko (approved):
Reviewed the current head against the latest upstream DMG behavior. The core patch remains fail-soft, the layout-mode coverage is focused, and the optional multi-match hardening can remain a follow-up.

PR #1094, opened 2026-07-20
Fix managed Node launchers and preserve Nix runtime

rlakki08:
## Summary
- repair upstream npm/npx launchers that hard-code `/usr/bin/node` after copying a managed Node runtime
- leave Nix-patched store launchers unchanged, avoiding writes to immutable `/nix/store` paths
- stage the upstream DMG intelligence helper in update-builder bundles
- cover copied launchers, existing runtimes, and Nix-patched launchers in smoke tests

## Validation
- Nix package build exercised the fixed runtime path: managed Node copied from the Nix store, then completed DMG extraction, native-module setup, and patching without the former `sed` permission error or Node download fallback.

The rolling upstream DMG currently fails the repository's required UI-patch compatibility checks after that point. The committed flake pin is unchanged; CI builds its pinned input.

### DIFF
--- a/scripts/lib/node-runtime.sh
+++ b/scripts/lib/node-runtime.sh
@@ -108,8 +108,10 @@ node_toolchain_compatible() {
 # `#!/usr/bin/node` shebang.  That makes the managed runtime unexpectedly
 # depend on a system Node installation when a caller executes `npx` directly
 # (as the installer does while patching app.asar).  Keep the launchers inside
-# the managed runtime self-contained by pointing their shebangs at its node
-# binary.  Resolve symlinks first so npm/npx remain symlinks in bin/.
+# copied managed runtime self-contained by pointing their shebangs at its node
+# binary. Do not run this on pre-existing runtimes: Nix supplies those as
+# read-only store symlinks. Resolve symlinks first so npm/npx remain symlinks
+# in bin/.
 repair_managed_node_launchers() {
     local runtime_dir="$1"
     local node_path="$runtime_dir/bin/node"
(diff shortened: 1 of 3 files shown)


ilysenko:
Thanks for taking the time to investigate this and for iterating on the Nix behavior. We reviewed the final approach across the installer, managed runtime, package builders, updater, and Nix paths.

We’re going to close this PR rather than add more core runtime complexity right now. The existing standard path continues to work without this change, and users affected by an environment-specific Node lookup issue can install a system Node runtime as a workaround.

There are also several technical concerns with the current implementation:

- The pinned official Node v22.22.2 archive uses `#!/usr/bin/env node`, while the repair only matches `#!/usr/bin/node`, so it does not modify the production managed runtime. The smoke fixture does not match the real archive layout.
- The replacement writes an absolute candidate/build path into the shebang. That path becomes stale after candidate promotion, package staging, AppImage assembly, or copying the runtime into the update-builder bundle.
- Resolving npm/npx symlinks with `readlink -f` and editing the resolved target does not ensure that the target remains inside the copied runtime. An absolute symlink could cause an attempted write to an external or immutable file.
- The branch now conflicts with `main`, and the `upstream-dmg-intel.js` staging change is already present there.
- The final head has not run the relevant GitHub CI matrix; only the contributor-limit check completed.

The managed Node, installer, updater, packaging, and Nix paths are already tightly coupled and expensive to maintain. At this point, the additional shared behavior and cross-format support burden are not justified by the demonstrated need.

Thank you again for the investigation and the test work. Closing this is a scope and maintenance decision, not a reflection on the effort involved.

PR #1095, opened 2026-07-20
feat: keep thread history visible across model providers

Mizoreww:
## Summary

- add a default-off `cross-provider-thread-history` Linux feature that requests `modelProviders: []` for current Recent, Projects, archive, resolver, and subagent enumeration surfaces
- make cold resume select `Desktop special provider ?? current config model_provider ?? Codex default`, while keeping the stored thread provider as provenance; already-loaded threads still require a Desktop restart before rebinding
- document the safe split-auth setup: global ChatGPT login for subscription/Remote Control plus a provider-scoped `experimental_bearer_token` for custom model requests
- register the feature for Nix and the current-DMG watchdog so future upstream chunk drift is caught

This addresses a Linux Desktop failure mode where changing the active provider hid local threads and could later resume a historical thread against the wrong endpoint. No credentials, auth files, or Remote Control device keys are included.

No linked issue; this is based on a directly reproduced user report.

## Validation

- `node --test linux-features/cross-provider-thread-history/test.js` — 15/15 passed
- Node 24 focused feature/acceptance/patcher suite — 426/426 passed
- latest upstream DMG `26.715.52143` (`3e5055c1…fc5e5`) feature-enabled build — accepted with the existing six unrelated optional core warnings; feature descriptor applied exactly once
- extracted final ASAR — 5/2/1 enumeration markers, one resume marker in the manager bundle, ESM syntax valid
- isolated app-server A/B matrix — cold A→B resume sent only B's scoped dummy token to B while retaining A provenance
- final Debian package — ASAR byte-identical to the validated build; no private-state paths or reported key prefix
- Matt Pocock two-axis final review — Standards: 0 findings; Spec: 0 findings
- full Node 24 suite: 1359 passed, 1 skipped, 2 failures in unchanged `directory-only-working-tree-watch` timing tests; the remaining deterministic failure reproduces byte-for-byte on `origin/main` in a detached worktree

## Checklist

- [x] This does not fix upstream drift; it targets only the latest `CODEX.DMG` and adds no legacy fallback path.
- [ ] I added or updated relevant tests, ran the validation listed above, and confirmed that required CI checks pass. *(Local validation is complete; hosted required checks are pending.)*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Added
 
+- An opt-in `cross-provider-thread-history` Linux feature keeps local Recent,
+  Projects, archive, and lookup lists visible after the active model provider
+  changes. It requests an explicit unfiltered provider list and binds cold
+  resume to the active provider without rewriting persisted thread provenance
+  or authentication state.
 - A shared upstream DMG acceptance profile now produces the same structured
   decision for local installs, updater rebuilds, and scheduled CI. Scheduled
   rejections create one fingerprinted drift issue and supersede issues for
--- a/README.md
+++ b/README.md
@@ -215,6 +215,7 @@ workarounds.
 | Wrapper updater button | Opt-in | `codex-wrapper-updater` | [Docs](linux-features/codex-wrapper-updater/README.md) |
 | Conversation mode | Opt-in | `conversation-mode` | [Docs](linux-features/conversation-mode/README.md) |
 | Copilot reasoning effort defaults | Opt-in | `copilot-reasoning-effort` | [Docs](linux-features/copilot-reasoning-effort/README.md) |
+| Cross-provider thread history | Opt-in | `cross-provider-thread-history` | [Docs](linux-features/cross-provider-thread-history/README.md) |
 | Directory-only working-tree watch | Opt-in | `directory-only-working-tree-watch` | [Docs](linux-features/directory-only-working-tree-watch/README.md) |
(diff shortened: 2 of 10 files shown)


ilysenko (changes requested):
Thanks for the carefully tested feature. Before merging, please update the documentation for the new cold-resume behavior. Rebinding a persisted thread to the currently active provider may send that thread's existing context to a different third-party endpoint, so the README should state this explicitly before users opt in.

Please also make the split-auth example prefer the provider `env_key` mechanism. Codex's current configuration schema discourages storing `experimental_bearer_token` directly in `config.toml` for security reasons. If the plaintext option must remain documented, present it only as a clearly labelled fallback with its security trade-off.

Mizoreww:
Thanks for the review. Both points are correct.

After opening this PR, I found that the current CC Switch release already provides opt-in controls for the original use case:

1. preserving the official ChatGPT login while switching model traffic to a third-party provider; and
2. unifying/migrating official and third-party Codex history, with backups and an explicit warning that cross-provider continuation may fail for backend-bound `encrypted_content`.

That makes this Linux patch redundant for my setup, while its cold-resume rebinding also carries the data-routing risk you identified: existing thread context may be sent to the newly active third-party endpoint.

I also agree that, if split-auth guidance is documented, `env_key` should be the preferred mechanism. A plaintext `experimental_bearer_token` should only be a clearly labelled fallback with its security trade-off. No credentials or private auth data were committed in this PR.

Rather than expand and maintain a Linux patch for behavior already handled at the user-space provider-management layer, I am closing this PR. Thank you for catching both documentation/security issues.

PR #1102, opened 2026-07-20
Retarget Suggested Prompts to the current desktop bundles

PinguuSS:
## Summary

- retarget the Suggested Prompts app-page descriptor to the current active bundle
- retarget the General Settings row and Home content descriptors to their current assets
- remove the prior-DMG hashes from source and fixtures without adding broad matchers or compatibility fallbacks

## Context

This maintains the default-off Suggested Prompts tweak introduced in #1083 after the maintainer approved the one-tweak-at-a-time direction in #1038.

While isolating the opt-in configuration from the GNOME/X11 tray report in #1100, the current acceptance gate rejected the Suggested Prompts profile: the main-process descriptor still applied, but all three webview descriptors were `skipped-optional`. The minified contracts remain valid in the current artifact; only the owning asset names changed.

This PR is independent of the GNOME/X11 tray issue; #1100 remains unresolved.

## Current upstream identity

- repository base: `157ea148c704755351fe62868b3ff685b917c602` (merge of #1101)
- branch head: `afec3035497738c39ee864389ef1a49466d24063`
- ChatGPT Desktop: `26.715.52143`
- Electron: `42.3.0`
- DMG SHA-256: `3e5055c185b68369d3cee4f24b45eec30d11b4f0c848c0c7fe550103135fc5e5`
- DMG size: `618693003` bytes
- runtime: Ubuntu 25.10, KDE Plasma, Wayland

## Current-DMG contract

The existing atomic patch functions now target:

- app page: `app-initial~app-main~appgen-settings-page~page~appgen-library-page~appgen-page~appgen-setti~ogh9jurw-CrEakH0Y.js`
- General Settings: `general-settings-CsA3Lt9Z.js`
- Home content: `home-ambient-suggestions-content-C01Mwmkt.js`

The previous `Ccxu2qV_`, `Boi5S8Wz`, and `DNeFqrrf` assets are removed from the descriptors, and regression coverage asserts that those old filenames no longer match. Existing complete markers continue to prove idempotency; missing or mixed insertion points warn and leave the asset byte-identical.

## Runtime validation

The isolated Suggested Prompts identity was exercised on the current DMG with only this nested tweak enabled:

- General Settings row and Settings search result are visible
- the runtime toggle removes and restores cards without duplication
- project changes generate project-specific suggestions
- selecting a card populates the composer without submitting
- ChatGPT removes Codex project cards while retaining its own Home suggestions; returning to Codex restores them
- a 760x800 window stays within the viewport
[...]

PR #1103, opened 2026-07-20
Make reasoning label drift handling atomic

PinguuSS:
## Summary

- preflight all eight current Simplified Chinese reasoning-label keys before changing the locale asset
- reject mixed English/untranslated marker states without completing a partial patch
- add regression coverage proving missing and mixed contracts remain byte-identical

## Context

The post-#1102 UI Tweaks current-DMG matrix found an atomicity gap in `reasoning.keepEffortLabelsEnglish`. When one current label key was missing, the patch warned about that key after already replacing the labels it had found. `patchAssetFiles()` writes the returned source, so the candidate contained a partially modified locale asset and reported the descriptor as `applied-with-warnings`. Acceptance still protected the installed app, but the feature patch itself was not atomic.

This change gathers every replacement before writing any of them. A clean untranslated asset applies all eight labels, a complete English asset remains idempotent, and missing or mixed markers now warn once and return the original source. The drift report therefore becomes `skipped-optional` with byte-identical asset contents.

## Current upstream identity

- repository base: `54c11eca98a51fece5c520c1c30b601ae166beae` (merge of #1102)
- branch head: `7c4fdea`
- ChatGPT Desktop: `26.715.52143`
- Electron: `42.3.0`
- DMG SHA-256: `3e5055c185b68369d3cee4f24b45eec30d11b4f0c848c0c7fe550103135fc5e5`
- DMG size: `618693003` bytes
- runtime: Ubuntu 25.10, KDE Plasma, Wayland

## Current-DMG validation

- isolated build with only English Reasoning Labels enabled: `accepted_with_warnings`, zero feature blockers
- first pass: `feature:ui-tweaks:reasoning-effort-labels-english` is `applied`
- second complete patch pass: descriptor is `already-applied`; `zh-CN-DacW0uyu.js` stays byte-identical at `a2cc30652036fd65bc7f42639714f526c816f19763e1d703b630d37983c5e3e5`
- synthetic missing-key drift: one warning, source byte-identical
- synthetic mixed-marker drift: one warning, source byte-identical
- generated locale asset passes `node --check`
- runtime in Simplified Chinese keeps the surrounding UI translated while showing `Low`, `Medium`, `High`, `XHigh`, `Max`, and `Ultra`
- the labels persist across a cold restart; the isolated profile was restored to `Auto detect` after QA

## Tests

- `node --test linux-features/ui-tweaks/test.js scripts/patch-linux-window-ui.test.js` (`457/457`)
- all other Linux Feature modules excluding the two independently rerun modules (`632/632`)
- `node --test linux-features/shared-app-server-socket/test.js` (`24/24`)
- first- and second-pass `scripts/ci/validate-patch-report.js` checks
- `node --check` on source, test, and generated current asset
- `git diff --check`

The monolithic Linux Feature run produced five `directory-only-working-tree-watch` timing/resource failures and one parallel socket-permission failure. The same five watcher failures reproduce unchanged on the clean pre-branch baseline, while the socket module passes `24/24` in isolation. The focused suite, the other 632 feature tests, and smoke all pass.

## Scope

This changes only the feature-local English reasoning-label patch and its tests. It does not alter the patch engine, other UI Tweaks, committed feature defaults, or updater behavior.

### DIFF
--- a/linux-features/ui-tweaks/patches/reasoning-effort-labels.js
+++ b/linux-features/ui-tweaks/patches/reasoning-effort-labels.js
@@ -13,7 +13,7 @@ const ENGLISH_REASONING_LABELS = Object.freeze({
 });
 
 function warn(message) {
-  console.warn(`WARN: ${message} - skipping missing ui-tweaks reasoning label markers`);
+  console.warn(`WARN: ${message} - skipping ui-tweaks reasoning label patch`);
 }
 
 function reasoningLabelConfig(context) {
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head: the feature-local change preserves atomic fail-soft drift handling, adds focused regression coverage, and all required checks pass.

PR #1106, opened 2026-07-20
Fix Linux settings icon registration

colonelpanic8:
## Summary

- patch the settings icon registry when adding the Linux desktop settings section
- support the current upstream split-bundle layout
- add idempotency and integration coverage for both consolidated and split settings bundles

## Root cause

Upstream moved the settings slug-to-icon map out of the settings-page bundle. The Linux desktop patch continued adding the navigation entry but no longer registered an icon for it, so the settings renderer received an undefined component and failed with React error #130.

## User impact

Opening Settings works again, including the Linux desktop settings page.

## Validation

- `node --test scripts/patch-linux-window-ui.test.js` (403 passed)
- `git diff --check`
- patched the current packaged webview assets and confirmed only the intended settings icon bundle changed
- launched an isolated patched app instance and verified both Settings and Linux desktop settings render normally

The broader `bash tests/scripts_smoke.sh` run progressed through packaging and launcher checks before an unchanged NixOS-only harness issue: a restricted `/usr/bin:/bin` PATH could not resolve an `env bash` shebang.

### DIFF
--- a/scripts/patches/impl/keybinds-settings.js
+++ b/scripts/patches/impl/keybinds-settings.js
@@ -637,7 +637,7 @@ function collectLinuxDesktopIconMapPatches(extractedDir) {
 
   const candidates = fs
     .readdirSync(webviewAssetsDir)
-    .filter((name) => name.endsWith(".js"))
+    .filter((name) => /^use-visible-settings-sections-.*\.js$/.test(name))
     .sort()
     .filter((name) => {
       const source = fs.readFileSync(path.join(webviewAssetsDir, name), "utf8");
--- a/scripts/patches/impl/keybinds-settings.js
+++ b/scripts/patches/impl/keybinds-settings.js
@@ -1167,13 +1167,20 @@ function applyLinuxDesktopSettingsRoutePatch(
 }
 
 function applyLinuxDesktopSettingsIconPatch(currentSource) {
-  if (/[A-Za-z_$][\w$]*=\{"linux-desktop":[A-Za-z_$][\w$]*,"general-settings":[A-Za-z_$][\w$]*,/.test(currentSource)) {
+  const patchedIconPattern = /[A-Za-z_$][\w$]*=\{"linux-desktop":[A-Za-z_$][\w$]*,"general-settings":[A-Za-z_$][\w$]*,(?=[^;]{0,3000}"keyboard-shortcuts":[A-Za-z_$][\w$]*[,}])/g;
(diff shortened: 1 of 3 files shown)


ilysenko (changes requested):
Thanks for fixing the current split settings icon-map regression. The current-DMG path looks right, but this PR still retains and extends coverage for the obsolete consolidated settings bundle shape. Because this repository supports only the latest upstream CODEX.DMG, drift fixes should remove old compatibility shapes instead of preserving them. Please drop the consolidated settings-bundle icon-map coverage/support added here and keep the patch/tests focused on the current split icon-map bundle.

scripts/patches/impl/keybinds-settings.js
```diff
@@ -1125,6 +1166,22 @@ function applyLinuxDesktopSettingsRoutePatch(
...
+  if (/[A-Za-z_$][\w$]*=\{"linux-desktop":[A-Za-z_$][\w$]*,"general-settings":[A-Za-z_$][\w$]*,/.test(currentSource)) {
+    return currentSource;
+  }
+
+  const iconPattern = /([A-Za-z_$][\w$]*=\{)"general-settings":([A-Za-z_$][\w$]*),(?=[^;]{0,3000}"keyboard-shortcuts":[A-Za-z_$][\w$]*[,}])/;
+  if (!iconPattern.test(currentSource)) {
+    throw new Error("Required Keybinds settings patch failed: could not add Linux desktop icon");
+  }
+
+  return currentSource.replace(
```
avifenesh, line 1186: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

New ASAR bundle patch helpers need to either process every eligible occurrence or explicitly assert that the needle is unique. Here the idempotency check returns as soon as *any* icon map is already patched, and this non-global `replace` only rewrites the first matching `"general-settings"` map. If a split bundle carries two matching slug-to-icon maps in the same asset, the later one is left without `"linux-desktop"`, so the settings renderer can still receive an undefined icon component. Please collect/replace all matches, or count the matches and fail unless there is exactly one before patching.

avifenesh (commented):

---

I found one ASAR patching invariant issue in the new icon-map helper.

ilysenko (approved):
Reviewed the current head. The current-DMG cleanup, icon-map uniqueness invariant, and split-layout smoke fixture are addressed, and all required checks are green.

PR #1107, opened 2026-07-21
Add opt-in workaround for sidebar thread hover-card stalls

colonelpanic8:
## What changed

- add a disabled-by-default `disable-sidebar-thread-hover-cards` Linux feature
- use the existing local thread row `disableHoverCard` path instead of mounting the hover-card metadata view
- expose the feature through the supported Nix feature ID list
- parse ordinary XML `Info.plist` app versions directly, retaining Python only as a fallback for binary or unusual plists

Thread navigation, status indicators, context menus, and row actions remain available. Default builds are unchanged unless the feature is explicitly enabled.

## Why

The local thread hover card mounts repository, branch, worktree, and status metadata, including watcher-backed Git work. On affected Linux systems, moving through thread rows can dramatically stall the renderer and make the sidebar effectively unusable.

This may not be the ideal upstream fix: disabling the hover card is deliberately blunt, and fixing its watcher/query lifecycle would preserve the metadata UI. It may therefore be reasonable not to merge this workaround. However, keeping it as an opt-in feature gives affected users a usable app now without changing default behavior.

The plist change fixes the one real test failure exposed after rebasing: XML app-version extraction unnecessarily required `python3` on `PATH`, which is not guaranteed in narrow Nix or GUI environments.

## Validation

- reproduced the hover-card stall path in an isolated Linux desktop session
- verified that the patched renderer no longer opens the card and remains responsive
- applied the feature to the current installed webview bundle; it changes one disable assignment and is idempotent
- `node --test scripts/patch-linux-window-ui.test.js linux-features/disable-sidebar-thread-hover-cards/test.js scripts/lib/linux-features.test.js scripts/patches/descriptor.test.js scripts/patches/runner.test.js` — 424/424 passed in a clean worktree
- `nix build .#checks.x86_64-linux.nix-linux-features-evaluation --no-link`


colonelpanic8:
Closing because the underlying hypothesis was disproven in real use: disabling the thread hover-card metadata view did not resolve the severe sidebar mouseover slowdown. I am going back to renderer profiling around the broader sidebar hover path before proposing another fix.

PR #1101, opened 2026-07-20
Retarget Dock icon to the current desktop bundles

PinguuSS:
## Summary

- retarget the Dock Icon main-process contract to the current `main-D-AEKvtN.js` aliases
- retarget the Settings row and search descriptors to the active current-DMG assets
- update the tray through the core-registered `codexLinuxTray` instead of calling the current asynchronous tray factory as a getter
- refresh the fixtures and remove the prior-DMG asset names

## Context

This follows the maintainer's invitation to try a focused PR in #1100. A zero-feature build from current `main` registered a working StatusNotifier item on KDE/Wayland, so this PR does not claim to resolve the separate GNOME/X11 report.

The isolated Dock Icon profile exposed a concrete opt-in failure first: the current artifact was rejected because all three Dock descriptors had drifted. Runtime QA then caught that `dae` now owns asynchronous tray creation; calling `dae()?.tray` from the injected icon helper could start it without configuration and prevented reliable tray registration. The current core patch already owns the concrete Electron tray in `codexLinuxTray`, which is the object this feature now updates.

## Current upstream identity

- repo base: `a29387ebc1c76cc51c942d97c5f5c609a81cdccf`
- DMG SHA-256: `3e5055c185b68369d3cee4f24b45eec30d11b4f0c848c0c7fe550103135fc5e5`
- DMG size: `618693003` bytes
- app: `26.715.52143`
- Electron: `42.3.0`
- runtime: Ubuntu 25.10, KDE Plasma, Wayland

## Visible behavior

- Appearance shows the upstream Dock icon row with ChatGPT and Codex choices
- Settings search returns Dock icon
- switching Codex to ChatGPT updates the live window and StatusNotifier icon
- the selected icon survives a cold restart
- closing the window keeps the isolated tray process alive, activating the tray restores the window, and tray Quit exits the app cleanly

## Validation

- `node --test linux-features/ui-tweaks/dock-icon.test.js` (24/24)
- `node --test linux-features/ui-tweaks/test.js` (53/53)
- `node --test scripts/patch-linux-window-ui.test.js` (402/402)
- `node --test linux-features/*/test.js` (766/766 with inherited updater hook variables removed)
- direct first/second pass against `main-D-AEKvtN.js`; byte-stable second pass and `node --check`
- direct second pass against `general-settings-CsA3Lt9Z.js` and `settings-page-BOdFN1v1.js`; one complete marker each
[...]

PR #1113, opened 2026-07-21
Select UI Tweaks renderer assets by semantic contract

PinguuSS:
## Summary

- retarget Dock icon and Suggested Prompts to the current renderer assets
- add semantic, unique asset selection for hash-only renderer filename churn
- keep genuine current-DMG contract drift fail-closed and visible in acceptance
- record the concrete selected renderer asset in the patch report

## Problem

The current upstream DMG changed renderer asset hashes without changing the Dock icon or Suggested Prompts contracts. The exact filename descriptors from #1101 and #1102 therefore skipped five active renderer assets, and enabled-feature drift rejected the candidate.

Updating five hashes restores this DMG, but repeats the same maintenance-only retarget whenever upstream rebuilds an otherwise compatible chunk. This PR keeps the current-DMG semantic contracts strict while treating the final renderer hash as non-semantic.

## Approach

Webview descriptors may now provide `assetMatch(source, assetName, context)`. The patch engine scans only the descriptor's stable filename family, then requires exactly one complete semantic contract:

| Semantic matches | Result |
| --- | --- |
| zero | warn, preserve every candidate byte, enabled-feature acceptance rejects |
| one current contract | apply once and report the concrete `assetName` |
| one complete patched contract | report `already-applied` with the same `assetName` |
| multiple | warn with the ambiguous filenames, preserve every candidate byte, acceptance rejects |

Dock icon selectors require one complete current or patched settings/search needle. Suggested Prompts selectors reuse the same current/patched/drifted contract classifiers as their atomic apply functions. Mixed markers, partial patches, changed counts, renamed chunk families, and ambiguous ownership remain drift. No previous-DMG contract or compatibility fallback is retained.

## Current DMG

- SHA-256: `447f3bf318ba01822371db1ec01c1b9553224e7f8cb5ae577e17da441aaa3a30`
- size: `618781423` bytes
- HTTP fingerprint: `etag=0x8DEE6ED04CFB50F|last_modified=Tue, 21 Jul 2026 05:57:59 GMT|content_length=618781423`
- app: `26.715.61943`
- Electron: `42.3.0`
- repository base: `5882958bf7d2a54837d2a6d8ed80b30c8cb6c318`
- environment: Ubuntu 25.10, KDE Plasma, Wayland

## Validation

- core patch suite: `409/409`
- focused UI Tweaks and descriptor/runner tests: `90/90`
- `node --check` on every changed patch module and every selected real renderer asset
- `git diff --check`
- first real-ASAR pass: all 7 Dock/Suggested descriptors `applied`
- second real-ASAR pass: all 7 descriptors `already-applied`
- patch report records the five concrete active renderer filenames on both passes
- isolated current-DMG candidate: `accepted_with_warnings`, zero blockers, all 13 enabled `ui-tweaks` descriptors applied
- side-by-side runtime used an isolated app id, state, cache, user data, and webview port; the installed daily driver remained untouched
- General Settings rendered Suggested Prompts and Appearance rendered the Dock icon controls
- changing ChatGPT to Codex updated KDE StatusNotifier immediately from `2048x2048` to `956x956`
- cold restart preserved the Codex selection, `956x956` tray image, Dock icon controls, and Suggested Prompts row

The aggregate `node --test linux-features/*/test.js` run from the original retarget advanced through test 673 but did not terminate because unrelated runtime-feature handles remained open. The affected feature, shared patch engine, runner, and full core patch suites above all complete cleanly.

## Maintenance commitment

I will actively supervise Dock icon and Suggested Prompts against each supported current upstream DMG and maintain both tracked opt-ins when renderer ownership or semantic contracts change. This selector handles hash-only filename churn; real upstream drift intentionally remains visible and blocking so it can be reviewed and retargeted with current-DMG evidence.

## Scope

This PR is limited to current-DMG UI Tweaks asset selection and the smallest generic patch-engine extension required to make that selection safe. It does not change feature defaults, updater provenance, renderer cache invalidation, nested feature-picker persistence, or unrelated optional core patches.


ilysenko (changes requested):
Thanks for making renderer asset selection semantic and ambiguity-safe. The new `assetMatch(source, assetName, context)` callback is not wrapped with the feature-scoped context in `wrapFeaturePatchDescriptor`, unlike `apply`, `enabled`, and the other descriptor callbacks. As a result, repository and local feature matchers cannot read the documented `context.feature.settings`; they receive the raw engine context instead. Please wrap `assetMatch` with `featureContext(context, feature)` and add a regression test whose matcher depends on feature settings.

ilysenko (approved):
Reviewed the current head. The previously requested feature-context fix is implemented and covered by a focused regression test; all required checks are green.

Issue #1001, opened 2026-07-15
Native package wrapper updater silently cannot discover updates


PinguuSS:
## Summary

Native package installs can expose and persist the **Check for ChatGPT Desktop for Linux updates** toggle while being unable to discover any wrapper update with the default configuration.

This affects the optional `codex-wrapper-updater` path that tracks this repository. It is separate from the normal updater path that tracks the upstream OpenAI DMG.

With the toggle enabled, `codex-update-manager check-wrapper` runs successfully, recognizes the installed wrapper metadata, but records no candidate because the default packaged `builder_bundle_root` is a frozen directory without `.git`. The in-app Update button therefore remains hidden, with no indication that wrapper detection is unavailable.

## Current environment

- audited repository `main`: `c06cd5ab3d658becb1cb57f1728c850b3b4a7dd0`
- package: `codex-desktop_2026.07.14.182239_amd64.deb`
- installed wrapper: `0.9.7` at `94c670a70a3894ce3076a6db914ae193edf7d446`
- available wrapper during the control test: `0.10.1` at `c06cd5ab3d658becb1cb57f1728c850b3b4a7dd0`
- upstream app: ChatGPT Desktop `26.707.72221`
- Electron: `42.1.0`
- DMG SHA-256: `40e34814e74e30943c209ebd4da94cd4de3581a52c5bffbe2bcf2e488d6361c6`
- Ubuntu `25.10`, KDE Plasma Wayland

No open issue or pull request was found for this exact detection gap. #860 fixed the current-DMG runtime and UI patching. #941 concerns the rebuild trust boundary and should be considered before choosing an implementation, but it does not make native-package wrapper detection functional with the current defaults.

## Reproduction

1. Install a native package built with `codex-wrapper-updater` enabled.
2. Leave `~/.config/codex-update-manager/config.toml` absent so packaged defaults are used.
3. Enable **Settings -> Linux desktop -> Check for ChatGPT Desktop for Linux updates**.
4. Confirm that `settings.json` contains:

   ```json
   {
     "codex-linux-wrapper-updates-enabled": true
   }
   ```

5. Run:

   ```bash
   codex-update-manager check-wrapper --json
   ```

The command exits successfully and identifies the installed wrapper, but records no candidate:

```json
{
  "installed_wrapper_version": "0.9.7",
  "installed_wrapper_commit": "94c670a70a3894ce3076a6db914ae193edf7d446",
  "candidate_wrapper_version": null,
[...]

PR #1117, opened 2026-07-21
Preserve updater build source provenance

PinguuSS:
## Summary

- Capture the wrapper source Git identity before the updater copies the builder into its no-Git workspace.
- Stage that identity through the existing `.codex-linux/source-info.json` contract so generated app build info and packaged update-builder metadata retain the same wrapper commit.
- Preserve packaged metadata for no-Git sources, reject identity inherited from a parent checkout, retain dirty state, and sanitize supported network remotes.
- Strip credentials from HTTP(S), SSH, and SCP-like remotes; omit local paths and custom transports from installed metadata.
- Verify the same provenance handoff through fake Debian, RPM, and pacman package builders.
- Bump `codex-update-manager` to `0.10.2`.

This addresses the reproducible state where an updater-built package contained `source.commit=null` and `source.provenance=unknown`, leaving `installed_wrapper_commit` unavailable even though the updater candidate checkout had a known wrapper commit.

The change does not alter wrapper update discovery/comparison or current-DMG patch targeting. Related discussions in #1001 and #1049 focused on those separate detection mechanisms.

## Reproduction Identity

- App version: `26.715.70719`
- Electron: `42.3.0`
- SHA-256: `95c66f30ad0ee946cfec2b83c50e68744700be5da76898c5de484a755000f4a3`
- Size: `618776575` bytes
- HTTP fingerprint: `etag=0x8DEE74C91EAF288|last_modified=Tue, 21 Jul 2026 17:21:58 GMT|content_length=618776575`

That exact DMG reproduced the missing wrapper provenance. The candidate was not promoted because an unrelated enabled feature drifted, so this PR makes no claim to fix that feature drift or the acceptance verdict.

The branch is rebased onto current `main` at `aa1b7b8`, which pins upstream app `26.715.72359` with Electron `42.3.0` in #1128. The provenance transport is independent of DMG bundle shape.

## Validation

- `cargo test -p codex-update-manager` (275 passed)
- `CI_CACHE_DIR=/tmp/codex-desktop-linux-ci TMPDIR=/tmp ./scripts/ci-local.sh pr`
  - core: updater, Node/patch, and script smoke suites passed
  - Debian, RPM, and pacman package jobs passed
  - no-updater variants passed

The focused regression tests cover credential-bearing SSH and SCP-like remotes, local/custom remote rejection, and metadata output from all three fake native package builders.

No daily-driver installation or runtime replacement was performed. The change is confined to updater builder metadata transport and is covered by the full package matrix.

## Checklist

- [x] This pull request is ready for review.
- [x] I followed `CONTRIBUTING.md` and kept the change focused.
- [x] The reproduction identity and current-main base are documented above.
- [x] I reviewed the final diff and validation results.
- [x] Tests cover clean, dirty, nested-checkout, no-Git packaged-metadata, remote privacy, cross-format package output, and complete updater-build paths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- Updater rebuild workspaces now retain the Git identity of the wrapper source
+  after `.git` is stripped, so installed build metadata and packaged
+  update-builder metadata report the wrapper commit instead of `unknown`.
 - V2 pets now look toward the live pointer position after successful Linux
   Computer Use click, scroll, and drag actions, then return to their normal
   animation. The bridge is isolated per app instance and fails softly when its
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -560,7 +560,7 @@ dependencies = [
 
 [[package]]
 name = "codex-update-manager"
-version = "0.10.1"
+version = "0.10.2"
 dependencies = [
  "anyhow",
  "chrono",
--- a/updater/Cargo.toml
+++ b/updater/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "codex-update-manager"
-version = "0.10.1"
+version = "0.10.2"
 edition = "2021"
 
 [dependencies]
--- a/updater/src/builder.rs
+++ b/updater/src/builder.rs
@@ -346,22 +346,50 @@ fn sanitize_git_remote(remote: Option<String>) -> Option<String> {
         || Path::new(&value).is_absolute()
         || value.starts_with("./")
         || value.starts_with("../")
+        || value.starts_with('~')
+        || value.contains('\\')
     {
         return None;
     }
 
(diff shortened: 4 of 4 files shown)


ilysenko (changes requested):
Thanks for the focused provenance fix. I found two blockers:

1. `sanitize_git_remote` removes credentials only from HTTP(S) URLs, so SSH/custom remotes and local-path forms can persist credentials or private path information into installed and packaged metadata. Please safely sanitize or reject every supported remote form and add tests for credential-bearing SSH/SCP-like and local remotes.
2. `builds_update_with_fake_bundle` always reads `dist/package-source-info.json`, but only the fake Debian builder creates it. The test therefore fails on supported RPM and pacman hosts. Please make every fake package builder emit the asserted metadata or make the assertion format-independent.

ilysenko (approved):
Both previously requested blockers are fixed on the current head. Re-reviewed the focused updater, sanitization, and cross-format test changes; final-head checks are green.

PR #1118, opened 2026-07-21
Fix frameless titlebar right padding

wy3z:
Restores the standard 8px right padding for header controls when the opt-in `frameless-titlebar` feature is enabled.

Default titlebar builds are unchanged.

Test: `node --test linux-features/frameless-titlebar/test.js`


ilysenko (approved):
Reviewed the current head. The frameless-titlebar fix is scoped, tested, and safe to merge.

PR #1119, opened 2026-07-21
Recover stale shared app-server sockets after crashes

moxw:
## Summary

- record the Desktop PID and Linux process start identity in the shared socket lock
- reclaim dead-owner locks and unbound stale socket inodes while preserving live endpoints
- guard async startup and disposal, with focused coverage for recovery races

## Why

After abnormal Desktop termination, the app-server process can be gone while the `.lock` file and/or Unix socket pathname remain. Subsequent launches then fail with `shared app-server socket is already owned` until those paths are cleaned up manually.

Recovery remains fail-closed:

- matching, live, or unverifiable owners are preserved
- connectable sockets are preserved
- only a dead owner with an absent or unbound socket is reclaimed
- socket and lock inode identities are checked before unlinking
- legacy metadata-free locks remain protected for 15 seconds

## Validation

- `node --test linux-features/shared-app-server-socket/test.js` (28/28)
- `TMPDIR=/tmp node --test scripts/patch-linux-window-ui.test.js` (409/409)
- latest-DMG local rebuild and Arch package installation
- controlled termination test leaving both the ownership lock and an unbound socket inode; the next Desktop launch recovered automatically


ilysenko (approved):
Reviewed the current head. The stale-owner recovery is fail-closed for live or unverifiable endpoints, preserves inode identity checks, and is covered by focused tests. Looks good to merge.

PR #1121, opened 2026-07-22
Fix Chrome app-server proxy inheritance on Linux

No1Hoo:
## Summary

- route the Linux Chrome native-host manifests through an owner-trusted launcher in the managed Chrome plugin cache
- recover `HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`, and their lowercase variants from the systemd user environment without overriding inherited values
- fall back to GNOME manual proxy settings when GUI-launched Chrome has no exported proxy environment
- keep native-host arguments and the managed `latest` extension-host target unchanged

Chrome starts native messaging hosts outside the desktop launcher environment. On proxied Linux desktops this allowed the side panel to start, but its Codex app-server attempted a direct connection and remained stuck in `Thinking`. The generated native-host launcher restores the user-session proxy environment before executing the bundled extension host.

This stays separate from the bundled-plugin trust/materialization fix in #1105: it does not bless or rewrite plugin source trees and only changes the manifest target used after the existing cache trust checks pass.

## Validation

- RED: `bash tests/scripts_smoke.sh` exited 1 after Chrome cache sync because the direct extension-host received no proxy variables
- GREEN: `bash tests/scripts_smoke.sh` — `All script smoke tests passed`
- `node --test scripts/patch-linux-window-ui.test.js scripts/lib/linux-features.test.js scripts/patches/descriptor.test.js scripts/patches/runner.test.js linux-features/*/test.js` — 1200 passed, 0 failed
- `node --check scripts/patches/impl/main-process/browser.js`
- live environment: Ubuntu 24.04, GNOME/X11, `.deb`, Desktop 26.715.71837; Chrome side-panel replies resumed after restart with the native host using the GNOME proxy

## Checklist

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -973,7 +973,10 @@ write_chrome_native_host_launcher() {
         'run_proxy_probe() {' \
         '    local output_file timeout_file probe_pid watchdog_pid rc=0 use_process_group=0' \
         '    output_file="$(mktemp "${TMPDIR:-/tmp}/codex-chrome-proxy-probe.XXXXXX")" || return 1' \
-        '    timeout_file="$output_file.timeout"' \
+        '    timeout_file="$(mktemp "${TMPDIR:-/tmp}/codex-chrome-proxy-timeout.XXXXXX")" || {' \
+        '        rm -f -- "$output_file"' \
+        '        return 1' \
+        '    }' \
         '    if command -v setsid >/dev/null 2>&1; then' \
         '        setsid "$@" >"$output_file" 2>/dev/null &' \
         '        use_process_group=1' \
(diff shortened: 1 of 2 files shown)


ilysenko (changes requested):
Thanks for the focused fix and regression coverage. I found three blockers before this is safe to merge:

1. An executable `native-host` pre-seeded while the cache root was unsafe is retained after permissions are hardened because the marker is trusted without regenerating the file. Please force regeneration after unsafe-cache detection and add a regression case.
2. Proxy restoration needs to preserve `NO_PROXY`/`no_proxy` and GNOME `ignore-hosts`, and honor GNOME’s HTTP-to-HTTPS/`use-same-proxy` semantics. The current default manual GNOME shape can still leave HTTPS direct or proxy bypassed destinations.
3. The new smoke fixture stages only the x64 host, while the wrapper selects arm64 on ARM64 systems. Please make the fixture architecture-aware or stub `uname`.

ilysenko (approved):
Verified the requested fixes and the final clean, green head.

PR #1125, opened 2026-07-22
Fix Linux desktop settings visibility on current DMG

magicJie:
## Summary

- patch the current standalone settings visibility catalog so the generated Linux desktop section is not filtered out
- collect the visibility bundle alongside route, navigation, metadata, and ordering assets while retaining support for combined bundle shapes
- add regression coverage for both modern and split settings assets

## Validation

- `node --test --test-name-pattern="Linux desktop settings|Linux desktop section" scripts/patch-linux-window-ui.test.js` (10 passed)
- `node --test linux-features/codex-wrapper-updater/test.js` (14 passed)
- `git diff --check`
- rebuilt and installed the current `26.715.72028` DMG; acceptance completed with the existing unrelated optional-patch drift warnings, and the installed visibility bundle contains the `linux-desktop` case

## Known unrelated test result

The full `scripts/patch-linux-window-ui.test.js` run completed with 409 passes and one failure in `materializes trusted Linux bundled plugins through a private staging root`. The failure is the sandbox temporary-directory trust check and does not exercise the settings assets changed here.


ilysenko (changes requested):

- Please remove the legacy combined-bundle visibility paths, including the visibility helper calls from the section/navigation patchers and the combined visibility fixtures. Current DMG 26.715.72028 keeps this switch only in `use-visible-settings-sections-*`, and CONTRIBUTING.md requires current-DMG-only patch shapes.
- Please make standalone visibility discovery fail-soft but observable: require exactly one current visibility asset/match, and report zero or ambiguous matches as `skipped-optional` with a warning instead of silently returning `matched: true`. Please add regression coverage for that drift behavior.

The core fix itself looks right once these are addressed.

magicJie:
Addressed both blockers in `dbb7010`:

- removed visibility patching from the section/navigation patchers and removed combined-bundle visibility fixture shapes
- restricted visibility discovery to exactly one current `use-visible-settings-sections-*` asset and exactly one visibility match; zero/ambiguous discovery now warns and reports `skipped-optional` without writing assets
- added regressions for zero assets, multiple assets, and multiple matches within one asset

Validation: relevant core tests 13/13 passed, wrapper updater tests 14/14 passed, and `git diff --check` passed. The full core suite remains 411/412 with only the previously noted unrelated sandbox staging-root trust failure.

magicJie:
Follow-up `8fb9767` updates the script smoke fixture to the current split visibility-bundle shape. The prior CI failure was the old fixture omitting `use-visible-settings-sections-*` visibility logic, so the newly strict discovery correctly skipped the patch. The focused smoke test now passes locally; shell syntax and diff checks also pass.

ilysenko (approved):
The requested current-DMG cleanup, fail-soft drift reporting, and smoke fixture updates are all addressed. CI and the upstream-DMG build pass.

PR #1127, opened 2026-07-22
fix(chrome-plugin): resolve canonical path and normalize permissions for Chrome native host

sanhee:
## Summary

Fixes an issue where launching the official [ChatGPT Chrome Extension](https://chromewebstore.google.com/detail/chatgpt/hehggadaopoacecdllhhajmbjkdcmajg) (Extension ID: `hehggadaopoacecdllhhajmbjkdcmajg`) fails with `Unable to start ChatGPT: Codex app-server manifest entry is missing required path parent`.

### Root Cause
1. **Symlink Traversal in Parent Chain**: The Chrome Native Messaging Host manifest (`com.openai.codexextension.json`) was generated referencing `.../chrome/latest/...`, where `latest` is a symlink pointing to the current version directory (e.g. `26.715.72028`). Upstream OpenAI Rust backend (`codex-chrome-extension-host`) executes `validate_trusted_parent_chain()`, which rejects any parent directory path containing symlinks (`is_symlink() == true`), throwing `required_path_error("parent")`.
2. **Unsafe Group Write Permissions (`0775`)**: Extracted vendor plugins and standalone packages under `~/.codex/plugins` and `~/.codex/packages` were created with `0775` permissions. `validate_trusted_parent_chain()` checks `has_unsafe_write_permissions(&metadata)` (`mode & 0o022 != 0`) and rejects directories with group/others write bits enabled.

### Fix
- **Canonical Path Resolution**: Update Chrome Native Messaging Host manifests and `chrome-native-hosts-v2.json` via `installManifest.mjs` patch to resolve paths to their real canonical directories instead of symlink aliases.
- **Permission Normalization**: Ensure strict `0755` permissions (`mode: 0o755` / `493`) when `installManifest.mjs` creates Native Host manifest directories during installation and staging.

## Validation

- **Parent Chain Diagnostics**: Executed custom ancestor traversal script against all manifest paths. Confirmed error count dropped from 9 `UNSAFE WRITE PERMISSIONS` errors to `0`.
- **Chrome Native Host IPC Handshake**: Tested stdio IPC handshake with `codex-chrome-extension-host` for [ChatGPT Chrome Extension](https://chromewebstore.google.com/detail/chatgpt/hehggadaopoacecdllhhajmbjkdcmajg) via canonical path. Verified successful JSON-RPC 2.0 response without `missing required path parent` error.
- **Unit Tests**: Ran project unit tests (`node --test scripts/lib/patch-chrome-plugin.test.js`) and verified core patch behavior remains intact.

## Checklist

### DIFF
--- a/scripts/lib/patch-chrome-plugin.js
+++ b/scripts/lib/patch-chrome-plugin.js
@@ -347,6 +347,33 @@ patchFileFirstMatch(path.join(scriptsDir, "installManifest.mjs"), {
     'linux:[".config/google-chrome/NativeMessagingHosts",".config/google-chrome-beta/NativeMessagingHosts",".config/google-chrome-unstable/NativeMessagingHosts",".config/BraveSoftware/Brave-Browser/NativeMessagingHosts",".config/chromium/NativeMessagingHosts"]',
 });
 
+patchFileFirstMatch(path.join(scriptsDir, "installManifest.mjs"), {
+  label: "Linux browser native host manifest realpath import",
+  oldTexts: [
+    'import{existsSync as A}from"node:fs";',
+  ],
+  newText:
+    'import{existsSync as A,realpathSync as Z}from"node:fs";',
+});
+
+patchFileFirstMatch(path.join(scriptsDir, "installManifest.mjs"), {
+  label: "Linux browser native host manifest canonical path resolution",
+  oldTexts: [
+    'let e=P.resolve(t).split(P.sep),o=e.lastIndexOf("cache");return o<1||e[o-1]!=="plugins"||e.length<=o+3?t:P.resolve(t,"..","latest")',
+  ],
+  newText:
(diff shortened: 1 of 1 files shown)


ilysenko (changes requested):
Thanks for digging into this. I found one blocking issue: the patch does not address the parent-chain failure it describes. The Linux host canonicalizes every runtime path before validating its ancestors, and the existing regression test explicitly accepts an executable reached through a symlink. The added `mode:493` applies only to newly created browser `NativeMessagingHosts` directories; it does not repair existing group-writable parents under `~/.codex/plugins` or `~/.codex/packages`.

Please reproduce the failure on latest `main` and the current DMG, identify the exact rejected path, fix its owning materialization/runtime path, and add a focused regression test that fails before the fix. Please also make the patch test exercise and assert all new current-DMG needles, since CI for this head predates DMG `26.715.72359`.

PR #1130, opened 2026-07-22
Make Linux close button exit instead of hiding to tray

grosserknallkopf:
## What

- Make the Linux close button close the last window instead of hiding it to the system tray.
- Keep minimize as the way to leave the app running with its tray icon and scheduled work available.
- Gate Linux tray startup on `codex-linux-system-tray-enabled`.
- Resolve `node:fs` and `node:path` inside the generated settings helpers so minified aliases from unrelated local scopes cannot break settings reads and writes.

## Why

The current main-bundle patch includes Linux in the Windows close-to-tray branch. Clicking X therefore calls `preventDefault()` and hides the window, while the tray-readiness predicate keeps the Electron process alive. A later launch can leave users with an inaccessible background process, and any update waiting for the app to exit remains blocked.

The Linux settings persistence patch also infers bundle-wide minified aliases. In the current DMG, that lookup can select `r` and `f` from unrelated function-local `require()` calls. The generated top-level settings helper then calls the wrong or undefined binding, so an explicit `codex-linux-system-tray-enabled: false` may not be readable at runtime.

## Behavior after this change

- **X:** closes the last Linux window and lets the normal Electron quit lifecycle run.
- **Minimize:** keeps the same app instance alive, including the tray and scheduled tasks.
- **Relaunch while minimized:** continues to use the existing required single-instance lock and second-instance handoff.
- **Update:** the existing updater quit path can finish without a hidden window process remaining behind.
- **Shutdown/logout:** the existing `before-quit`/`will-quit` cleanup destroys the registered tray and drains app state normally.

## Checks

- `node --test scripts/patch-linux-window-ui.test.js` — 391 passed
- Isolated `test_linux_tray_patch_smoke` — passed
- Applied the patch to the current `main-2laZZXle.js` DMG bundle shape and verified idempotence — passed
- `git diff --check` — passed
- Full `tests/scripts_smoke.sh` progressed through packaging, launcher, updater, and warm-start coverage, then stopped in the unrelated CLI-resolution fixture because this host resolves `/usr/bin/codex` before the fixture's Linuxbrew binary.


ilysenko:

ilysenko:
Thanks for the contribution and for documenting the behavior clearly.

We are going to close this PR because keeping the application running when its last window is closed is standard upstream behavior. On macOS, closing a window is intentionally different from quitting the application, and the current upstream bundle applies the corresponding close-to-tray behavior on Linux as well.

As stated in our contribution guidelines, this project preserves existing upstream and platform behavior unless there is a compelling Linux compatibility need to change it. At the moment, there is no linked issue, reproducible critical failure, or evidence of a widespread problem that would justify maintaining this divergence.

The review also identified the following technical blockers:

- The PR mixes the close-button behavior change with independently releasable changes to tray startup, Linux settings persistence, and module resolution. These concerns should not be combined in one PR.
- The new tray-setting startup condition is added to the `linux-tray` patch, which is classified as `required-upstream`. Drift in this optional behavior would therefore reject an otherwise usable build instead of producing a warning.
- The settings-persistence changes retain legacy already-patched bundle handling, older bundle shapes, and fallback tests. This conflicts with the repository's latest-`CODEX.DMG`-only policy, which requires obsolete compatibility paths to be removed when this patch logic is changed.

There is also a direct build-reliability cost. If a future DMG changes any of the new minified bundle patterns, the build and candidate acceptance may fail instead of merely reporting a warning. Package builds and updater rebuilds could then remain blocked until the patch is inspected, rewritten, and validated against the new DMG.

The upstream DMG changes very frequently, sometimes daily. Every additional patch against the minified main-process bundle creates recurring review, testing, and maintenance work. That permanent cost and failure surface are not justified without a demonstrated critical need.

This is a repository policy and product decision, not a request for a different implementation of the same behavior. We do not accept changes to standard upstream behavior based solely on user preference, even when the implementation is technically sound and well tested. Please do not reopen this PR or submit a replacement PR that makes the Linux close button exit the application. We would reconsider this behavior only after a maintainer-approved issue demonstrates a reproducible critical Linux compatibility problem that cannot reasonably be solved while preserving the upstream lifecycle.

If there is a concrete case where the tray is unavailable, the application becomes unreachable, or updates remain blocked, please open an issue with the distro, desktop environment, package format, and exact reproduction steps. We can then evaluate a narrower solution based on verified behavior.

Closing this for now. Thank you again for the work.

PR #1131, opened 2026-07-23
Fix updater workspace cleanup race

magicJie:
## Summary

- guard startup cache cleanup with the updater check lock
- reload persisted updater state after locking before pruning workspaces
- keep the existing check-specific lock message and add regression coverage for active and stale-state workspaces

## Problem

A second updater process could run startup cache cleanup while another process was rebuilding a DMG. The second process would later notice the held check lock, but by then it could already have removed the active workspace builder, app, and package directories. The running build then failed with missing update-builder files.

## Validation

- `env -u XDG_CONFIG_HOME cargo test -p codex-update-manager` (277 passed)
- built and installed a local `.deb`; the updater recovered the failed candidate to `idle` with no error

### DIFF
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -264,6 +264,25 @@ fn maybe_prune_caches(config: &RuntimeConfig, state: &PersistedState) {
     maybe_prune_generated_artifacts(config);
 }
 
+fn maybe_prune_caches_if_check_idle(
+    config: &RuntimeConfig,
+    state: &mut PersistedState,
+    paths: &RuntimePaths,
+) -> Result<()> {
+    let Some(_check_lock) = try_acquire_check_lock(paths)? else {
+        info!("skipping updater cache cleanup because another check is already active");
+        return Ok(());
+    };
+
+    // A long-lived daemon or a concurrently launched check-now process may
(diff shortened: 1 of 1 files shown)


ilysenko (changes requested):
Thanks for the focused race fix and regression coverage. I found several blockers:

- The cleanup lock is acquired after startup operations that can persist the entire stale `PersistedState`. This leaves an interleaving where a second process overwrites an active or retryable workspace state before the helper reloads it, and cleanup then removes that workspace. Please move the serialization/reload boundary before those writes, or merge the state safely, and add an entrypoint-level regression test covering the production order.
- Using `check.lock` for cleanup-only work allows a forced `check-now` to return successfully because it assumes the lock holder is performing an upstream check, while that holder may only prune and then exit through the `--if-stale` path. Please preserve forced-check semantics and add concurrency coverage.
- Please keep startup cleanup fail-soft: lock/open/reload failures should warn and skip cleanup rather than abort daemon startup.
- Per `CONTRIBUTING.md`, please bump the `codex-update-manager` patch version in `updater/Cargo.toml` and `Cargo.lock`, and document the fix under `Unreleased / Fixed`.

PR #1139, opened 2026-07-23
Fix upstream DMG drift

ilysenko:
## Summary
- repair Linux compatibility drift for the current upstream DMG
- remove obsolete compatibility paths replaced by the current DMG shape

## Validation
- exact current-DMG acceptance recorded by watchdog v2
- GitHub CI must pass, including Nix Package Builds


spencer-thompson (approved):
Generated bundle remains in PR history. The latest commit deletes the ~14 MB upstream bundle, but it still exists in earlier commit 9b90772. A merge-commit or rebase merge would preserve it in main history; rewrite the branch or require a squash merge.

Otherwise it looks good!

Don't have write access but reviewed with 5.6-sol max

PR #1141, opened 2026-07-24
Fix updater workspace cleanup race

magicJie:
## Summary

- serialize daemon startup maintenance, forced checks, periodic checks, reconciliation, state reloads, and cache cleanup under the updater check lock
- make forced `check-now` wait for startup maintenance while stale checks remain non-blocking
- keep daemon startup lock/open/state-reload failures fail-soft
- bump `codex-update-manager` to 0.10.3 and document the fix

## Problem

A second updater process could persist stale state or prune cache workspaces while another process was rebuilding a DMG. The active builder then lost its update-builder, app, or package files and failed late in `install.sh`.

This replaces #1131 and addresses all owner review blockers there.

## Validation

- `git diff --check`
- `env -u XDG_CONFIG_HOME cargo test -p codex-update-manager -- --test-threads=1` (281 passed)
- focused forced-check, daemon-startup, reconciliation, and fail-soft regression tests
- GitHub CI: Rust and Smoke Tests, Debian, RPM, Pacman, and Nix builds all passed
- previously built and installed a local `.deb`; the updater recovered the failed candidate to `idle` with no error

### DIFF
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -66,6 +66,11 @@ pub async fn run(cli: Cli) -> Result<()> {
     }
     let mut state =
         PersistedState::load_or_default(&paths.state_file, effective_auto_install(&config))?;
+    #[cfg(test)]
+    wait_for_process_test_barrier(
+        "CODEX_UPDATE_MANAGER_TEST_ENTRYPOINT_LOADED",
+        "CODEX_UPDATE_MANAGER_TEST_ENTRYPOINT_CONTINUE",
+    )?;
     if !matches!(
         &cli.command,
         Commands::Daemon | Commands::CheckNow { .. } | Commands::InstallReady
@@ -74,6 +79,8 @@ pub async fn run(cli: Cli) -> Result<()> {
         state.installed_version = install::installed_package_version();
         persist_if_changed(&paths, &state, &original_state)?;
     }
+    #[cfg(test)]
+    signal_process_test_marker("CODEX_UPDATE_MANAGER_TEST_ENTRYPOINT_PRE_DISPATCH")?;
 
     match cli.command {
         Commands::Daemon => run_daemon(&config, &mut state, &paths).await,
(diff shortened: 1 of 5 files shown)


magicJie:
Follow-up to the owner review on #1131:

- The serialization boundary now precedes state reloads and every full-state startup write; daemon startup and reconciliation have production-order stale-state regression tests.
- Forced `check-now` waits for the active startup/check flow, while `--if-stale` remains non-blocking; concurrency coverage verifies the forced check still reaches upstream.
- Startup lock/open/reload failures warn and skip maintenance, with dedicated fail-soft tests.
- The updater version is bumped to 0.10.3 in both manifests and the fix is documented under Unreleased / Fixed.

The branch is rebased onto current main and all CI checks are green.

Yo-DDV (commented):
Thanks for the follow-up. I found one remaining concurrency case in the shared updater state, with the concrete interleaving inline.

updater/src/app.rs
```diff
@@ -66,9 +66,11 @@ pub async fn run(cli: Cli) -> Result<()> {
     }
     let mut state =
         PersistedState::load_or_default(&paths.state_file, effective_auto_install(&config))?;
-    let original_state = state.clone();
-    state.installed_version = install::installed_package_version();
-    persist_if_changed(&paths, &state, &original_state)?;
+    if !matches!(&cli.command, Commands::Daemon | Commands::CheckNow { .. }) {
+        let original_state = state.clone();
+        state.installed_version = install::installed_package_version();
```
Yo-DDV, line 79: `InstallReady` is still outside the serialization boundary. The daemon reconciliation path can hold `check.lock` and load `ReadyToInstall`, while a concurrent `install-ready` command loads the same state without that lock. If both reach `trigger_install()` before either observes the other's `Installing` write, each starts its own `pkexec` for the same package.

Could you put the install entrypoint behind a reload-after-lock boundary, or use a dedicated install lock? An entrypoint-level regression test can race `reconcile_pending_install_from_disk()` against `run_install_ready()` and assert that only one install command is launched.

magicJie: Addressed in 47a1e98. `install-ready` now waits on the shared updater-flow lock, reloads state only after acquiring it, and skips the pre-dispatch full-state write that previously happened outside the lock. I also added an entrypoint-level regression test that starts daemon reconciliation in a controlled fake `pkexec`, races `run_install_ready()` against it, and asserts exactly one install command is launched. The install path now honors the existing notification setting, which keeps the concurrency test independent of a desktop D-Bus runtime. Targeted `install_ready`, `check_lock`, and `daemon_reconcile` tests pass, as do `cargo check` and Clippy.

ilysenko (approved):
Verified the current head after adding deterministic public-entrypoint, cross-process updater serialization regressions and bounded subprocess cleanup. All required checks are green.

PR #1142, opened 2026-07-24
Add opt-in Codex Micro Linux support

pierce403:
## Summary

- Add a default-off `codex-micro` Linux feature that reuses the Codex Micro
  service and controls already bundled with the upstream desktop app.
- Verify and stage the matching Linux x64 or arm64 `node-hid@3.3.0` binding.
- Install narrowly scoped USB and Bluetooth hidraw access rules for native
  packages, with corresponding Nix/NixOS support.
- Leave disabled builds and unrelated HID devices unchanged.

The upstream app contains the Work Louder integration but not the Linux native
binding expected by its nested `node-hid` loader. Linux also requires udev
permission for the device's vendor hidraw collection. This feature supplies
those Linux-specific pieces without emulating the device or reimplementing its
protocol.

Refs #631

## Validation

- `node --test linux-features/codex-micro/test.js scripts/lib/linux-features.test.js scripts/ci/upstream-dmg-acceptance.test.js`
- JavaScript syntax checks for the feature and feature-framework files
- Shell syntax checks for the installer, launcher template, shared helpers, and
  package builders
- `git diff --check`
- Debian, RPM, pacman, and AppImage smoke paths, including feature-disabled
  packages, permissions, dependencies, udev reload behavior, and symlink-escape
  checks
- Accepted side-by-side app build using the latest upstream desktop
  `26.721.30844` / Electron `42.3.0`; both Codex Micro patches applied and the
  acceptance verdict was `accepted_with_warnings` with no feature blockers
- Built and installed Debian package from desktop `26.715.72359`
- Physical USB and Bluetooth verification of input, Codex status, and lighting
  on desktop `26.715.72359`

Nix is unavailable on the validation host, so `nix flake check` was not run
locally. `bash tests/scripts_smoke.sh` reaches and passes every affected package
phase, then exits at the final unrelated launcher validation; the same exit
reproduces from a clean current `origin/main` archive.

## Checklist

### DIFF
--- a/linux-features/codex-micro/patch.js
+++ b/linux-features/codex-micro/patch.js
@@ -1,0 +1,128 @@
+"use strict";
+
+const childProcess = require("node:child_process");
+const path = require("node:path");
+
+const {
+  extractedAppPatch,
+  webviewAssetPatch,
+} = require("../../scripts/patches/descriptor.js");
+
+const CODEX_MICRO_GATE_ID = "3207467860";
+const CODEX_MICRO_GATE_MARKER = "codexLinuxCodexMicroGateOverride";
+const JS_IDENT = "[A-Za-z_$][\\w$]*";
+
+function escapeRegExp(value) {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+}
+
+function exportedFeatureGateHook(source) {
+  const exportStart = source.lastIndexOf("export{");
+  const exportEnd = exportStart < 0 ? -1 : source.indexOf("}", exportStart);
+  if (exportStart < 0 || exportEnd < 0) {
+    return null;
+  }
+
+  const exportBlock = source.slice(exportStart, exportEnd + 1);
+  const candidates = new RegExp(
+    `function (${JS_IDENT})\\((${JS_IDENT})\\)\\{return ` +
+      `(${JS_IDENT})\\(\\),(${JS_IDENT})\\((${JS_IDENT}),\\2\\)\\}`,
+    "g",
+  );
+  const exportedCandidates = [];
+  for (const match of source.matchAll(candidates)) {
+    const hookName = match[1];
+    const exportedAsGateHook = new RegExp(
+      `(?:\\{|,)${escapeRegExp(hookName)} as ${JS_IDENT}(?:,|\\})`,
+    );
+    if (exportedAsGateHook.test(exportBlock)) {
+      exportedCandidates.push({
+        source: match[0],
+        hookName,
+        argumentName: match[2],
+        contextHookName: match[3],
+        atomReadName: match[4],
+        gateAtomName: match[5],
+      });
+    }
+  }
+  return exportedCandidates.length === 1 ? exportedCandidates[0] : null;
+}
+
+function matchesCodexMicroFeatureGateContract(source) {
+  if (typeof source !== "string") {
+    return false;
+  }
+  if (source.includes(CODEX_MICRO_GATE_MARKER)) {
+    return true;
+  }
+  return source.includes("useFeatureGate hook failed to find a valid StatsigClient") &&
+    exportedFeatureGateHook(source) != null;
+}
+
+function applyCodexMicroFeatureGatePatch(source) {
+  if (typeof source !== "string" || source.includes(CODEX_MICRO_GATE_MARKER)) {
+    return source;
+  }
+
+  const hook = exportedFeatureGateHook(source);
+  if (hook == null) {
+    if (source.includes("useFeatureGate hook failed to find a valid StatsigClient")) {
+      console.warn(
+        "WARN: Could not find the current exported feature-gate hook - " +
+          "skipping Codex Micro gate override",
+      );
+    }
+    return source;
+  }
+
+  const replacement =
+    `function ${hook.hookName}(${hook.argumentName}){return ` +
+    `${hook.contextHookName}(),${hook.atomReadName}(${hook.gateAtomName},${hook.argumentName})||` +
+    `${hook.argumentName}===\`${CODEX_MICRO_GATE_ID}\`/*${CODEX_MICRO_GATE_MARKER}*/}`;
+  return source.replace(hook.source, replacement);
+}
+
+function stageNativeBinding(extractedDir) {
+  const helper = path.join(__dirname, "native-binding.js");
+  const output = childProcess.execFileSync(process.execPath, [helper, "--stage", extractedDir], {
+    encoding: "utf8",
+    env: process.env,
+    maxBuffer: 16 * 1024 * 1024,
+  });
+  return JSON.parse(output);
+}
+
+module.exports = {
+  CODEX_MICRO_GATE_ID,
+  CODEX_MICRO_GATE_MARKER,
+  applyCodexMicroFeatureGatePatch,
+  exportedFeatureGateHook,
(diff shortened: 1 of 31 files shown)


ilysenko (changes requested):
Thanks for the substantial Codex Micro work. Before this can merge, please address these blockers:

1. Restore the Debian GTK fallback to `libgtk-3-0t64 | libgtk-3-0`; the current `libgtk-3.0-0` typo breaks every non-t64 `.deb`, even when the feature is disabled.
2. Make the USB hidraw udev rule self-contained. The rule matches `ID_VENDOR_ID`, `ID_MODEL_ID`, and `ID_USB_INTERFACE_NUM`, but those properties are not guaranteed for hidraw unless `usb_id` is imported. Please validate the resulting rule with `udevadm test`.
3. Split the reusable `packageResources`/`packageDependencies` and native-package lifecycle framework into a prerequisite PR, then rebase the Codex Micro feature on it. The current PR mixes a broad shared packaging extension with an optional hardware feature.
4. Validate the enabled-feature paths with real Debian, RPM, and pacman package builds, and provide or document the `libudev.so.1`/`libusb-1.0.so.0` runtime requirements for AppImage and source/user-local installs.
5. Remove the currently unreachable source-build fallback and its toolchain/lockfile, or move it to a separate change with a concrete current-DMG need. Both supported architectures already use pinned prebuilds.

scripts/lib/package-common.sh
```diff
@@ -165,6 +165,152 @@ run_linux_feature_package_hooks() {
...
+    local package_format="$2"
+    local helper="$REPO_DIR/scripts/lib/linux-features.js"
+    local node_bin
+
+    [ -d "$staging_root" ] || error "Missing package staging root: $staging_root"
+    [ -f "$helper" ] || error "Missing Linux features helper: $helper"
+    node_bin="$(package_node_binary)"
+    "$node_bin" "$helper" --stage-package-resources "$package_format" "$staging_root"
+}
+
```
Yo-DDV, line 179: **The package feature set can diverge from the app that was actually built.** This command reads the live feature config at packaging time. The generated app already records its build-time set in `.codex-linux/build-info.json` (`linuxFeatures.enabled`), but `ensure_app_layout` never compares them. The new Debian smoke path demonstrates the mismatch by packaging `{"enabled":["codex-micro"]}` against the generic app fixture, which contains neither the feature-gate patch nor the HID binding, and still treating the package as valid. A package-only run after editing `features.json` can therefore ship udev/dependencies without the feature, or the feature without its package resources. Please make the prerequisite packaging framework consume or validate the staged app feature snapshot and add tests for both mismatch directions.

linux-features/codex-micro/patch.js
```diff
@@ -0,0 +1,128 @@
...
+function matchesCodexMicroFeatureGateContract(source) {
+  if (typeof source !== "string") {
+    return false;
+  }
+  if (source.includes(CODEX_MICRO_GATE_MARKER)) {
+    return true;
+  }
+  return source.includes("useFeatureGate hook failed to find a valid StatsigClient") &&
+    exportedFeatureGateHook(source) != null;
+}
```
Yo-DDV, line 61: **This contract can pass after the Codex Micro gate has drifted.** `assetMatch` checks only the generic Statsig warning and the presence of one exported gate hook; it does not verify gate `3207467860` or a Codex Micro-specific callsite/marker. A bundle containing the generic hook but no Micro reference still matches, gets modified, and is reported as `applied`, so enabled-feature acceptance will not reject that class of upstream drift. Please add a Micro-specific current-bundle contract (same asset or a separate probe) and a negative test where the generic hook remains but the Micro gate/callsite is absent.

Yo-DDV (commented):
Two additional correctness gaps remain on `63f91d7de8`; details inline. They are independent of the five blockers already listed and should be carried into the prerequisite framework/rebased feature work.

ilysenko (changes requested):
Thanks for taking this on. The Codex Micro integration is useful, but the current PR combines two independently reviewable changes: a shared native-package feature framework and the device-specific implementation. Please split the work into two sequential PRs.

**Stage 1 — generic package framework**

Create a prerequisite PR containing only the reusable packaging contract:

- Add the minimal generic support for feature-owned package resources and runtime dependencies.
- Keep all Codex Micro, node-hid, udev, and feature-specific Nix logic out of this PR.
- Do not add udev-specific helpers to shared package code. If lifecycle integration needs another extension point, make it a small generic hook that other features can reuse.
- Preserve the existing behavior when no feature is enabled, including restoring the Debian dependency to `libgtk-3-0t64 | libgtk-3-0`.
- Test both the empty plan and a generic sample feature with real `.deb`, RPM, and pacman builders, not only rendered templates or stubbed packagers.

This PR can be reviewed and merged independently as a shared framework change.

**Stage 2 — Codex Micro feature**

Rebase the Codex Micro work onto the merged framework PR and keep this stage focused on `linux-features/codex-micro/` plus the smallest necessary Nix integration:

- Reuse the upstream Codex Micro service and stage only the verified `node-hid@3.3.0` x64/arm64 prebuilds required by the current DMG.
- Remove the speculative source-build fallback, toolchain, lockfile, and provenance code. It is unreachable with the current supported architectures and can be introduced separately if a future current-DMG release actually requires it.
- Make the USB hidraw rule self-contained by importing or otherwise reliably matching the required USB properties. Validate it with `udevadm test`, not only string assertions.
- Declare the runtime dependencies for native packages and document the required `libudev`/`libusb` packages for AppImage, source, and user-local installs.
- Build real feature-enabled `.deb`, RPM, and pacman artifacts and keep the Nix watchdog build enabled.

If keeping core changes near zero is preferred, Stage 2 can initially bundle the udev rule as a feature resource and document its manual installation for every format. Automatic system-wide udev installation can then be added in a later focused PR.

Before pushing the split updates, please verify both PRs against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant tests locally, review each complete diff independently with a local LLM, and wait for every required CI check to pass before requesting maintainer review or merge.

ilysenko:
One more bit of context: this is a useful PR, and the request to split and isolate the work is not about making you jump through unnecessary hoops or being overly critical of your implementation.

The upstream `CODEX.DMG` changes very frequently—sometimes almost every day. Every additional patch, package hook, updater dependency, and core touchpoint becomes another contract that must be checked and repaired after future upstream releases. When device-specific behavior reaches into core packaging or update paths, a change intended for a small opt-in feature can break normal installs or updater rebuilds even when that feature is disabled. Over time, compatibility workarounds and interactions accumulate and become increasingly difficult to reason about.

Core and updater paths are therefore the most expensive places to add integration-specific behavior. The goal is to keep the generic extension point as small and independently reviewable as possible, while keeping Codex Micro behavior inside its own disabled-by-default feature. If upstream drift breaks the integration later, the feature can then be rejected or disabled without destabilizing launch, installation, packaging, or updates for everyone else.

Please do not read the split request as a rejection of the idea or as criticism directed at you. The feature is useful. The stricter structure is required because the repository will have to maintain it across every future DMG release, long after this PR is merged.

pierce403:
Awesome, thanks for the detailed review, I'll start integrating the feedback now.

niogyn:
I was going to hop in and help - I'm very much looking forward to this.  @pierce403 & @ilysenko

ilysenko:

PR #1144, opened 2026-07-24
Fix Browser Use node_repl hook compatibility

maxwellfortney:
## Summary

- add a Linux Browser Use compatibility shim for node_repl builds without `addAfterSubmittedCodeHook`
- keep the shim alongside the existing nodeRepl config shim so frozen nodeRepl objects can still be patched via the prototype
- extend Browser and Chrome staging smoke coverage for the hook fallback

## Validation

- focused `test_browser_plugin_renamed_upstream_staging` and `test_chrome_plugin_staging`
- `node --test scripts/patches/impl/main-process/browser.test.js scripts/lib/browser-client-node-repl-runtime.test.js`

### DIFF
--- a/scripts/lib/bundled-plugins.sh
+++ b/scripts/lib/bundled-plugins.sh
@@ -1321,7 +1321,9 @@ value = match.group("value")
 shim = r'''
 function codexLinuxBrowserUseConfigShim() {
   let repl = globalThis.nodeRepl;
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head: the Browser Use node_repl compatibility shim is focused, safe on the frozen Linux runtime object, and all required checks pass.

PR #1148, opened 2026-07-24
Fix enabled feature drift on the current DMG

PinguuSS:
## Summary

- retarget the Dock icon main/settings contracts and the Suggested Prompts app-page asset to the current desktop bundles
- select the Record & Replay dictation asset by its current semantic contract and preserve the new upstream Skysight tray guard outside Linux
- reject partial Chronicle callback drift byte-identically before any bridge mutation

## Why #1113 did not cover this

#1113 made UI Tweaks renderer selection resilient to hash churn inside stable filename families and semantic contracts. DMG 26.721.31836 changed the bundle topology to `app-initial-*` and changed the Dock and Chronicle source contracts, so the existing descriptors correctly failed closed instead of guessing at new shapes.

## Current DMG

- app version: `26.721.31836`
- Electron: `42.3.0`
- size: `593891632` bytes
- SHA-256: `ff6e8ac9985aec44caa305787552e4ea517a7c745aef283bd4cbcab992de64b7`

## Validation

- `node linux-features/record-and-replay/test.js` (32/32)
- `node linux-features/ui-tweaks/dock-icon.test.js` (24/24)
- `node linux-features/ui-tweaks/test.js` (54/54)
- `node scripts/patch-linux-window-ui.test.js` (386/386)
- real extracted DMG: every descriptor for all ten enabled features applied on the first pass; all 17 touched UI Tweaks and Record & Replay descriptors reported already-applied on the second pass
- isolated candidate acceptance: `accepted_with_warnings`, with no feature blockers and only the two existing optional Computer Use core warnings
- Plasma/Wayland side-by-side runtime: home and Settings loaded without the error fallback; Suggested Prompts and both Dock choices rendered
- Dock selection updated the KDE StatusNotifier immediately (`956x956` Codex to `2048x2048` ChatGPT) and remained selected after a cold restart

## Maintenance

I will continue monitoring and maintaining the Dock icon and Suggested Prompts opt-ins as upstream desktop contracts change.

### DIFF
--- a/docs/record-and-replay-linux.md
+++ b/docs/record-and-replay-linux.md
@@ -50,8 +50,8 @@ The workflow is:
 
 The important compatibility point is the contract: the bundled plugin launches
 an `event-stream` MCP server and the durable output is a Codex skill. The
-official macOS bundle supplies that server through
-`SkyComputerUseClient event-stream mcp`; the Linux feature supplies
+current macOS bundle supplies that server through
+`computer-use-client-launcher event-stream mcp`; the Linux feature supplies
 `SkyLinuxComputerUseClient event-stream mcp`, backed by the Rust
 `codex-record-replay-linux` backend.
 
--- a/linux-features/record-and-replay/patch.js
+++ b/linux-features/record-and-replay/patch.js
@@ -22,12 +22,18 @@ function hasRecordReplayPluginGate(source) {
   const pluginGateArray = findBundledPluginGateArray(source);
   const target = pluginGateArray?.text ?? source;
   return new RegExp(
-    String.raw`\{(?:[^{}]*,)?installWhenMissing:!0,name:${pluginNameExpressionRegex(RECORD_REPLAY_PLUGIN_NAME)},(?:isEnabled|isAvailable):`,
+    String.raw`\{(?:[^{}]*,)?installWhenMissing:!0,name:${pluginNameExpressionRegex(RECORD_REPLAY_PLUGIN_NAME)},isAvailable:`,
   ).test(target);
 }
 
-function buildRecordReplayDescriptor(availabilityProp) {
-  return `{installWhenMissing:!0,name:\`${RECORD_REPLAY_PLUGIN_NAME}\`,${availabilityProp}:({platform:e})=>e===\`linux\`}`;
+function hasObsoleteRecordReplayPluginGate(source) {
+  return new RegExp(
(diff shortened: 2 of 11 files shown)


PinguuSS:
Supplemental current-DMG composition/runtime validation:

- All enabled feature suites passed together: 321/321 tests across appshots, wrapper updater, frameless titlebar, open-target discovery, persistent status panel, Record & Replay, remote control UI/mobile, UI Tweaks, and X11 Computer Use.
- The accepted side-by-side candidate was copied under a trusted 0755 path before plugin validation, matching the bundled-plugin path security contract.
- Candidate Browser Use initialized its Linux in-app-browser backend and per-user socket successfully; the installed in-app browser also completed an example.com navigation smoke.
- The candidate bundled Computer Use binary passed doctor/window discovery on Plasma Wayland. Window targeting, portal click, and literal text input were exercised against the candidate. Host-level key-combo injection remains unavailable because Ubuntu currently provides ydotool 0.1.8; the existing compatibility guard correctly rejects that unsupported CLI.
- Record & Replay completed a start/stop capture against the candidate with accessibility, screenshot, and window metadata enabled.
- Remote Control settings/status RPCs and desktop-owner cold-start wiring loaded correctly. Enabling a second remote-control host in the candidate was correctly rejected while the installed daily driver owned the singleton host, so exclusive candidate remote-host E2E was intentionally not attempted.

These checks supplement the patch-report/idempotency evidence. The ydotool host dependency and Remote Control singleton topology are unrelated to the UI Tweaks and Record & Replay retargets in this PR.

ilysenko (changes requested):
Thanks for retargeting these features and validating the current DMG. Before merge, please complete the repository’s latest-DMG-only cleanup:

- Remove `upgradeRecordReplayBridgeSource`, the old `linux-record-replay-doctor` bundle-upgrade branch, and the corresponding “old patched bundles” test. Keep only strict idempotency for a complete current patched bundle.
- Narrow the bundled plugin availability contract from `isEnabled|isAvailable` to the current DMG’s `isAvailable` shape, with coverage that rejects the obsolete form.

These legacy fallback contracts should not remain in the same drift-sensitive feature being updated.

ilysenko (approved):
Reviewed the current head after the requested latest-DMG cleanup. The obsolete Record & Replay contracts are removed or rejected, current-DMG staging and diagnostics are aligned, and all required checks pass.

PR #1153, opened 2026-07-25
fix(remote-mobile-control): preserve connections and recovery settings

GuillaumeCisco:
## Summary

- preserve saved per-host Remote Control auto-connect choices across Desktop restarts
- auto-connect only the Remote Control environment owned by the local Desktop
- keep the Remote Control settings reachable on Linux when client authorization must be renewed
- update the feature contract, changelog, and regression coverage

## Root cause

The Linux Remote Control enablement bridge iterated over every discovered
Remote Control environment at startup. It enabled the environment matching the
local Desktop installation and explicitly wrote `autoConnect: false` for every
other environment, overwriting intentional choices such as keeping another
enrolled workstation connected.

Separately, the Linux visibility override still applied the upstream
`accessRequired` gate. If the controller enrollment became unavailable or
needed renewed authorization, the settings entry that starts the authorization
flow disappeared at the same time.

The bridge now writes only the local Desktop environment, leaving every other
saved choice untouched. On Linux, the settings entry remains visible so the
user can complete authorization recovery in-app. Non-Linux visibility behavior
is unchanged.

## User impact

Remote hosts that the user chose to auto-connect remain enabled after restarting
Codex Desktop. If Remote Control authorization expires or is revoked, the user
can still reach the settings and authorize it again instead of losing the
Remote Control UI entirely.

## Validation

- `node --test linux-features/remote-mobile-control/test.js` — 107 passed
- `node --test scripts/patch-linux-window-ui.test.js` — 386 passed
- build against Codex Desktop `26.721.41059` with `remote-mobile-control`
  enabled — accepted with no blockers or warnings
- generated webview asset contains the Linux authorization-recovery visibility
  guard and the patch report records it as applied


ilysenko (changes requested):
Thanks for the focused fix. The auto-connect change does not currently reach the latest upstream DMG. `applyLinuxRemoteControlEnablementBridgePatch` still targets the obsolete `[remote-connections/slingshot-gate-bridge]` shape, while 26.721.41059 uses `[remote-connections/gate-bridge]` with a new two-gate function. The patch silently returns unchanged, the report misclassifies it as already applied, and the generated asset contains neither the bridge marker nor the new auto-connect logic. Please retarget this descriptor and its behavioral fixture to the current DMG shape, remove the obsolete shape rather than keeping a fallback, and warn when the current anchors are missing.

GuillaumeCisco:
Closed because this change is intended to be reviewed only within the contributor fork, not submitted upstream.

PR #1156, opened 2026-07-25
fix(computer-use): support Arch ydotool raw CLI

kortylokai-web:
## Summary

- Recognize the raw-event CLI shape shipped by Arch Linux's `ydotool` 1.0.4 package.
- Verify required wheel and stdin-typing semantics through an isolated datagram socket before enabling the backend, rejecting incomplete `ydotool` 1.0.0 and 1.0.1 releases.
- Preserve capability probing when `XDG_RUNTIME_DIR` is absent by using a validated temporary base and a private, randomly named probe directory.
- Document `ydotool` 1.0.2 as the minimum compatible raw-event release.

The previous command-list fingerprint required a top-level `stdin` command that Arch's current package does not expose, even though the backend-compatible `type --file -` behavior is available. The semantic probes are effect-free: wheel deltas are zero, typing receives empty stdin, and both commands target a private dummy socket rather than the live `ydotoold` socket.

## Validation

- `TMPDIR=/tmp cargo test -p codex-computer-use-linux` (`376` passed)
- `tests/scripts_smoke.sh`
- Live `doctor` probe on Arch Linux with `ydotool 1.0.4-2`: compatible raw-event CLI detected and no readiness blockers
- Final diff reviewed by coding agents, including the no-XDG follow-up's filesystem and Unix-socket handling

The probe uses `XDG_RUNTIME_DIR` when it is safe and otherwise falls back to a validated temporary base. It still fails closed when it cannot allocate a safe isolated socket location or when `/proc/self/fd` is unavailable, leaving the portal fallback available instead of enabling an unverified raw-input backend.

## Checklist

- [x] This is not an upstream-drift fix; it changes only the Linux Computer Use helper and its dependency documentation.
- [x] I added or updated relevant tests, ran the validation listed above, and confirmed that required local checks pass.

### DIFF
--- a/computer-use-linux/src/ydotool.rs
+++ b/computer-use-linux/src/ydotool.rs
@@ -1,10 +1,15 @@
 use std::{
-    env, fs,
-    os::unix::net::UnixDatagram,
+    env,
+    ffi::OsStr,
+    fs, io,
+    os::unix::{
+        ffi::OsStrExt,
+        fs::{DirBuilderExt, MetadataExt, PermissionsExt},
+        net::UnixDatagram,
+    },
     path::{Path, PathBuf},
     process::{self, Command, Stdio},
     sync::OnceLock,
-    time::{SystemTime, UNIX_EPOCH},
 };
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


ilysenko (changes requested):
Thanks for adding Arch ydotool 1.0.4 support and semantic probing. One blocker remains: `ProbeSocket::bind()` now requires `XDG_RUNTIME_DIR`, while the existing runtime intentionally supports explicit `YDOTOOL_SOCKET`, `/run/user/<uid>`, and `/tmp/.ydotool_socket` fallbacks when that variable is absent. This makes `ensure_supported()` reject and cache an error for previously working ydotool setups. Please preserve the no-XDG path with a safely isolated probe socket and add regression coverage for a supported raw CLI with `XDG_RUNTIME_DIR` unset.

computer-use-linux/src/ydotool.rs
```diff
@@ -1,12 +1,51 @@
...
+const UNSUPPORTED_MESSAGE: &str = "unsupported ydotool CLI; Computer Use requires ydotool 1.0.2 or newer with raw key events, wheel movement, stdin typing, and absolute mouse movement";
+
+struct ProbeSocket {
+    _socket: UnixDatagram,
+    path: PathBuf,
+}
+
+impl ProbeSocket {
+    fn bind() -> Result<Self, String> {
+        let runtime_dir = env::var_os("XDG_RUNTIME_DIR")
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`ensure_supported()` is called before every ydotool invocation and by backend availability, but this new probe now fails before testing CLI semantics whenever `XDG_RUNTIME_DIR` is unset. The runtime socket selection still intentionally supports explicit `YDOTOOL_SOCKET`, `/run/user/<uid>/.ydotool_socket`, and `/tmp/.ydotool_socket` without requiring that env var; on those valid setups `ProbeSocket::bind()` returns an error and the `OnceLock` caches ydotool as unsupported even though the later runtime path could connect. Please choose an isolated probe socket using the same no-XDG fallback behavior (or another safe temp dir) rather than making `XDG_RUNTIME_DIR` mandatory.

avifenesh (commented):

---

I found one regression in the new ydotool capability probe.

Yo-DDV:
@kortylokai-web I picked up the remaining no-XDG blocker from the maintainer review and prepared one focused follow-up commit on top of the current PR head (`4b0d5f6`):

[`4842dc7`](https://github.com/Yo-DDV/codex-desktop-linux/commit/4842dc7cb6d211e35d373142e240444937257b36)

The live `YDOTOOL_SOCKET` resolution is unchanged. Only the capability probe is adjusted:

- it uses `XDG_RUNTIME_DIR` when that directory is safe;
- otherwise it falls back to a validated temporary base;
- it creates a randomly named private `0700` directory and binds the dummy Unix datagram socket inside it;
- it rejects overlong Unix socket paths and cleans up both the socket and its directory;
- it includes an end-to-end regression test for a supported raw CLI with `XDG_RUNTIME_DIR` absent.

The complete CI matrix passed on the follow-up commit: formatting, Clippy with warnings denied, workspace checks and tests, script smoke tests, and Debian, RPM, Pacman, and Nix package builds.

You can apply it directly with:

```bash
git cherry-pick 4842dc7cb6d211e35d373142e240444937257b36
```

After the cherry-pick, the final paragraph of the PR description should also be adjusted: an absent `XDG_RUNTIME_DIR` is now supported. The probe still fails closed when it cannot allocate any safe isolated socket location or when `/proc/self/fd` is unavailable.

kortylokai-web:
@Yo-DDV Thank you for preparing the focused follow-up. I applied `4842dc7` directly so the original authorship and commit remain intact.

The branch now preserves no-`XDG_RUNTIME_DIR` probing through a validated temporary base and a private randomly named `0700` directory, including the end-to-end supported-raw-CLI regression test. I also updated the PR description to remove the obsolete no-XDG limitation.

Local verification on the resulting PR head:

- `TMPDIR=/tmp cargo test -p codex-computer-use-linux` (`376` passed)

The repository CI matrix is now running against `4842dc7`. @ilysenko this addresses the requested no-XDG blocker and regression coverage.

ilysenko (approved):
Reviewed the current head. The previous no-XDG ydotool probe blocker is fixed with a safe temporary-directory fallback and regression coverage, and all required checks are green.

PR #1147, opened 2026-07-24
test(launcher): add resident window-reopen behavior harness

TanguyDeTaxis:
## Summary

Add an isolated launcher harness for the healthy resident reopen handoff contract.

The harness builds a controlled fake Desktop runtime with its own Electron identity, packaged webview, runtime markers, Unix handoff socket, and observable process state. It verifies that reopening a healthy resident:

- forwards `--new-chat` through the launcher IPC handoff;
- preserves the resident PID and keeps exactly one controlled Electron main process;
- retains the runtime markers;
- leaves a same-app-id decoy process untouched; and
- emits no user-visible error.

Mutation mode now has a dedicated, machine-checkable result: the injected resident-replacement regression must exit with status `86` and emit `resident-replacement-detected`. Setup failures, timeouts, probe failures, and unrelated non-zero exits remain ordinary test failures. A paired control runs the same mutation path without installing the source mutation and must preserve the resident without emitting exit `86` or the mutation sentinel.

The harness also checks pidfd cleanup capability before creating its workspace or starting any fixture process. Unsupported hosts exit with the explicit status `77`; unexpected probe failures remain hard failures. Cleanup revalidates process identity before signalling and fails the test if bounded TERM/KILL cleanup cannot complete.

The previous synthetic focus state has been removed. This test covers launcher IPC and resident preservation only; it does not claim to exercise the production window-focus receiver.

This is a test-only change and does not modify production launcher behavior.

## Validation

- `bash -n tests/launcher_window_reopen_behavior.sh tests/scripts_smoke.sh`
- `git diff --check`
- Nominal harness: resident PID preserved, one controlled main process, `--new-chat` acknowledged, markers retained, decoy untouched, no user-visible error
- Mutation harness: dedicated exit `86` and `resident-replacement-detected` outcome
- No-mutation control: same mutation path preserves the resident and cannot emit exit `86` or the mutation sentinel
- No-pidfd path: dedicated exit `77` before workspace creation or process spawn
- Probe setup failure: hard failure, not a mutation result or unsupported-host skip
- Full repository CI passed on `0f1648a66786fe0fa8e0fb24aca14ae600765a9b`: Rust formatting, Clippy, checks and tests, script smoke tests, Debian, RPM, Pacman, and Nix validation

The existing ephemeral-port selection can still fail safely if another process claims the selected port before the fixture binds it; that produces a test failure rather than a false pass.

## Checklist

- [x] This pull request is ready for maintainer review.
[...]

PR #1160, opened 2026-07-26
fix(lifecycle): prevent windowless warm-start survivors

kortylokai-web:
## Summary

Fix Linux explicit Quit leaving an Electron process alive after its window and main-process services have already been disposed.

The current upstream `will-quit` handler prevents the initial quit, drains lifecycle work, disposes the application context, and calls `app.quit()` again. When cleanup settles during Electron's event-emission microtask checkpoint, that second call can occur before the original quit has finished unwinding. Electron ignores the re-entrant quit, leaving a windowless process holding the single-instance lock.

This change:

- contains synchronous cleanup failures and bounds the Linux drain to three seconds;
- preserves the existing bounded application-context cleanup;
- logs rejected cleanup work without allowing it to suppress termination;
- finishes the Linux explicit-quit path with `app.exit(0)`, after cleanup, instead of re-entering `app.quit()`;
- leaves the upstream non-Linux path and launcher handoff behavior unchanged;
- adds semantic matching, drift, idempotence, failure-path, and platform regression coverage.

This is complementary to #1147. Healthy resident instances must still retain their PID and accept reopen handoffs. This PR changes only explicit Quit after teardown has begun; the #1147 resident-reopen harness continues to pass.

Source-of-truth files:

- `scripts/patches/impl/main-process/quit-lifecycle.js`
- `tests/scripts_smoke.sh`
- `docs/windowless-warm-start-fix-report.md`

No existing issue is linked. The failure was reproduced while investigating packaged Linux warm-start behavior.

## Validation

Automated:

- `node --test scripts/patch-linux-window-ui.test.js` — 398/398 passed
- `bash tests/scripts_smoke.sh` — passed, including the #1147 healthy-resident reopen harness
- `bash -n install.sh launcher/start.sh.template scripts/lib/*.sh scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh scripts/build-appimage.sh`

Manual packaged-app validation on Arch Linux with Hyprland/Wayland:

- An equivalent build retaining the final `app.quit()` reproduced the surviving windowless primary process.
- The `app.exit(0)` build completed ten consecutive Quit/relaunch cycles.
- Every cycle reached zero primary processes, zero packaged helper processes, no launch-action socket, and no compositor window before relaunch.
- An eleventh launch opened a healthy window after the tenth exit.

Manual reproduction has not been repeated on other distributions, desktop environments, or X11. The affected lifecycle patch is shared by Linux package formats, but this PR does not claim universal reproduction across Linux environments.

The intentional tradeoff is that cleanup still running after the bounded deadline cannot delay explicit termination indefinitely.

## Checklist

### DIFF
--- a/docs/windowless-warm-start-fix-report.md
+++ b/docs/windowless-warm-start-fix-report.md
@@ -1,0 +1,127 @@
+# Windowless Warm-Start Fix
+
+## Status
+
+The Linux explicit-quit path now guarantees bounded cleanup followed by process
+exit. A later desktop launch no longer attaches to a surviving Electron process
+that has no window or launch-action socket.
+
+The change is confined to the existing main-process lifecycle patch and its
+regression coverage. Launcher handoff behavior is unchanged.
+
+## Failure
+
+Choosing **Quit** from the application menu could remove the last window and
+partially dispose the main-process application context without terminating the
+Electron process. A later launch then lost the Electron single-instance lock to
+that process. The surviving process could not present a window because the
+services needed by its launch handler had already been disposed.
+
+The warm-start handoff exposed the failure but did not cause it. Reopening a
+window or extending the launch-action socket lifetime after teardown begins
+would route work into a partially destroyed application context.
+
+## Root Cause
+
+The current upstream main bundle has one targeted lifecycle `will-quit` handler
+with two cleanup branches:
+
+- a reduced branch stops Codex Micro and flushes tracing;
+- a full branch also flushes global state and settings.
+
+Both branches call `preventDefault()`, run lifecycle disposers, wait with
+`Promise.allSettled()`, dispose the application context and shared disposable
+collection, and then call `app.quit()` again.
+
+That sequence was not total:
+
+- a synchronous lifecycle-disposer or drain-setup exception could occur before
+  the promise continuation was installed;
+- a stalled drain had no deadline;
+- a context or shared-disposable failure could skip the terminal action;
+- rejected `Promise.allSettled()` members were not observable;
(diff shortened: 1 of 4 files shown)


Yo-DDV (approved):
Reviewed the current head against the lifecycle contract, Electron 42.3.0 shutdown path, #1147 handoff coverage, and the current-DMG artifact. I did not find a blocking issue.

The patch now requires one semantic `will-quit` owner and validates that the reduced and full branches share the same event, lifecycle managers, drain work, and finalizer. Missing, malformed, or duplicate owners fail the required descriptor instead of drifting silently. On Linux, synchronous setup failures, rejected work, deadline expiry, context cleanup failure, and shared-disposable failure all converge on one terminal `app.exit(0)`; non-Linux remains on the upstream `app.quit()` path. I also exercised double handler delivery and a drain that rejected after losing the timeout race; neither produced a second terminal action or an unhandled rejection.

Electron's `Browser::Exit()` sets the exit code, destroys remaining windows when necessary, and still reaches `Shutdown()`/`OnQuit`, while intentionally avoiding another `before-quit`/`will-quit` pass. That matches the reproduced re-entrant-quit failure. The packaged A/B result and green current-DMG acceptance provide the critical Linux compatibility evidence that was missing from the close-button behavior proposal in #1130.

No new untrusted input, privilege, or persistence boundary is introduced. The bounded deadline can abandon cleanup that exceeds the limit, but the PR states that trade-off explicitly, and terminating a committed quit is safer than retaining a partially disposed primary process and its single-instance lock. Approving.

PR #1161, opened 2026-07-26
Add per-connection SSH command wrappers

yanis-falaki:
## Summary

Adds an opt-in `ssh-command-wrapper` Linux feature that allows each saved SSH connection to define an argv-based remote command wrapper in **Settings -> Connections -> SSH**.

I often work with HPC clusters that route incoming SSH sessions to randomly selected login nodes without allowing a specific node to be chosen before connecting. This can cause multiple Codex app-server processes to run concurrently on different login nodes. Because these nodes share a filesystem, the redundant processes are wasteful and can race or conflict with one another.

This feature lets me route all Codex remote operations through a consistent login node so only one app-server environment is used. More generally, it allows any saved SSH connection to route Codex operations through a user-specified argv command.

Codex appends its generated remote command as the wrapper’s final argument. For example:

```text
ssh -T target-host --
```

The feature:

- Supports saved SSH aliases and manually configured hosts.
- Applies to remote probes, installation commands, and the app-server proxy.
- Persists the wrapper as an argv array rather than a shell snippet.
- Supports POSIX-style quoting and escaping while rejecting unquoted shell operators, newlines, malformed quoting, and oversized input.
- Includes the wrapper in the SSH startup-gate identity.
- Preserves upstream behavior when no wrapper is configured.
- Remains disabled by default.

No issue is currently linked.

## Validation

Passed:

- `node --test linux-features/ssh-command-wrapper/test.js`
  - 9/9 focused tests passed.
  - 387/387 tests passed.
- `git diff --check`
- Latest upstream DMG rebuild and acceptance:
  - App version: `26.721.41059`
  - Both feature descriptors applied.
  - No acceptance warnings or blockers.
- Verified the final patch against pristine JavaScript extracted from the latest upstream DMG:
  - Main-process and webview bundles have valid syntax.
  - The generic `ssh -T target-host --` placeholder is present.
  - Both patches are idempotent.
  - The feature composes with the `remote-mobile-control` patches.
- Manual end-to-end test:
  - Configured an SSH connection with `ssh -T login1 --`.
  - Connected through Codex and ran `hostname` from the remote task.
  - Confirmed the app-server was running on `login1`.

Broader local validation encountered two reproducible environment-dependent failures unrelated to this change:

  - 816/817 tests passed.
  - The `shared-app-server-socket` integration test failed because the locally installed Codex app-server exited before creating its test socket.
  - Reached launcher CLI-resolution validation, then failed because this environment resolved `/usr/bin/codex` instead of the test’s expected Linuxbrew installation.

Known risk: this optional feature patches the current upstream minified SSH transport and settings bundles. Future upstream bundle drift may cause the fail-soft descriptors to skip until their current-DMG needles are updated.

## Checklist

- [x] If this fixes upstream drift, it targets only the latest `CODEX.DMG` and removes obsolete fallback code and tests from the affected area. *(Not applicable: this is a new opt-in feature, validated against only the latest DMG.)*
- [x] I added or updated relevant tests, ran the validation listed above, and confirmed that required CI checks pass. *(Tests were added and run locally; awaiting GitHub CI.)*

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,6 +232,7 @@ workarounds.
 | Read Aloud MCP | Opt-in | `read-aloud-mcp` | [Docs](linux-features/read-aloud-mcp/README.md) |
 | Remote Control UI gates | Opt-in | `remote-control-ui` | [Docs](linux-features/remote-control-ui/README.md) |
 | Experimental Remote Mobile Control | Opt-in | `remote-mobile-control` | [Docs](linux-features/remote-mobile-control/README.md) |
+| SSH command wrapper | Opt-in | `ssh-command-wrapper` | [Docs](linux-features/ssh-command-wrapper/README.md) |
 | Thorium Chrome Plugin Support | Opt-in | `thorium-chrome-plugin` | [Docs](linux-features/thorium-chrome-plugin/README.md) |
 | UI tweaks | Opt-in | `ui-tweaks` | [Docs](linux-features/ui-tweaks/README.md) |
 | X11/EWMH Computer Use adapter | Opt-in | `x11-ewmh-computer-use` | [Docs](linux-features/x11-ewmh-computer-use/README.md) |
--- a/linux-features/ssh-command-wrapper/README.md
+++ b/linux-features/ssh-command-wrapper/README.md
@@ -1,0 +1,59 @@
+# SSH Command Wrapper
+
+This opt-in feature adds a **Remote command wrapper** field to each saved SSH
+connection in **Settings → Connections → SSH**. Codex appends its generated
+remote command as the wrapper's final argument and runs the result on the
+configured SSH host.
+
+For a connection that should route Codex operations through a specific target
+host, configure:
+
+```text
+ssh -T target-host --
+```
+
+Codex still opens the configured outer connection, but every remote probe,
+installation command, and app-server proxy is then routed through
+`target-host`.
+The wrapper is also part of the SSH startup-gate identity, so connections with
+different wrappers do not share startup sequencing.
+
+The field accepts POSIX-style argv text, including single quotes, double
+quotes, and backslash escaping. It is deliberately not a shell-snippet field:
+unquoted shell operators (`;`, `&`, `|`, `<`, and `>`) and newlines are
+rejected. The limit is 4096 input characters and 64 arguments. Wrapper
+failures use the existing SSH connection error path and abort setup.
+
+An empty wrapper leaves upstream SSH behavior unchanged. Saved discovered
+aliases and manually entered hosts both support wrappers.
+
+## Enable
+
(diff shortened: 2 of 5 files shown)


Yo-DDV (changes requested):
Thanks for the careful implementation and the end-to-end cluster test. The default-off feature boundary is appropriate, and I did not find a command-injection problem in the argv parser or quoting path. I did find two blocking correctness issues before this can merge.

1. The patch can report success while only part of the SSH path is wrapped. `replaceRequired()` accepts any occurrence and uses first-match `String.replace()`, while both patch functions treat the helper marker alone as already applied. I reproduced a main bundle with two valid `Gx(e,s)]` owners: one was wrapped, the other remained untouched, the descriptor was still classified as applied, and the second pass became byte-identical because `MAIN_MARKER` was present. A helper-only partial state is likewise accepted; the webview patch has the same failure mode. This can split management/probe and proxy behavior across different hosts while acceptance reports success. Please require unique or semantically owned targets, validate all postconditions before accepting an already-applied state, and add duplicate-target and partial-marker tests for both descriptors.

2. The editor/storage contract does not round-trip all values it accepts. `ssh "` plus 1,024 apostrophes plus `"` is only 1,030 characters, so `parseCommandWrapper()` accepts and persists it. On reload, `formatCommandWrapper()` expands each apostrophe to the standard `'\''` sequence, producing 4,102 characters; the same parser then rejects it because of the 4,096-character text limit. A valid saved connection therefore reopens as invalid and cannot be saved unchanged. Please make the parser, persisted validator, and formatter share a closed round-trip contract, and add boundary coverage for quote expansion.

The green Upstream Build run does not close these gaps: its acceptance artifact has `enabledFeatures: []`, so it never exercised either new descriptor against the current DMG. Given the number of current-minified-bundle owners involved, please add a feature-enabled current-DMG/watchdog gate that proves both descriptors apply, remain idempotent, and compose. Please also either register the feature in the Nix allowlist/tests or state explicitly that pure-flake selection is unsupported. The README should warn that wrapper argv is stored with the saved connection and must not contain passwords, tokens, private-key material, or similar secrets.

@ilysenko, the implementation is correctly isolated as an opt-in feature, but the repository-ownership question still needs your call: this adds recurring maintenance over both SSH transport and settings bundles for a specialized cluster workflow, with no linked issue. I would keep the review at changes requested until the two correctness blockers and feature-enabled acceptance are resolved, then decide tracked repository feature versus local feature based on that scope.

ilysenko (approved):
Reviewed the current head as an isolated, disabled-by-default Linux feature. The reported duplicate-marker and extreme quote-expansion cases are non-blocking hardening follow-ups for this optional integration; required checks are green.

PR #1162, opened 2026-07-27
Harden SSH command-wrapper patching

yanis-falaki:
## Summary

Follow-up to #1161, which introduced the opt-in `ssh-command-wrapper` feature.

Hardens the merged feature in response to review feedback.

- Requires the complete, unambiguous main-process and webview patch state before accepting a descriptor as applied; partial or duplicate targets fail safely as feature drift.
- Uses semantically specific SSH launch targets and validates full postconditions, with regression coverage for duplicate targets and helper-only partial states.
- Enforces a closed argv/editor round-trip contract for quote expansion, including apostrophe-heavy boundary coverage.
- Adds a feature-enabled current-upstream-DMG CI gate that requires both descriptors to apply and verifies a second pass is idempotent.
- Registers the feature in the Nix allowlist/watchdog profile and documents that saved wrapper argv must not contain secrets.

## Validation

Passed:

- `node --test linux-features/ssh-command-wrapper/test.js` — 14/14
- `node --test scripts/patch-linux-window-ui.test.js` — 387/387
- `git diff --check`
- Current-DMG rebuild with `ssh-command-wrapper` enabled: accepted; both descriptors applied.
- Current-DMG extracted-ASAR second pass: both descriptors reported `already-applied`.
- Workflow YAML and feature-profile JSON validation.

## Checklist

- [x] If this fixes upstream drift, it targets only the latest `CODEX.DMG` and removes obsolete fallback code and tests from the affected area. *(Not an upstream-drift fix; validated against the current DMG.)*
- [x] I added or updated relevant tests, ran the validation listed above, and confirmed that required CI checks pass. *(Tests were added and run; CI is pending.)*

### DIFF
--- a/.github/workflows/upstream-build-app.yml
+++ b/.github/workflows/upstream-build-app.yml
@@ -189,7 +189,14 @@ jobs:
         run: |
           set -euo pipefail
 
-          first_report="$GITHUB_WORKSPACE/upstream-acceptance/patch-report.json"
+          first_report="$(node -e '
+            const decision = require(process.argv[1]);
+            const reportPath = decision?.checks?.patchReport?.reportPath;
+            if (typeof reportPath !== "string" || reportPath.length === 0) {
+              throw new Error("acceptance decision does not record a patch report path");
+            }
+            process.stdout.write(reportPath);
+          ' "$GITHUB_WORKSPACE/upstream-dmg-decision.json")"
           second_report="$GITHUB_WORKSPACE/upstream-acceptance/ssh-command-wrapper-second-pass.json"
           node scripts/ci/validate-patch-report.js "$first_report" \
             --require-enabled-feature ssh-command-wrapper \
--- a/.github/workflows/upstream-build-app.yml
+++ b/.github/workflows/upstream-build-app.yml
@@ -11,6 +11,7 @@ on:
       - scripts/patch-linux-window-ui.test.js
       - scripts/patches/**
       - scripts/ci/validate-patch-report.js
+      - scripts/ci/ssh-command-wrapper-features.json
       - scripts/ci/upstream-dmg-*.js
       - scripts/validate-upstream-dmg.js
       - scripts/lib/candidate-install.sh
@@ -34,6 +35,7 @@ on:
       - scripts/patch-linux-window-ui.test.js
       - scripts/patches/**
       - scripts/ci/validate-patch-report.js
+      - scripts/ci/ssh-command-wrapper-features.json
       - scripts/ci/upstream-dmg-*.js
       - scripts/validate-upstream-dmg.js
       - scripts/lib/candidate-install.sh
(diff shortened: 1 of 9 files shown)


ilysenko (changes requested):
Thanks for addressing the earlier feedback. I want to correct the scope of my previous review: asking for a feature-enabled current-DMG/watchdog gate was too broad for this change. There is no need to take the validation that far.

`ssh-command-wrapper` is a niche, optional feature and must remain disabled by default. The shared Upstream Build and general CI workflows must not enable it or depend on it. Drift in this feature must never fail the default workflow or affect the repository-wide upstream-DMG acceptance result. We do not need to add or expand GitHub Actions for this feature.

Please remove:

- the `upstream-build-app.yml` changes;
- `scripts/ci/ssh-command-wrapper-features.json`;
- the addition to the shared watchdog feature profile and the test expectations added only for that global gate.

The Nix allowlist may remain if it only permits users to enable the feature explicitly and leaves the default Nix output unchanged. Focused feature-local tests are sufficient here.

One code-level issue from the review still needs to be addressed: the new complete-state detector validates helper function names, but not the exact current helper implementations. A stale or corrupted helper body can therefore be accepted as `already-applied` without a warning. Please validate the exact helper source/postcondition and add regression coverage for stale and damaged helper bodies.

I also found a pre-existing difference between the exported parser and the injected webview parser for escaped backticks inside double quotes. This PR did not introduce or worsen it, so it is a non-blocking follow-up and does not need to expand this change.

Once the shared workflow additions are removed, my earlier comments about the patch-report path and second-pass CI idempotency are no longer applicable.

Thanks again for the careful follow-up, and apologies for making the previous validation request broader than this optional feature warrants.

yanis-falaki:
> Thanks for addressing the earlier feedback. I want to correct the scope of my previous review: asking for a feature-enabled current-DMG/watchdog gate was too broad for this change. There is no need to take the validation that far.
> 
> `ssh-command-wrapper` is a niche, optional feature and must remain disabled by default. The shared Upstream Build and general CI workflows must not enable it or depend on it. Drift in this feature must never fail the default workflow or affect the repository-wide upstream-DMG acceptance result. We do not need to add or expand GitHub Actions for this feature.
> 
> Please remove:
> 
>     * the `upstream-build-app.yml` changes;
> 
>     * `scripts/ci/ssh-command-wrapper-features.json`;
> 
>     * the addition to the shared watchdog feature profile and the test expectations added only for that global gate.
> 
> 
> The Nix allowlist may remain if it only permits users to enable the feature explicitly and leaves the default Nix output unchanged. Focused feature-local tests are sufficient here.
> 
> One code-level issue from the review still needs to be addressed: the new complete-state detector validates helper function names, but not the exact current helper implementations. A stale or corrupted helper body can therefore be accepted as `already-applied` without a warning. Please validate the exact helper source/postcondition and add regression coverage for stale and damaged helper bodies.
> 
> I also found a pre-existing difference between the exported parser and the injected webview parser for escaped backticks inside double quotes. This PR did not introduce or worsen it, so it is a non-blocking follow-up and does not need to expand this change.
> 
> Once the shared workflow additions are removed, my earlier comments about the patch-report path and second-pass CI idempotency are no longer applicable.
> 
> Thanks again for the careful follow-up, and apologies for making the previous validation request broader than this optional feature warrants.

Cool cool, will do.

yanis-falaki:
Alright I’ve addressed the remaining feedback:

- Removed the `ssh-command-wrapper` changes from the shared Upstream Build workflow.
- Removed the feature-specific CI configuration and shared watchdog profile additions.
- Removed the test expectations that existed only for that global gate.
- Kept the Nix allowlist entry so users can explicitly select the feature; the default Nix output remains unchanged.
- Updated complete-state detection to require the exact current helper source, not only the expected function names.
- Added regression tests for stale main-process and damaged webview helper implementations.

The escaped-backtick parser difference remains unchanged as requested.

Validation now passes:

- `node --test linux-features/ssh-command-wrapper/test.js` — 16/16
- `node --test scripts/patch-linux-window-ui.test.js` — 395/395
- `node --test scripts/ci/upstream-dmg-acceptance.test.js` — 15/15
- `git diff --check`

The final diff no longer changes shared workflows, watchdog configuration, or general CI expectations.

ilysenko (approved):
Reviewed the current head. The shared CI/watchdog changes have been removed, the feature remains explicit opt-in, and exact helper-body validation plus regressions address the prior request.

PR #1174, opened 2026-07-29
fix(remote-mobile-control): retarget current gate bridge

MatsumotoMorami:
## Summary

- retarget the opt-in Remote Mobile Control patch from the retired slingshot gate bridge to the current gate bridge and dual-gate contract
- report recognizable current-bundle drift instead of treating it as already patched
- enable auto-connect only for the local Desktop host while preserving saved selections for other hosts
- expand regression coverage and update the feature documentation/changelog

## Validation

- node --test linux-features/remote-mobile-control/test.js (110/110)
- TMPDIR=<private 0700 directory> node --test scripts/patch-linux-window-ui.test.js (401/401)
- built a side-by-side app from Codex Desktop 26.721.81911 and verified Android QR pairing, session listing, thread resume/hydration, and a streamed live turn on Linux with Codex CLI 0.146.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- Remote mobile control now patches the current 26.721 dual-gate enablement
+  bridge instead of reporting it as already applied. Startup auto-connects the
+  environment owned by this Desktop without overwriting saved choices for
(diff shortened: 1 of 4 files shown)


ilysenko (approved):
Reviewed the current head: the feature-only gate-bridge drift fix is focused, preserves upstream dual-gate behavior and saved host choices, and all required checks pass.

PR #1176, opened 2026-07-29
Fix current Linux open target command lookup

### LINKED ISSUE #1170: Linux open-target-discovery lists VS Code and Terminal but cannot open projects
## Summary

When the opt-in `open-target-discovery` feature is enabled, Linux open targets correctly discover and display VS Code and Terminal. However, selecting either target fails immediately with “cannot open project”. The File Manager target works.

This appears to be a bug in the current Linux feature integration rather than a missing VS Code or Terminal installation.

## Environment

- Repository: `ilysenko/codex-desktop-linux`
- Repository commit: `c6d76231f0623c3ef0b18c7e9158697c96bdcf9f`
- ChatGPT Desktop / upstream app version: `26.721.41059`
- DMG SHA-256: `ae864e2def7db56d0bb77a876a5cbe4e4c2f554ccc654cec921b946892583c0a`
- Distribution: CachyOS Linux, rolling, Arch-based
- Desktop environment: KDE Plasma 6.7.3
- Window manager: KWin 6.7.3
- Session type: native Wayland
- Installation format: pacman package (`.pkg.tar.zst`)
- Installed VS Code executable: `/usr/bin/code`
- VS Code desktop entry: `/usr/share/applications/code.desktop`

Relevant enabled feature:

```json
{
  "enabled": [
    "open-target-discovery"
  ]
}
```

Other enabled features are listed in the attached `linux-features-staged.json`.

## Reproduction

1. Enable the `open-target-discovery` Linux feature.
2. Build and install the current Linux package from the latest supported DMG.
3. Open a project in ChatGPT Desktop.
4. Open the project/file open-target menu.
5. Select `VS Code` or one of the discovered `Terminal` targets.

## Expected behavior

- Selecting `VS Code` should launch `/usr/bin/code` with the project path.
- Selecting `Terminal` should launch the detected terminal with the project directory as its working directory.
- The File Manager target should continue to work.

## Actual behavior

- VS Code and Terminal are visible in the open-target menu.
- Selecting VS Code fails without launching VS Code.
- Selecting Terminal also fails without launching the terminal.
- File Manager opens the project successfully.

The application log contains:

```text
[open-in-service] Failed to open file errorMessage="Unknown open target \"vscode\"" errorName=Error errorStack="Error: Unknown open target \"vscode\""
```

The error occurs before `/usr/bin/code` is executed.

## Additional evidence

The installed build metadata confirms that `open-target-discovery` is enabled.

The generated patch report also shows both relevant feature patches as applied:

- `feature:open-target-discovery:main-bundle-open-target-discovery`
- `feature:open-target-discovery:webview-native-open-target-selection`

The installed main bundle contains Linux VS Code and Terminal target definitions, including:

```text
linux:codexLinuxIdePlatform
linux:{label:`Terminal`
```

However, the current upstream open-in service still contains a private lookup method equivalent to:

```js
async#t(e) {
  let {command:t} = await this.#n()({
    method: `get-target-command`,
    params: T0(this.settingsStore, e)
  });
  if (t == null) {
    throw Error(`Open target "${e}" is not available`);
  }
  return t;
}
```

The Linux feature patch currently handles the older `getOpenInTargetCommand` path, but does not appear to patch this current `open-in-service` private method. As a result, the service asks the old worker registry for `vscode`, which returns:

```text
Unknown open target "vscode"
```

The same path likely explains the Terminal failure.

## Validation

The feature’s existing tests pass:

```bash
node --test linux-features/open-target-discovery/test.js
```

Result:

```text
38 tests passed
```

However, the existing tests cover the older `getOpenInTargetCommand` shape and do not cover the current `open-in-service` private `#t()` lookup path.

## Possible fix direction

On Linux, the current `open-in-service` target-command lookup should use the Linux target registry already introduced by the feature, for example:

```js
codexLinuxOpenTargetRegistryCommand(this.settingsStore, target)
```

A regression test should cover the current `async#t(e)` service shape and verify that Linux does not call the old worker `get-target-command` path for discovered targets.

Related historical issues:

- [#123 — Linux open targets only show file manager and miss installed IDEs](https://github.com/ilysenko/codex-desktop-linux/issues/123)
- [#161 — Add opt-in Linux open target discovery](https://github.com/ilysenko/codex-desktop-linux/issues/161)
- [#171 — Refine Linux open target discovery](https://github.com/ilysenko/codex-desktop-linux/issues/171)

This report is specifically about the current feature being able to discover targets but failing to launch them.

PinguuSS:
## Summary

- route the current private open-target command lookup through the feature-owned Linux registry while preserving the upstream worker path outside Linux
- validate the complete current command shape before any main-bundle mutation, including the Linux platform guard, and report missing, partial, or corrupted markers as enabled-feature drift with byte-identical output
- cover current-DMG first pass, second pass, drift rejection, patch-report behavior, and the corrupted-guard regression; update the changelog

## Root cause

The feature patched target detection and availability, but the current `async#t(e)` command lookup still sent `get-target-command` to the worker registry. Linux-only targets were therefore visible in the menu but failed at selection with `Unknown open target`.

## Current upstream

- Base: `origin/main@586c20e802b0`
- ChatGPT Desktop: `26.721.81911`
- Electron: `42.3.0`
- DMG SHA-256: `7b3722e4f5864202b1dd69e6e6062f2f8c4388d215091503c73e1957b4cbf97a`
- DMG size: `593881254`
- ETag: `0x8DEED41E83B4B62`

## Validation

- `node linux-features/open-target-discovery/test.js` (42/42)
- `node --test --test-concurrency=1 linux-features/*/test.js` (833/833 before the final detector-only hardening; the affected feature remains 42/42 afterward)
- `node scripts/patch-linux-window-ui.test.js` (401/401 before the final detector-only hardening)
- `./scripts/ci-local.sh pr`
- patch-report validation passed for an isolated `open-target-discovery` candidate and the normal 10-feature composition; both acceptance verdicts were `accepted` with zero blockers/warnings and clean Git provenance
- active generated main/webview assets passed `node --check`
- KDE/Wayland side-by-side runtime listed the current targets; selecting Terminal spawned `/usr/bin/x-terminal-emulator` with the project as process and shell cwd, showed no error toast, and remained correct after a cold restart

## Scope

Source changes are limited to the opt-in feature patch/test and the changelog. The daily driver was not modified. Account/server rollout and the upstream worker implementation are unchanged.

Fixes #1170

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- Open Target Discovery now resolves the selected Linux editor or terminal
+  through the current private open-target command path. Command-path drift is
+  reported before the feature changes the main bundle, so enabled-feature
+  acceptance cannot mistake a partially patched bundle for success.
 - Remote mobile control now patches the current 26.721 dual-gate enablement
   bridge instead of reporting it as already applied. Startup auto-connects the
   environment owned by this Desktop without overwriting saved choices for
--- a/linux-features/open-target-discovery/patch.js
+++ b/linux-features/open-target-discovery/patch.js
@@ -624,42 +624,15 @@ function findCurrentOpenTargetCommandMatch(source) {
   );
 }
 
-function findPatchedOpenTargetCommandBlock(source) {
-  const marker = "let _codexLinuxOpenTargetCommand=await codexLinuxOpenTargetRegistryCommand(";
-  const markerIndex = source.indexOf(marker);
-  if (markerIndex === -1) {
-    return null;
-  }
-
-  const methodStart = source.lastIndexOf("async#", markerIndex);
-  const blockStart = methodStart === -1 ? -1 : source.indexOf("{", methodStart);
-  const block = findBalancedBlock(source, blockStart);
-  const signature = source.slice(methodStart, blockStart).match(
-    /^async#[A-Za-z_$][\w$]*\(([A-Za-z_$][\w$]*)\)$/u,
+function findPatchedOpenTargetCommandMatch(source) {
+  return source.match(
+    /async#([A-Za-z_$][\w$]*)\(([A-Za-z_$][\w$]*)\)\{if\(process\.platform===`linux`\)\{let _codexLinuxOpenTargetCommand=await codexLinuxOpenTargetRegistryCommand\(this\.settingsStore,\2\);if\(_codexLinuxOpenTargetCommand==null\)throw Error\(`Open target "\$\{\2\}" is not available`\);return _codexLinuxOpenTargetCommand\}let\{command:([A-Za-z_$][\w$]*)\}=await this\.#([A-Za-z_$][\w$]*)\(\)\(\{method:`get-target-command`,params:([A-Za-z_$][\w$]*)\(this\.settingsStore,\2\)\}\);if\(\3==null\)throw Error\(`Open target "\$\{\2\}" is not available`\);return \3\}/u,
   );
-  if (block == null || signature == null || block.end <= markerIndex) {
-    return null;
-  }
-
-  const targetVar = signature[1];
-  const linuxCommand =
-    `let _codexLinuxOpenTargetCommand=await codexLinuxOpenTargetRegistryCommand(this.settingsStore,${targetVar})`;
-  const unavailable =
-    `if(_codexLinuxOpenTargetCommand==null)throw Error(\`Open target "\${${targetVar}}" is not available\`)`;
-  if (
-    !block.text.includes(linuxCommand) ||
-    !block.text.includes(unavailable) ||
-    !block.text.includes("return _codexLinuxOpenTargetCommand") ||
-    !block.text.includes("method:`get-target-command`")
-  ) {
-    return null;
-  }
-  return block;
 }
 
 function applyOpenInTargetCommandPatch(currentSource) {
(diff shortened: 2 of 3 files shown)


PinguuSS:
Final contract review found one actionable gap in the already-applied detector: a patched command method with a corrupted platform guard could still be accepted as complete. I am tightening the complete-marker match, adding a byte-identical negative regression, and will rerun the focused/current-DMG evidence before requesting review. The PR will remain at three logical commits.

PinguuSS:
The complete-marker hardening is now in `92425cc`. The detector requires the exact Linux guard and preserved non-Linux worker fallback; a guard-only corruption is left byte-identical and reported as partial drift. Focused coverage is 42/42, the current active DMG bundle passes clean first/second passes plus the negative drift case, and all GitHub checks on `a9a757e` are green (upstream DMG, Rust/smoke, Debian, RPM, Pacman, and Nix). The branch remains exactly three commits and is ready for review.

ilysenko (changes requested):
Thanks for the focused fix and thorough current-DMG coverage. One cleanup blocker remains:

- Please remove the remaining `getOpenInTargetCommand` legacy-shape check from `applyOpenInTargetRegistryCommandPatch`. This PR updates drift handling to the current private command lookup, and the repository's current-DMG-only policy requires obsolete DMG markers and fallback diagnostics to be removed in the same change.

Once that marker is removed, the feature-local fix looks ready.

ilysenko (approved):
Reviewed the updated head. The requested current-DMG cleanup is complete and all required checks are green.

Issue #1175, opened 2026-07-29
AppView RPC registration fails because Linux Sparkle shim lacks getDownloadProgressPercent


NicolasFerec:
## Summary

On startup, the packaged Linux build repeatedly fails to register the AppView RPC services because the Linux Sparkle replacement does not provide `getDownloadProgressPercent`:

```text
Failed to register AppView RPC services errorMessage="this.options.sparkleManager.getDownloadProgressPercent is not a function" errorName=TypeError errorStack="TypeError: this.options.sparkleManager.getDownloadProgressPercent is not a function
    at dge.getAppUpdateViewState (/opt/codex-desktop/resources/app.asar/.vite/build/main-DXmJ7M03.js:1260:20889)
    at dge.registerAppView (/opt/codex-desktop/resources/app.asar/.vite/build/main-DXmJ7M03.js:1260:15957)"
```

The message is emitted multiple times during a single launch.

## Environment

- OS: Ubuntu 26.04 LTS, amd64
- Session: GNOME Wayland
- Package: `codex-desktop 2026.07.25.020936+ae864e2d amd64`
- Upstream desktop app version reported by Crashpad: `26.721.41059`
- Electron: `42.3.0`
- Installation path: `/opt/codex-desktop`

The issue is reproducible with both the launcher's default rendering profile and `CODEX_LINUX_RENDERING_MODE=wayland-gpu`, so it does not appear related to the selected Ozone backend.

## Steps to reproduce

1. Install and launch the packaged `.deb` build.
2. Inspect `~/.cache/codex-desktop/launcher.log`.
3. Search for `Failed to register AppView RPC services`.

## Expected behavior

The Linux Sparkle/update-manager compatibility layer should implement the interface expected by the current upstream bundle, including `getDownloadProgressPercent`, or otherwise adapt `getAppUpdateViewState` so AppView RPC service registration completes without a `TypeError`.

ilysenko:
Thank you for the report. I am investigating this now.

PR #1179, opened 2026-07-29
fix(updater): complete Linux Sparkle menu contract

MatsumotoMorami:
## Summary

- add hasUpdater and getUnavailableReason to the Linux package updater shim, matching the methods used by the current 26.721 Check for Updates menu
- report updater availability from the existing codex-update-manager probe and return a stable reason while the probe is unavailable
- derive the menu-facing contract from a current consumer fixture and execute the generated manager through successful and failed probe paths

## Context

Follow-up to #1177 / #1175. That change completed the AppView and RPC getters, but the current menu consumer also calls hasUpdater and getUnavailableReason before checkForUpdates.

## Validation

- focused updater bridge tests: 4/4
- full scripts/patch-linux-window-ui.test.js: 403/403 with a private TMPDIR
- node --check on both changed JavaScript files
- git diff --check


ilysenko (approved):
Reviewed the current Linux Sparkle menu contract and probe behavior. The focused follow-up is covered by green upstream and cross-format CI.

Issue #1073, opened 2026-07-18
GNOME/X11: top-bar right-click still disables mouse input after PR #1072


psypeal:
## Summary

Right-clicking the Codex top/title bar still disables mouse interaction on GNOME/X11 after the fix from #1069 / #1072 is installed.

After the right-click, the terminal/sidebar controls and buttons inside the active thread remain visible but cannot be clicked with the mouse. This makes the conversation UI effectively unusable.

## Environment

- Ubuntu 24.04.4 LTS
- GNOME Shell 46.0
- X11 session
- Package: `codex-desktop 2026.07.18.111000`
- Upstream app: `26.715.31925`
- Electron: `42.3.0`
- Installed wrapper source: `f7de27215d637a41f615423f436263257da9016b`
- Package format: `.deb`
- `frameless-titlebar` is not enabled

## Fix-presence verification

The installed wrapper source contains merge commit `cc610ca3f66a2aeacbcb9436c15ce5a8a84866d7` from PR #1072. The installed patch report also reports `linux-native-titlebar` as applied.

This is therefore a post-fix reproduction, not a stale package that predates #1072.

PR #1072 noted that live GNOME/X11 validation still required a desktop rebuild/retest. This report supplies that validation and shows that the `system-context-menu` suppression is not sufficient for the current top-bar path.

## Reproduction

1. Launch ChatGPT Desktop on Ubuntu GNOME/X11.
2. Open any conversation thread.
3. Right-click the Codex top/title bar.
4. Move the pointer to the terminal/sidebar controls or any button in the thread conversation.
5. Attempt to click the controls.

## Actual behavior

Mouse clicks no longer operate the terminal/sidebar controls or buttons in the conversation thread. The UI remains visible, but pointer interaction is stuck after the top-bar right-click.

## Expected behavior

Right-clicking the top bar should not leave an active X11/GNOME input grab or block pointer events. Whether the system context menu is suppressed or displayed, mouse interaction with Codex and the rest of the desktop must remain usable afterward.

## Related work

- #496 — original GNOME/X11 title-bar input-lockup report.
- #649 — later Ubuntu/X11 reproduction.
- #1069 — current-generation reproduction, closed by #1072.
- #1072 — added Linux `system-context-menu` prevention but was not live-validated on GNOME/X11 before merge.

## Acceptance criteria

- [ ] Rebuild and live-test the default title-bar path on Ubuntu GNOME/X11.
[...]

PR #1180, opened 2026-07-29
Fix GNOME/X11 titlebar context-menu patch false positive

MatsumotoMorami:
## Summary

Addresses #1073.

### Root cause

PR #1072 added Linux `system-context-menu` suppression through the bundle-wide
`linux-menu` patch, but the current upstream WindowManager does not use the
Windows-only shape that patch expected:

```js
(process.platform === "win32" || process.platform === "linux") && N.removeMenu()
```

The browser-comment popup in the same bundle still used a Windows-only
`e.removeMenu()` expression. As a result, the old global matcher patched only
that auxiliary popup and still reported `linux-menu: applied`; the registered
primary `BrowserWindow` behind the top/title bar never received the listener.
This omission is present in pristine core bundles and does not require an
optional Linux feature to alter the patch path.

### Fix

- Add a required-upstream `linux-managed-window-system-context-menu`
  descriptor before the auxiliary global menu patch.
- Identify the unique managed `BrowserWindow` semantically within
  `WindowManager.createWindow`: the current `appearance` contract, a
  `BrowserWindow` construction, and the same alias passed to
  `this.registerWindow(...)` must all agree.
- Replace only the current combined Linux/Windows `removeMenu()` contract with
  Linux `system-context-menu` prevention plus `removeMenu()`, while preserving
  the Windows-only `removeMenu()` behavior.
- Fail closed on missing or ambiguous WindowManager candidates, malformed
  methods, non-canonical or duplicate listeners, and extra `removeMenu()`
  targets. The required patch records a strategy in the patch report so future
  upstream drift is diagnosable instead of silently masked by a popup match.
- Keep the browser-comment popup handling in the separate `linux-menu` patch,
  teach that patch the current combined contract, and remove obsolete legacy
  menu fallback shapes in accordance with the fresh-DMG-only policy.
- Add fixture, runtime, descriptor-report, CLI smoke, and real-bundle
  idempotence coverage.

## Validation

Automated checks:

- `TMPDIR=/tmp node --test scripts/patch-linux-window-ui.test.js`
  - 412 passed, 0 failed.
- `TMPDIR=/tmp node --test scripts/ci/upstream-dmg-acceptance.test.js`
  - 15 passed, 0 failed.
- `TMPDIR=/tmp node --test linux-features/frameless-titlebar/test.js linux-features/ui-tweaks/test.js linux-features/ui-tweaks/dock-icon.test.js linux-features/ui-tweaks/suggested-prompts.test.js`
  - 97 passed, 0 failed.
- `bash scripts/ci/run-node-checks.sh` in the Ubuntu CI container
  - 1,445 passed, 2 skipped, 0 failed (1,447 total).
- `bash tests/scripts_smoke.sh` in the Ubuntu CI container
  - all script and packaging smoke tests passed, including the new managed
    titlebar context-menu CLI/report/SHA idempotence smoke.
- `git diff --check`

The focused tests cover:

- Primary WindowManager versus nested browser-comment popup ownership.
- Minified alias drift (`N` and `M`).
- Missing, malformed, ambiguous, duplicate, and non-canonical contracts.
- Required-patch failure reporting and strategy telemetry.
- First application, half-patched upgrade, and byte-idempotent second
  application.
- Runtime behavior on simulated Linux, Windows, and macOS Electron
  `BrowserWindow` instances.
- Preservation of unrelated non-Darwin modal menu behavior.

Real upstream bundle checks:

| Upstream build | Source state | Result |
| --- | --- | --- |
| `26.715.31925` | pristine issue build | applied, then byte-identical `already-applied` |
| `26.715.52143` | pristine later issue build | applied, then byte-identical `already-applied` |
| `26.721.81911` | pristine latest build | applied, then byte-identical `already-applied` |
| `26.721.81911` | existing half-patched installed-bundle copy | upgraded, then byte-identical `already-applied` |

For every bundle above:

- the managed WindowManager candidate was unique;
- the final main-process bundle parsed successfully;
- the primary window had exactly one listener;
- the browser-comment popup had exactly one listener;
- no old combined menu target remained; and
- a second full patch pass preserved the main bundle and extracted-tree SHA.

An independent maximum-reasoning diff review also checked target ownership,
atomic failure behavior, platform semantics, idempotence, and real-bundle
results. Its two findings (alternate listener APIs/quotes and an extra Windows
menu target) were addressed with fail-closed validation and regression tests;
the final review found no remaining actionable issue.

### Ubuntu GNOME/X11 VM interaction

I built both packages from the same official `26.721.81911` upstream artifact
and tested them in the same disposable KVM guest:

- Ubuntu `24.04.4`, kernel `6.8.0-124`, GNOME Shell `46.0`.
- GDM forced to Xorg; `loginctl` reported `Type=x11`, and Electron's GPU and
  renderer processes reported `--ozone-platform=x11`.
- Default feature configuration (`enabledFeatures: []`).
- Exact `origin/main` baseline `042adb7b23b1f47ca8d6a09da0b1abc5e00ff4a7`,
  package SHA-256
  `0c3585a4e5721b273e7ead589f9fcd47a0da09572518fd21faebd1253316728e`.
- PR commit `5b74648bff603a5284e07b96b512359d7cb2fc04`, package
  SHA-256
  `2cf91f7e3c1d0ec2e6d50374a98e982805b5cabf494e47289bb0c6d252c55d0a`.

The X11 grab helper was first self-tested: one process acquired the pointer
(`GrabSuccess`), a concurrent probe detected it (`AlreadyGrabbed`), and the
probe returned to `GrabSuccess` after release.

The initial interaction matrix was run against both baseline and PR packages:

- 203 recorded checks per package across windowed and maximized states.
- Left, center, and right points in the actual CSS drag strip, ten repetitions
  per point.
- Right-click followed by an outside dismissal, then grab probes at 100 ms,
  500 ms, and 2 s.
- First post-dismiss window-control clicks verified through
  `_NET_WM_STATE_HIDDEN`.
- Zero post-dismiss grab failures; the final idle probe was `GrabSuccess`.

I also attached a read-only observer through Electron's main-process inspector.
It was registered after the application's listeners and never called
`preventDefault()`:

- The baseline's visible registered `BrowserWindow` had zero pre-existing
  `system-context-menu` listeners. Its observed top-bar event had
  `defaultPrevented: false`, and the GNOME window menu was displayed.
- The PR package's same window had exactly one pre-existing listener. Its
  observed event had `defaultPrevented: true`.

Finally, an additional PR-only edge matrix used CDP to verify that `y=4`,
`y=16`, and `y=28` were one continuous `-webkit-app-region: drag` run, then
tested the upper and lower edges at three horizontal points:

- 181 recorded checks; 36 right-click sequences across windowed and maximized
  states.
- All 36 main-process events reached the managed window and all 36 were already
  default-prevented by the PR listener.
- All 108 post-dismiss grab probes returned `GrabSuccess`.
- After every dismissal, one physical click on an instrumented in-app button
  produced exactly one trusted `pointerdown`, `mousedown`, `pointerup`,
  `mouseup`, and `click`: 180 trusted events total.
- The app, Electron processes, GNOME Shell, and Xorg remained alive, with no
  matching fatal/crash entry in the captured app log.

#### Scope boundary

The current signed-out VM scenario did **not** reproduce #1073's persistent
post-dismiss lock: the unpatched baseline also released its visible GNOME menu
and passed the 203-check stress matrix. The VM therefore provides direct
runtime proof that the baseline missed the managed-window listener, that this
PR installs and executes it, and that the PR path has no observed X11 input
regression; it is not a claim that the reporter's authenticated-thread failure
was reproduced and resolved A/B.

The guest deliberately remained unauthenticated, and no host credentials or
private task data were copied into it. Consequently, the issue's exact
sidebar/thread/composer controls and the reporter's unavailable private
`browser-multi-tabs` feature could not be tested. A reporter-equivalent
authenticated retest is still required before closing the issue. This PR
therefore intentionally says **Addresses #1073**, not `Fixes` or `Closes`.

## Checklist

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -1099,6 +1099,14 @@ test("default core patch descriptors are grouped and unique", () => {
     "optional",
   );
   assert.equal(
+    descriptors.find(
+      (descriptor) =>
+        descriptor.id === "linux-managed-window-system-context-menu",
+    )?.ciPolicy,
+    "optional",
+    "GNOME/X11 titlebar mitigation drift should warn without blocking install or updater candidates",
+  );
+  assert.equal(
     descriptors.find((descriptor) => descriptor.id === "linux-computer-use-native-desktop-apps")?.ciPolicy,
     "opt-in",
   );
@@ -1137,7 +1145,6 @@ test("default core patch descriptors are grouped and unique", () => {
   );
   for (const id of [
     "linux-window-options",
-    "linux-managed-window-system-context-menu",
     "linux-native-titlebar",
     "linux-opaque-background",
     "linux-avatar-overlay-mouse-passthrough",
@@ -3380,11 +3387,40 @@ function windowsAndLinuxMenuSnippet(windowAlias) {
   );
 }
 
+function gnomeX11SystemContextMenuListenerSnippet(
+  windowAlias,
+  eventAlias = "e",
+) {
+  return (
+    "process.platform===`linux`&&" +
+    "(process.env.XDG_SESSION_TYPE??``).trim().toLowerCase()===`x11`&&" +
+    "/(^|:)gnome(:|$)/i.test((process.env.XDG_CURRENT_DESKTOP??``).trim())&&" +
+    `${windowAlias}.on(\`system-context-menu\`,${eventAlias}=>` +
+    `${eventAlias}.preventDefault()),`
+  );
+}
+
 function canonicalLinuxMenuSnippet(windowAlias, eventAlias = "e") {
+  return (
+    gnomeX11SystemContextMenuListenerSnippet(windowAlias, eventAlias) +
+    `process.platform===\`linux\`&&${windowAlias}.removeMenu(),` +
+    `process.platform===\`win32\`&&${windowAlias}.removeMenu(),`
+  );
+}
+
+function legacyAllLinuxMenuSnippet(windowAlias, eventAlias = "e") {
   return (
     `process.platform===\`linux\`&&(${windowAlias}.on(\`system-context-menu\`,` +
     `${eventAlias}=>${eventAlias}.preventDefault()),${windowAlias}.removeMenu()),` +
     `process.platform===\`win32\`&&${windowAlias}.removeMenu(),`
+  );
+}
+
+function scopedManagedLinuxMenuSnippet(windowAlias, eventAlias = "e") {
+  return (
+    gnomeX11SystemContextMenuListenerSnippet(windowAlias, eventAlias) +
+    `(process.platform===\`win32\`||process.platform===\`linux\`)&&` +
+    `${windowAlias}.removeMenu(),`
   );
 }
 
@@ -3416,12 +3452,23 @@ test("patches the managed WindowManager window before the browser-comment popup"
   assert.equal(
     (managedPatched.slice(popupStart).match(/system-context-menu/g) ?? []).length,
     0,
-    "the required managed-window patch must not claim success by patching only the popup",
+    "the managed-window patch must not claim success by patching only the popup",
   );
-  assert.match(managedPatched, new RegExp(escapeRegExp(canonicalLinuxMenuSnippet("N"))));
+  assert.match(
+    managedPatched,
+    new RegExp(escapeRegExp(scopedManagedLinuxMenuSnippet("N"))),
+  );
 
   const fullyPatched = applyLinuxMenuPatch(managedPatched);
   assert.equal((fullyPatched.match(/system-context-menu/g) ?? []).length, 2);
+  assert.match(
+    fullyPatched,
+    new RegExp(escapeRegExp(scopedManagedLinuxMenuSnippet("N"))),
+  );
+  assert.match(
+    fullyPatched,
+    new RegExp(escapeRegExp(canonicalLinuxMenuSnippet("e"))),
+  );
   assert.equal(
     applyLinuxMenuPatch(
       applyLinuxManagedWindowSystemContextMenuPatch(fullyPatched),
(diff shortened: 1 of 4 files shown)


ilysenko (changes requested):
Thanks for the thorough implementation and validation. I found one scope/policy blocker:

- The new managed-window patch is `required-upstream` and suppresses the system context menu on every Linux desktop, while the available evidence is limited to GNOME/X11 and the documented baseline did not reproduce the lockup. This makes harmless matcher drift reject all package and updater candidates and changes KDE/Wayland behavior without supporting evidence. Please make this patch fail-soft and scope the runtime suppression to the affected GNOME/X11 surface, or provide default-feature reproduction plus cross-desktop validation that justifies unconditional critical behavior.

MatsumotoMorami:
### Review follow-up

Thanks for the review. I followed the first requested path in `c9c327539b386d00e209017652515b21183c6ef9`.

#### Implementation

- **Fail-soft policy**
  - Changed `linux-managed-window-system-context-menu` from `required-upstream` to `optional`.
  - Matcher failure now preserves the original bundle bytes and reports `skipped-optional`.
  - It contributes no required-core warning and does not reject package or updater candidates.

- **GNOME/X11 runtime scope**
  - The `system-context-menu` listener is registered only when:
    - `process.platform === "linux"`;
    - normalized `XDG_SESSION_TYPE === "x11"`; and
    - `XDG_CURRENT_DESKTOP` contains an exact colon-delimited `GNOME` token, including values such as `ubuntu:GNOME`.
  - Existing Linux and Windows `removeMenu()` behavior is unchanged; only listener registration is scoped.

- **Fail-soft fallback safety**
  - The later generic `linux-menu` patch now uses the same GNOME/X11 guard.
  - If managed-window discovery drifts but the broader menu matcher still succeeds, it cannot reintroduce unconditional suppression on KDE, Wayland, or unknown Linux sessions.

- **Existing patched bundles**
  - Added migration for the previous unconditional managed-window and generic listener forms.
  - Rerunning the patcher now narrows existing outputs instead of accepting the broader behavior as already applied.

#### Validation

- `TMPDIR=/tmp node --test scripts/patch-linux-window-ui.test.js`: 415/415 passed.
- Targeted file-manager and titlebar context-menu smoke tests passed, including:
  - first application;
  - patch-report policy;
  - byte-identical second pass.
- Runtime coverage verifies:
  - the listener is enabled on GNOME/X11 and `ubuntu:GNOME` X11;
  - the listener is absent on GNOME/Wayland, KDE/X11, unknown Linux sessions, Windows, and macOS;
  - Linux `removeMenu()` behavior remains unchanged.
- On an app derived from the official `26.721.81911` artifact:
  - the managed patch was reported as optional and applied successfully;
  - both final `system-context-menu` sites carried the GNOME/X11 guard;
  - the next full pass applied zero descriptors;
  - the extracted-tree SHA remained byte-identical.

#### Scope

The earlier GNOME/X11 VM evidence remains applicable to the affected runtime surface because the new guard evaluates true there.

This follow-up does not claim cross-desktop justification for unconditional behavior, and it does not change the documented reproduction caveat.

Please re-review when convenient.

ilysenko (changes requested):
Thanks for addressing the previous review. The GNOME/X11 runtime scope and fail-soft policy now look correct. One policy blocker remains: please remove the `semanticLegacy*`/`legacy*` matchers, the `legacy-unscoped` migration paths, and their migration tests. They only support previously patched historical output, while this repository’s patch contract is fresh-DMG-only. Final-form idempotence is already covered by the scoped `already-applied` path.

ilysenko (approved):
Reviewed the updated current-DMG implementation. The legacy migration paths are removed, the managed-window matcher is verified against 26.727.40816, drift remains fail-soft, and the final head is clean and green.

PR #1181, opened 2026-07-29
Add opt-in OpenAI dictation quality feature

tishmen:
## Summary

- add a disabled-by-default `dictation-capture-quality` Linux feature
- request stereo capture with Chromium echo cancellation, noise suppression, and AGC disabled
- route completed WebM/Opus recordings through a main-process bridge to the user's own OpenAI API key using `gpt-4o-transcribe` and `language=en`
- keep the key in Secret Service and out of the renderer, source, environment, and logs
- restrict paid requests to primary Codex windows, allow one request at a time, bound payload size/time, and sanitize upstream errors
- declare the feature's conflict with `conversation-mode` and native `secret-tool` dependencies where package names are portable

## Why

On the affected Linux microphone path, controlled capture comparisons isolated Chromium WebRTC processing as the major accuracy regression. The clean stereo capture produced accurate transcription, while Codex's normal processed path did not. Codex's built-in streaming route also cannot use the user's selected API key, model, and fixed English hint.

This remains opt-in because it changes capture processing, disables the built-in streaming route, sends completed audio to the user-billed OpenAI Audio Transcriptions API, and is intentionally English-only and batch-only.

## Security and support boundaries

The renderer can only pass bounded base64 WebM data through the existing trusted bridge. The main process checks the originating primary window, canonical base64, MIME type, size, and EBML signature before reading the Codex-owned Secret Service credential. It uses a fixed OpenAI endpoint/model/language, never returns upstream bodies, combines caller cancellation with a 120-second timeout, and clears mutable audio/secret buffers best-effort.

Debian and pacman packages declare the portable `secret-tool` provider dependency. RPM, AppImage, and other self-built installs must provide `/usr/bin/secret-tool` and an unlocked Secret Service implementation, as documented.

## Validation

- `node --test linux-features/dictation-capture-quality/test.js` — 20/20 passed
- `node --test linux-features/*/test.js` — 849 passed, 1 intentional skip, 0 failed
- `node --test scripts/patch-linux-window-ui.test.js` — 403/403 passed
- `node --test scripts/lib/linux-features.test.js` — 28/28 passed
- `CODEX_DMG_REFRESH_MODE=pinned make build-app` — accepted with zero blockers or warnings; both feature descriptors applied to Codex 26.721.81911 / Electron 42.3.0
- staged diff and syntax/JSON checks passed; no credential-shaped strings found

### DIFF
--- a/linux-features/dictation-capture-quality/README.md
+++ b/linux-features/dictation-capture-quality/README.md
@@ -1,0 +1,91 @@
+# OpenAI Dictation Quality
+
+Opt-in Linux dictation that uses a verified clean capture path while keeping the
+user's OpenAI API key out of the renderer, source tree, package, environment,
+and logs.
+
+The feature changes only the existing Codex composer dictation path:
+
+- requests two-channel audio;
+- disables Chromium echo cancellation, noise suppression, and automatic gain
+  control;
+- disables Codex's built-in streaming transcription so it cannot bypass this
+  feature's OpenAI batch endpoint;
+- passes base64 WebM audio through Codex's trusted `vscode://codex` bridge;
+- accepts bridge calls only from primary Codex windows and permits one paid
+  transcription request at a time;
+- retrieves a Codex-owned key from the freedesktop.org Secret Service in the
+  Electron main process;
+- sends the unchanged WebM/Opus recording to OpenAI with
+  `model=gpt-4o-transcribe` and `language=en`.
+
+The key is never sent to the webview. Before opening Secret Service or making a
+network request, the bridge checks the caller, canonical base64 encoding,
+declared WebM/Opus MIME type, 25 MiB size limit, and WebM EBML signature. The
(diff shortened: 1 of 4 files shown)


ilysenko (changes requested):
Thanks for the careful implementation and thorough feature-local tests. I found two blockers:

1. The demonstrated audience is currently too narrow for permanent repository ownership: this combines device-specific capture constraints with an English-only, user-billed BYO-key transcription path, without linked multi-user/device evidence. Please keep it under `linux-features/local/` for now, or demonstrate broader reusable demand and separate the capture-quality concern from the BYO-key transcription product.
2. Cancellation is not wired through the production renderer/main bridge. The renderer posts without a signal, the current dispatcher does not inject one, and the test passes a signal directly to the raw handler. As a result, abandoned paid uploads can continue until timeout and hold the global slot. Please wire real lifecycle cancellation and cover the full bridge path end-to-end.

ilysenko:

Issue #1171, opened 2026-07-29
[Bug] Clicking the GNOME dock/application icon after hiding to tray triggers a conflict and breaks the running instance


Naerelyth:
1## Environment

- OS: Fedora Linux
- Desktop: GNOME 50.3
- Session: Wayland
- Electron: 42
- Tested independently with:
  - **AppImage installed through Gear Lever**
- Build command:
  ```bash
  PACKAGE_WITH_UPDATER=0 make build-app
  make appimage
  ```

## Description

When the main window is closed, the application correctly hides to the tray. Clicking the tray icon restores the window as expected.

However, clicking the application's `.desktop` icon from the GNOME application overview or a pinned dock/panel icon does not reliably restore the existing window.

## Reproduction steps

1. Build and install the AppImage through Gear Lever.
2. Start ChatGPT Desktop from its `.desktop` entry.
3. Close the main window so it hides to the tray.
4. Click the ChatGPT icon from the GNOME application overview or dock.
5. Observe the application window state.

~The same behavior was reproduced independently with the RPM package.~

## Actual behavior

After the second click:

- The existing window remain hidden and may become unhealthy.
- The running instance reports various errors.

<img width="800" alt="Image" src="https://github.com/user-attachments/assets/ead052c9-85a2-4ebc-af35-df71608b9fb8" />

- PID, socket, and webview state may be removed or replaced.
- The application may recover after relaunching or manually deleting:

```bash
rm -rf \
  ~/.config/codex-desktop \
  ~/.local/state/codex-desktop \
  ~/.cache/codex-desktop \
  ~/.config/codex-update-manager \
  ~/.local/state/codex-update-manager \
  ~/.cache/codex-update-manager
```

## Expected behavior

Clicking the application icon after hiding the window to the tray should:

- Reuse the existing Electron instance.
- Restore and focus the existing window.
- Preserve the existing webview server, PID marker, socket, and application state.
- Treat different temporary AppImage mount points as the same installation.
- Avoid destructive cleanup of the running instance.

ilysenko:
Thank you for the report. I am investigating this now.

ilysenko:
Thanks again for the detailed report. After investigating, this appears to be a very narrow AppImage/Gear Lever case: a second activation can use a different transient FUSE mount for the same AppImage.

Handling that generically in core would add substantial process-discovery and state-reconciliation complexity to an already complex launcher, so we are not going to land a broad launcher change from the maintainer side right now.

[...]

PR #1183, opened 2026-07-29
fix: restore AppImage window from desktop activation

Naerelyth:
closes：#1171

## Summary

- Fix AppImage desktop activation after the main window is closed to the tray.
- Handle Gear Lever launches where the same AppImage is started from a new FUSE mount: identify the resident process through the stable `APPIMAGE` path rather than the mount-specific Electron path.
- Send desktop activation to the resident launch-action socket before launcher state reconciliation can discard that socket.
- Normalize desktop activation to `--show` and serialize launch actions to prevent concurrent activation handling from corrupting the running instance.
- Preserve app-id and side-by-side instance isolation.

## Validation

- Manually built the AppImage and tested on Fedora GNOME with Gear Lever integration:
  - Close the window to the tray.
  - Restore it through the Applications overview and a pinned dock/panel icon.
  - Repeated activations restore the existing window without creating or corrupting a second instance.
- `git diff --check`
- Shell syntax checks for the launcher and smoke scripts.
- Targeted launch-action patch test:
  - `node --test --test-name-pattern='adds Linux launch actions through current setSecondInstanceArgsHandler bundle' scripts/patch-linux-window-ui.test.js`
- Launcher window-reopen integration test is present, but was skipped locally because the environment does not provide `pidfd` support.
- Full CI has not been run locally.

## Checklist

- [x] This does not fix upstream DMG drift.
- [x] I reviewed the final diff with my coding agent, addressed the findings, and reran the relevant targeted tests.


ilysenko (changes requested):
Thanks for working on the AppImage remount case. The stable `APPIMAGE` identity and explicit `--show` action are useful building blocks, but the current implementation still has several merge blockers.

1. The new socket-first path runs before resident identity, cross-install, and webview health checks. Any socket returning `ok` can swallow a cold launch, including a stale/unowned listener or another installation with the same app id. This already breaks the focused warm-start recovery test: the launcher exits before Electron starts. It also bypasses `recover_unhealthy_running_app`.

2. Cross-mount discovery and recovery use different identity rules. `pid_matches_running_app` accepts a resident through the stable `APPIMAGE` path, but `terminate_stale_electron_with_pidfd` still requires the executable from the current temporary mount. If the socket is unavailable and the webview is unhealthy, the resident is detected but cannot be recovered safely.

3. The serialized action callbacks no longer preserve the previous fallback-on-error behavior. When window creation, navigation, focus, or `--show` rejects, the error is reported but the upstream focus fallback is not invoked.

4. The required regression coverage is currently broken:
   - the reopen harness checks `SECOND_LOG`, while the launcher writes the message to `APP_LOG`;
   - it invokes `start.sh` without `--show` but expects the socket payload to be `["--show"]`;
   - the socket-first exit prevents the mutation test from reaching its mutation;
   - the generated-handler smoke assertion still expects the old handler shape.

A simpler and safer implementation would reuse the existing verified warm-start path instead of adding a second IPC path:

- Remove the unconditional pre-reconciliation socket handoff.
- Keep the existing sequence: detect the resident, verify identity and webview health, perform recovery when needed, and only then use the launch-action socket.
- Preserve the exact executable-path check as the fast path for native packages. Only read `APPIMAGE` and additional `/proc/<pid>/environ` data when the executable path differs because of an AppImage remount.
- Use one resident-identity contract for discovery, active-state checks, cross-install protection, IPC handoff, and pidfd termination.
- Keep the AppImage desktop `--show` action, but retain the activation queue only if a behavioral test demonstrates the concurrent-action race. If the queue remains, rejected actions must still run the previous fallback.

This should also avoid adding startup latency. The current early socket attempt can start Python and wait for up to one second before falling back when a stale or unresponsive socket exists. The current shared matcher also adds environment reads to the normal fixed-path case. Both costs can be avoided with the fast-path/fallback structure above.

Please add focused behavioral coverage for:

- an unowned or stale socket during the first cold launch;
- a different AppImage file using the same app id;
- the same AppImage file mounted at a different `APPDIR`;
- a remounted resident with an unhealthy webview or missing socket;
- the production-equivalent `AppRun --show` path;
- overlapping activation requests and fallback after a rejected action.

Please also verify the complete change against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md). Before pushing the next update, run the relevant tests locally and perform an independent review of the complete diff with a local LLM. After updating the PR, wait until every required CI check has completed successfully before requesting maintainer review or merge.

PR #1184, opened 2026-07-29
Keep current feature composition idempotent on second pass

PinguuSS:
## Summary

- keep current core and enabled Linux feature descriptors idempotent when a fully composed ASAR is processed again
- fail candidate acceptance when a grouped Chrome runtime write cannot prove that rollback restored the original bytes
- keep optional `frameless-titlebar` composition knowledge inside the feature, using runner-validated descriptor metadata and stable core completion markers as the generic seam
- reject partial titlebar, safe-area, external-open, generated-settings, and Chrome completion states without accepting damaged bytes
- keep the new Chrome, generated-settings, and cross-feature composition coverage in owner-focused suites
- keep the 26.727 comma-separated open-target declaration scoped to its exact `fileManager` binding so the first pass does not rewrite the preceding Emacs target and the second pass remains byte-identical

## Invariant

A second full descriptor pass over a complete composed app applies zero descriptors and preserves every asset byte-for-byte. Mixed, damaged, or unverified rollback states remain byte-identical when possible and otherwise fail candidate acceptance.

## Base and upstream identity

- Merge base and current `origin/main`: `689fc5b138c2e9b33a64e4b6d3d66410f173930f`
- Head: `eae84f0b0aecf422b0cd289743b75197527ca343`
- ChatGPT Desktop: `26.727.40816`
- Electron: `42.3.0`
- DMG SHA-256: `fb93a239c811c7639cf45a90ff36c262fa0290640140cd12da3fdc60b62255ae`
- DMG size: `609303023`

The original 16 commits were rebased without conflicts. `git range-diff` reports all 16 as equivalent; `eae84f0` is the only new commit and is limited to the 26.727 open-target declarator boundary plus its regression test.

## Current-DMG evidence

The final side-by-side build completed successfully with these ten enabled features:

- appshots
- codex-wrapper-updater
- frameless-titlebar
- open-target-discovery
- persistent-status-panel
- record-and-replay
- remote-control-ui
- remote-mobile-control
- ui-tweaks
- x11-ewmh-computer-use

The first 26.727 probe exposed a real second-pass mutation: open-target discovery rewrote the preceding Emacs Linux property on pass one and the intended file-manager property on pass two. The final head scopes the declaration block to the exact `fileManager` binding.

On the rebuilt final candidate, a full second pass reported:

- `129 already-applied`
- `0 applied`
- `5 skipped-optional`
- zero integrity failures
- tree hash before: `c19786504ac3070531a8b293ad893563f35fc32c6bb7170fe3902f8718bcf4f7`
- tree hash after: `c19786504ac3070531a8b293ad893563f35fc32c6bb7170fe3902f8718bcf4f7`

The Emacs target remained unchanged and the intended file-manager target contained the Linux discovery implementation after both passes.

Candidate acceptance correctly returned `rejected` because three enabled opt-in descriptors have independent current-DMG drift:

- `feature:ui-tweaks:appearance-dock-icon-main-process`
- `feature:record-and-replay:record-replay-dictation-transcript`
- `feature:record-and-replay:record-replay-global-dictation-transcript`

The decision also records two non-blocking core warnings: `linux-bundled-plugin-copy-permissions` and `linux-computer-use-ui-availability`. This PR does not retarget those unrelated current-DMG surfaces or weaken enabled-feature acceptance.

No candidate was promoted or installed.

## Validation

- focused and surrounding Node suites: 588 passed, 0 failed
- `cargo test -p codex-update-manager`: 334 passed
- `bash tests/scripts_smoke.sh`: passed
- second-pass patch report validation: passed for the upstream-build profile
- `git diff --check origin/main...HEAD`: passed
- final `git range-diff`: original 16 commits equivalent, one scoped 26.727 follow-up commit

## Review follow-up

The maintainer completion-state findings and later independent-audit findings remain addressed:

- titlebar completion validates the helper, primary branch, zoom owner/consumer, and native-theme sync owner/consumer
- safe-area completion validates all insets, slot/padding consumers, browser gate, and Linux chrome mapping
- external-open completion validates both current core-owned targets
- feature composition authorization is derived only from an active descriptor plan with a valid owner, matching phase, later order, and exactly one delegate
- transactional rollback integrity is an engine-level `failed-integrity` blocker independent of descriptor policy
- generated Linux Settings validates all four rendered toggles and the build-info panel consumer
- report-level mutations cover complete-owner deletion, restored consumers, unauthorized/inactive delegates, enabled-feature rejection, and byte preservation

## Scope

No generated app, package, updater state, or daily-driver installation is committed or modified. The three independent enabled-feature drifts listed above remain out of scope and visible in the candidate-acceptance result.


ilysenko (changes requested):
Thanks for the work on this. Centralizing the Chrome multi-asset operation and separating its transactional write/rollback logic is a good architectural direction. The blockers are the new completion-marker paths used by the titlebar and safe-area patches.

`applyLinuxNativeTitlebarPatch()` currently treats one marker plus the helper function as proof that the required patch is complete. On the current PR head, removing the patched Linux zoom consumer while preserving the marker produces no warning: the required `linux-native-titlebar` descriptor is reported as `already-applied`, and `criticalFailuresFromReport()` remains empty.

`applyLinuxWindowControlsSafeAreaPatch()` has the same underlying problem: one marker causes an immediate return without verifying the inset, slot property, and padding consumers. Marker presence proves only that a patch ran previously; it does not prove that the current bytes still satisfy its contract.

Please add report-level regressions that preserve each marker while damaging a required consumer, and verify that these states cannot be reported as successfully applied.

A simpler architecture would be to extend the existing descriptor runner into the single source of truth for the complete patch plan, instead of adding separate completion protocols inside individual patches. Most of the required infrastructure already exists in the runner, descriptors, feature registry, patch report, and candidate acceptance logic.

The global patch plan could use:

- the current upstream identity: app version, DMG SHA-256, and Electron version;
- the Linux target and enabled feature set;
- a fingerprint of the exact registered descriptor plan;
- final hashes of every modified asset.

After a successful full pass, the runner could store that information in the patch report or a dedicated patch-state manifest. On a later invocation:

- if the upstream identity, patch-plan fingerprint, enabled features, and all final asset hashes match, skip the complete plan as already applied;
- if state exists but any final hash differs, treat the candidate as modified or damaged and rebuild from the fresh DMG instead of guessing which patches should be reapplied;
- if no state exists, require a fresh current-DMG input, run one ordered transactional pass, validate the final result, and then write the state.

The DMG version should be used for identity and diagnostics, not for compatibility branches. This repository supports only the latest upstream DMG, so the global patcher should not retain old-version patch paths.

Individual descriptors should still own their current-DMG transforms and validation rules, but they should not independently decide that a fully composed app is valid solely because a local marker remains. Enabled features would participate automatically through the existing feature descriptor registry and would be included in the global plan fingerprint.

If full-plan re-entry is not required by a real production workflow, an even simpler contract would be:

`fresh current DMG -> one ordered descriptor pass -> final validation -> candidate acceptance`

Per-patch idempotency can remain, while mixed or previously composed input is rejected rather than supported through increasingly complex marker state machines. The Chrome transactional write and verified rollback protection should remain because it protects a real first-pass failure path.

Please verify the update against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant tests locally, run an independent complete-diff review with a local LLM, and wait for every required CI check to complete successfully before requesting another maintainer review or merge.

Yo-DDV (changes requested):
Thanks for the work here. One additional blocking false-completeness case remains in the external-open path, separate from the titlebar and safe-area findings already raised.

`hasCompleteLinuxExternalOpenEnvPatch()` validates only the target markers that are still present; it does not prove that every current core-owned target still has a marker. Starting from a fully patched source with two external-open targets, restoring one entire marker-plus-wrapper expression to a plain `require("electron")` while leaving the other target and helper intact still reports the patch complete. The descriptor can therefore return `already-applied` for a partially restored core asset.

The existing regression removes a wrapper while preserving its marker, so it does not cover disappearance of a complete target. Please validate the expected semantic owners or otherwise prove that every current target is present and wrapped, then add a report-level mutation test that removes one full marker-plus-wrapper target and confirms the state cannot be accepted as successfully applied.

@ilysenko, this is independent of the two marker-completeness blockers in the current maintainer review. It should remain blocking unless the proposed global patch-plan validation supersedes the local completion check and covers this restoration case.

PinguuSS:
@ilysenko I addressed the completion-state blockers and the follow-up audit findings in e5b4da0.

- Titlebar completion now validates the helper default, primary and zoom consumers, and the native-theme sync path; partial first-pass mutations are atomic.
- Safe-area completion validates every inset owner plus the slot, padding, and header consumers, including non-numeric restored owners.
- External-open completion validates both current core-owned targets.
- Feature composition is delegated through descriptor metadata and a runner-derived owner allowlist; core no longer contains frameless feature knowledge.
- Transactional rollback integrity is now a generic engine-level failed-integrity blocker, independent of descriptor policy.
- Report-level regressions cover damaged native and delegated states, unauthorized delegates, enabled-feature rejection, and byte preservation.

Validation:

- All required GitHub checks passed: Rust and Smoke Tests, Debian, RPM, Pacman, Nix, and the current upstream DMG build.
- Local Node suite: 1,354 passed, 0 failed; one host-inotify-limited watcher test was skipped.
- Local Rust suite: 334 passed.
- Local scripts_smoke: passed.
- Current-DMG second pass with 10 enabled features: 134 already-applied, only linux-bundled-plugin-copy-permissions skipped-optional, zero integrity findings, and identical tree hash 2b9e660c0115faa93d71f0e11151c5504a2c51aa9cedc0fb0a312f293ba1158d.
- Independent complete-diff correctness/security and maintainability reviews found zero P0-P2 blockers.

The PR is ready for another maintainer review.

PinguuSS:
Updated this PR to `origin/main@689fc5b` and pushed `eae84f0`.

- The original 16 commits rebased without conflicts and remain equivalent in `git range-diff`.
- The first 26.727 second-pass probe exposed a real current-bundle issue in `open-target-discovery`: the comma-separated `fileManager` declarator included the preceding Emacs target, so pass one rewrote Emacs and pass two rewrote the intended target.
- `eae84f0` scopes the block to the exact binding while keeping helper insertion at statement scope, with a 26.727 regression fixture.
- Final real-candidate second pass with 10 enabled features: 129 `already-applied`, 0 `applied`, zero integrity failures, and identical tree hash `c19786504ac3070531a8b293ad893563f35fc32c6bb7170fe3902f8718bcf4f7` before/after.
- Local validation: 588 Node tests passed, updater Rust tests 334 passed, and `tests/scripts_smoke.sh` passed.
- All six required GitHub checks passed on the new head.

The local candidate build succeeded, but acceptance correctly remained `rejected` for three independent enabled-feature drifts on 26.727 (`ui-tweaks` Dock icon and two Record & Replay dictation descriptors). This PR does not retarget or suppress those unrelated blockers; the updated PR body records them explicitly.

ilysenko (approved):
Final maintainer review complete on 063f042d. All blocking completion, composition, and integrity findings are fixed; independent full-diff reviews are clean and all required checks pass.

PR #1197, opened 2026-08-01
Fix Chrome plugin installation cache collision

ilysenko:
## Summary
- keep the Linux Chrome runtime fallback outside the app-managed plugin cache so plugin installation cannot replace the files used by the launcher
- support the current upstream `extension-ids.json` metadata shape
- harden the official installed Chrome plugin cache before native-host registration
- make the generated native-host wrapper prefer the trusted official app-server plugin host so its executable identity matches the v2 registry entry, with the Linux runtime copy retained as a safe fallback
- add regression coverage for cache replacement, current extension metadata, permissions, and registered-host identity selection

## Tests/checks run
- `node --test scripts/patches/impl/chrome-plugin.test.js`
- `TMPDIR=/tmp node --test scripts/patch-linux-window-ui.test.js`
- `tests/scripts_smoke.sh` (using BusyBox `stat` first on PATH because the host uutils `stat` hangs)
- `TMPDIR=/tmp ./install.sh ./Codex.dmg`

## Notes / risks
- The original install failure was caused by the launcher runtime sharing the cache subtree replaced by app-server during plugin install.
- The later Chrome connection failure had two parts: writable cache-directory permissions rejected by the native host, followed by the launcher executing the fallback host while the app-server v2 registry referenced the official cached host. The wrapper now selects that registered official host only after validating and hardening its path; otherwise it falls back safely.
- Chrome integration remains optional and disabled by default.
- Draft only; do not merge yet.


ilysenko:

ilysenko:

PR #1199, opened 2026-08-01
fix(chrome): explain untrusted manifest paths

diegomarzaa:
## Summary

The Chrome native host validates every app-server manifest path, but it previously collapsed all validation failures into `missing required path file`. That message hid which manifest field failed and suggested a missing file when the path could exist with unsafe permissions.

This change preserves the existing trust checks and carries the manifest field name through file, directory, and parent-chain validation. The error now identifies the field and tells the user to check ownership, executability, and group/world-write permissions.

The regression test covers a group-writable parent for `codexCliPath`.

## Validation

- `git diff --check`
- `env TMPDIR=/tmp cargo test -p codex-computer-use-linux --bin codex-chrome-extension-host chrome_runtime` (34 passed)

### DIFF
--- a/computer-use-linux/src/chrome_runtime.rs
+++ b/computer-use-linux/src/chrome_runtime.rs
@@ -1244,19 +1244,20 @@ fn validate_runtime_entry_for_host(
             "Matching manifest entry is malformed",
         ));
     }
-    validate_owned_dir(&entry.paths.codex_home, true)?;
-    validate_owned_dir(&entry.paths.resources_path, false)?;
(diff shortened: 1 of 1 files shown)


ilysenko (changes requested):
Thanks for improving the manifest-path diagnostics. One blocker remains: the shared error says paths may be owned by the user or root, but `codexHome` is validated with `require_user_owner=true` and rejects root ownership. Please make the ownership guidance accurate for each validation policy and add coverage for `codexHome` and allowed sticky root-owned ancestors.

PR #1205, opened 2026-08-02
feat(linux): support external app-server attachment

kortylokai-web:
## Summary

- Add a disabled-by-default `external-app-server-attachment` Linux feature so Codex Desktop can attach to an existing, same-user local Codex app-server without starting, stopping, locking, reclaiming, or unlinking the external authority.
- Validate the attachment descriptor, its parent directory, the Unix socket endpoint, ownership, permissions, file type, and replacement races. After attachment is selected, unsafe or stale state fails closed instead of falling back to Desktop-managed app-server startup.
- Extend the generic Linux-feature framework with manifest-declared capabilities and conflicts, deterministic capability aggregation in build info, and an early, side-effect-free launcher `--print-build-info` surface. The new feature conflicts with `shared-app-server-socket`; that existing feature remains byte-identical to `origin/main`.
- Keep the default build and behavior unchanged. User-visible behavior changes only when `external-app-server-attachment` is explicitly enabled.
- Source-of-truth changes are limited to `linux-features/external-app-server-attachment/`, `scripts/lib/linux-features.js`, `scripts/lib/build-info.js`, `launcher/start.sh.template`, adjacent tests, and `docs/linux-features-architecture.md`. No generated output or package-format-specific payload is changed.
- No linked issue: a current upstream search found no open issue or pull request covering external app-server attachment. Known limitations are intentional: one trusted same-user local Unix socket descriptor is supported, and the feature is `required-upstream`, so enabled-feature patch drift blocks candidate promotion. No follow-up work is intentionally left out.

## Validation

- Validation environment: `TMPDIR=/tmp`; `CODEX_CLI_PATH` pointed to a validated `codex-cli 0.146.0` binary.
- `node --test linux-features/shared-app-server-socket/test.js linux-features/external-app-server-attachment/test.js scripts/lib/linux-features.test.js scripts/patch-linux-window-ui.test.js`: 540 passed, 0 failed, 0 skipped, including both real-Codex socket integrations.
- `./scripts/ci-local.sh pr`: exit 0. The core Node matrix reported 1,585 passed, 0 failed, and 3 skipped; Rust suites and Debian, RPM, and pacman package gates passed.
- Live acceptance against ChatGPT Desktop `26.727.51351` and Codex CLI `0.146.0` using a standalone same-user app-server: Desktop attached through `codex app-server proxy`, observed the same task universe, reflected live message and goal updates, reconnected after Desktop quit without restarting the external authority, fell back normally without a descriptor, and failed before Electron startup when the configured socket was unavailable.
- `git diff --check origin/main...HEAD`: passed.
- `git diff --exit-code origin/main -- linux-features/shared-app-server-socket` and tree-object comparison: passed; both trees are `f0622b4b55062293dee59a78be149717bb6aeb46`.
- Maximum-reasoning final diff review: 0 Critical, 0 Important, and 0 Minor findings. The reviewer reran the focused matrix at 540 passed, 0 failed, and 0 skipped.
- Environment limitation observed during review: a repository-local long `TMPDIR` exceeded the Unix-domain socket path limit in the unchanged shared-feature integration test. Re-running with `TMPDIR=/tmp` passed both real-Codex integrations; this is a test-environment path-length constraint, not a product regression.

## Checklist


kortylokai-web:
The `Rust and Smoke Tests` failure appears unrelated to this PR.

`killed_npm_supervisor_cleans_descendants_before_lock_release` timed out waiting for `cli-preflight` to exit. This PR does not modify `updater/`, `Cargo.toml`, or `Cargo.lock`, and the preceding updater unit suite passed 316/316.

Running the exact test locally consistently fails at a different process-lifecycle boundary: the npm process group disappears before the liveness assertion. This suggests the existing test is timing-sensitive rather than exposing a regression from this PR.

I don’t have permission to re-run the failed job. Could a maintainer please re-run it?

Failed run: https://github.com/ilysenko/codex-desktop-linux/actions/runs/30770278887/job/91556136249

ilysenko (changes requested):
Thank you very much for the substantial amount of work that went into this PR. The implementation is clearly thoughtful, and we especially appreciate the care taken around descriptor validation, lifecycle ownership, and test coverage.

After reviewing both the technical implementation and the long-term product fit, we do not currently see enough evidence that external app-server attachment is a broadly needed repository feature that the project should commit to supporting. It serves a specialized workflow built around an externally supervised app-server, and even as a disabled-by-default feature it would add a significant permanent maintenance surface: more than 2,000 lines, a current-DMG-sensitive ASAR patch, launcher changes, and new generic build-info/capability APIs.

For that reason, we believe the best home for this integration is `linux-features/local/external-app-server-attachment/`, where it can remain available for your setup without creating an ongoing compatibility and support obligation for every future upstream release. This is not a criticism of the quality of the work; it is primarily a scope and ownership decision.

There are also several technical blockers in the current version:

1. A present but invalid descriptor does not actually fail closed. The reader exits nonzero, but the generic launcher logs the launcher-hook failure and continues without `ATTACH_ONLY`, allowing Desktop to fall back to its ordinary managed app-server. That contradicts the feature’s main safety guarantee.
2. The transport matcher accepts a no-`hasConnected` lifecycle that is not present in the current 26.727.51351 DMG. Keeping this additional upstream shape conflicts with the repository’s current-DMG-only policy.
3. The generic `linuxCapabilities`, build-info, and `--print-build-info` changes are independently meaningful core APIs and are separable from the attachment transport itself. They should not be introduced as part of this specialized integration.
4. Rewriting the socket as `/proc/<pid>/fd/<fd>/<basename>` can make an otherwise valid Unix socket path exceed the AF_UNIX pathname limit.

Given the product-fit concern, we do not want to ask you to spend more time resolving all of these points for a repository merge unless broader demand and ownership are established first. We would be grateful if you kept the integration as a local feature for the workflow that needs it. Thank you again for the careful contribution and for understanding the project-maintenance tradeoff.

kortylokai-web:
@ilysenko, sorry for the late reply, and thanks for the detailed review. I understand the concern about adding and maintaining this integration upstream.

I have addressed the three actionable implementation issues in separate follow-up commits on the source branch:

- a present malformed, unsafe, unreadable, or unsupported descriptor now fails closed before ordinary transport selection;
- the transport matcher now requires the current `this.hasConnected=!0` lifecycle shape;
- the proxy uses the validated socket parent as its working directory and receives only the socket basename, avoiding Unix socket path-length expansion while preserving the validated parent identity.

I retained the bounded capability/build-info surface because [Codex Session Control](https://github.com/agentlehub/codex-session-control) uses it to verify Desktop compatibility before enabling attachment. Codex Session Control is an open-source MCP server for observing and controlling live Codex sessions, and I maintain it under the `agentlehub` GitHub organization. If you would like to see it in action, you are welcome to try it; the README includes the setup instructions.

The fixes are available on the [original source branch](https://github.com/kortylokai-web/codex-desktop-linux/tree/feat/external-app-server-attachment). I will maintain this integration in the [`agentlehub/codex-desktop-linux`](https://github.com/agentlehub/codex-desktop-linux) fork so Codex Session Control users have a supported Desktop build. If Codex Session Control gains broader adoption and there is clearer evidence of user demand for this integration, I may ask whether you would be open to revisiting this decision.

PR #1206, opened 2026-08-03
fix(updater): serialize native installs with app launches

MatsumotoMorami:
## Summary

Fix an updater race that could replace hashed webview assets underneath an Electron renderer reopened while native package authorization was still pending.

### Root cause

The affected renderer had loaded the old `app-initial-DRyZ1Lin.js` bundle and later lazy-loaded:

`model-picker-power-slider-impl-CrCCLhyI.js`

While the app was running, the native package upgrade replaced `/opt/codex-desktop/content/webview` with a newer asset set containing:

`model-picker-power-slider-impl-DmqcN1z7.js`

The old hashed chunk no longer existed, so opening the model picker—observed while switching from GPT-5.3 Spark—triggered the full-screen error boundary with:

`Failed to fetch dynamically imported module`

The webview server already sent `Cache-Control: no-store`; this was an install/launch race rather than a browser-cache issue.

### Changes

- Acquire the same `launcher.lock` used by the desktop launcher before beginning a native package installation.
- Hold that lock through polkit authorization and package replacement, preventing a normal desktop launch from publishing a new Electron process during installation.
- Recheck Electron liveness after acquiring the launcher lock.
- Return the update to `WaitingForAppExit` if a launch wins the race or the lock cannot be acquired safely.
- Pass the configured application executable to the privileged Debian, RPM, and pacman helpers.
- Check application liveness before package stabilization and again immediately before invoking the package manager.
- Return a machine-recognizable `CODEX_UPDATE_APP_RUNNING` marker when the privileged guard detects a reopened app.
- Continue recognizing an existing process when `/proc/<pid>/exe` ends in ` (deleted)` after its installed executable has been replaced.
- Document the new install/launch serialization behavior and add the changelog entry.

## Validation

  - 324 unit tests passed
  - 5 integration tests passed
  - Passed in an isolated Docker PID namespace
  - Covered Debian, RPM, pacman, AppImage, warm-start, window-reopen, launcher-lock, and packaging paths
- Ubuntu 24.04 VM acceptance:
  - A running real Electron app caused the privileged helper to return `CODEX_UPDATE_APP_RUNNING` in approximately 30 ms.
  - The running app remained alive and its executable checksum was unchanged.
  - A live process whose executable appeared as ` (deleted)` in `/proc/<pid>/exe` was still detected.
  - The same path produced no false positive after the process exited.
  - While `launcher.lock` was held, the real launcher timed out safely after approximately 5.38 seconds and created no Electron process.
  - After releasing the lock, the same launcher started Electron successfully.
- On the originally affected runtime, both old and new model-picker chunks returned HTTP 200 under the live recovery asset root, with no new matching dynamic-import errors afterward.

`cargo clippy` was not available in the host Rust installation. GitHub CI should run the repository-wide clippy job.

## Checklist

- [x] If this fixes upstream drift, it targets only the latest `CODEX.DMG` and removes obsolete fallback code and tests from the affected area. Not applicable: this fixes an updater race, not upstream drift.

### DIFF
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -31,7 +31,6 @@ const PROMPT_INSTALL_CLI_CANCELLED_EXIT_CODE: i32 = 10;
 const PROMPT_INSTALL_CLI_NO_BACKEND_EXIT_CODE: i32 = 11;
 // Nonzero so `Restart=on-failure` relaunches the daemon on the new binary.
 const BINARY_REPLACED_RESTART_EXIT_CODE: i32 = 12;
-const LAUNCHER_INSTALL_GATE_WAIT_SECONDS: u64 = 10;
 const POLKIT_AUTH_AGENT_PROCESS_TOKENS: &[&str] = &[
     "budgie-polkit",
     "cinnamon-polkit",
@@ -1430,7 +1429,12 @@ async fn reconcile_pending_install(
 
             if state.auto_install_on_app_exit && liveness::is_app_running(config)? {
                 if !graphical_polkit_auth_agent_is_likely_available() {
-                    defer_install_for_manual_auth(state, paths, &package_path)?;
+                    defer_install_for_manual_auth(
+                        state,
+                        paths,
+                        &package_path,
+                        &config.app_executable_path,
+                    )?;
                     maybe_notify_manual_install_required(state, paths, config.notifications)?;
                     return Ok(());
                 }
@@ -1473,7 +1477,12 @@ async fn reconcile_pending_install(
 
             if liveness::is_app_running(config)? {
                 if !graphical_polkit_auth_agent_is_likely_available() {
-                    defer_install_for_manual_auth(state, paths, &package_path)?;
+                    defer_install_for_manual_auth(
+                        state,
+                        paths,
+                        &package_path,
+                        &config.app_executable_path,
+                    )?;
                     maybe_notify_manual_install_required(state, paths, config.notifications)?;
                     return Ok(());
                 }
@@ -1494,7 +1503,12 @@ async fn reconcile_pending_install(
             }
 
             if !graphical_polkit_auth_agent_is_likely_available() {
-                defer_install_for_manual_auth(state, paths, &package_path)?;
+                defer_install_for_manual_auth(
+                    state,
+                    paths,
+                    &package_path,
+                    &config.app_executable_path,
+                )?;
                 maybe_notify_manual_install_required(state, paths, config.notifications)?;
                 return Ok(());
             }
@@ -1640,9 +1654,14 @@ async fn run_install_ready_locked(
 
     if liveness::is_app_running(config)? {
         if !graphical_polkit_auth_agent_is_likely_available() {
-            defer_install_for_manual_auth(state, paths, &package_path)?;
+            defer_install_for_manual_auth(
+                state,
+                paths,
+                &package_path,
+                &config.app_executable_path,
+            )?;
             maybe_send_manual_install_required_notification(config.notifications);
-            print_manual_install_required(&package_path);
+            print_manual_install_required(&package_path, &config.app_executable_path)?;
             return Ok(());
         }
         clear_install_auth_required_event(state, paths)?;
@@ -1659,9 +1678,9 @@ async fn run_install_ready_locked(
     clear_install_auth_required_event(state, paths)?;
     state.waiting_for_app_exit_auto_install = false;
     if !graphical_polkit_auth_agent_is_likely_available() {
-        defer_install_for_manual_auth(state, paths, &package_path)?;
+        defer_install_for_manual_auth(state, paths, &package_path, &config.app_executable_path)?;
         maybe_send_manual_install_required_notification(config.notifications);
-        print_manual_install_required(&package_path);
+        print_manual_install_required(&package_path, &config.app_executable_path)?;
         return Ok(());
     }
     trigger_install(state, paths, config, &package_path, config.notifications).await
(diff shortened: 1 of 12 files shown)


updater/src/app.rs
```diff
@@ -1980,7 +2016,7 @@ async fn trigger_install(
     );
 
     let current_exe = std::env::current_exe().context("Failed to resolve updater binary path")?;
-    let output = install::pkexec_command(&current_exe, package_path)
+    let output = install::pkexec_command(&current_exe, package_path, &config.app_executable_path)
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

**[P1] Keep the launcher gate alive through Debian package replacement.** For daemon-driven `.deb` updates, this process does not survive the transaction: `packaging/linux/codex-update-manager.prerm:22-25` unconditionally runs `systemctl --user stop codex-update-manager.service`, including on upgrade, and Debian runs the old `prerm` before unpacking the new files. That terminates this user-service process and drops `_launcher_lock` while the root package transaction is still outside the user manager's control, so a launcher can acquire the lock during unpack/replacement. Please either keep the lock in a process the maintainer script cannot stop or avoid stopping the service on upgrade until replacement is complete.

updater/src/liveness.rs
```diff
@@ -3,24 +3,111 @@
...
     Ok(config::resolve_app_state_dir()?.join("app.pid"))
 }
 
+/// Tries to acquire the same lock the launcher holds until Electron is spawned
+/// and its PID is published. `None` means a launch is currently in progress.
+pub fn try_acquire_launcher_lock() -> Result<Option<LauncherLock>> {
+    let state_dir = config::resolve_app_state_dir()?;
+    fs::create_dir_all(&state_dir)
+        .with_context(|| format!("Failed to create {}", state_dir.display()))?;
+    let lock_path = state_dir.join("launcher.lock");
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

**[P1] Use an app-wide gate that also covers multi-instance launches.** The daemon normally resolves this to the base app state directory, but `configure_multi_launch_instance` rewrites the launcher's `APP_STATE_DIR` to `$base_state_dir/instances/$CODEX_LINUX_INSTANCE_ID` (`launcher/start.sh.template:257-270`), so `--new-instance` locks a different `launcher.lock`. Such a launch therefore does not contend with this install gate and can spawn after the privileged helper's final liveness check but before the package manager replaces files. Every launcher instance needs to acquire the same installation gate.

updater/src/wrapper_apply.rs
```diff
@@ -404,9 +404,13 @@ async fn apply_packaged(
     .context("wrapper package rebuild failed")?;
 
     let current_exe = std::env::current_exe().context("Failed to resolve updater binary path")?;
-    let output = install::pkexec_command(&current_exe, &artifacts.package_path)
-        .output()
-        .context("Failed to launch pkexec for wrapper update installation")?;
+    let output = install::pkexec_command(
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

**[P2] Serialize packaged wrapper installs as well.** This native-package path now uses the guarded helper but never acquires `LauncherLock`. In particular, the wrapper feature invokes it from its after-exit hook (`linux-features/codex-wrapper-updater/apply-pending.sh:168`), while the launcher released its lock immediately after spawning Electron (`launcher/start.sh.template:4422`) and does not reacquire it for after-exit hooks. A concurrent user launch can therefore pass between the helper's final process check and package replacement. Route this path through the same install gate (or have the after-exit path hold the app-wide gate through `pkexec`).

avifenesh (commented):

---

The new serialization has three gaps where a launcher can still overlap native package replacement.

ilysenko (changes requested):
Thanks for tackling this updater race. Please address these blockers before merge:

- Keep one install-wide launch gate held for the full native package transaction, including Debian upgrades where the old `prerm` stops `codex-update-manager.service`. Add coverage proving the gate remains held through package replacement.
- Make every launcher namespace, including `--new-instance` and alternate app-state roots, contend on the same install-wide gate. Add a cross-namespace regression test.
- Bump the updater patch version in `updater/Cargo.toml` and `Cargo.lock` as required by `CONTRIBUTING.md`.

MatsumotoMorami:
@ilysenko, I’ve addressed the requested blockers and rebased the branch onto `df725ad`. Could you please re-review?

## Review blockers addressed

- Replaced the app-state-specific `launcher.lock` with one stable, per-user gate keyed by the canonical app executable. Normal launches, `--new-instance`, alternate app IDs, alternate instance IDs, and alternate `XDG_STATE_HOME` roots now contend on the same gate.
- Passed the exact locked file descriptor to `pkexec` through stdin, allowing the privileged package transaction and its package-manager children to retain the lease. If an old Debian `prerm` stops `codex-update-manager.service`, the package transaction continues holding the gate until replacement completes.
- Routed packaged wrapper updates, rollback installs, and terminal/manual native-package installs through the same gate.
- Bumped `codex-update-manager` from `0.11.0` to `0.11.1`.

## Regression coverage

- Added cross-namespace coverage proving alternate app-state roots and `--new-instance` resolve to the same gate.
- Added a process-level regression that kills the updater parent while a simulated privileged package transaction continues, proving launches remain blocked until package replacement finishes.
- Added inherited-lease lifetime and shell `flock` interoperability coverage.
- Made post-transaction release assertions use the same bounded-wait semantics as the production updater.

## Validation

- [GitHub CI run](https://github.com/ilysenko/codex-desktop-linux/actions/runs/31111891926): Rust and Smoke Tests, Debian, RPM, Pacman, and Nix all passed.
- `cargo test -p codex-update-manager`: 339 unit tests and 5 integration tests passed.
- The install-gate test group passed 100 consecutive local runs.

ilysenko (changes requested):
Thank you for all the work you have put into this PR. We genuinely appreciate the time, testing, and effort involved.

We are going to close the PR for now and ask that you pause work on it. The current version still has several blockers:

- The native package transaction changes a system-wide installation, but the launch gate is per-user, so another user can still start the app during replacement.
- The launcher releases the gate before the child process is confirmed to have execed Electron, leaving the original launch/install race open.
- AppImage launches derive the gate from a transient mount path, so concurrent launches of the same AppImage use different locks.
- The wrapper prelaunch path attempts to reacquire a gate already held by its launcher and therefore times out.

If you decide to return to this later, please prepare a new PR only after the complete solution is ready. Before opening it, verify the change against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run all relevant tests locally, and have your local coding model perform several independent review passes over the complete diff. Please address every finding from those passes and wait for all required CI checks to finish successfully before requesting maintainer review.

This should help ensure that the next maintainer review is focused on final verification rather than discovering the same architectural and concurrency issues again.

Thank you again for the contribution and for understanding.

PR #1208, opened 2026-08-03
fix(updater): use scope-safe module imports

### LINKED ISSUE #1207: Linux updater bridge captures function-scoped CommonJS bindings
## Summary

The generated Linux updater bridge can capture function-scoped CommonJS variables from the patched main bundle and reference them outside their lexical scope. This makes the update-manager probe fail at runtime even though `/usr/bin/codex-update-manager` exists and works from the terminal.

## Environment

- Repository commit: `ec38ca664c7fd18fa33744216850924bd3132872`
- Upstream app version: `26.727.51351`
- Electron: `42.3.0`
- Distribution: Fedora Linux 44
- Package: `.rpm`
- Installation path: `/opt/codex-desktop`
- Installed wrapper version: `0.10.4`

## Steps to reproduce

1. Enable `codex-wrapper-updater` in `linux-features/features.json`.
2. Run `make install-native` and install the generated RPM.
3. Start the packaged app.
4. Select **Check for Updates** from the application menu.

## Actual behavior

The app displays:

```text
Automatic updates are unavailable right now.
Updater initialization skipped: Linux package update manager unavailable
```

The launcher log contains:

```text
[sparkle] Check for updates requested via menu.
[sparkle] Desktop updater unavailable; init likely skipped. reason="Linux package update manager unavailable"
```

At the same time, `codex-update-manager diagnose --json` reports `ok: true`, the user service is active, and `/usr/bin/codex-update-manager --help` exits successfully.

## Root cause

`scripts/lib/linux-update-bridge-patch.js` uses `requireName` to find identifiers assigned from `require(...)`. The regex is not scope-aware and can select a nested binding such as:

```js
function codexLinuxLaunchExternalUrl(url) {
  // ...
  let __codexChild = require(`node:child_process`).spawn(...);
}
```

The generated bridge later emits code equivalent to:

```js
__codexChild.execFile(...)
```

outside that function. The probe rejects with a `ReferenceError`, and the bridge records the update manager as unavailable. The same risk applies to nested filesystem and path bindings.

## Expected behavior

The generated bridge should use scope-safe module access and successfully invoke the packaged update manager from the current main bundle.

## Proposed fix

Generate direct CommonJS module expressions for `node:child_process`, `node:fs`, and `node:path` instead of reusing identifiers found by a scope-unaware regex. Add regression assertions ensuring that generated updater calls do not reference nested minified or injected bindings.

boommasterxd:
## Summary

- replace scope-unaware CommonJS binding discovery in the Linux updater bridge with direct module expressions for `node:child_process`, `node:fs`, and `node:path`
- prevent generated updater code from referencing function-scoped or injected minified bindings outside their lexical scope
- add regression coverage with deliberately nested module bindings matching the current patched bundle failure

This restores the packaged **Check for Updates** path when `codex-update-manager` is installed and healthy.

Closes #1207.

## Validation

- `node --check scripts/lib/linux-update-bridge-patch.js`
- focused updater bridge tests: 5/5 passed
- full `scripts/patch-linux-window-ui.test.js`: 411/411 passed
- `bash tests/scripts_smoke.sh`: packaging, installer, updater, and launcher checks passed until the warm-start recovery test; that test could not start its fixture because the daily-driver `/opt/codex-desktop` process was running with the same `codex-desktop` app ID. No running-app guard was bypassed.

## Checklist

- [x] I added or updated relevant tests, ran the validation listed above, and confirmed that required focused checks pass.

### DIFF
--- a/scripts/lib/linux-update-bridge-patch.js
+++ b/scripts/lib/linux-update-bridge-patch.js
@@ -1,11 +1,6 @@
 const fs = require("fs");
 const path = require("path");
 
-function requireName(source, moduleName) {
-  const escaped = moduleName.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
-  return source.match(new RegExp(`([A-Za-z_$][\\w$]*)=require\\([\\\`'"]${escaped}[\\\`'"]\\)`))?.[1] ?? null;
-}
-
 function buildUpdateManagerEnvSource() {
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head. The scope-safe built-in module imports fix the updater bridge regression, and the relevant CI checks are green.

PR #1209, opened 2026-08-03
Clean up orphaned shared app-server authorities

moxw:
## Summary

- stage a feature-local orphan reaper for the shared app-server socket
- run it after Electron exits and before the next cold start
- preserve live owners and unknown listeners while stopping only an exact same-user, reparented `codex app-server --listen` process tree

## Root cause

After an update, Desktop could exit while its Node wrapper and native `codex app-server` child remained alive under PID 1. The existing stale-lock recovery correctly preserved the still-listening socket, so the next Desktop launch failed with `shared app-server socket is already owned`.

## User-visible behavior

A later launch now recovers from that exact orphaned authority automatically. Ambiguous listeners, changed inodes, mismatched command lines, and live Desktop owners remain untouched.

## Source of truth

- `linux-features/shared-app-server-socket/orphan-reaper.js`
- the feature's declarative resource and runtime-hook entries
- `socket-env.sh`
- focused feature documentation and tests

## Validation

- PASS: `env -u CODEX_CLI_PATH node --test linux-features/shared-app-server-socket/test.js` (32 passed, 1 opt-in real-CLI integration skipped)
- PASS: `bash tests/scripts_smoke.sh` (all cross-format script smoke tests passed)
- PASS: `node --test scripts/patch-linux-window-ui.test.js` passes 409 tests in the sandbox; the remaining filesystem-ownership-sensitive test passes outside sandbox ownership virtualization
- LIVE: reproduced the post-update failure, verified a dead Desktop lock owner with a same-user reparented wrapper/native authority, stopped only that exact orphan, and confirmed a fresh Desktop owner/socket handled requests


ilysenko (dismissed):
Thanks for the focused fix. One destructive edge case still needs to be addressed before merge: a Unix pathname can have multiple live listener inodes after unlink/rebind. The reaper currently selects the first `/proc/net/unix` match, but later unlinks the current filesystem socket, which can belong to a replacement listener. Please bind cleanup to the authority represented by the stale lock, refuse ambiguous pathname matches, and add a regression test covering two listener inodes with the same path.

ilysenko (commented):
Thanks for addressing the authority binding and duplicate-inode case. One non-blocking follow-up remains: the launcher and after-exit reapers can run concurrently, and a delayed reaper can use cached socket state to unlink a replacement socket created by the next launch. The feature is opt-in and disabled by default, so this does not block the current merge, but please serialize the feature-local cleanup across both hooks (or use an equivalent race-safe protocol) and add a focused concurrent-hook regression test in a follow-up.

ilysenko (approved):
Reviewed current head 913899ac406af878f7ea4ff26933945ee9aaddc2. The original safety blocker is fixed; the remaining opt-in concurrency concern is documented as a non-blocking follow-up.

PR #1210, opened 2026-08-03
Retarget enabled Linux features to Codex 26.727

moxw:
## Summary

- retarget `project-group-last-updated-sort` to the current Codex 26.727 project-group sorter and call-site symbols
- refresh `shared-app-server-socket` detection for the current SSH WebSocket readiness and adapter lifecycle
- update both feature fixtures while keeping drift fail-soft and second-pass patching byte-identical

## Root cause

Codex `26.727.51351` changed the minified project sidebar symbols and inlined project recency calculation. It also inserted explicit WebSocket readiness handling and a connection-state assignment between SSH keepalive setup and adapter construction.

Both opt-in feature descriptors therefore skipped their current insertion points. Because the features were enabled locally, upstream-DMG acceptance correctly rejected candidate promotion with `enabled-feature-drift` blockers.

## User-visible behavior

Users who enable either feature can rebuild from the current upstream DMG without those descriptors being rejected as stale. Non-updated project sort modes still preserve upstream saved ordering, and unsupported future bundle shapes remain unchanged with warnings.

## Validation

- `env -u CODEX_CLI_PATH node --test linux-features/project-group-last-updated-sort/test.js linux-features/shared-app-server-socket/test.js` — 35 passed, 1 opt-in real-CLI integration skipped
- Codex DMG `26.727.51351` — both patches change their real extracted assets and produce byte-identical output on a second pass

## Scope

This PR contains only the four descriptor and fixture files for the two enabled-feature drift blockers. The separate shared app-server orphan-recovery change remains in #1209.

The full transactional native rebuild was not rerun to completion after the local handoff; validation uses the focused suites and freshly extracted real DMG assets.

### DIFF
--- a/linux-features/project-group-last-updated-sort/patch.js
+++ b/linux-features/project-group-last-updated-sort/patch.js
@@ -1,14 +1,14 @@
 "use strict";
 
 const currentGroupSorter =
-  "function h2o({groups:e,items:t,projectOrder:n}){let r=new Map(t.map(e=>[e.task.key,e.recencyAt]));return iZi(e.map((e,t)=>({group:e,index:t,recencyAt:y2o(e,r)})).sort((e,t)=>t.recencyAt-e.recencyAt||e.index-t.index).map(({group:e})=>e),n)}";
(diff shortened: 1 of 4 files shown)


ilysenko (approved):
Reviewed the current head. The scoped current-DMG retarget is acceptable for these opt-in Linux features, and the relevant checks are green.

PR #1211, opened 2026-08-03
fix(updater): preserve feature picker settings

PinguuSS:
## Summary

- preserve the effective Linux feature configuration object whenever the updater feature picker saves a selection
- replace only the `enabled` list, keeping nested feature settings and unknown local keys intact on both first save and later saves
- bump `codex-update-manager` to `0.10.5` and add regressions for both persistence paths

## Problem

The updater feature picker rewrote `linux-features.json` as only:

```json
{"enabled":["..."]}
```

That silently removed settings owned by enabled features. In the reported update, `ui-tweaks` remained enabled but its nested Dock Icon and Suggested Prompts opt-ins disappeared, so the following rebuild reported all seven related descriptors as `skipped-disabled`.

The initial fix preserved an existing per-user `linux-features.json`, but the first picker save still started from an empty object. At that point the effective configuration comes from the bundled `linux-features/features.json`, so nested settings would still be lost when the per-user file was first created.

## Invariant

Completing the picker may change the selected feature IDs, but it must not mutate configuration owned by those features or other local top-level keys. Cancelled and invalid selections retain their existing no-write behavior.

`updater/src/feature_picker.rs` remains the only behavioral owner changed. The writer now seeds from `effective_feature_config_path`: the saved per-user object when it exists, otherwise the bundled feature config. It then replaces only `enabled` and writes the result to the stable per-user path. Missing, malformed, or non-object effective configuration retains the existing fresh-object fallback.

The first-save regression also proves that the bundled source file remains unchanged and that an unknown enabled feature ID survives the selection.

## Validation

Regression oracle before the original fix:

```text
selection_preserves_existing_feature_settings_and_unknown_keys ... FAILED
left: Null
right: Object {"alpha": Object {"nested": Object {"enabled": Bool(true)}}}
```

Final checks:

- `cargo test -p codex-update-manager feature_picker::tests` (15 passed)
- `cargo test -p codex-update-manager` (336 passed across 15 suites on the host)
- isolated `codex-update-manager pick-features --json` subprocess with no per-user feature config: preserved nested Dock Icon, Suggested Prompts, metadata, and an unknown local feature ID while leaving the bundled source unchanged
- the sandboxed full Rust suite was expectedly unable to bind wiremock sockets; the same suite passed outside the sandbox

Current upstream build evidence from the original implementation commit (`480b956`):

- DMG SHA-256: `45ec006a0f3f0fa004b6fd4d6d5529979a05361f995ca2c51de3a3b04deee123`
- ETag: `0x8DEEF605E2457C7`
- Last-Modified: `Sat, 01 Aug 2026 00:03:50 GMT`
- Size: `609312441` bytes
- App: `26.727.51351`
- Electron: `42.3.0`
- acceptance: `accepted_with_warnings`, no blockers or integrity failures
- patch report: 111 applied, 18 already applied, one unrelated optional Computer Use UI descriptor skipped
- enabled `ui-tweaks`: 12 descriptors applied, including all three Dock Icon and all four Suggested Prompts descriptors

The review follow-up changes only updater config persistence, its tests, and updater package metadata; it does not alter ASAR patch descriptors or the candidate app payload.

## Scope

This does not change feature defaults, selection validation, config paths, updater state, package formats, or the behavior for malformed configuration. It does not attempt to repair unrelated optional feature drift.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- The updater feature picker now changes only the enabled feature list, preserving
+  nested feature settings and other local configuration keys across rebuilds.
 - The opt-in Dock icon tweak now targets the current upstream main-process
   bundle, restoring Linux window, tray, and desktop icon synchronization.
 - The opt-in shallow repository watcher now patches both current app bundles
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -561,7 +561,7 @@ dependencies = [
 
 [[package]]
 name = "codex-update-manager"
-version = "0.10.4"
+version = "0.10.5"
 dependencies = [
  "anyhow",
  "chrono",
--- a/updater/Cargo.toml
+++ b/updater/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "codex-update-manager"
-version = "0.10.4"
+version = "0.10.5"
 edition = "2021"
 
 [dependencies]
--- a/updater/src/feature_picker.rs
+++ b/updater/src/feature_picker.rs
@@ -475,9 +475,14 @@ fn write_feature_config(enabled: &[String]) -> Result<()> {
         std::fs::create_dir_all(dir)
             .with_context(|| format!("Failed to create {}", dir.display()))?;
     }
-    let value = serde_json::json!({ "enabled": enabled });
+    let mut object = std::fs::read_to_string(&path)
+        .ok()
+        .and_then(|content| serde_json::from_str::<serde_json::Value>(&content).ok())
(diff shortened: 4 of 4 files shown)


ilysenko (changes requested):
Thanks for the update. I found two blockers:

- On the first picker save, the per-user config does not exist yet, so the writer starts from an empty object and still drops settings from the effective bundled feature config. Please seed from `effective_feature_config_path` and add a first-save regression with no per-user file.
- Please bump the `codex-update-manager` patch version, including `Cargo.lock`, as required for updater fixes by `CONTRIBUTING.md`.

ilysenko (approved):
Reviewed the current head. Both requested changes are addressed: the first save now preserves the effective bundled feature configuration, the regression covers that path, and the updater patch version is bumped consistently. All relevant checks are green.

PR #1215, opened 2026-08-03
fix(chrome): stage legacy extension metadata

antonkulaga:
## Summary
- generate the legacy `scripts/extension-id.json` adapter from canonical `extension-ids.json` when staging the Chrome plugin
- include the generated metadata in Chrome cache freshness checks
- add a staging regression test for its contents

The Electron bundle currently reads singular `extension-id.json`, while the current Chrome plugin supplies only plural `extension-ids.json`. Without the adapter, Chrome extension synchronization fails before a browser session can attach.

## Validation
- `bash -n scripts/lib/bundled-plugins.sh launcher/start.sh.template`

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -1330,6 +1330,7 @@ sync_chrome_bundled_plugin_cache() {
             scripts/check-extension-installed.js \
             scripts/check-native-host-manifest.js \
             scripts/chrome-is-running.js \
+            scripts/extension-id.json \
             scripts/extension-ids.json \
             scripts/installed-browsers.js \
             scripts/open-chrome-window.js; do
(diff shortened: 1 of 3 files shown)


ilysenko (changes requested):
Thanks for the focused PR. One blocker: the current supported DMG 26.727.51351 does not contain any `extension-id.json` consumer. Its Electron main/app-server/webview code and the staged Chrome plugin all use `scripts/extension-ids.json`; PR #1197 intentionally migrated the launcher to that current contract. Generating and cache-tracking the singular file therefore reintroduces an obsolete compatibility shape contrary to the repository’s latest-DMG-only policy, while the added tests do not reproduce the reported synchronization failure. Please identify and cover the exact current-DMG consumer that requires the singular file, or remove this adapter and diagnose the failure against the existing plural-only flow.

PR #1218, opened 2026-08-04
fix(api-key-model-visibility): match refactored upstream model gate

agentixsoftware:
The api-key-model-visibility feature patch no longer matches the current upstream app bundle: vite split the app main webview chunk (`app-initial~app-main~*.js` -> `app-initial-*.js`) and the allowlist gate was refactored out of the catalog filter function into a per-model visibility helper with an `additionalAvailableModels` parameter. As a result the descriptor was skipped on current DMGs and API-key-authenticated custom providers had their models hidden behind the desktop Statsig allowlist (only models overlapping the allowlist, e.g. gpt-5.6-luna, appeared in the picker).

Changes:
- Descriptor pattern now matches both `app-initial~app-main~*.js` and `app-initial-*.js`.
- Patch targets the refactored helper gate, adding `&&authMethod!==\`apikey\`` to bypass the allowlist for API-key hosts (same treatment already given to non-ChatGPT hosts).
- Legacy inline-gate pattern kept as a fail-soft fallback for older bundle shapes.
- Idempotency guard (marker) and `(?=[,;?])` lookahead so we never double-patch or patch mid-expression.
- Tests updated to the current bundle shape; 8/8 pass, including hidden-model exclusion, ChatGPT/amazonBedrock behavior preservation, and compose-with-service-tier.

Verified end-to-end on Fedora 44 with a custom OpenAI-compatible provider: model picker now lists the full provider catalog.

### DIFF
--- a/linux-features/api-key-model-visibility/patch.js
+++ b/linux-features/api-key-model-visibility/patch.js
@@ -8,7 +8,34 @@ function warn(message, patchName) {
 }
 
 function applyApiKeyModelVisibilityPatch(source) {
-  const modelVisibilityPattern = new RegExp(
+  if (source.includes(PATCH_MARKER)) {
+    return source;
+  }
+
+  // Current upstream shape (refactored): the allowlist gate lives in a
+  // per-model visibility helper, e.g.
+  //   function q$r({additionalAvailableModels:e,authMethod:t,availableModels:n,model:r,useHiddenModels:i}){return e?.has(r.model)===!0||(i&&t!==`amazonBedrock`?n.has(r.model):!r.hidden)}
+  // Bypass the allowlist for API-key authenticated hosts the same way it is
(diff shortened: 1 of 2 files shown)


ilysenko (changes requested):
Thanks for updating this feature for the current upstream bundle. Please address these blockers:

- This drift fix must target only the latest `CODEX.DMG`. Remove the legacy inline matcher, the old `app-initial~app-main~*` filename path, and the legacy fixture/composition coverage.
- Constrain the helper needle to the complete current ternary allowlist expression. The current `[,;?]` boundary also modifies semicolon- and comma-drifted helpers instead of failing soft; please add byte-identical near-miss tests for those shapes.

Once addressed, please rerun the focused feature tests.

ilysenko (approved):
Re-reviewed head 8acb0159: the requested latest-DMG cleanup and exact fail-soft helper matching are implemented, focused/current-asset validation passes, and all required checks are green.

PR #1226, opened 2026-08-05
fix(remote-control): keep outbound tab visible on Linux

huaixv:
## Summary
- bypass the upstream `782640499` outbound remote-control tab hide gate on Linux only
- keep the existing remote-control settings tab wrapping and non-Linux behavior unchanged
- add a regression test for the current minified `B=!l` gate shape that hides `Control other devices`

## Testing
- `node --test --test-name-pattern='settings UX patch|outbound tab hide gate' linux-features/remote-mobile-control/test.js`
- `git diff --check`

## Notes
- `origin/main` at `ee3f570` did not include an equivalent fix when this branch was created.


ilysenko (changes requested):

- Please remove the obsolete `codexLinuxRemoteControlSettingsTabs` / `wrapRemoteControlTabs` fallback and its legacy fixtures while retargeting this drift fix. Both wrapper patterns match zero times in the current 26.730.61639 asset, and the repository’s current-DMG-only policy requires old fallback paths to be removed in the same change.
- Please emit an optional drift warning when rollout gate `782640499` is present but its current consumer needle does not match, and add a regression test for that case. The current implementation silently no-ops while the combined descriptor can still be reported as applied.

The new gate matcher itself does match the current bundle and is idempotent. Thank you!

huaixv:
Addressed the review feedback in the amended commit cec29bb.

Changes:
- removed the obsolete remote-control tabs wrapper fallback and its legacy fixtures
- kept the fix scoped to the current rollout gate consumer
- added an optional drift warning when gate 782640499 is present but the consumer needle no longer matches
- added a regression test for that drift warning

Local verification:
- `node --test --test-name-pattern='settings UX patch|outbound tab gate|SSH release handling drifts|participates in ASAR patching' linux-features/remote-mobile-control/test.js`
- `git diff --check`

ilysenko (approved):
Reviewed the updated head. Both prior blockers are resolved: obsolete tab-wrapper compatibility is removed, and outbound-gate consumer drift now reports a fail-soft warning with regression coverage. Current-DMG and CI validation are clean.

PR #1230, opened 2026-08-05
Fix Chrome plugin env compatibility and stale cache refresh

jadabreu:
## Summary

- guard every direct Browser/Chrome client `nodeRepl.env[...]` read when staging the current upstream plugin
- atomically refresh an existing same-version app-server Chrome plugin cache from the patched bundled source on cold start
- keep the app-server-owned install separate from the hardened Linux native-host/runtime cache
- add regression coverage for the current minified security-state code and stale managed-cache behavior

## Root cause

The current upstream Browser client snapshots security settings through several direct `env` reads. The Linux fallback `node_repl` runtime does not expose `env`, while the compatibility patch only guarded the older global helper read. Browser setup therefore crashed before it could connect to Chrome.

After fixing the staged client, existing installations could still load the unpatched JavaScript because app-server keys its managed plugin cache by the unchanged upstream plugin version. Installing a rebuilt native package did not invalidate that same-version cache.

## User impact

Chrome browser control can initialize with the Linux fallback runtime even when `nodeRepl.env` is absent. Native package upgrades also repair an existing stale same-version Chrome install during cold-start cache synchronization instead of requiring users to delete `~/.codex` state manually.

## Validation

- `node --test scripts/lib/browser-client-node-repl-runtime.test.js`
- real staged Browser and Chrome clients imported through the bundled Linux `node_repl` runtime
- transactional candidate rebuild accepted against the current upstream DMG
- RPM built and inspected for both the env-guard marker and managed-cache refresh logic
- `bash tests/scripts_smoke.sh` passed the relevant Chrome cache/permission probes; its final warm-start recovery check could not run because the installed desktop app was intentionally still running and was detected as a conflicting instance

The fix was also verified end-to-end against a running Chrome extension session after refreshing the stale managed cache.

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -1251,6 +1251,7 @@ sync_chrome_bundled_plugin_cache() {
     local managed_cache_plugin
     local managed_cache_client
     local managed_tmp_plugin
+    local managed_backup_plugin
     local marketplace_root
     local marketplace_plugins_dir
     local marketplace_plugin_link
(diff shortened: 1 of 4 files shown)


ilysenko (changes requested):
Thanks for the focused fix. One blocker remains:

Please make the same-version managed Chrome cache refresh failure-safe. It currently removes the live version directory before promotion; a failed move or concurrent destination recreation can leave the cache missing or incorrectly nested and abort cold start under `set -euo pipefail`. Preserve or restore the previous cache until replacement succeeds, and add regression coverage for promotion failure or concurrent destination recreation.

The current-DMG env guard and runtime coverage otherwise look good.

jadabreu:
Thanks for catching this. I updated the cache refresh so it keeps the existing cache as a backup until the replacement is safely in place, and restores it if the update fails. I also added a test for that failure case. All checks are now passing.

ilysenko (approved):
Reviewed the updated head. The managed Chrome cache promotion now preserves the previous cache, restores it on failure, avoids destination nesting with mv -T, and keeps cold-start failures handled. Current-DMG and package checks are green.

PR #1232, opened 2026-08-05
feat(linux): integrate Watchbound working-tree watches

gadicc:
## Summary

Replace the opt-in `directory-only-working-tree-watch` feature’s bespoke JavaScript recursive watcher with Watchbound 2.0.0. Codex continues to own the application policy around it: Git and user exclusions, Git metadata refresh, logical-path mapping, conservative invalidation, focus recovery, and root replacement.

This also fixes an upstream drift regression in Desktop 26.730.61639. Upstream added a Linux-specific working-tree route that called `@parcel/watcher` directly, bypassing `startFileWatch()` and therefore silently bypassing this enabled feature. The patch now reroutes exactly that local working-tree call through Watchbound.

Route detection correlates semantic roles across minifier-generated aliases instead of pinning their spellings. Missing, duplicate, partial, mismatched, or ambiguous contracts still fail closed and preserve the original bundle bytes.

The integration additionally:

- Pins and verifies the Watchbound wrapper, loader, and GNU/Linux x64 or ARM64 native package using registry integrity, archive/native hashes, complete file inventories, ELF architecture, and runtime capability checks.
- Builds the pinned native target from source under Nix without fetching npm packages inside the derivation.
- Makes bundle mutation and native-package staging transactional, with rollback and `failed-integrity` handling when restoration cannot be proven.
- Remains optional, disabled by default, and mutually exclusive with `shallow-repository-watches`.

### Approach comparison

| Approach | Trade-off |
| --- | --- |
| Shallow repository watches | Uses the fewest native watches and no additional runtime, but deep changes may be deferred until focus recovery. |
| Upstream Parcel | Mature, typed, and cross-platform, but its public API does not expose a process-wide watch budget, coverage/loss state, active exclusion replacement, or reconciliation. Exact 2.5.6 Linux conformance also reproduced silent overflow and root/moved-tree coverage gaps. |
| Watchbound | GNU/Linux x64/ARM64 only, but retains live deep invalidations under a bounded recursive budget and reports partial or uncertain coverage so Codex can invalidate conservatively and recover on focus. |

Historical seven-trial tmpfs feasibility measurements—not a fresh Watchbound 2.0.0 benchmark—recorded:

| 10,001-directory startup | Recorded Watchbound | Parcel 2.5.6 | Previous Codex JS watcher |
| --- | ---: | ---: | ---: |
| Cold | 32.31 ms / 3.19 MiB | 49.11 ms / 4.45 MiB | 273.05 ms / 30.26 MiB |
| Warm | 36.59 ms / 3.00 MiB | 47.42 ms / 4.50 MiB | 239.37 ms / 9.38 MiB |

The 1,001-directory cold ranges overlapped, and the implementations have different batching policies, so these figures are feasibility evidence rather than a general performance claim.

Even after adding the package verifier and transactional staging, the feature’s JavaScript implementation and tests fall from 10,140 to 8,523 lines (-16%) and from 364 KB to 291 KB (-20%). The runtime patch itself falls from 2,287 to 1,723 lines (-25%), while `test.js` falls from 7,853 to 5,301 lines (-32%) as engine-level topology and lifecycle coverage moves into Watchbound.

## Validation

- `node --test linux-features/directory-only-working-tree-watch/test.js` — 108/108 passed
- `node --test scripts/patch-linux-window-ui.test.js` — 411/411 passed
- `git diff --check main...HEAD`
- Published Watchbound 2.0.0 source, package, archive, and native hashes independently verified
- Published x64 package loaded under the managed Node 24 runtime; physical symlink-root mapping verified
- Cross-format packaging smoke suite previously passed. The current-tip rerun completed its packaging checks and stopped at launcher warm-start validation because an installed `/opt/codex-desktop` instance was already running.
- Not run locally: full Nix derivation builds, because Nix is unavailable on this host

## Checklist

## Final remarks

I admit I'd be slightly reluctant to blindly accept a PR that adds a new dep which is itself a new package of the same contributor.  However, Watchbound was purpose-built to address the exact optimal runtime surface Codex needs, has a big platform test matrix in CI, and at the end of the day, still remains an opt-in feature in Codex.  Both Watchbound and our use of it in Codex underwent multiple adversarial reviews with gpt-5.6 `ultra`, and have enjoyed daily use by me throughout the development cycle until changes stabilized.


gadicc:
Noted the two failing CI above - will investigate and report back:

* [CI / Rust and Smoke Tests (pull_request)](https://github.com/ilysenko/codex-desktop-linux/actions/runs/31002454075/job/92294333653?pr=1232)
* [CI / Nix Package Builds (pull_request)](https://github.com/ilysenko/codex-desktop-linux/actions/runs/31002454075/job/92294333590?pr=1232)

Everything else passed.  Unchecking 'ready for review' for now.

ilysenko (changes requested):
Thank you for the substantial amount of work and care that went into this PR. The current-DMG regression is real, and we would like to take the narrowly scoped fix that reroutes the current Parcel working-tree handoff through the existing feature.

Please split this PR so that the current-DMG routing fix can be reviewed on its own. The Watchbound dependency migration, the global Electron/runtime qualification changes, and the broader transactional/Nix infrastructure should not be part of that fix. If you would like to pursue the Watchbound migration afterward, please open it as a separate PR once its product value, platform support, and maintenance model can be evaluated independently.

The remaining blockers should also be addressed in the appropriate follow-up PRs:

1. Non-Nix builds now execute the downloaded or cached Electron binary during every build without cryptographically authenticating it first, including when Watchbound is disabled. Please either verify the runtime before executing it or keep this qualification work out of the shared build path.
2. Nix currently fails on SC2155 before any Watchbound-enabled output is built. Please fix the Nix build, include the new Watchbound manifest and Cargo lock pin sources in the CI guard, run the feature-enabled watchdog output, and verify the Nix wrapper/loader against the shared artifact manifest.
3. The ambiguous-rename ownership proof is not reliable under inode reuse. The current test can classify a foreign destination as owned and returns an untagged error instead of PATCH_INTEGRITY_FAILURE. Please make the ownership/rollback guarantee reliable and restore the required test.
4. Reusing the existing feature ID drops previously supported ARMv7/armhf behavior while the repository still supports those architectures and has no platform constraint in the feature picker. Please preserve that behavior or introduce an explicit generic platform-constraint and migration path in a separate change.

We would be happy to review the focused current-DMG fix once it is isolated. Thank you again for investigating the CI failures and for documenting the integration thoroughly.

gadicc:
Ah sorry, I did the most recent push before seeing your comment.  Just heading out now but will review and address accordingly.  Thanks for taking a look!

gadicc:
Thanks for the review... all very valid feedback :)  Closing for now to keep the repo clean, will submit the split PRs - one at a time :) - when they are ready.

PR #1233, opened 2026-08-05
updater: add automatic build toggle

moxw:
## Summary

- add an independent **Build updates automatically** Linux setting, enabled by default
- when disabled, background checks download and verify the upstream DMG, persist an `update_available` state, and notify without starting the native/Electron/package build
- let an explicit **Check for updates** action build a pending update while keeping install-on-exit as a separate preference
- update the current Linux settings generator, optional wrapper-updater composition, tests, and updater documentation

## Validation

- `cargo test -p codex-update-manager` (321 unit tests and 5 concurrency tests passed)
- `node --test scripts/patch-linux-window-ui.test.js` (411 passed)
- `node --test linux-features/codex-wrapper-updater/test.js scripts/patches/impl/keybinds-settings.test.js` (24 passed)
- `git diff --check`
- `bash tests/scripts_smoke.sh` exercised native and AppImage packaging successfully, then encountered an unrelated temporary-directory cleanup race in launcher recovery (`Directory not empty`)

## Checklist

- [x] I followed CONTRIBUTING and kept the change focused on source files.
- [x] If this fixes upstream drift, it targets the latest supported DMG shape without retaining legacy fallbacks.
- [x] I reviewed the final staged diff with a coding agent before committing.


ilysenko (changes requested):
Thanks for the thoughtful updater improvement. Separating update detection, package building, and installation is useful, but I found several blockers that should be addressed before merge.

1. **Revalidate the pending DMG before building it.**  
   `UpdateDetected` and `UpdateAvailable` are currently handled before fetching fresh upstream metadata. After candidate A is deferred, background checks stop checking upstream, and an explicit **Check for updates** builds A even if candidate B has since been published. This conflicts with the repository’s latest-`CODEX.DMG`-only policy. Please refresh the upstream identity under the existing check lock, replace a stale pending candidate, and reuse the cached DMG only after confirming it is still current.

2. **Recover a missing deferred DMG in the same check cycle.**  
   A deferred artifact is stored under the user’s cache and may legitimately be removed. Currently, the first explicit check marks the update as failed when the cached DMG is missing; only a second check redownloads it. The same explicit check should redownload the current candidate and continue into the build path.

3. **Isolate this non-default workflow as a disabled-by-default Linux feature.**  
   Please keep only the smallest generic mechanism in updater core and move the setting key, Linux Settings control, and activation into a dedicated `linux-features/<id>/` feature. When disabled, its descriptors and settings must not be loaded, and the updater must preserve the existing automatic-build behavior.

   The feature patch must remain fail-soft: a missing patch target should be reported as optional drift rather than crashing the patcher or affecting ordinary manual/native builds where the feature is disabled. If a user-enabled feature drifts, the existing acceptance policy may reject candidate promotion and preserve the working installation, but it must not leave a partially updated app. Please also handle the case where the feature is disabled while a persisted `UpdateAvailable` candidate already exists.

4. **Apply the required updater version change.**  
   This PR adds user-visible updater behavior but leaves `codex-update-manager` at `0.10.5`. Please apply the version bump required by `CONTRIBUTING.md`, update `Cargo.lock` and the relevant changelog/release documentation, and explicitly resolve compatibility of the new serialized `update_available` status. An older updater binary cannot currently deserialize that value, so either preserve downgrade compatibility or classify the version change accordingly.

Please add focused tests covering:

- deferred candidate A followed by newer upstream candidate B;
- an explicit check building B rather than stale A;
- recovery of a deleted deferred DMG in one invocation;
- a clean disabled-feature state;
- disabling the feature with an existing persisted `UpdateAvailable` candidate.

Thank you!

moxw:
Addressed the requested changes in 2d0d4bd:

- Pending deferred DMGs now re-enter the normal upstream metadata/download flow under the check lock, so candidate B replaces stale candidate A before an explicit build.
- A missing cached deferred DMG is redownloaded and proceeds into the build path in the same check invocation.
- Deferred builds now live in the disabled-by-default `deferred-update-build` Linux feature. The feature owns the setting key, settings toggle, policy declaration, and fail-soft patch; core only loads generic updater policy from enabled features. Disabling the feature restores automatic builds, including for persisted deferred state.
- The updater is bumped to 0.11.0. Deferred state retains `update_detected` plus an optional marker for 0.10.x downgrade compatibility, and prerelease `update_available` state migrates back to `update_detected`.

Added focused A→B, explicit-current-candidate, deleted-cache single-invocation, disabled-feature, persisted legacy-state, feature drift, and composition tests.

Local validation: Rust 1.97 clippy clean; updater suite 324/324 after rerunning its ownership-sensitive host check outside the sandbox; focused feature/settings tests 29/29; patcher suite 411/411 after the same host ownership verification; cross-format packaging smoke passed.

ilysenko (changes requested):
Thank you for the substantial work addressing the previous review. The latest update resolves the original blockers: the feature is now isolated and disabled by default, stale candidates and missing cached DMGs are handled, and updater version/state compatibility is preserved. The remaining issue does not affect users who leave the feature disabled.

This is a large updater and persisted-state change, so the final behavior needs careful review. This is the last issue found in the current full review:

A stable deferred candidate is still treated as an interrupted retry. `update_check_should_retry` returns true for every `UpdateDetected`, including `deferred_build=true`. Native launches run `check-now --if-stale`, so this bypasses freshness and the unchanged-fingerprint/same-SHA shortcuts. With the feature enabled and automatic builds off, the full DMG can therefore be downloaded again on every app launch. An offline launch can also turn the pending state into `Failed`.

Please treat `UpdateDetected + deferred_build=true` as a stable pending state for `--if-stale`. After confirming that upstream is unchanged, reuse the valid cached DMG and perform GET only when the candidate changed or the cache is missing.

Before requesting another review, please verify the final change against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant checks locally, perform an independent review of the complete diff with a local LLM, and wait for every required CI check to complete successfully after pushing the update.

Thank you again for the effort put into this PR.

moxw:
Addressed the final deferred-candidate review on exact head `9db2ba5`:

- `UpdateDetected + deferred_build=true` is now stable for background checks. Fresh app-launch checks do not make an upstream DMG request.
- Stale checks use HEAD and reuse the cached DMG when the fingerprint is unchanged; GET occurs only when upstream changed or the cache is missing.
- Offline app-launch and periodic-daemon checks preserve the pending candidate instead of replacing it with `Failed`.
- Re-enabling automatic builds or disabling the feature bypasses freshness and resumes the deferred build immediately.

Added focused coverage for fresh deferred launch checks, unchanged-HEAD cache reuse, missing/changed candidates, resumed automatic builds, and offline app-launch plus daemon refreshes.

Validation:
- updater suite: 328/328 effective pass (327 in the sandbox plus the ownership-sensitive `/usr/bin` test passing with host ownership visible)
- complete PR diff independently reviewed locally; the follow-up review found two edge cases, both fixed and re-reviewed with no remaining actionable findings
- exact-head GitHub CI: 6 successful, 1 expected skip, 0 pending/failing

Ready for re-review.

ilysenko (approved):
Reviewed the current head after the requested updater fixes. The deferred-candidate behavior is now isolated, fail-safe, current-DMG compliant, and covered by green exact-head checks. Approved for merge.

PR #1236, opened 2026-08-05
fix(shared-app-server-socket): support systemd user adoption

kortylokai-web:
## Summary

Normal systemd user sessions can adopt a detached shared app-server authority under `systemd --user` rather than directly under PID 1. Current orphan cleanup requires the authority's direct parent to be PID 1, so it refuses this valid topology and leaves the orphaned authority running.

This pull request:

- accepts authorities adopted directly by PID 1 or by a verified same-user `systemd --user` manager whose own parent is PID 1
- preserves the existing lock owner, authority PID/start-time, UID, command-line, socket identity, listener ownership, and pre-signal race checks
- keeps unrelated parents and invalid systemd-user identities fail-closed
- adds deterministic coverage for direct PID-1 adoption, conventional and Nix systemd-user paths, invalid adopter identities, unrelated parents, and adopter changes before signaling
- updates the feature documentation to describe the supported adoption topologies

The source-of-truth files changed are:

- `linux-features/shared-app-server-socket/orphan-reaper.js`
- `linux-features/shared-app-server-socket/test.js`
- `linux-features/shared-app-server-socket/README.md`

There is no linked issue.

The verified systemd-user allowance requires a live, stable adopter with the same UID, PID 1 as its parent, `systemd` as its process name, an absolute executable argument whose basename is `systemd`, and an exact `--user` argument. It does not rely on `/proc/<pid>/exe`, which can be unreadable under normal host procfs policy.

## Validation

- Reproduced on untouched `upstream/main` at `20825ea9a907adbce8e5a10e466543704c5aaef6`:
  - `node --test linux-features/shared-app-server-socket/test.js`
  - 31 passed, exactly 2 timed out, and 1 optional integration test was skipped
  - both failures waited for direct PID-1 adoption while the authority was adopted by the same-user `systemd --user` manager
- Invoked untouched main's production orphan reaper against the reproduced topology and confirmed that it refused cleanup and left the authority running
- `node --test linux-features/shared-app-server-socket/test.js`
  - 39 passed, 0 failed, and 1 optional integration test was skipped
- `node --test --test-name-pattern='orphan reaper (stops an exact reparented authority|refuses two live listener inodes)' linux-features/shared-app-server-socket/test.js`
  - 2 passed under a normal systemd user session
- `./scripts/ci-local.sh pr`
  - passed the complete core, Debian, RPM, and pacman PR matrix
  - passed

The optional real-Codex proxy integration test was not run because `CODEX_CLI_PATH` was not configured.

## Checklist

### DIFF
--- a/linux-features/shared-app-server-socket/README.md
+++ b/linux-features/shared-app-server-socket/README.md
@@ -37,14 +37,15 @@ than the authority startup timeout, before they can be reclaimed when no socket
 exists.
 
 The launcher also cleans up a live authority orphaned by a terminated Desktop
-process. Cleanup is limited to a same-user, reparented `codex app-server
---listen unix://PATH` process serving the exact locked socket. Once the authority
-is ready, its PID and process-start identity are recorded in the ownership lock.
-The lock owner, socket inode, listener identity, command line, and process start
-identities are rechecked before signaling it. Unknown listeners, live Desktop
-owners, changed identities, and pathnames with multiple live listener inodes
-remain untouched. The same cleanup runs after Electron exits and before a later
-cold start.
+process. Cleanup is limited to a same-user `codex app-server --listen unix://PATH`
+process serving the exact locked socket after direct PID 1 adoption or adoption
+by a verified same-user `systemd --user` manager whose own parent is PID 1. Once
+the authority is ready, its PID and process-start identity are recorded in the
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed the current head. The systemd user adoption path remains feature-local, fail-closed for mismatched identities, and is covered by the relevant tests.

PR #1240, opened 2026-08-06
fix(chrome): support concurrent browser clients

### LINKED ISSUE #1239: Chrome native host evicts an active @Chrome task when a second Codex chat connects
## Summary

Two concurrent Codex chats using `@Chrome` cannot keep independent browser bindings on Linux. When the second browser client connects to the Chrome native host, the first client is deliberately drained and its Unix socket is closed. The first task then fails with `Browser is not available`; reconnecting it evicts the second task, producing a repeatable ping-pong takeover.

This is distinct from the IAB-vs-Chrome socket-discovery bug fixed by #843: both clients in this report are Chrome extension clients.

## Environment

- Distro: Manjaro Linux, GNOME on Wayland
- Package: pacman `codex-desktop 2026.07.22.204125-1`
- Wrapper source version: `0.10.2` (`c370485`)
- Upstream DMG app build: `26.715.72359`
- Chrome extension: `1.2.27236.6274`
- Chrome profile: one `Default` profile with one extension instance
- Native host: `~/.codex/plugins/cache/openai-bundled/chrome/latest/extension-host/linux/x64/extension-host`

The app and extension were both current on their configured update channels when reproduced.

## Reproduction

1. Start Codex task A and use `@Chrome` to claim/read tab A.
2. Start independent Codex task B and use `@Chrome` to claim/read tab B.
3. Ask task A to read its already-bound tab again.
4. Reconnect task A, then ask task B to read its tab again.

No task modifies the other task's page. The failure happens at the browser-binding layer.

## Actual behavior

The browser handles alternate between valid and unavailable:

| UTC time | Event |
| --- | --- |
| 11:55:29 | Task A connects and receives browser id `-419e…` |
| 11:55:52 | Task B reconnects and receives browser id `-d87f…` |
| 11:56:04 | Task A fails: `Browser is not available: -419e…` |
| 11:56:14–21 | Task A discovers/rebinds to `-e54d…` |
| 11:56:22 | Task B fails: `Browser is not available: -d87f…` |

Tab titles, URLs, windows, and Chrome tab groups do not affect the failure.

## Expected behavior

Multiple authorized Codex browser clients should remain connected to the same extension host so independent chats can issue interleaved commands to their own claimed tabs. A new client must not invalidate an active client's browser binding.

## Current source-level cause

Current `main` still makes replacement explicit:

- `HostState::replace_with_client` drains every existing client and clears both pending-request maps.
- `accept_clients` calls it for every newly accepted Unix client and closes all evicted sockets.
- Chrome-originated requests use `select_single_client_id`, so the reverse direction currently assumes exactly one client.

Relevant locations on current `main`:

- `computer-use-linux/src/bin/codex-chrome-extension-host.rs:120-130`
- `computer-use-linux/src/bin/codex-chrome-extension-host.rs:493-502`
- `computer-use-linux/src/bin/codex-chrome-extension-host.rs:743-789`

Client-originated requests already retain `client_id` in `PendingChromeRequest`, so Chrome responses can be returned to the correct requesting client. The missing pieces appear to be connection retention and reverse-direction routing.

## Suggested minimal design

1. Add authorized clients without draining existing ones.
2. Learn `session_id -> client_id` from normal client requests, whose params already carry `session_id` and `turn_id`.
3. Route Chrome-originated requests/notifications carrying `session_id` to that client.
4. Preserve current single-client fallback when only one client is connected.
5. Return a clear ambiguity error for an unscoped Chrome-originated request when multiple clients exist; do not evict either client.
6. Remove only the disconnecting client's session ownership and pending correlations.
7. Add a bounded maximum-client count and regression tests for interleaved requests, response correlation, disconnect cleanup, and ambiguous reverse requests.

## Acceptance criteria

- Two clients can connect without either socket being closed.
- Interleaved client-to-Chrome requests receive responses only on the originating client.
- Disconnecting one client leaves the other usable.
- A scoped Chrome-to-client request reaches the session owner.
- An ambiguous unscoped reverse request fails explicitly without disconnecting clients.
- Existing peer-credential checks, pending-request limits, TTL cleanup, and IAB socket isolation remain intact.

Eeeeye:
## Summary

- keep multiple authorized Codex browser clients connected instead of evicting the previous client whenever a new task connects;
- learn `session_id -> client_id` ownership from client requests and route scoped Chrome requests and notifications back to that owner;
- move Unix socket writes to bounded per-client writer queues, so a stalled or disconnected peer is removed without holding the global `HostState` lock or blocking healthy clients;
- fan out the extension global heartbeat to available clients and accept only the first valid JSON-RPC `pong`, so an error, malformed response, or unresponsive oldest client cannot disable healthy clients;
- bound pending requests per client and globally in both directions, while cleaning up only the disconnecting client routes and correlations;
- add regression coverage for interleaved response correlation, reverse routing, heartbeat failure isolation, writer queue byte and message limits, shared broadcast frames, disconnect cleanup, and all four pending-request limit paths.

This preserves the single-client fallback and returns an explicit ambiguity error for other unscoped reverse requests when multiple clients are connected.

Fixes #1239.

## Review follow-up

- Each client writer now has both a 64-message secondary bound and a serialized-byte budget large enough for one maximum accepted frame. The budget includes queued and in-flight bytes and is released when each frame is dropped.
- Broadcasts are serialized once into an `Arc<[u8]>` shared by all recipient queues; heartbeat fanout mutates only the generated correlation ID instead of deep-cloning the full JSON value per client.
- Invalid heartbeat responses remove only their own correlation. Sibling correlations remain live until a valid `{"jsonrpc":"2.0","result":"pong"}` arrives or all recipients have returned invalid responses.
- Tests independently cover byte-limit isolation, message-limit isolation, shared serialized broadcast storage, and error-first/malformed-second/healthy-third heartbeat handling.
- The independent `turn_aborted` / `turn/completed` rollout handling remains outside this PR.

## Validation

- `git diff --check`
- `cargo +nightly check --workspace --all-targets`
- `cargo +nightly test -p codex-computer-use-linux -- --test-threads=1`
- extension-host tests: 76 passed
- `cargo +nightly clippy --workspace --all-targets -- -D warnings -A clippy::nonminimal_bool` (the allow is limited to two unrelated pre-existing simplification warnings outside this PR)

The machine Cargo configuration points crates.io at the legacy Git index, so local validation used a command-line sparse-index override. No repository or user Cargo configuration was changed, and no generated artifacts are included.

## Checklist

- [x] This is not an upstream-drift fix; it changes only the Linux Chrome native-host source and targets the current supported runtime.
- [ ] I reviewed the final diff with my coding agent using maximum reasoning effort. The final diff was reviewed in this session, but no separate maximum-effort review was run.

### DIFF
--- a/computer-use-linux/src/bin/codex-chrome-extension-host.rs
+++ b/computer-use-linux/src/bin/codex-chrome-extension-host.rs
@@ -14,7 +14,10 @@ use std::{
     },
     path::{Path, PathBuf},
     process,
-    sync::{Arc, Mutex},
+    sync::{
+        mpsc::{sync_channel, Receiver, SyncSender, TrySendError},
+        Arc, Mutex, Weak,
+    },
     thread,
     time::{Duration, Instant, SystemTime, UNIX_EPOCH},
 };
@@ -42,8 +45,10 @@ const MAX_PENDING_REQUESTS_PER_DIRECTION: usize = 1024;
 const MAX_PENDING_REQUESTS_PER_CLIENT_PER_DIRECTION: usize = 256;
 /// Bounds retained string IDs to about 4 MiB per direction at the entry cap.
 const MAX_PENDING_REQUEST_ID_STRING_BYTES: usize = 4 * 1024;
-/// Bounds simultaneously connected Codex browser clients and their reader threads.
+/// Bounds simultaneously connected Codex browser clients and their I/O threads.
 const MAX_CONNECTED_CLIENTS: usize = 64;
+/// Prevents a slow client socket from accumulating unbounded outbound messages.
+const CLIENT_WRITE_QUEUE_CAPACITY: usize = 64;
 /// Bounds retained browser-session routing state.
 const MAX_TRACKED_SESSIONS: usize = 1024;
 const MAX_SESSION_ID_BYTES: usize = 1024;
@@ -52,11 +57,10 @@ const INVALID_REQUEST_ERROR_CODE: i64 = -32600;
 
 type SharedState = Arc<Mutex<HostState>>;
 type SharedChromeWriter = Arc<Mutex<Box<dyn Write + Send>>>;
-type SharedClientWriter = Arc<Mutex<UnixStream>>;
 
-#[derive(Clone)]
 struct Client {
-    writer: SharedClientWriter,
+    sender: SyncSender<Value>,
+    shutdown: UnixStream,
 }
 
 struct PendingChromeRequest {
@@ -70,6 +74,7 @@ struct PendingChromeRequest {
 struct PendingClientRequest {
     client_id: usize,
     chrome_request_id: Value,
+    fanout_group: Option<u64>,
     created_at: Instant,
 }
 
@@ -126,7 +131,7 @@ impl HostState {
         }
     }
 
-    fn add_client(&mut self, writer: SharedClientWriter) -> Option<usize> {
+    fn add_client(&mut self, client: Client) -> Option<usize> {
         if self.clients.len() >= MAX_CONNECTED_CLIENTS {
             return None;
         }
@@ -136,12 +141,15 @@ impl HostState {
             id = id.checked_add(1).unwrap_or(1);
         }
         self.next_client_id = id.checked_add(1).unwrap_or(1);
-        self.clients.insert(id, Client { writer });
+        self.clients.insert(id, client);
         Some(id)
     }
 
-    fn remove_client(&mut self, client_id: usize) {
-        self.clients.remove(&client_id);
+    fn remove_client(&mut self, client_id: usize) -> bool {
+        let Some(client) = self.clients.remove(&client_id) else {
+            return false;
+        };
+        let _ = client.shutdown.shutdown(Shutdown::Both);
         self.session_owners
             .retain(|_, owner_client_id| *owner_client_id != client_id);
         remove_pending_requests_for_client(
@@ -149,6 +157,7 @@ impl HostState {
             &mut self.pending_client_requests,
             client_id,
         );
+        true
     }
 
     fn track_session_owner(&mut self, client_id: usize, message: &Value) {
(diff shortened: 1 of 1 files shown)


ilysenko (changes requested):
Thanks for tackling concurrent Chrome clients and for fixing the previous Clippy failure. The current head still needs a few changes before it is safe to merge:

- Move client socket writes out of the global `HostState` lock and use bounded per-client writer queues so one stalled peer cannot block every client.
- Make the global heartbeat failure-isolated instead of permanently routing it to the lowest client ID; a healthy client must remain usable when the oldest client stops responding.
- Add boundary tests that actually exercise the per-client and global pending-request guards in both directions.
- Please split the independent `turn_aborted`/`turn/completed` rollout cleanup into a separate PR.

Eeeeye:
Addressed all four review points in the rebased head:

- client socket writes now use bounded per-client queues and dedicated writer threads, with queue failure isolated to that client;
- global heartbeat requests fan out and use the first healthy response instead of selecting the lowest client ID;
- boundary tests now independently exercise per-client and global guards in both bridge directions, plus a full-writer-queue isolation test;
- the unrelated rollout completion/abort changes have been removed from the PR diff.

Validation passed: formatting, diff check, full `codex-computer-use-linux` check/test suite, 73 extension-host tests, and Clippy `-D warnings` for the affected target.

ilysenko (changes requested):
Thanks for addressing all four previous review points. Two issues in the new queue and heartbeat fanout implementation still need changes:

- Bound writer queues by serialized bytes, not only message count, and avoid deep-cloning the full JSON value for every recipient. With 64 MiB frames, the current 64-message queue can retain about 4 GiB for one stalled client before eviction. Please add byte-limit isolation coverage.
- Only a successful expected heartbeat response should cancel sibling correlations. A JSON-RPC error or malformed response must not beat and discard a later healthy `pong`; please add an error-first/healthy-second regression test.

Eeeeye:
Addressed the two latest review points in 57c536a:

- writer queues now reserve actual serialized frame bytes (queued plus in-flight), retain the 64-message secondary bound, and disconnect only the over-budget client;
- broadcasts serialize once and share an `Arc<[u8]>` instead of deep-cloning JSON per recipient;
- heartbeat errors and malformed/non-`pong` responses remove only their own correlation; only a valid JSON-RPC `pong` cancels siblings;
- added byte-limit isolation, message-limit isolation, shared-frame, and error-first/malformed-second/healthy-third regression coverage.

Local formatting, diff checks, workspace check, package tests, 76 extension-host tests, and Clippy for the changed code passed. The hosted checks are running on the new head.

ilysenko (approved):
Reviewed the current head. The previously requested writer-queue byte bounds, shared serialization, heartbeat response validation, capacity coverage, and scope split are all addressed, and the required checks pass.

PR #1243, opened 2026-08-07
fix(patches): realign core patches with ChatGPT.dmg 26.803.41515

jurgenmahn:
## What changed

Realigns five core patch targets with the current upstream DMG. Two are `required-upstream`, so today's DMG makes candidate promotion fail outright. Two more were silently degrading to `skipped-optional`. The fifth reported success while emitting invalid JavaScript, which crashed the app on launch.

| Patch | Status before | Upstream change |
| --- | --- | --- |
| `linux-explicit-quit-drain-timeout` | `failed-required` | skip-drain branch now flushes global state |
| `main-process-ui` | `failed-required` | (rollup of the above) |
| `linux-avatar-overlay-mouse-passthrough` | `failed-required` | `applyPointerInteractivityPolicy` gained an `applyInputShape` early return |
| `browser-annotation-screenshot` | `skipped-optional` | `comment-preload.js` folded into `browser-page-preload.js` |
| `linux-browser-use-external-availability` | `skipped-optional` | browser registry + guard + family helpers reminified |
| `linux-tray` | `applied`, but emitted broken JS | `new Tray()` gained a second argument containing a nested call |

### 1. `will-quit` drain sequence

The reduced (skip-drain) branch used to settle only the stop and trace promises; it now flushes global state as well:

```js
// before
Promise.allSettled([p(),m()]).then(t)
// 26.803.41515
Promise.allSettled([d.flush(),p(),m()]).then(t)
```

`parseCurrentWillQuitDrainBody` required the old shape literally, so `currentWillQuitDrainCandidates` returned zero candidates and the patch bailed. The reduced-branch needle, its replacement, and the already-applied postcondition in `hasAppliedWillQuitCleanupPostcondition` now all carry `globalState.flush()`, and the reduced/full binding cross-check asserts both branches flush the same object.

This one is the reason the build aborts: without the bounded drain the Linux app can hang on quit instead of reaching `app.exit(0)`.

### 2. Avatar overlay pointer policy

Upstream added its own input-shape path (`BrowserWindow.isInputShapeSupported()` / `win.setInputShape()`), which inserted a new early return into the method the patch replaces:

```js
if(this.applyInputShape(e))return;
```

The needle now matches the current method, and the patched replacement keeps that upstream call so the native path still wins where it exists. With the stock Electron headers used here `isInputShapeSupported()` is absent, `supportsInputShape` is `false`, and `applyInputShape` returns `false` immediately — so the existing `setShape` recovery path continues to drive Linux behavior exactly as before.

### 3. Annotation screenshot preload bundle

`.vite/build/comment-preload.js` no longer exists; the annotation code now ships in `browser-page-preload.js`. The anchor patch itself still matches unchanged — only the file lookup was stale. `patchCommentPreloadBundle` is renamed to `patchBrowserPagePreloadBundle` along with its warning text, and the descriptor is rewired.

### 4. Browser Use external availability contracts

The three asset contracts guarding the transactional write were fingerprinting exact minified names that all rotated this release:

- main caller: `async function sne(...)` / `async function lne(...)` / `function Ol(`
- main registry: `function fb(e){return Object.hasOwn(ob,e)}` and the `So` → `ob` export
- renderer: `function Yl(e){return Object.hasOwn(Xl,e)}` and `Object.keys(Xl).filter(Yl)`

The availability gate and registry literal both still match, so only the guards were stale. They now resolve the registry binding and its `Object.hasOwn` guard from the registry literal itself (`currentBrowserUseRegistryBinding`) and match the browser-family helpers by parameter shape rather than name. Same assets, same writes, but the contracts survive the next reminification instead of turning into silent optional drift.

### 5. Tray constructor argument list

This one reported `applied` and still broke the app. Upstream now passes a second argument to the tray constructor:

```js
new l.Tray(t.defaultIcon,process.platform===`win32`&&l.app.isPackaged?dEe(e.buildFlavor):void 0)
```

The retention patch captured the argument list with `([^;)]+)`, which stops at the first `)` — the one closing the nested `dEe(...)` call. The needle matched, the patch reported success, and the emitted bundle was:

```js
n=codexLinuxRegisterTray(new l.Tray(t.defaultIcon,...?dEe(e.buildFlavor)):void 0);
```

Unbalanced parentheses, so the packaged app died on launch with `ChatGPT failed to start. Unexpected token ')'`.

`findMatchingParen` is added next to the existing `findMatchingBrace`, and the construction span is now resolved by balancing parentheses across the upstream shape and both already-applied shapes.

Worth calling out: this drift was invisible before, because the two `failed-required` patches above rejected the candidate first. It only becomes reachable once promotion succeeds, and no check in the pipeline parses the patched bundle — a needle can match, report `applied`, and still emit invalid JavaScript. A syntax gate over the patched bundles would have caught it; that felt like a separate change, so it is not in this PR.

## Why

`ChatGPT.dmg` published 2026-08-07 01:24 GMT reshaped these targets.

| | |
| --- | --- |
| URL | `https://persistent.oaistatic.com/codex-app-prod/ChatGPT.dmg` |
| Size | 600747074 bytes |
| ETag | `0x8DEF4229FE5F388` |
| Electron | 42.3.0 |

Before this change, `install.sh` on that DMG ends with:

```
Critical patch failures (3):
  - linux-explicit-quit-drain-timeout (failed-required)
  - linux-avatar-overlay-mouse-passthrough (failed-required)
  - main-process-ui (failed-required)
Aborting: these patches are required for a working Linux app.
[ERROR] Candidate was not installed (verdict: rejected)
```

Per the maintainer rule, the old DMG shapes are replaced rather than kept alongside the new ones — no fallback paths were added.

- `scripts/patches/impl/main-process/quit-lifecycle.js`
- `scripts/patches/impl/avatar-overlay.js`
- `scripts/patches/impl/webview/index.js`
- `scripts/patches/core/all-linux/extracted-app/browser-annotation/patch.js`
- `scripts/patches/impl/main-process/tray.js`
- `scripts/patches/lib/minified-js.js`
- `tests/scripts_smoke.sh`

## How it was tested

- `node --test scripts/patch-linux-window-ui.test.js` — 414/414 pass (411 before, plus 3 new)
- `node --test scripts/workspace-root-open-targets-patch.test.js` — 7/7 pass
- Existing fixtures updated to the 26.803.41515 shapes; new regression tests cover the browser-page preload lookup (hit and miss) and the name-agnostic Browser Use binding resolution using deliberately different minified identifiers than the current DMG.
- Every touched patch verified directly against the real 26.803.41515 bundle and confirmed idempotent on a second pass.
- Every emitted bundle in `.vite/build` parsed with `node --check`, and the main bundle additionally with `acorn`, before and after the tray fix, to confirm the syntax error is gone.
- `bash tests/scripts_smoke.sh` passes with only `test_launcher_warm_start_recovery` skipped (see below); its fixtures for the will-quit drain, the annotation preload bundle, and the tray constructor were all updated to the current shapes.
- Full `./install.sh` against the DMG above: verdict `accepted`, `"blockers": []`, `"warnings": []`, candidate promoted. Previously `rejected`.

Not run, and why:

- `cargo check` / `cargo test -p codex-update-manager` — no Rust toolchain in this environment. No Rust code is touched by this change.
- `test_launcher_warm_start_recovery` inside `tests/scripts_smoke.sh` — fails here on `timed out waiting for first Electron`. Reproduced identically on unmodified `df725ad` in a clean worktree, so it is environmental (Electron cannot start in this headless sandbox) and not a regression from this change. Every other smoke test passes.
- Package builds (`build-deb.sh` / `build-rpm.sh` / `build-pacman.sh`) — no packaging code touched.

## Risks and follow-ups

- The packaged app was rebuilt and installed from this branch and starts again; the `Unexpected token ')'` crash is gone.
- Environment: verified on x86_64 Ubuntu (kernel 7.0), source install path. Not verified on `.deb` / `.rpm` / pacman / AppImage / Nix, though nothing outside the ASAR patch layer changed.
- The avatar overlay patch now sits behind upstream's `applyInputShape`. On a runtime that does implement `setInputShape` the upstream path would take priority and the `codexLinux*` passthrough recovery timer would not start. That is the intended upstream behavior and is inert on stock Electron, but it is a behavioral branch worth knowing about.
- Out of scope, same DMG: the bundled Chrome/Browser Use plugin scripts (`scripts/lib/patch-chrome-plugin.js`, `patch-browser-client-iab-socket-scope.js`) also drifted — 8 of 13 targets across `installed-browsers.js`, `chrome-is-running.js`, `check-native-host-manifest.js`, `open-chrome-window.js`, `installManifest.mjs`, `SKILL.md`, plus the `browser-client.mjs` site_status allowlist and the IAB socket listing. Those are fail-soft and do not block promotion, so they are left for a separate focused PR.


ilysenko:
Thanks a lot for jumping on this! We have a bot that handles upstream DMG drift automatically, so next time please don’t feel you need to spend time on it or rush to submit a fix—the bot will take care of it.

PR #1249, opened 2026-08-07
Fix Nix PipeWire microphone support

0xdeafbeef:
## Summary

- include PipeWire's ALSA module in the Nix Electron runtime
- set `ALSA_PLUGIN_DIR` so ALSA can load `libasound_module_pcm_pipewire.so`
- add a deterministic Nix check for the plugin and wrapper environment

Nix-packaged Electron could load `libasound.so.2`, but ALSA could not locate
the dynamically loaded PipeWire PCM module. As a result, microphone startup
failed with `NotReadableError`/`TrackStartError`, while the UI misleadingly
reported that another application was using the microphone.

## Validation

- `nix flake check --no-build --no-update-lock-file`
- `nix build .#checks.x86_64-linux.nix-pipewire-alsa-wrapper --no-link`
- `nix run .#codex-desktop` — microphone dictation verified on Fedora with PipeWire

## Checklist

- [x] This change does not add compatibility handling for an older DMG shape.
- [x] I added a targeted check, ran the validation listed above, and confirmed the microphone works in the built application.
- [x] I reviewed the final diff with my coding agent using maximum reasoning effort.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -292,6 +292,7 @@ jobs:
             .#codex-desktop-computer-use-ui
             .#codex-desktop-remote-mobile-control
             .#codex-desktop-computer-use-ui-remote-mobile-control
+            .#checks.x86_64-linux.nix-pipewire-alsa-wrapper
             .#checks.x86_64-linux.nix-gsettings-schema-wrapper
             .#checks.x86_64-linux.watchdog-linux-features
           )
--- a/flake.nix
+++ b/flake.nix
@@ -327,6 +327,7 @@
           mesa
           libgbm
           alsa-lib
+          pipewire
           libX11
           libXcomposite
           libXdamage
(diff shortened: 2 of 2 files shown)


ilysenko (changes requested):

- Please avoid unconditionally overwriting `ALSA_PLUGIN_DIR`. NixOS and users may already supply a single symlink-joined plugin directory, and `--set` hides those plugins. Use default/fallback semantics (for example, `--set-default`) and extend the check to verify that a pre-set value is preserved.
- Please add `.#checks.x86_64-linux.nix-pipewire-alsa-wrapper` to the Nix CI output list. `nix flake check --no-build` only evaluates this derivation, so the new assertions are not currently run by GitHub CI.

Thank you!

0xdeafbeef:
@ilysenko done

ilysenko (approved):
The requested changes are verified on the current head. The wrapper now preserves an explicit ALSA_PLUGIN_DIR, and the targeted Nix check is exercised by CI. All current checks pass.

PR #1250, opened 2026-08-07
Fix Linux browser runtime hook cloning

moxw:
## Summary

- preserve Linux node_repl notification hooks when the current Browser Use client clones the runtime object
- apply the clone shim while staging both Browser and Chrome bundled plugins
- cover the current 26.803 minified shape in runtime and staging tests

## Root cause

The Linux compatibility methods are installed on the non-extensible node_repl object through its prototype. The current upstream Browser Use client creates a runtime with an object spread, which drops those inherited methods before registering the after-submitted-code hook.

## Validation

- real packaged node_repl import of patched Browser and Chrome 26.803.41515 clients
- node --test scripts/lib/browser-client-node-repl-runtime.test.js
- bash tests/scripts_smoke.sh
- bash syntax checks and git diff --check

The broad patcher suite still has one pre-existing local failure in the private bundled-plugin staging-root test; it reproduces unchanged on origin/main.

### DIFF
--- a/scripts/lib/browser-client-node-repl-runtime.test.js
+++ b/scripts/lib/browser-client-node-repl-runtime.test.js
@@ -167,6 +167,64 @@ test("guards every Browser client nodeRepl env read", () => {
   }
 });
 
+test("keeps Browser notification hooks on the cloned nodeRepl runtime", async () => {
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed the current head: the Browser and Chrome runtime clone shim is focused, preserves the Linux node_repl hook on the current upstream client, and all required checks pass.

PR #1248, opened 2026-08-07
Retarget project Last updated sorting to Codex 26.803

moxw:
## Summary

- Retarget the `project-group-last-updated-sort` feature to the exact project-group sorter and call-site symbols in Codex `26.803.41515` (`app-initial-Biw83Aiz.js`).
- Refresh the current-DMG fixture while preserving Last updated recency behavior and upstream saved order for Manual and Priority modes.
- Restore upstream-DMG candidate acceptance when this optional feature is enabled.

This intentionally replaces the Codex 26.730 matcher shape rather than retaining a compatibility fallback. The source-of-truth changes are limited to the feature patch descriptor and its test fixture.

## User-visible behavior

Users with `project-group-last-updated-sort` enabled can rebuild from the current upstream DMG without candidate promotion being rejected for enabled-feature drift. Last updated continues to order project groups by their newest task, while Manual and Priority keep the upstream saved project order.

## Validation

- `node --test linux-features/project-group-last-updated-sort/test.js` — 8 passed.
- Real Codex 26.803 asset application — first pass changed the exact sorter and call site once; second pass was byte-identical.
- `./scripts/rebuild-candidate.sh ./Codex.dmg` from commit `4fd435a0` — Codex `26.803.41515` candidate verdict `accepted_with_warnings`, zero blockers, and `project-group-last-updated-sort` reported `applied`. The remaining Computer Use UI availability warning is an unrelated optional core patch.
- `git diff --check origin/main...HEAD` — passed.
- Broader suites were attempted: the core patch suite passed 413/414 tests, with the bundled-plugin staging test rejected by the sandbox trust check; the Linux feature suite passed 879/881 tests, with the real app-server integration exiting before socket creation and Thorium cleanup unable to write the sandboxed `~/.codex/hooks.json`.
- Native system package installation was not run; validation used the side-by-side transactional candidate.

## Checklist

- [x] This pull request is ready for review and is no longer a draft.
- [x] I followed [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), kept the change focused, edited source files rather than generated output, and removed unrelated changes.
- [x] If this fixes upstream drift, it targets only the latest `CODEX.DMG` and removes obsolete fallback code and tests from the affected area.
- [x] I added or updated relevant tests, ran the validation listed above, and confirmed that required CI checks pass.
- [x] I reviewed the final diff with my coding agent using maximum reasoning effort, addressed all findings, and reran the relevant tests.

PR #1251, opened 2026-08-07
fix: re-target two Linux patches missed by the 26.803 drift fixes

jurgenmahn:
## What changed

Two Linux patches that ChatGPT.dmg 26.803.41515 broke, and that the drift fixes already on `main` (#1242, #1245, #1247, #1248) did not cover. Both are fail-soft, so neither shows up as a build failure — they just stop doing their job.

### 1. IAB socket listing filter

`patch-browser-client-iab-socket-scope.js` filters `extension-` prefixed sockets out of the Unix socket listing so in-app-browser discovery does not pick up extension-host sockets. The listing changed shape:

```js
// pattern on main: zero-argument helper closing over the socket directory
CV=async()=>(await yP(Cb)).map(e=>wP.resolve(Cb,e))
// 26.803.41515: resolves the directory from an options argument
e4=async e=>{let t=ys(e.platform);return(await BE(t)).map(n=>NE.resolve(t,n))}
```

On `main` this reports `Expected one IAB Browser socket listing target, found 0` and leaves discovery unfiltered. The pattern now matches the current shape with named groups. The Windows listing has a similar shape but declares its own pipe prefix and already filters, so it still cannot match and the exactly-one-target guard is preserved.

### 2. Computer Use settings availability

`linux-computer-use-ui-availability` has two halves and discards both if either fails:

```js
if (availabilityChanged && cardChanged && …) return patchedSource;
console.warn("… skipping Linux Computer Use UI availability patch");
return currentSource;
```

The settings component is now compiled with the React Compiler, so the plugin card is no longer derived from a declaration chain ending in a bare `computerUsePlugin;` declarator, but inside memo-cache branches:

```js
let x=ee(u,b),S=Ut(u),…,F,I;
…
t[9]!==S||t[10]!==x.availablePlugins?(re=Y(x.availablePlugins,lr,S),…):re=t[11];
```

The card half matched nothing, which took the working availability half down with it. The net effect on Linux is that `available:!0` is never set and Computer Use reports itself unavailable even when the backend built fine.

The card target is now resolved from the derivation itself: find the `selector(plugins.availablePlugins, name, path)` call whose name identifier holds the `computer-use` slug, then inject the synthetic bundled-marketplace card at the end of the declaration producing the plugins query. A new `findStatementEnd` locates that boundary while ignoring separators nested in calls, literals, and template substitutions.

Anchoring on the plugin slug also makes this stricter than the pattern it replaces: the bundle contains a second, structurally identical derivation for a different plugin, and the old shape-only pattern could have taken it.

The drift work already merged targets patches that fail loudly — `failed-required` entries that block candidate promotion, and the optional ones surfaced in the patch report. These two are invisible to that process:

- The IAB filter lives in `scripts/lib/`, outside the ASAR patch report, and its warning is emitted during bundled-plugin staging.
- The Computer Use patch is `ciPolicy: opt-in` and gated behind `enableComputerUseUi`, so it never runs in CI and only warns for users who opted in.

Both were still drifting after `main` reached 26.803 parity.

## Upstream DMG

| | |
| --- | --- |

- `scripts/lib/patch-browser-client-iab-socket-scope.js`
- `scripts/lib/patch-browser-client-iab-socket-scope.test.js`
- `scripts/patches/impl/computer-use.js`
- `tests/scripts_smoke.sh`

## How it was tested

- Both patches run against the real 26.803.41515 sources extracted from the DMG: each applies, and re-applying is a no-op.
- The patched Computer Use settings asset parses as ESM (`acorn --module`), and both halves are present in the installed asset.
- `node --test scripts/lib/patch-browser-client-iab-socket-scope.test.js` — 6/6 pass; fixtures moved to the current shapes.
- `bash tests/scripts_smoke.sh` — passes with only `test_launcher_warm_start_recovery` skipped, which fails identically on unmodified `main` in this headless environment.
- Full `./install.sh` on 26.803.41515: verdict `accepted` with no patch-drift warnings, where `main` gives `accepted_with_warnings`.

## Risks and follow-ups

- The Computer Use change is the larger of the two: it replaces a pattern rather than adjusting one, because the React Compiler output has no equivalent anchor. The injection point and the resulting parse were both checked against the real asset.
- Out of scope: nothing in the pipeline parses a patched bundle, so a needle can match, report success, and still emit invalid JavaScript — the tray constructor drift in #1247 was exactly that. A syntax gate over patched bundles would catch that class of failure and deserves its own PR.


ilysenko (approved):
Reviewed current-DMG drift cleanup and exact-head checks; both patches remain fail-soft and current-only.

ilysenko (changes requested):
Thanks for the update. One blocker remains:

- `main` advanced with #1250, and this PR now has a content conflict in `tests/scripts_smoke.sh`. Please update the branch and preserve both #1250's Browser runtime-clone fixture change and this PR's current 26.803 socket-listing fixture, then let the exact-head CI rerun.

jurgenmahn:
Thanks for the review — picking this up now.

Will rebase onto current `main` and resolve the `tests/scripts_smoke.sh` conflict so that #1250's Browser runtime-clone fixture and the 26.803 socket-listing fixture both survive, then let the exact-head CI rerun.

jurgenmahn:
Rebased onto current `main` (d48fa56).

The `tests/scripts_smoke.sh` conflict was on the single-line browser-client fixture. Resolved by taking #1250's line and re-applying only the socket-listing change on top of it, so `setupAtlasRuntime` / the runtime-clone fixture and the 26.803 listing shape both survive. The diff against `main` for that file is now exactly one changed line.

Local: full smoke suite passes, every `*.test.js` under `scripts/`, `tests/`, and `linux-features/` passes, and `git diff --check` is clean.

ilysenko (approved):
Reviewed the rebased current-DMG fix. The prior smoke-fixture conflict is resolved, exact-head checks pass, and both patches remain current-only and fail-soft.

PR #1252, opened 2026-08-07
chore(plugins): drop Chrome plugin patches upstream now ships

jurgenmahn:
## What changed

Removes nine bundled Browser Use plugin patches that ChatGPT.dmg 26.803.41515 made unnecessary. No behaviour change intended — this is drift cleanup, and it is deliberately separate from the two functional patch fixes for the same DMG.

Upstream replaced the plugin's hardcoded, Chrome-only Linux paths, commands, and process names with a data-driven browser registry (`scripts/extension-ids.json`, exposed as `browser.linux.*`) covering Chrome, Chromium, Chrome Beta, Chrome Unstable, Brave, Edge, Opera, and Vivaldi. Nine of this repo's thirteen plugin patch targets existed to add exactly that support.

Each was checked against the current plugin sources before deleting, so these are removals because upstream took the work over — not because a needle stopped matching:

| Patch | Why it is obsolete |
| --- | --- |
| `installManifest.mjs` — native messaging host locations | The registry flatMaps `nativeMessagingManifestDirectories` across every browser, so Brave and Chromium are already covered |
| `check-native-host-manifest.js` — locations | Upstream added a `process.platform === "linux"` branch |
| `check-native-host-manifest.js` — fallback | Same branch, via `resolveLinuxNativeMessagingManifestPath`, which is browser-aware and honours `XDG_CONFIG_HOME` |
| `installed-browsers.js` — browser inventory | `KNOWN_BROWSERS` is gone; the list is built from `extension-ids.json` |
| `chrome-is-running.js` — running-process detection | `CHROME_PROCESS_NAMES_BY_PLATFORM` is gone; detection reads `browser.linux.processNames` |
| `check-extension-installed.js` — extension-aware profile fallback | Replaced by the shared browser-aware user data directory resolver |
| `open-chrome-window.js` — default-browser profile fallback | Same resolver |
| `open-chrome-window.js` — browser window command | Resolved from `browser.linux.commands` with a `commandPath` availability check |
| `browser-client.mjs` — site_status allowlist fallback | `throwIfBlocksUrl` now wraps the check in `try { … } catch { log; return }`, so a failed site_status check already fails open |

The last row is the one worth a reviewer's attention, since it is the only removal with a security-relevant shape. Upstream now does:

```js
try{c=await this.isBlocked(t,s,n)}catch(l){ig(l,n);return}
```

where `ig` records `outcome:"error_fail_open", reason:"site_status_unavailable"`. That is the same fail-open the removed patch constructed by hand.

### Re-targeted, not removed

`control-chrome/SKILL.md` — profile launch guard. An earlier revision of this PR deleted it on the claim that the skill was gone. It is not: it ships from the chrome plugin at `plugins/chrome/skills/control-chrome/SKILL.md`, which my DMG extraction never covered. What is true is narrower — the skill text was rewritten, so the anchor drifted.

It is restored and re-targeted. The enumerate-and-match paragraph is dropped, because 26.803.41515 now instructs the opposite (bind the first connected extension instance without calling `agent.browsers.list()` first) and re-adding it would fight upstream guidance. The Linux tab-creation guard is kept, since nothing upstream covers it.

The patcher runs against both the chrome and browser plugins and only chrome ships this skill, so it now skips when the skill directory is absent rather than warning about a file that was never there.

The four running-browser profile preference and resolver patches on `check-extension-installed.js` and `open-chrome-window.js` still apply cleanly and are untouched.

The drift work already on `main` (#1242, #1245, #1247, #1248) targets the ASAR patch report, where drift is surfaced per patch. These live in the bundled-plugin staging path in `scripts/lib/`, which is outside that report — its warnings only appear in the staging log and never reach a patch status.

On current `main`, staging the bundled plugins against 26.803.41515 emits:

```
WARN: Could not find Browser Use site_status allowlist fallback insertion point — leaving browser-client.mjs unchanged
WARN: installManifest.mjs missing patch target for Linux browser native host manifest locations
WARN: check-native-host-manifest.js missing patch target for Linux native host manifest locations
WARN: check-native-host-manifest.js missing patch target for Linux browser native host manifest fallback
WARN: installed-browsers.js missing patch target for Linux browser inventory
WARN: chrome-is-running.js missing patch target for Linux browser running-process detection
WARN: check-extension-installed.js missing patch target for Linux extension-aware browser profile fallback
WARN: open-chrome-window.js missing patch target for Linux default-browser profile fallback
WARN: open-chrome-window.js missing patch target for Linux browser window command
```

None of it blocks promotion, which is exactly why it is worth clearing: a permanently noisy staging log makes real drift easy to miss.

Per the maintainer rule, obsolete drift workarounds are deleted rather than kept behind fallbacks.

## Upstream DMG

| | |
| --- | --- |

- `scripts/lib/patch-chrome-plugin.js`
- `scripts/lib/patch-chrome-plugin.test.js`
- `tests/scripts_smoke.sh`

## How it was tested

- Each removal verified against the pristine plugin sources extracted from the DMG, confirming upstream provides equivalent or broader Linux coverage.
- `node scripts/lib/patch-chrome-plugin.js <pristine plugin>` — the four surviving patches apply, zero warnings, and a second run reports all four already applied.
- `bash tests/scripts_smoke.sh` — passes with only `test_launcher_warm_start_recovery` skipped, which fails identically on unmodified `main` in this headless environment. Smoke coverage for the removed patches is removed with them; coverage for the surviving four is unchanged.
- Patcher run against both real plugins extracted from the DMG: chrome gets five patches including the re-targeted skill guard, browser gets four, and neither emits a warning.
- Chrome smoke fixture moved to the real 26.803.41515 layout (`control-chrome` with the rewritten text, registry-shaped `extension-ids.json`), with assertions retained for the upstream registry surviving staging and for the re-targeted guard.
- Full `./install.sh` on 26.803.41515: the bundled plugin staging log is free of drift warnings. Merged together with #1251 and #1255 the whole build reports verdict `accepted` with zero patch-drift warnings, every emitted bundle parses, and all three fixes are present in the installed app.
- `git diff --check` clean.

## Risks and follow-ups

- The real risk in a removal PR is deleting something whose upstream replacement is narrower — and the skill guard above is exactly that failure, caught in review. The table is the audit; `open-chrome-window.js` command resolution and the site_status fail-open are the two most worth a second opinion.
- Depends on #1255 (Thorium current-registry migration), which must land first: the Thorium feature's needles keyed on the core patch output removed here.


ilysenko (changes requested):
Thanks for the careful current-DMG audit. I found three blockers:

1. The required Rust and Smoke Tests check is failing because the Thorium feature tests still depend on the legacy core patch shapes removed here. Please land the current-registry Thorium migration as a separate feature-owned PR first, then rebase this cleanup; please do not silence the drift warnings or weaken the assertions.
2. Please update the Chrome smoke fixture to match the actual 26.803.41515 layout and retain assertions for the upstream registry behavior. The pristine DMG ships `chrome/skills/control-chrome/SKILL.md`, while this fixture substitutes `control-in-app-browser` and keeps legacy Chrome-only script shapes.
3. Please remove the extra blank line at EOF reported by `git diff --check` in `scripts/lib/patch-chrome-plugin.js`.

jurgenmahn:
Thanks for the careful review — all three blockers accepted, and the second one is a straight mistake on my side.

**2. `control-chrome/SKILL.md`.** You are right, and my justification for that removal was wrong. I extracted only `plugins/browser/skills/` from the DMG, saw `control-in-app-browser`, and concluded the skill was gone. It ships from the chrome plugin — `plugins/chrome/skills/control-chrome/SKILL.md` — which my extraction never looked at.

What is true is narrower: the needle no longer matches, because the skill text was rewritten and ``Use the browser bound to `browser` for tasks in this skill.`` is no longer present. So that patch is drifted and needs re-targeting, not deleting. I will restore it, re-target it against the current skill, and fix the Chrome smoke fixture to the real 26.803.41515 layout instead of substituting `control-in-app-browser`.

**1. Thorium.** Understood — the current-registry migration lands first as its own feature-owned PR, then this cleanup rebases on top. No warnings silenced and no assertions weakened. Worth noting for that PR: those seven patch targets can collapse into a single `extension-ids.json` browser entry, the same shape upstream now uses for Chrome, Brave, Edge, Opera, and Vivaldi.

**3.** Trailing blank line will go.

Order I am planning: rebase #1251 first, then the Thorium migration PR, then re-work and rebase this one. I will convert this to draft in the meantime so it is not sitting in your queue.

jurgenmahn:
Parking this until #1255 (the Thorium current-registry migration) is merged, since you asked for that to land first and this cleanup cannot go green before it. The branch stays on my fork; I will reopen with the rebase once #1255 is in.

Closing rather than leaving it as a draft because the contributor limit counts drafts, and it auto-closed #1255 while all three were open.

PR #1256, opened 2026-08-07
feat(thorium): register Thorium in the current browser registry

jurgenmahn:
Feature-owned prerequisite for #1252, as requested in review there.

## What changed

`thorium-chrome-plugin` extends the bundled Chrome plugin to recognise Thorium. It did that with seven patches against Chrome-only constants spread over the plugin's scripts. ChatGPT.dmg 26.803.41515 replaced those constants with a browser registry in `scripts/extension-ids.json`, read as `browser.linux.*`, so all seven targets drifted at once:

```
WARN: installManifest.mjs missing patch target for Thorium native host manifest location
WARN: check-native-host-manifest.js missing patch target for Thorium native host manifest fallback
WARN: installed-browsers.js missing patch target for Thorium browser inventory
WARN: chrome-is-running.js missing patch target for Thorium running-process detection
WARN: check-extension-installed.js missing patch target for Thorium extension-aware browser profile fallback
WARN: open-chrome-window.js missing patch target for Thorium default-browser profile fallback
WARN: open-chrome-window.js missing patch target for Thorium browser window command
```

That is reproducible on current `main`: run `scripts/lib/patch-chrome-plugin.js` and then the feature's patcher against the plugin extracted from the current DMG, and six of the seven miss, with the seventh unable to find `installManifest.mjs` at all. The feature's own tests stayed green because their fixtures still described the pre-26.803 script shapes.

Registering one Thorium entry in `browserDiagnostics` replaces all seven, because that registry is now where upstream reads each of them from:

| Old patch target | Registry field |
| --- | --- |
| native host manifest location + fallback | `linux.nativeMessagingManifestDirectories` |
| browser inventory | the entry itself (`displayName`, `linux.commands`) |
| running-process detection | `linux.processNames` |
| extension-aware profile fallback | `linux.userDataDirectorySegments` |
| default-browser profile fallback | `linux.userDataDirectorySegments` |
| browser window command | `linux.commands` |

Thorium inherits Chrome's `extensionIds` and `storeUrl` from the registry's own Chrome entry rather than hardcoding them, because it is a Chromium fork loading the same extension build. `macos` and `windows` blocks are populated too, since `installed-browsers.js` maps them unconditionally when it builds its list.

The settings/status patch (`patch.js`) and the stage hook's manifest-path handling are unchanged — only the plugin patcher moves.

## Why it is a separate PR

Requested in the #1252 review: the current-registry Thorium migration lands first as a feature-owned change, then that cleanup rebases on top. The two are otherwise coupled, because #1252 removes the core patches whose output the old Thorium needles keyed on.

Worth stating plainly: this is a fix, not just a rebase enabler. On `main` today the feature is inert against the current DMG — enabling it changes nothing about how Thorium is detected.

## Upstream DMG

| | |
| --- | --- |

- `linux-features/thorium-chrome-plugin/patch-chrome-plugin.js`
- `linux-features/thorium-chrome-plugin/test.js`
- `linux-features/thorium-chrome-plugin/README.md`

## How it was tested

- Patcher run against the chrome plugin extracted from the current DMG: registers Thorium with the expected `linux` fields, and a second run reports it already patched.
- `node --test linux-features/thorium-chrome-plugin/test.js` — 6/6 pass. Fixtures moved to the current registry shape; the staging test asserts the registry entry and the inherited extension identity instead of seven per-script edits.
- Every `*.test.js` under `scripts/`, `tests/`, and `linux-features/` passes.
- `bash tests/scripts_smoke.sh` passes with only `test_launcher_warm_start_recovery` skipped, which fails identically on unmodified `main` in this headless environment.
- `git diff --check` clean.

## Risks and follow-ups

- The feature stays disabled by default; `features.example.json` is untouched.
- The registry entry is the whole contract now, so a wrong field is a silent behaviour change rather than a warning. The Linux values are carried over verbatim from the patches they replace; the macOS and Windows blocks are new and follow the shape of the sibling Chromium-fork entry, since nothing on those platforms consumes them for Thorium today.
- Verified on x86_64 Ubuntu, source install path.

### DIFF
--- a/linux-features/thorium-chrome-plugin/README.md
+++ b/linux-features/thorium-chrome-plugin/README.md
@@ -18,9 +18,16 @@ core Linux port does not regularly test. Enable it by adding the feature id to
 When enabled, the feature:
 
 - adds Thorium native-messaging manifest locations for the generated launcher
-- patches the staged Chrome plugin scripts to detect Thorium installs, profiles,
-  running processes, default-browser desktop IDs, and launch commands
+- registers Thorium in the staged Chrome plugin's browser registry
+  (`scripts/extension-ids.json`), which is where upstream reads install
+  commands, process names, profile directories, and native-messaging manifest
+  locations from
 - adds Thorium to the Electron-side Chrome extension settings/status helper
+
+Thorium inherits the Chrome extension identity, because it loads the same
+extension build. Upstream moved this contract into the registry in
+26.803.41515; before that the same coverage needed seven separate patches
+against Chrome-only constants in individual plugin scripts.
 
 Run the focused tests with:
 
--- a/linux-features/thorium-chrome-plugin/patch-chrome-plugin.js
+++ b/linux-features/thorium-chrome-plugin/patch-chrome-plugin.js
@@ -4,69 +4,133 @@
 const fs = require("node:fs");
 const path = require("node:path");
 
+const THORIUM_BROWSER_FAMILY = "thorium";
+
+// 26.803.41515 replaced the per-script Chrome-only Linux constants with a
+// browser registry in extension-ids.json. Adding Thorium there covers manifest
+// locations, the browser inventory, running-process detection, profile
+// resolution, and window command selection in one entry, which is what the
+// seven source patches this replaces did by hand.
+const THORIUM_BROWSER_ENTRY = {
+  browserFamily: THORIUM_BROWSER_FAMILY,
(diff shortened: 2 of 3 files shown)


ilysenko (approved):
Reviewed the current head. This is an isolated, disabled-by-default Thorium feature migration to the current browser registry; required checks are green and no merge blockers remain.

PR #1258, opened 2026-08-08
Recognize DankMaterialShell polkit agent

Cameudis:
## Summary

- Recognize the DankMaterialShell QuickShell process as a graphical polkit authentication agent.
- Cover the DMS process path and an unrelated QuickShell path in the updater matching test.

## User-visible behavior

niri sessions using the DMS built-in polkit agent no longer fall back to the manual sudo install path.

## Validation

- `cargo test -p codex-update-manager polkit_agent_process_matching_ignores_daemon_and_matches_agents`
- `git diff --check`

### DIFF
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -45,6 +45,7 @@ const POLKIT_AUTH_AGENT_PROCESS_TOKENS: &[&str] = &[
     "polkit-dde-agent",
     "polkit-gnome-authentication-agent",
     "polkit-kde-authentication-agent",
+    "quickshell/dms",
     "soteria",
     "ukui-polkit",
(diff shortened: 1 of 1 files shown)


ilysenko:

ilysenko (changes requested):
Thanks for adding DMS polkit-agent detection. Please handle the supported `DMS_DISABLE_POLKIT=1` mode before this is merged. The current unconditional `quickshell/dms` match treats the process as an active agent even when DMS has disabled the agent, so the updater can enter the `pkexec` flow and wait/retry instead of using the manual sudo fallback. Please make the DMS match respect the disabled-agent case and add a regression test for it.

PR #1260, opened 2026-08-08
fix(linux): stage parcel watcher for git repository watches

ahatasham5:
## Summary

- Stage the upstream-declared `@parcel/watcher` package and its Linux binding while rebuilding the Linux app.
- Install it in an isolated temporary npm prefix so the extracted upstream app package manifest, which contains `workspace:*` dependencies, is never parsed by npm.

## Root cause

The current upstream app imports `@parcel/watcher` for Git repository watches, but the Linux package did not include it. Installing it directly into the extracted app failed because npm does not support the app manifest’s `workspace:*` dependencies in this standalone context.

## Validation

- `bash -n scripts/lib/native-modules.sh`
- Rebuilt and launched the Debian package locally; the Git watcher `ERR_MODULE_NOT_FOUND` is absent from the launcher log.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -292,6 +292,7 @@ jobs:
             .#codex-desktop-computer-use-ui
             .#codex-desktop-remote-mobile-control
             .#codex-desktop-computer-use-ui-remote-mobile-control
+            .#checks.x86_64-linux.parcel-watcher-staged-runtime
             .#checks.x86_64-linux.nix-pipewire-alsa-wrapper
             .#checks.x86_64-linux.nix-gsettings-schema-wrapper
             .#checks.x86_64-linux.watchdog-linux-features
--- a/.github/workflows/update-codex-hash.yml
+++ b/.github/workflows/update-codex-hash.yml
@@ -35,6 +35,7 @@ jobs:
         .#codex-desktop-computer-use-ui
         .#codex-desktop-remote-mobile-control
         .#codex-desktop-computer-use-ui-remote-mobile-control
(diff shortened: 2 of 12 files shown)


ilysenko (changes requested):
Thanks for the update. I found three blockers:

- Please stage the complete `@parcel/watcher` runtime dependency closure in the Nix/prebuilt path and add a test that loads the watcher from the staged app tree; currently only the `@parcel` scope is copied.
- Please add `@parcel/watcher` to the current-DMG Nix pin validation/refresh workflow so its version cannot silently drift.
- Please revert or split the unrelated Chrome plugin synchronization change into a separate focused PR with its own rationale and validation.

ahatasham5:
Thanks for the review — I addressed all three blockers:

- The Nix/prebuilt path now stages the complete `@parcel/watcher` runtime dependency closure, and both the Nix check and smoke coverage load the watcher from the staged app tree.
- `@parcel/watcher` is now part of current-DMG Nix pin validation and refresh. The validator reads the exact version from the app manifest because the current DMG declares the watcher there but does not bundle its package directory in the ASAR.
- I reverted the unrelated Chrome plugin synchronization change, so the final PR diff contains no Chrome synchronization changes.

Validation is green: Nix Package Builds, Rust and Smoke Tests, Build App Against Upstream DMG, and the Debian/RPM/Pacman package builds all pass.

ilysenko (approved):
Reviewed the current head after the requested fixes. The complete watcher runtime closure is staged and runtime-tested, current-DMG pin refresh/validation is wired, the unrelated Chrome scope is gone, and all required checks are green.

PR #1263, opened 2026-08-08
chore(plugins): drop Chrome plugin patches upstream now ships

jurgenmahn:
Replaces #1252, which was closed while its prerequisite was in flight. #1256 (Thorium current-registry migration) has since merged, so this rebases onto it cleanly.

## What changed

Removes nine bundled Browser Use plugin patches that ChatGPT.dmg 26.803.41515 made unnecessary. No behaviour change intended — this is drift cleanup, and it is deliberately separate from the two functional patch fixes for the same DMG.

Upstream replaced the plugin's hardcoded, Chrome-only Linux paths, commands, and process names with a data-driven browser registry (`scripts/extension-ids.json`, exposed as `browser.linux.*`) covering Chrome, Chromium, Chrome Beta, Chrome Unstable, Brave, Edge, Opera, and Vivaldi. Nine of this repo's thirteen plugin patch targets existed to add exactly that support.

Each was checked against the current plugin sources before deleting, so these are removals because upstream took the work over — not because a needle stopped matching:

| Patch | Why it is obsolete |
| --- | --- |
| `installManifest.mjs` — native messaging host locations | The registry flatMaps `nativeMessagingManifestDirectories` across every browser, so Brave and Chromium are already covered |
| `check-native-host-manifest.js` — locations | Upstream added a `process.platform === "linux"` branch |
| `check-native-host-manifest.js` — fallback | Same branch, via `resolveLinuxNativeMessagingManifestPath`, which is browser-aware and honours `XDG_CONFIG_HOME` |
| `installed-browsers.js` — browser inventory | `KNOWN_BROWSERS` is gone; the list is built from `extension-ids.json` |
| `chrome-is-running.js` — running-process detection | `CHROME_PROCESS_NAMES_BY_PLATFORM` is gone; detection reads `browser.linux.processNames` |
| `check-extension-installed.js` — extension-aware profile fallback | Replaced by the shared browser-aware user data directory resolver |
| `open-chrome-window.js` — default-browser profile fallback | Same resolver |
| `open-chrome-window.js` — browser window command | Resolved from `browser.linux.commands` with a `commandPath` availability check |
| `browser-client.mjs` — site_status allowlist fallback | `throwIfBlocksUrl` now wraps the check in `try { … } catch { log; return }`, so a failed site_status check already fails open |

The last row is the one worth a reviewer's attention, since it is the only removal with a security-relevant shape. Upstream now does:

```js
try{c=await this.isBlocked(t,s,n)}catch(l){ig(l,n);return}
```

where `ig` records `outcome:"error_fail_open", reason:"site_status_unavailable"`. That is the same fail-open the removed patch constructed by hand.

### Re-targeted, not removed

`control-chrome/SKILL.md` — profile launch guard. An earlier revision of this PR deleted it on the claim that the skill was gone. It is not: it ships from the chrome plugin at `plugins/chrome/skills/control-chrome/SKILL.md`, which my DMG extraction never covered. What is true is narrower — the skill text was rewritten, so the anchor drifted.

It is restored and re-targeted. The enumerate-and-match paragraph is dropped, because 26.803.41515 now instructs the opposite (bind the first connected extension instance without calling `agent.browsers.list()` first) and re-adding it would fight upstream guidance. The Linux tab-creation guard is kept, since nothing upstream covers it.

The patcher runs against both the chrome and browser plugins and only chrome ships this skill, so it now skips when the skill directory is absent rather than warning about a file that was never there.

The four running-browser profile preference and resolver patches on `check-extension-installed.js` and `open-chrome-window.js` still apply cleanly and are untouched.

The drift work already on `main` (#1242, #1245, #1247, #1248) targets the ASAR patch report, where drift is surfaced per patch. These live in the bundled-plugin staging path in `scripts/lib/`, which is outside that report — its warnings only appear in the staging log and never reach a patch status.

On current `main`, staging the bundled plugins against 26.803.41515 emits:

```
WARN: Could not find Browser Use site_status allowlist fallback insertion point — leaving browser-client.mjs unchanged
WARN: installManifest.mjs missing patch target for Linux browser native host manifest locations
WARN: check-native-host-manifest.js missing patch target for Linux native host manifest locations
WARN: check-native-host-manifest.js missing patch target for Linux browser native host manifest fallback
WARN: installed-browsers.js missing patch target for Linux browser inventory
WARN: chrome-is-running.js missing patch target for Linux browser running-process detection
WARN: check-extension-installed.js missing patch target for Linux extension-aware browser profile fallback
WARN: open-chrome-window.js missing patch target for Linux default-browser profile fallback
WARN: open-chrome-window.js missing patch target for Linux browser window command
```

None of it blocks promotion, which is exactly why it is worth clearing: a permanently noisy staging log makes real drift easy to miss.

Per the maintainer rule, obsolete drift workarounds are deleted rather than kept behind fallbacks.

## Upstream DMG

| | |
| --- | --- |

- `scripts/lib/patch-chrome-plugin.js`
- `scripts/lib/patch-chrome-plugin.test.js`
- `tests/scripts_smoke.sh`

## How it was tested

- Each removal verified against the pristine plugin sources extracted from the DMG, confirming upstream provides equivalent or broader Linux coverage.
- `node scripts/lib/patch-chrome-plugin.js <pristine plugin>` — the four surviving patches apply, zero warnings, and a second run reports all four already applied.
- `bash tests/scripts_smoke.sh` — passes with only `test_launcher_warm_start_recovery` skipped, which fails identically on unmodified `main` in this headless environment. Smoke coverage for the removed patches is removed with them; coverage for the surviving four is unchanged.
- Patcher run against both real plugins extracted from the DMG: chrome gets five patches including the re-targeted skill guard, browser gets four, and neither emits a warning.
- Chrome smoke fixture moved to the real 26.803.41515 layout (`control-chrome` with the rewritten text, registry-shaped `extension-ids.json`), with assertions retained for the upstream registry surviving staging and for the re-targeted guard.
- Full `./install.sh` on 26.803.41515, rebased onto current `main`: verdict `accepted` with zero patch-drift warnings.
- Local mirror of the CI job (shell syntax, `scripts/ci/run-node-checks.sh`, cargo fmt/clippy/check/test, `git diff --check`, smoke suite): everything passes except three tests that fail identically on unmodified `main` in this environment — `appshots` XInput2 (no real X input stream) and two `directory-only-working-tree-watch` cases (inotify `ENOSPC`) — plus the launcher warm-start smoke test, which needs a non-headless Electron.
- `git diff --check` clean.

## Risks and follow-ups

- The real risk in a removal PR is deleting something whose upstream replacement is narrower — and the skill guard above is exactly that failure, caught in review. The table is the audit; `open-chrome-window.js` command resolution and the site_status fail-open are the two most worth a second opinion.
- The Thorium coupling that blocked the previous revision is resolved: #1256 moved that feature onto the browser registry, so it no longer keys on the core patch output removed here. Its test suite passes on this branch (6/6).


ilysenko (approved):
Reviewed the final current-DMG cleanup, fail-soft behavior, and cross-format validation coverage.

PR #1267, opened 2026-08-09
Fix Chrome app-server registry path lifetime

### LINKED ISSUE #1266: Chrome app-server registry paths do not survive managed plugin cache replacement
## Summary

On Linux, the Chrome plugin app-server registration stores `extensionHostPath` and `browserClientPath` through a `.codex-linux-runtime` symlink inside the app-server-managed installed plugin cache. When that cache root is replaced during plugin reconciliation, the symlink disappears while the v2 registry entry still points to it.

The trusted runtime copy under `plugins/linux-runtime-cache/.../latest` remains intact, but the Chrome side panel can no longer find a compatible app-server entry and shows:

```text
Open or update the ChatGPT desktop app
No compatible Codex app-server entry was found
```

This reproduces with Google Chrome. It is not a request for an additional Chromium-family browser integration.

## Environment

- Source/test reproduction: latest `main`, `6192a9c554d2166b94e7aba4dcfacad52199670e`
- Original end-to-end Chrome observation: wrapper baseline `de575d0`, using the same current DMG; the affected runtime helper is unchanged on `6192a9c`
- Upstream Desktop DMG: `26.803.41515`
- Distribution: Arch Linux, x86_64
- Desktop/session: KDE Plasma, Wayland
- Installation method used for the end-to-end observation: pacman package
- Optional Linux Features: disabled (`linux-features/features.json` absent)
- Browser: Google Chrome with the official ChatGPT extension

## Reproduction

1. Build the current supported DMG from the latest repository `main` and start ChatGPT Desktop.
2. Let the bundled Chrome plugin register its app-server entry.
3. Observe that `codexLinuxChromePluginRuntimeConfig()` validates the durable runtime under:

   ```text
   $CODEX_HOME/plugins/linux-runtime-cache/openai-bundled/chrome/latest
   ```

   but returns a bridge under:

   ```text
   $CODEX_HOME/plugins/cache/openai-bundled/chrome/.codex-linux-runtime
   ```

4. Replace or remove the managed installed plugin cache, as plugin reconciliation may do.
5. Inspect `chrome-native-hosts-v2.json` or retry the Chrome side panel connection.

A focused regression test against `6192a9c` reproduces the path-lifetime failure before the fix: 11 tests pass and the two durability assertions fail. The actual registered path is under `plugins/cache/.../.codex-linux-runtime`; the expected path is under `plugins/linux-runtime-cache/.../latest`. Removing the installed cache makes both registered host/client paths disappear even though the validated runtime files still exist.

## Expected behavior

The app-server registry should use the already validated durable Linux runtime path. Replacing the installed plugin cache should not invalidate `extensionHostPath` or `browserClientPath`.

## Actual behavior

The registry entry follows a bridge stored inside the replaceable installed cache. Cache reconciliation can therefore leave a syntactically valid registry entry whose host and browser client paths no longer exist.

## Root cause

`codexLinuxChromePluginRuntimeConfig()` already verifies all of the following for the durable runtime:

- the `latest` link is owned by the current user;
- it resolves to the installed plugin version;
- the ancestor chain and runtime tree are owner-trusted and not group/world writable;
- the architecture-specific extension host exists and is executable;
- extension IDs are valid.

After those checks, it creates a second bridge inside the managed installed cache and returns that bridge as `pluginRoot`. The second bridge shortens the lifetime of otherwise durable paths without adding another trusted runtime.

## Related work

- #1178 identified the browser-independent cache-lifetime class while investigating Helium. The maintainer requested that it be separated and demonstrated on current `main` with Chrome, Brave, or Chromium; this report uses Google Chrome.
- #1197 moved the Linux runtime outside the app-server-managed plugin cache, but retained the cache-local registration bridge.
- #1230 refreshes stale same-version installed plugin caches. That repairs cached contents, but it does not make registry paths survive replacement of the cache containing the bridge.

## Proposed regression coverage

- Register host/client paths from the trusted durable runtime.
- Remove the installed plugin cache and verify both registered paths remain readable.
- Verify the same behavior when `CODEX_HOME` is exposed through a symlink.
- Retain the existing owner, permission, version, executable, and extension-ID rejection tests.

NaruSora:
## Summary

Fixes #1266.

- register the Chrome app-server host and browser client through the validated `plugins/linux-runtime-cache/.../latest` runtime so managed plugin-cache reconciliation cannot invalidate v2 registry or native-messaging manifest paths
- extend Linux current-DMG explicit removal across both lifecycle resources: the matching durable v2 registry entry and the native-messaging manifest that points to the durable host
- preserve existing managed-cache cleanup, non-Linux behavior, unrelated paths/native hosts, and cross-plugin cache roots
- cover production-shaped `Sq() -> nJ() -> rJ()` manifest removal and `Eq() -> kq()` registry removal with normal and symlinked `CODEX_HOME`

Registration, cache reconciliation, and explicit plugin removal have different ownership rules. Registration uses the version-matched and validated durable runtime; reconciliation may replace `plugins/cache/...` without invalidating that registration. Explicit disable/uninstall now maps only the exact Linux `openai-bundled/chrome` managed root through the real `CODEX_HOME`, reuses the upstream containment predicate against the corresponding durable root, and removes both lifecycle resources.

Source-of-truth files changed:

- `scripts/patches/impl/chrome-plugin.js`
- `scripts/patches/impl/chrome-plugin.test.js`
- `scripts/patches/test-fixtures/current-dmg.js`

No generated app or package output is committed.

## Validation

- registry cleanup test-first reproduction on `210609e`: 12/14 passed; normal and symlinked durable registry entries were retained
- manifest cleanup test-first reproduction on `c1e507f`: 13/15 passed; normal and symlinked durable native-messaging manifests were retained
- `node scripts/patches/impl/chrome-plugin.test.js` — 15/15 passed
- `node --check` — passed for all three changed files
- `git diff --check` — passed
- final `app.asar` inspection confirmed the patched `Sq() -> nJ() -> rJ()` manifest chain, the `Eq() -> kq()` registry chain, and exactly one durable-path helper
- `make MAX_BUILD_THREADS=4 build-app` — passed against upstream Desktop `26.803.41515` (`f8a5a74acb38aab4a59ac0ad7fef3bb4bb4f280a7ed25f3cb7ef6145df5e8747`); acceptance verdict `accepted`, zero blockers, zero warnings
- fresh independent `codex review --uncommitted` with `gpt-5.6-sol`, `xhigh`, read-only — no correctness issues found; it reviewed the final diff, both cleanup lifecycles, real `app.asar` output, and patch idempotency
- required GitHub CI for `baba642` — current-DMG, Debian, RPM, Pacman, Nix, and Rust/Smoke all passed; drift reconciliation skipped as designed

Build-time `npm audit` still reports the existing `1 moderate severity vulnerability`; this change adds no dependency.

## Checklist

- [x] I independently reviewed the final diff with my own model at maximum reasoning effort; no correctness findings remained.


ilysenko (changes requested):
Thanks for the focused durability fix. One blocker remains:

- The current DMG removes v2 registry entries only when `extensionHostPath` is lexically inside the managed plugin cache. This PR moves that path to `plugins/linux-runtime-cache/.../latest`, so disabling or uninstalling the Chrome plugin no longer removes its registry entry. Please preserve the durable path for reconciliation while extending the Linux current-DMG patch and regression coverage so explicit plugin removal also deletes the corresponding durable entry, including a symlinked `CODEX_HOME`.

NaruSora:
Addressed the requested removal lifecycle in `c1e507f`:

- kept the durable `linux-runtime-cache/.../latest` registry paths across managed plugin-cache reconciliation
- extended the Linux current-DMG removal predicate to delete the corresponding durable v2 entry during explicit Chrome plugin disable/uninstall
- preserved legacy managed-cache cleanup, unrelated plugin/native-host entries, and non-Linux behavior
- added regression coverage for normal and symlinked `CODEX_HOME`, cross-plugin scope, and fail-closed path resolution

I also verified against the current-DMG bundle that the patched cleanup predicate is called only through the explicit-removal chain `Sq() -> Eq() -> kq()`, so reconciliation does not use the broadened predicate.

Validation is green: Chrome patch tests pass 14/14, the current-DMG build is accepted with zero blockers or warnings, and all required GitHub Actions checks pass. Could you take another look?

ilysenko (changes requested):
Thanks for addressing the registry cleanup feedback. One lifecycle blocker remains:

- Registration now writes the Chrome native-messaging manifest with the durable `extensionHostPath`, but current-DMG explicit removal discovers manifests only when their `path` is inside `pluginCacheRoot` (`Sq -> nJ -> rJ`). The durable path is outside that root, so disable/uninstall removes the registry entry but leaves a functional manifest pointing to the durable host. Please extend the Linux current-DMG cleanup to recognize and remove the matching durable manifest as well, preserving managed-cache, non-Linux, and cross-plugin behavior, and add production-shaped `Sq/nJ/rJ` coverage including symlinked `CODEX_HOME`.

Before pushing the next revision, please also run a fresh independent code review with your own model against the final diff, focusing on both registry-entry cleanup and native-messaging manifest cleanup lifecycles.

NaruSora:
Addressed the remaining native-messaging manifest lifecycle blocker in `baba642` without rewriting the earlier reviewed commits:

- kept the original managed-cache manifest/registry cleanup paths
- generalized the validated durable-root containment to a path-level helper while retaining the registry wrapper
- passed `codexHome` through the production `Sq() -> nJ() -> rJ()` chain so explicit Linux Chrome removal also discovers and deletes manifests pointing to the corresponding durable host
- kept the durable extension host itself under wrapper staging ownership; this change removes the manifest and registry entry, not the durable runtime cache
- preserved non-Linux behavior, exact `openai-bundled/chrome` scope, cross-plugin isolation, unrelated paths/native hosts, and fail-closed realpath handling

The production-shaped regression moved from 13/15 on `c1e507f` to 15/15 and covers managed plus durable manifests, non-Linux behavior, cross-plugin scope, and symlinked `CODEX_HOME`. The existing registry lifecycle coverage remains green.

Before pushing, I ran a fresh independent read-only review with Codex `gpt-5.6-sol` at `xhigh` effort against the final diff, explicitly covering both registry-entry and native-messaging manifest lifecycles. It reported no correctness issues and also checked the real patched `app.asar` and patch idempotency.

The current-DMG `26.803.41515` build is accepted with zero blockers or warnings, and all required CI for `baba642` is green (current-DMG, Debian, RPM, Pacman, Nix, Rust/Smoke; drift reconciliation skipped as designed). Could you take another look?

ilysenko (approved):
Reviewed the current head after both requested lifecycle fixes. Durable registry and native-messaging manifest cleanup are now covered, including symlinked CODEX_HOME and behavior boundaries. All relevant checks are green; no blockers remain.

PR #1274, opened 2026-08-09
fix(updater): read sparkleManager from this.sparkleManager in getAppUpdateViewState

appotry:
## Problem

The upstream Zoe class (app-update state service) stores the sparkleManager in `this.sparkleManager` (a constructor argument), but `getAppUpdateViewState()` reads `this.options.sparkleManager`. On macOS the base class options happens to carry the field, so the bug is hidden; on Linux it is `undefined` and the app crashes with `this.options.sparkleManager.getDownloadProgressPercent is not a function`, leaving the app stuck on the splash screen.

## Fix

- `applyLinuxAppUpdateViewStatePatch`: rewrite `getAppUpdateViewState` to read `this.sparkleManager` (the field actually stored in the constructor) with null-safe fallbacks
- scan bundles containing `getAppUpdateViewState` in addition to the existing menu/bridge markers
- regression test: with a manager the state is returned; without one the method returns safe defaults instead of throwing

## Testing

- `node --test scripts/patch-linux-window-ui.test.js`: 415 pass
- Real app.asar extraction + patch applied, then launched: `getDownloadProgressPercent is not a function` no longer appears, app reaches main UI

### DIFF
--- a/scripts/lib/linux-update-bridge-patch.js
+++ b/scripts/lib/linux-update-bridge-patch.js
@@ -121,6 +121,24 @@ function applyLinuxAppUpdaterMenuPatch(currentSource) {
   return currentSource.replace(menuRegex, "$1=$2.$3.shouldIncludeSparkle($4,process.platform,process.env)||process.platform===`linux`");
 }
 
+function applyLinuxAppUpdateViewStatePatch(currentSource) {
+  // 上游 bug: Zoe 类构造时把 sparkleManager 存入 this.sparkleManager，
+  // 但 getAppUpdateViewState() 读取 this.options.sparkleManager。
+  // macOS 上基类 options 恰好也有 sparkleManager，Linux 上没有，导致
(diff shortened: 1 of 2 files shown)


ilysenko:
Thank you for the contribution. In its current form, we do not think this change is needed.

The reason is straightforward: current `main` already injects a Linux updater manager that implements all six required methods. This PR patches a different class shape that does not actually exist in the current DMG and only hides a potential error by returning safe `null`, `false`, and `idle` values.

In practical terms:

- there is no demonstrated benefit for a fresh build from current `main`;
- the failure you observed may have come from an older build created before PR #1245;
- this adds another fragile ASAR regex that would need ongoing maintenance whenever the upstream DMG changes;
- if a real regression occurs, this patch could hide it and report an incorrect updater state instead of fixing the underlying cause.

We are therefore closing this PR. If you can reproduce the crash on a fresh build from current `main`, please share the new logs and exact reproduction steps. We can then investigate the actual root cause and consider a focused fix.

Issue #1048, opened 2026-07-17
Side bar UI feels laggy.


basilk15:
Hi, since I recently updated the app to the newer version of codex as per the repo guidelines, the UI feels a bit laggy. Specially the side bar when i am scrolling through the projects. 
Am using linux, ubuntu 26 LTS.

colonelpanic8:
yeah im seeing this too. In particular, the tooltips that show when hovering over a convo/different project seem to cause long lags sometimes.

ilysenko:
Thanks for the report. Could you share a bit more detail so we can reproduce and measure this?

- The exact Codex Desktop build and package format, plus the Build information JSON from Settings > Linux desktop.
- Roughly how many projects and conversations are shown, and whether the lag happens only when a project/conversation tooltip appears.
- A short screen recording with an approximate delay, and any enabled Linux features or custom Electron flags.
- Whether the same lag occurs after fully quitting the app and testing with optional features disabled, on both X11 and Wayland if available.

colonelpanic8:
I can say for sure that it does not happen on:

```
{
  "schemaVersion": 1,
  "generatedAt": "2026-07-15T17:16:35.000Z",
  "appIdentity": {
    "id": "codex-desktop",
    "displayName": "ChatGPT"
  },
  "upstreamDmg": {
    "fileName": "Codex.dmg",
    "sizeBytes": 615731826,
    "sha256": "40e34814e74e30943c209ebd4da94cd4de3581a52c5bffbe2bcf2e488d6361c6",
    "appVersion": "26.707.72221"
  },
  "electronVersion": "42.1.0",
  "source": {
    "commit": "4f0470de99f842b44fb848a169a191352e907cd1",
    "shortCommit": "4f0470de99f8",
    "version": "0.10.1",
    "branch": null,
    "remote": "https://github.com/ilysenko/codex-desktop-linux.git",
    "commitUrl": "https://github.com/ilysenko/codex-desktop-linux/commit/4f0470de99f842b44fb848a169a191352e907cd1",
    "describe": null,
    "dirty": false,
    "provenance": "git"
  },
  "linuxTarget": {
    "summary": "unknown/unknown",
    "distro": {
      "id": "",
      "idLike": [],
      "versionId": "",
      "versionMajor": null,
      "prettyName": ""
    },
    "packageFormat": "unknown",
    "packageManager": "unknown",
    "arch": "x64",
    "desktop": {
      "raw": "",
[...]

PR #1276, opened 2026-08-09
fix(linux): reduce sidebar scroll layout invalidation

nxank4:
## Problem

Continuous scrolling in a large ChatGPT Desktop sidebar stutters badly on Linux. The main renderer was spending almost the entire scroll-event budget updating style/layout rather than running the React handler or painting conversation content.

Related to #1048.

## Root cause

The main sidebar scroll container uses `vertical-scroll-fade-mask`, whose `edge-fade` animation drives `--top-fade` and `--bottom-fade` from `animation-timeline: scroll(self y)`.

On Electron 42 / Chromium on Linux, advancing that scroll timeline causes `UpdateLayoutTree` to revisit essentially the entire sidebar subtree on each scroll delivery, despite the fade properties being declared non-inherited. The worst stack is the semantic sidebar `onScroll` site because the synchronous style update is charged to event dispatch; the React scroll-state update and tooltip dismissal were ruled out by separate A/B traces.

## Fix

Add one optional all-Linux core webview patch that:

- semantically identifies the unique main sidebar container using the `sidebarScroll`, fade-mask, containment, header-mask, ref, and `currentTarget.scrollTop` handler contract;
- disables only that container's scroll-driven animation/timeline;
- preserves the existing React `onScroll` handler and header mask;
- preserves the footer affordance as the same static 40px fade;
- applies once, is idempotent, and leaves bytes unchanged with an explicit warning on ambiguity, partial markers, or upstream drift.

No global animations, GPU features, tooltips, history rows, or integrations are disabled.

## Performance

The DevTools comparison is a same-renderer/same-fixture A/B: both passes used the same process, 5,038px scroll height, 707px viewport, 3,770 sidebar descendants, 4,206 document elements, normal scroll handler, four identical synthesized gestures, and the same 40px footer fade. Only `edge-fade / scroll(self y)` versus `none / auto` changed.

| Metric | Before | After |
|---|---:|---:|
| Renderer CPU average (`pidstat`, clean candidate run) | 75.70% | 30.38% |
| Renderer CPU peak | 107% | 111% |
| Minor faults / second | ~7,394 | 4,461.75 |
| Major faults | 0 | 0 |
| Scroll event average | 38.817 ms | 0.416 ms |
| Scroll event p50 | 39.047 ms | 0.328 ms |
| Scroll event p95 | 46.345 ms | 0.538 ms |
| `UpdateLayoutTree` average | 37.824 ms | 0.236 ms |
| `UpdateLayoutTree` p95 | 45.161 ms | 0.377 ms |
| `UpdateLayoutTree` max | 55.734 ms | 2.959 ms |
| Average invalidated elements | 3,244.87 | 2.97 |
| Maximum invalidated elements | 3,340 | 48 |
| Broad invalidations (>1,000 elements) | 137 | 0 |
| Long tasks (>=50 ms) | 56 | 2 |
| Presented-frame gap p50 | 49.218 ms | 8.298 ms |
| Presented-frame gap p95 | 105.599 ms | 10.224 ms |

The original field baseline was 28.66ms average scroll handling, 28.03ms average `UpdateLayoutTree`, ~2,342 invalidated elements/event, and 75.7% renderer CPU. The controlled A/B above used an intentionally expanded, fixed sidebar DOM and reproduced the same pathology.

The patched renderer accepted 918 scroll deliveries over 8.16s versus 141 over 8.44s before, consistent with no longer blocking input behind ~38ms style updates. Peak one-second process CPU remains noisy because the responsive renderer now services far more events/frames; average CPU fell by 59.9%. RSS remained bounded in the clean patched sample (about 555MB average). The preferred <30% CPU target was missed by 0.38 percentage points, while the frame and layout targets were exceeded by a wide margin.

## Testing

- `node --test --test-isolation=none --test-name-pattern='sidebar|default core patch descriptors' scripts/patch-linux-window-ui.test.js` — 6 passed
- Full `scripts/patch-linux-window-ui.test.js` — new tests pass; five known host-sensitive failures reproduced before this change
- `./scripts/ci-local.sh pr` in the supported Ubuntu 24.04 Docker profile — passed; 1,575 Node tests passed, 2 skipped, all Rust, script-smoke, and package checks passed
- `./scripts/rebuild-candidate.sh` — accepted ChatGPT Desktop 26.803.41515 / Electron 42.3.0 candidate; patch applied exactly once, no required or integrity failures
- `git diff --check` — passed
- Manual Wayland and X11 candidate smoke — conversation/project selection, expand/collapse, active highlight, sidebar scrolling/restoration, keyboard navigation, tooltip, context menu, and search passed
- Wayland and X11 traces both eliminated repeated broad invalidation

## Compatibility

The patch targets a stable semantic construct rather than generated offsets, hash names, or minified symbol names. The matcher requires the co-occurrence of the main-sidebar ownership, fade/containment classes, header mask, ref, and semantic `scrollTop` handler shape. Zero or multiple matches are non-mutating optional drift; partial marker states also fail closed. The patch report therefore exposes upstream changes without corrupting a future asset.

## Trace evidence

The original trace attributed about 97.8% of scroll processing to `UpdateLayoutTree`, with affected-element count and duration correlated at approximately r=0.976. In the controlled A/B, the baseline had 137 broad invalidations and the patch had zero; `UpdateLayoutTree` p95 fell from 45.161ms to 0.377ms. No private conversation text or trace artifact is attached.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -192,12 +192,14 @@ const {
   applyLinuxI18nGatePatch,
   applyLinuxOpaqueWindowsDefaultPatch,
   applyLinuxSettingsSearchVisibilityPatch,
+  applyLinuxSidebarScrollPerformancePatch,
   applyLinuxSkillsListDedupePatch,
   applyLinuxThreadSidePanelNativeTooltipPatch,
   applyLinuxTooltipWindowControlsCollisionPatch,
   applyLinuxWindowControlsSafeAreaPatch,
   applySubagentNicknameMetadataPatch,
   codexLinuxWatchBrowserWebviewAttachment,
+  matchesLinuxSidebarScrollPerformanceContract,
 } = require("./patches/impl/webview/index.js");
 const {
   findCodexRequestWebviewAsset,
@@ -1034,6 +1036,7 @@ test("default core patch descriptors are grouped and unique", () => {
     "linux-projectless-xdg-documents-dir",
     "linux-workspace-root-open-targets",
     "linux-settings-search-visibility",
+    "linux-sidebar-scroll-performance",
     "linux-i18n-gate",
     "automation-schedule-multi-time-rrule",
     "automation-update-eager-tool",
@@ -1141,6 +1144,14 @@ test("default core patch descriptors are grouped and unique", () => {
     descriptors.find((descriptor) => descriptor.id === "linux-terminal-user-path")?.ciPolicy,
     "optional",
   );
+  const sidebarScrollPerformance = descriptors.find(
+    (descriptor) => descriptor.id === "linux-sidebar-scroll-performance",
+  );
+  assert.equal(sidebarScrollPerformance?.ciPolicy, "optional");
+  assert.equal(sidebarScrollPerformance?.pattern.test("app-initial-Biw83Aiz.js"), true);
+  assert.equal(sidebarScrollPerformance?.pattern.test("app-DuLjgNkx.css"), false);
+  assert.equal(sidebarScrollPerformance?.assetMatch(sidebarScrollFixture()), true);
+  assert.equal(sidebarScrollPerformance?.assetMatch("function unrelated(){}"), false);
   assert.equal(
     descriptors.find((descriptor) => descriptor.id === "linux-computer-use-avatar-cursor")?.ciPolicy,
     "optional",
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed the current head. The Linux sidebar performance patch is narrowly scoped, idempotent, fail-soft on drift, and covered by the current-DMG build and regression tests.

PR #1278, opened 2026-08-10
feat(linux-features): add opt-in automation tool schema flatten for strict providers

xjimlinx:
## Summary

Adds a disabled-by-default opt-in Linux feature, `automation-schema-flatten`, that normalizes the `codex_app__automation_update` tool schema at build time so strict OpenAI-compatible providers stop rejecting model requests.

The upstream app emits `automation_update` with a root-level `oneOf` JSON Schema and no top-level `"type": "object"` (`type` is `null`). OpenAI's own Responses endpoint tolerates this, but strict providers that validate function schemas server-side (DeepSeek, Azure structured outputs) return:

```text
Invalid schema for function 'codex_app__automation_update':
schema must be a JSON Schema of 'type: "object"', got 'type: null'
```

Background:

- [openai/codex#37786](https://github.com/openai/codex/issues/37786) — issue filed upstream for the same problem on the DeepSeek official integration
- [openai/codex#30132](https://github.com/openai/codex/issues/30132) — same failure class on Azure, closed as completed but does not cover strict custom providers
- [cc-switch#5447](https://github.com/farion1231/cc-switch/issues/5447) — proxy-side oneOf flattening workaround

## Changes

Only additive, no core code touched:

- `linux-features/automation-schema-flatten/feature.json` — manifest, `defaultEnabled: false`
- `linux-features/automation-schema-flatten/patch.js` — `webview-asset` patch descriptor targeting `app-initial-*.js`; rewrites the `automation_update` emitter line to merge the root `oneOf` variants' `properties` into a top-level `type: "object"` schema, preserving `required` and enabling `additionalProperties`
- `linux-features/automation-schema-flatten/README.md` — usage, testing, known risks
- `linux-features/automation-schema-flatten/test.js` — 6 self-contained tests (unit + framework integration + patch report)
- `README.md` — one row in the Opt-In Linux Features table

The patch is defensive: if upstream renames or restructures the emitter line, it warns and skips instead of failing the build.

## Validation

- `node --test linux-features/automation-schema-flatten/test.js` — 6/6 pass
- Applied to the current upstream bundle (26.803.41515): old emitter line replaced, normalizer present, full bundle passes `node --input-type=module --check`
- End-to-end on Arch Linux: app builds, launches, and DeepSeek (Responses API, `deepseek-v4-flash`) no longer returns the schema 400

## Risks

- The merged wire schema is more permissive than the original `oneOf`; runtime argument validation (`safeParse`) is unchanged and still rejects invalid calls app-side
- Not a substitute for an upstream fix; this only adapts the wire-level schema on Linux builds

## Test instructions

```bash
# in a checkout:
cp linux-features/features.example.json linux-features/features.json
# add "automation-schema-flatten" to "enabled", then:
node --test linux-features/automation-schema-flatten/test.js
./install.sh
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,7 @@ workarounds.
 |---|---|---|---|
 | Record and Replay (alpha) | Opt-in alpha | `record-and-replay` | [Docs](linux-features/record-and-replay/README.md) |
 | Agent Workspaces | Opt-in | `agent-workspace` | [Docs](linux-features/agent-workspace/README.md) |
+| Automation tool schema flatten | Opt-in | `automation-schema-flatten` | [Docs](linux-features/automation-schema-flatten/README.md) |
 | API key model visibility | Opt-in | `api-key-model-visibility` | [Docs](linux-features/api-key-model-visibility/README.md) |
 | API key service tier | Opt-in | `api-key-service-tier` | [Docs](linux-features/api-key-service-tier/README.md) |
 | Linux AppShots | Opt-in | `appshots` | [Docs](linux-features/appshots/README.md) |
--- a/linux-features/automation-schema-flatten/README.md
+++ b/linux-features/automation-schema-flatten/README.md
@@ -1,0 +1,58 @@
+# Flatten automation_update schema for strict providers
+
+The Codex desktop app registers an `automation_update` tool (scheduled
+automations) whose emitted JSON Schema uses a root-level `oneOf` and has no
+top-level `"type": "object"` (`type` is `null`). OpenAI's own Responses
+endpoint tolerates this, but strict OpenAI-compatible providers that validate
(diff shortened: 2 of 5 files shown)


ilysenko (changes requested):
Thanks for the focused opt-in feature and tests. I found a blocking schema-correctness issue: the flattening discards root `$defs` while retaining `$ref` properties, overwrites shared discriminators with the last `oneOf` branch, and unions mutually exclusive `required` fields. As a result, valid view/create/update/delete calls are no longer representable and some providers may reject the dangling references. Please use a semantics-preserving normalization and add executable regression tests against a representative current `automation_update` schema, including `$defs` and every operation variant, rather than checking only the injected source text. Please also validate and document the exact provider schema profile being targeted.

ilysenko:

PR #1281, opened 2026-08-10
fix(linux): reduce app-shell tab layout stalls

### LINKED ISSUE #1279: perf(linux): reduce session-switch stalls from app-shell tab layout work
## Problem

Opening or switching sessions in the Linux desktop app can stall for several seconds when a session has a large rendered conversation or retained Browser / Computer Use tabs. Browser guest views must remain mounted because work can continue while their tab is hidden, so discarding them is not a safe optimization.

## Reproduction / profile

Tested with an isolated instance built from the same Codex DMG as the installed app:

- App: 26.803.41515
- Electron: 42.3.0
- Controlled profiling run: accessibility disabled only in the isolated candidate
- Eight-session baseline sweep: about 2.0–6.8 s to settle
- Worst observed long task: 3.6 s
- Largest sampled document: 6,674 elements
- Browser state remained intact: all guest views stayed connected and hidden as expected

A renderer CPU profile on a heavy switch isolated two app-shell tab hotspots:

- the ResizeObserver overflow callback synchronously reading `scrollWidth` / `clientWidth`: about 423 ms self time
- Framer Motion `measureScroll`, triggered by every tab entering from zero width after a remount: about 361 ms self time

The trace also contained genuine long-conversation render work, so this issue does not claim every source of long-session latency is fixed.

## Root cause

The app-shell tab strip currently combines two expensive behaviors during a session remount:

1. Every tab replays its enter-from-zero width animation, even though the tab is already established.
2. ResizeObserver callbacks synchronously read overflow geometry while that width/layout cascade is still in progress.

This amplifies layout work as the number of retained tabs and mounted content grows.

## Proposed fix

- Disable only the app-shell tab's mount-time enter-from-zero animation.
- Preserve normal animate, exit, drag, close, and overflow-fade behavior.
- Defer and coalesce overflow measurement to at most one `requestAnimationFrame` per label element.
- Keep Browser / Computer Use webviews mounted and connected.
- Implement the Linux patch as optional, semantic, idempotent, and fail-soft on upstream bundle drift.

## Measured result

On the same heavy-session scenario:

- overflow measurement fell from about 423 ms to about 14 ms self time (roughly 97% lower)
- `measureScroll` fell from about 361 ms to about 40 ms with the mount-only change, then dropped below the final profile's top-self-time cutoff after the combined patch
- two overflowing labels still showed both fade indicators after two animation frames
- 3/3 Browser guest views remained connected, with 2 hidden

The end-to-end settle-time sweep was noisy because live session content continued changing, so the PR should report hotspot reductions rather than an unsupported wall-time percentage.

## Rejected approaches

- Reducing conversation virtualization overscan from 2 to 1: no material A/B improvement.
- Broad CSS containment changes: too much behavioral risk for weak evidence.
- Destroying or suspending hidden Browser views: incompatible with ongoing Browser / Computer Use work.

This is separate from #1163, which tracks sustained main-process watcher CPU rather than renderer session-switch layout stalls.

nxank4:
## Summary

Fixes #1279.

Session switches with long conversations and retained Browser / Computer Use tabs can trigger an app-shell tab resize cascade on Linux. Renderer profiling identified two avoidable costs in that cascade:

- every existing tab replays its enter-from-zero-width animation when the tab strip remounts;
- its ResizeObserver callback synchronously reads `scrollWidth` / `clientWidth` while layout is still changing.

This PR adds one optional semantic webview patch that:

- disables only the tab's mount-time enter animation while preserving animate, exit, drag, close, and controls;
- defers and coalesces overflow measurement to one animation frame per label element;
- leaves retained Browser / Computer Use guest views mounted and connected;
- remains idempotent and fails soft if the latest upstream bundle contract drifts or becomes ambiguous.

Source-of-truth changes are limited to the webview patch implementation and descriptor, focused regression tests, and the implementation plan.

## Performance evidence

| Profiled renderer work | Before | After |
| --- | ---: | ---: |
| App-shell overflow callback self time | ~423 ms | ~14 ms |
| Framer Motion `measureScroll` self time | ~361 ms | below the final top-self-time cutoff |

The overflow callback reduction is about 97%. The intermediate mount-only profile reduced `measureScroll` to about 40 ms before the combined patch removed it from the final profile's hot-self list.

The live end-to-end settle-time sweep was noisy because session content continued changing, so this PR intentionally does not claim a wall-time percentage. Long-conversation rendering remains visible in the trace and is outside this focused patch.

## Validation

- `node --test --test-name-pattern='app-shell tab' scripts/patch-linux-window-ui.test.js`
- Transform applied successfully to the current installed `app-initial-*.js` contract; the second application was byte-identical
- Same-DMG isolated candidate accepted with no required patch or integrity failures
- Heavy-session CPU profile repeated after the patch
- Overflow UI checked after two animation frames: 2/2 overflowing labels retained fade indicators
- Browser retention checked: 3/3 guest views remained connected, including 2 hidden views
- `./scripts/ci-local.sh pr`: passed
  - container test phase: 1,580 passed, 0 failed, 2 skipped
  - script smoke suite: passed
  - Debian, RPM, and pacman package matrices: passed
- `git diff --check`
- Independent read-only review found no confirmed defect

The direct host-only patch suite reports 419/424 passing; the same five environment-sensitive tests already fail on the unmodified host setup (cursor bridge timing, native desktop enumeration, bundled-plugin path trust, and two CLI stderr assertions). Those paths pass in the repository's isolated CI container.

## Compatibility and risk

- Targets only the current upstream DMG semantic contract.
- Optional/fail-soft behavior leaves unknown or partial future contracts unchanged.
- Does not change Browser / Computer Use lifetime policy.
- Does not add dependencies or modify generated app artifacts.
- Remaining risk is limited to upstream minified-bundle drift, which is surfaced as an optional patch skip.

## Checklist

- [x] This is not an upstream-drift compatibility fix; the semantic patch intentionally targets only the latest `CODEX.DMG`.
- [x] I added or updated relevant tests, ran the validation listed above, and confirmed that required local and GitHub CI checks pass.

### DIFF
--- a/.agents/plans/2026-08-10-linux-app-shell-tab-layout-performance.md
+++ b/.agents/plans/2026-08-10-linux-app-shell-tab-layout-performance.md
@@ -1,0 +1,59 @@
+# Linux app-shell tab layout performance
+
+## Goal
+
+Reduce stalls when opening or switching to long/running conversations with
+Browser tabs, without suspending Browser Use or Computer Use workloads.
+
+## Evidence and scope
+
+- An isolated same-DMG baseline reproduced 2.0-6.8 second settle times, a 3.6
+  second long task, and up to 6,674 DOM elements across eight session switches.
+- A focused heavy-session CPU profile attributed about 423 ms to the app-shell
+  tab overflow callback and 361 ms to Framer Motion `measureScroll` work.
+- App-shell tabs entered from zero width on every tab-strip mount. Their label
+  ResizeObserver synchronously read `scrollWidth` during the same resize/layout
+  cascade.
+- Retained hidden Browser guest views are intentional for active Browser Use.
+  Destroying or suspending them is out of scope.
+- Reducing thread overscan and adding per-block CSS containment were measured
+  against the same workload and rejected because they did not materially help.
+
+## Implementation
+
+1. Add one optional all-Linux webview-asset patch for the semantic app-shell tab
+   contract in the current `app-initial-*.js` bundle.
+2. Disable only the tab's enter-from-zero-width animation when a tab strip is
+   mounted; preserve its exit animation, drag behavior, and tab controls.
+3. Defer and coalesce each label's overflow read to one animation-frame callback
+   so it no longer forces layout during React/ResizeObserver delivery.
+4. Keep the overflow fade state and retained Browser webviews intact.
+5. Match one complete contract, remain idempotent, and fail soft on upstream
+   drift, ambiguity, or a partial patch.
+
+## Validation
+
+- Lock the transform with fixture tests for the complete contract, idempotence,
+  semantic rejection, ambiguity, and drift.
+- Rebuild from the same upstream DMG and require an accepted patch report.
+- Profile a light-to-heavy switch in an isolated candidate and verify the two
+  targeted layout-read hotspots fall materially.
+- Check after two animation frames that overflowing tab labels retain their fade
+  and all retained Browser webviews remain connected.
+- Run focused and full patch tests, smoke tests, local PR CI, diff checks, and an
+  independent read-only review before publication.
+
+## Publication
+
+- Correct issue #1279 with the confirmed resize/layout root cause and sanitized
+  measurements.
+- Commit only implementation, descriptor, tests, and this plan.
+- Push a focused branch, open one draft PR, and inspect CI to terminal status.
+
+## Stop conditions
+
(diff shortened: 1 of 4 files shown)


ilysenko (approved):
Reviewed the current head. The focused Linux webview patch is fail-soft, idempotent, and covered by the current-DMG and regression checks.

PR #1282, opened 2026-08-10
Add opt-in ChatGPT conversation deletion for Linux

peter1599:

## Summary

Adds opt-in `conversation-delete` Linux feature.

- Adds confirmed **Delete chat** action beside Archive.
- Uses upstream authenticated delete API with no request body.
- Removes deleted conversations from sidebar caches.
- Suppresses stale refetched entries during backend propagation.
- Disabled by default.
- Resolves sidebar conversation deletion request.

Known risk: relies on private upstream endpoint and minified bundle markers.

## Validation

- `node --test linux-features/conversation-delete/test.js` — 6 passed.
- `node --test linux-features/*/test.js` — 887 passed, 1 skipped, 0 failed.
- `node --test scripts/patch-linux-window-ui.test.js` — 414 passed.
- Patched latest upstream bundle successfully.
- `/usr/bin/node --check` passed on patched bundle.
- `git diff --check` passed.
- Required CI checks not run locally.

## Checklist

- [ ] If this fixes upstream drift, it targets only the latest `CODEX.DMG` and removes obsolete fallback code and tests from the affected area.

### DIFF
--- a/linux-features/conversation-delete/README.md
+++ b/linux-features/conversation-delete/README.md
@@ -1,0 +1,40 @@
+# Conversation Delete
+
+Optional Linux feature that adds **Delete chat** to ChatGPT conversations in
+Codex Desktop's sidebar.
+
+Feature is disabled by default because deletion is permanent. Enable it in
+`linux-features/features.json`:
+
+```json
+{
+  "enabled": ["conversation-delete"]
+}
+```
+
+Action asks for confirmation, then uses upstream's authenticated ChatGPT
+request client. It sends:
+
+```text
+DELETE /backend-api/conversation/id/<conversation_id>
+```
+
+Request has no body. Existing upstream client sends session authentication,
+handles `204 No Content`, removes conversation from sidebar cache, suppresses
+stale refetched list entries during backend propagation, and returns to home
+when deleting active chat.
+
+## Scope
+
+Feature targets ChatGPT server conversations shown in ChatGPT mode. It does not
+add deletion for local Codex threads or archived-chat bulk actions.
+
+## Test
+
+```bash
+node --test linux-features/conversation-delete/test.js
+```
+
+Known risk: endpoint and minified sidebar markers are private upstream
+contracts. Current-DMG drift leaves asset unchanged and emits warning; rebuild
+with feature enabled after upstream refresh.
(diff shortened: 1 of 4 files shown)


ilysenko (changes requested):
Thanks for the contribution. I found two blockers:

- Deleting the active conversation currently calls the navigator with that conversation row’s own route, so it stays on or reopens the deleted chat instead of returning home. Please navigate to the actual home/new-chat route and update the regression test accordingly.
- Tombstone filtering omits `listProjectConversations()`. A project conversation can therefore reappear on refetch while deletion propagates. Please filter that response using the deleted-ID set and add an executable project-conversation refetch test.

peter1599:
> Thanks for the contribution. I found two blockers:
> 
>     * Deleting the active conversation currently calls the navigator with that conversation row’s own route, so it stays on or reopens the deleted chat instead of returning home. Please navigate to the actual home/new-chat route and update the regression test accordingly.
> 
>     * Tombstone filtering omits `listProjectConversations()`. A project conversation can therefore reappear on refetch while deletion propagates. Please filter that response using the deleted-ID set and add an executable project-conversation refetch test.

Don't merge yet I still see 1 bug after fix.

ilysenko (changes requested):
Thanks for the update. The previous route and project-refetch issues are fixed, but one blocker remains:

- The injected delete `onSelect` captures `isActive`, while the upstream React-compiled menu factory is memoized without `isActive` in its dependency guard. After a row changes between inactive and active, the cached callback can use the stale value, so active deletion may not navigate home or inactive deletion may navigate unexpectedly. Please patch the memo guard for the new dependency and add a two-render regression test that exercises the compiled cache path.

peter1599:
> Thanks for the update. The previous route and project-refetch issues are fixed, but one blocker remains:
> 
>     * The injected delete `onSelect` captures `isActive`, while the upstream React-compiled menu factory is memoized without `isActive` in its dependency guard. After a row changes between inactive and active, the cached callback can use the stale value, so active deletion may not navigate home or inactive deletion may navigate unexpectedly. Please patch the memo guard for the new dependency and add a two-render regression test that exercises the compiled cache path.

Pushed and tested. Deleting currently active chat now actually navigates to new chat but i still see some strange things. Sometimes deleted chat re apperas for like 10 seconds then disappears? Any idea how can I improve more on that?

ilysenko (approved):
Reviewed the current head. The previously requested route, project refetch, and compiled memo-cache fixes are implemented and covered by tests; all required checks pass.

PR #1285, opened 2026-08-10
perf(linux): reduce chunked message decoder overhead

### LINKED ISSUE #1284: perf: reduce chunked message decoder object-write overhead
## Problem

Switching into content-heavy conversations can keep the renderer main thread busy reconstructing chunked host messages. The cost becomes visible alongside retained Browser tabs/webviews, where preserving the user's working set is important and unloading tabs is not an acceptable shortcut.

Environment profiled:

- ChatGPT Desktop 26.803.41515
- Electron 42.3.0 / Chrome 148
- Linux generated from current `main` (`afc4b9a`)
- 201-session sidebar with up to seven retained Browser webviews

## Evidence

A 34-session CPU profile repeatedly identified the host chunk receive path as a renderer hotspot. One baseline sweep recorded:

- `j6e` (message receive/assembly): 2.36 s self time
- `saveValue` (decoded object writes): 367 ms self time

The current assembler calls `Object.defineProperty` for every decoded object key. That preserves safe own-property behavior for the special `__proto__` key, but pays the same cost for ordinary keys.

A controlled benchmark through the real `window.postMessage` chunk listener decoded five 50,000-property transfers per sample. Three sequential A/B samples (concurrent samples excluded to avoid Electron CPU contention) showed:

| Metric | Baseline median | Candidate median | Change |
| --- | ---: | ---: | ---: |
| End-to-end decoder wall time | 777.9 ms | 728.5 ms | -6.35% |
| `saveValue` self time | 163.2 ms | 108.6 ms | -33.45% |

The candidate retains `Object.defineProperty` for `__proto__` and uses ordinary assignment only for other keys. Property descriptors for ordinary keys remain enumerable, configurable, and writable, and the decoded object's prototype remains unchanged for `__proto__`.

## Proposed scope

- Add an optional, semantic, fail-soft patch for the single current chunk assembler contract in `app-initial-*.js`.
- Preserve root and array writes.
- Preserve exact `__proto__` own-property/prototype behavior.
- Reject ambiguous, drifted, or incomplete bundle shapes without modifying assets.
- Add idempotency, descriptor, prototype-safety, ambiguity, and drift coverage.

## Acceptance criteria

- The current generated asset matches exactly once and patches idempotently.
- Normal object writes avoid unconditional `Object.defineProperty`.
- `__proto__` retains the existing safe data-property behavior.
- Browser webviews remain retained; no performance claim depends on unloading them.
- Full relevant regression and script-smoke suites pass.

nxank4:
## Summary

Fixes #1284.

Large host messages are reconstructed in the renderer by the chunk assembler. Its object path currently calls `Object.defineProperty` for every decoded key, so content-heavy session hydration pays the safe `__proto__` write cost for ordinary properties too.

This PR adds one optional semantic webview patch that:

- uses ordinary assignment for normal decoded object keys;
- retains the existing `Object.defineProperty` path for `__proto__`;
- preserves root writes, array writes, property descriptors, and decoded-object prototypes;
- selects exactly one complete current assembler contract and leaves drifted, duplicated, or partially patched assets byte-identical;
- remains idempotent and fail-soft as upstream minified symbols change.

The patch does not unload retained Browser / Computer Use webviews and adds no dependency or generated app artifact.

## Performance evidence

Environment: ChatGPT 26.803.41515, Electron 42.3.0 / Chrome 148, Linux, 201-session sidebar, up to seven retained Browser webviews.

A 34-session CPU profile identified the chunk receive/assembly path as a recurring renderer hotspot. One baseline sweep recorded 2.36s of `j6e` self time and 367ms of `saveValue` self time.

A controlled benchmark through the real in-app `window.postMessage` chunk listener decoded five 50,000-property transfers per sample. Three sequential A/B samples produced:

| Metric | Before median | After median | Change |
| --- | ---: | ---: | ---: |
| Decoder wall time | 777.9 ms | 728.5 ms | -6.35% |
| `saveValue` self time | 163.2 ms | 108.6 ms | -33.45% |

Concurrent samples were excluded because two Electron instances contended for CPU. A separate text-measurement candidate was also excluded after an exact-input 34-session sweep regressed end-to-end long-task time, so it is not part of this PR.

## Validation

- TDD red/green cycle: the focused tests first failed because the transform and descriptor did not exist, then passed after the implementation
- `node scripts/patch-linux-window-ui.test.js` outside the restricted sandbox: 430 passed, 0 failed
- `bash tests/scripts_smoke.sh`: passed, including Debian/RPM/pacman/AppImage smoke matrices
- `node --check` on the implementation, descriptor, and test file
- `git diff --cached --check`
- Current `app-initial-Biw83Aiz.js` semantic acceptance:
  - exactly one match
  - one byte change
  - second application byte-identical
  - source SHA-256 remained `426cdc7d62c25d470ef82ff184fba0826026edab2653d75256d7f4458455eb61`
  - patched SHA-256 `a4d603fd40fda8472f596d934e2275b2f71841f9f69021c2927ec1a5cdba1796`, matching the profiled candidate
- Browser retention during session sweeps: up to 7/7 guest views remained connected

`make build-dev-app` was attempted in a side-by-side target. The execution environment hit its filesystem quota while unpacking the managed Node runtime, before the patcher ran (`patchReport: missing`), so that build is reported as inconclusive rather than passing. The generated-asset transform and the profiled candidate were verified independently as listed above.

## Compatibility and risk

- Targets only the latest primary `app-initial-*.js` semantic contract.
- `__proto__` remains an enumerable, configurable, writable own data property and cannot mutate the decoded object's prototype.
- Ordinary `constructor` and `toString` keys retain the same own data-property descriptors.
- Unknown, ambiguous, incomplete, or future upstream shapes are skipped with an optional drift warning.
- Remaining risk is upstream bundle drift, which does not block installation and does not modify an unknown asset.

## Checklist

- [x] This is not an upstream-drift compatibility fix; the semantic patch intentionally targets only the latest `CODEX.DMG`.
- [ ] I added or updated relevant tests, ran the validation listed above, and confirmed that required CI checks pass. (Local validation is complete; GitHub CI is pending.)

### DIFF
--- a/.agents/plans/2026-08-10-linux-chunked-message-decoder-performance.md
+++ b/.agents/plans/2026-08-10-linux-chunked-message-decoder-performance.md
@@ -1,0 +1,63 @@
+# Linux Chunked Message Decoder Performance Implementation Plan
+
+> **For Codex:** Execute this plan sequentially with test-driven development. Keep the generated webview patch optional, semantic, fail-soft, and idempotent.
+
+**Goal:** Reduce renderer main-thread work while reconstructing large chunked host messages without changing decoded object semantics or weakening protection for the special `__proto__` key.
+
+**Architecture:** Add one all-Linux webview asset descriptor for the primary `app-initial-*.js` bundle. Its semantic matcher will locate exactly one current chunk assembler contract using the chunk protocol marker, assembler diagnostics, and object-write shape. The transform will use ordinary property assignment for normal keys and retain `Object.defineProperty` for `__proto__`, which preserves the existing own-property descriptor and prototype behavior.
+
+**Evidence baseline (2026-08-10, ChatGPT 26.803.41515 / Electron 42.3.0):** A 34-session CPU profile identified chunk decoding as a recurring renderer hotspot (`j6e` 2.36s, `saveValue` 367ms in one baseline sweep). A controlled in-app benchmark decoded five 50,000-property transfers through the real `window.postMessage` listener. Across three sequential A/B samples, median wall time fell from 777.9ms to 728.5ms (-6.35%) and median `saveValue` self time fell from 163.2ms to 108.6ms (-33.45%). Concurrent samples were excluded because the two Electron instances contended for CPU. A previous synchronous text-measurement candidate was rejected after an exact-input sweep regressed end-to-end long-task time by about 9%.
+
+---
+
+## Task 1: Lock the semantic and safety contract with failing tests
+
+**Files:**
+- Modify: `scripts/patch-linux-window-ui.test.js`
+
+1. Import the new matcher and transform from `scripts/patches/impl/webview/index.js`.
+2. Add a compact current-bundle fixture containing the chunk marker, validation/assembler diagnostics, object/array/root behavior, and the original unconditional `Object.defineProperty` write.
+3. Add tests that require:
+   - exactly one semantic contract match;
+   - idempotent transformation;
+   - ordinary keys use direct assignment;
+   - `__proto__` still becomes an enumerable, configurable, writable own data property while the decoded object's prototype remains unchanged;
+   - ordinary keys such as `constructor` and `toString` preserve own-property descriptors;
+   - array and root writes are unchanged;
+   - generic, drifted, duplicated, and incomplete candidates remain byte-identical and warn precisely.
+4. Add the descriptor id to the canonical core descriptor list and assert its filename pattern/optional policy.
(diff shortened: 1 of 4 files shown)


nxank4:
Closing this draft because the isolated decoder optimization improves the full decoder median by only 6.35% (about 49 ms in the controlled benchmark) and does not yet demonstrate a material end-to-end session-switch improvement. Issue #1284, the branch, tests, and measurements remain available while we profile the larger root cause. A replacement PR should be one coherent change with repeatable user-visible A/B evidence and regression coverage.

PR #1288, opened 2026-08-10
perf(linux): avoid replaying Markdown fade animations

### LINKED ISSUE #1287: perf(linux): avoid replaying Markdown fades when restoring active sessions
## Problem

Switching back to a long, still-running session can freeze the Linux renderer for hundreds of milliseconds or several seconds. The impact is most visible in Browser Use / Computer Use workflows with large timelines and retained browser tabs.

## Profiling evidence

A representative active turn mounted 2,484 descendants across 24 animated Markdown roots and 1,181 fade spans, including content outside the viewport. Chromium traces attributed the switch stall to renderer work including `UpdateLayoutTree`, `Layout`, `Layerize`, and `Commit`.

Disabling only the streaming Markdown text/list-marker fades produced the following same-renderer A/B result on the heavy switch:

| Metric | Baseline | Animation disabled |
| --- | ---: | ---: |
| Total long-task time | 2,608-2,881 ms | 884-930 ms |
| Longest task | 636-660 ms | 389-392 ms |

A fresh-app sweep over the exact same 25-session fingerprint sequence, with both variants retaining at most eight Browser Use webviews, measured:

| Metric | Baseline | Candidate | Change |
| --- | ---: | ---: | ---: |
| Total long-task time | 16,119 ms | 12,866 ms | -20.2% |
| Maximum task | 760 ms | 518 ms | -31.8% |
| Median longest task | 217 ms | 189 ms | -12.9% |

Reducing the chunked-message decode batch size did not materially improve these costs, so the bottleneck is the replayed per-item animations rather than decoder batching.

## Proposed behavior

On Linux, render streaming Markdown text and list markers immediately when a session is restored, while preserving content updates, scrolling, Browser Use webview retention, and the separate bounded image-enter animation.

## Acceptance criteria

- Avoid replaying text/list-marker fades across historical incomplete Markdown blocks.
- Preserve image-enter motion and all message content.
- Apply semantically, idempotently, and fail softly if upstream CSS drifts or becomes ambiguous.
- Include focused regression coverage and measured before/after evidence.

nxank4:
## Summary

- render streaming Markdown text and list markers immediately on Linux instead of replaying per-item fades when a running thread is restored
- preserve message content, streaming behavior, image-enter motion, and Browser Use webview retention
- add a semantic, idempotent, ambiguity-rejecting, fail-soft CSS patch and focused regression coverage

## Root cause

A profiled active turn mounted 24 animated Markdown roots and 1,181 fade spans, including offscreen content. Restoring it replayed those animations and caused renderer spikes in layout, layerization, and commit work. Reducing the chunk decoder batch size did not materially improve the stall.

## Results

Same-renderer A/B on the heavy session switch:

| Metric | Baseline | Patched | Result |
| --- | ---: | ---: | ---: |
| Total long-task time | 2,608-2,881 ms | 884-930 ms | about 66% lower |
| Longest task | 636-660 ms | 389-392 ms | about 39% lower |

Fresh 25-session sweep using the exact same fingerprint sequence; both variants retained at most eight Browser Use webviews:

| Metric | Baseline | Patched | Change |
| --- | ---: | ---: | ---: |
| Total long-task time | 16,119 ms | 12,866 ms | -20.2% |
| Maximum task | 760 ms | 518 ms | -31.8% |
| Median longest task | 217 ms | 189 ms | -12.9% |
| Median total long-task time | 417 ms | 417 ms | unchanged |

## Validation

- focused patch and descriptor tests: 5/5 passed
- exact current DMG stylesheet: matched, changed, idempotent, and image-enter animation preserved
- Node syntax checks and `git diff --check` passed
- full Node suite: 423/428; the remaining five failures were reproduced before this change and depend on the local desktop/plugin environment
- the full smoke test reached its warm-start safety gate, which correctly refused to launch a duplicate app while the installed `/opt/codex-desktop` instance was running

Closes #1287

### DIFF
--- a/.agents/plans/2026-08-10-linux-streaming-markdown-animation-performance.md
+++ b/.agents/plans/2026-08-10-linux-streaming-markdown-animation-performance.md
@@ -1,0 +1,54 @@
+# Linux streaming Markdown animation performance
+
+## Goal
+
+Remove the session-switch freeze caused by replaying word-level Markdown fade
+animations across a large in-progress turn, without changing message content,
+Browser Use webview retention, scrolling, or image enter motion.
+
+## Evidence and root cause
+
+- A representative running session mounted one 4,920 px turn containing 2,484
+  descendants, 24 animated Markdown roots, and 1,181 fade spans. Most animated
+  roots were outside the viewport.
+- Chromium traces attributed the switch stall to renderer program work plus
+  `UpdateLayoutTree`, `Layout`, `Layerize`, and `Commit`; reducing the chunked
+  message batch size did not improve these costs.
+- Same-renderer A/B on the heavy `21 -> 7` switch reduced total long-task time
+  from 2,881/2,608 ms to 930/884 ms and the longest task from 660/636 ms to
+  392/389 ms when only Markdown text/list-marker animation was disabled.
+- Fresh-app 25-session A/B used the same fingerprint sequence and reached the
+  same maximum of eight retained Browser Use webviews. Total long-task time
+  fell from 16,119 ms to 12,866 ms, maximum task duration from 760 ms to 518 ms,
+  and median longest task from 217 ms to 189 ms.
+
+## Implementation
+
+1. Add an optional, Linux-only webview CSS descriptor targeting the unique
+   streaming Markdown animation contract in the current `app-initial` asset.
+2. Render streaming text, block elements, and list markers immediately instead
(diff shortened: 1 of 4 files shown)


ilysenko:

ilysenko (approved):
Reviewed the current head. The default-on Linux Markdown performance patch is focused, fail-soft on upstream drift, covered by regression tests, and verified against the current upstream DMG.

PR #1292, opened 2026-08-10
fix(launcher): keep GPU compositing enabled on Wayland sessions

### LINKED ISSUE #1291: perf(launcher): stop disabling GPU compositing by default on Wayland sessions
## Summary

On GNOME Wayland multi-monitor systems, the launcher can select the stable X11/XWayland profile and still append `--disable-gpu-compositing` solely because the raw desktop session is Wayland. On a long-running, content-heavy workspace this makes Electron's GPU/Viz process consume a large amount of CPU and creates visible session/tab switching lag.

This is distinct from the native-Wayland/KWin freeze in #1008 and the selection-specific freeze in #1122, although the latter reports the same `--disable-gpu-compositing` launch flag as a possible amplifier.

## Environment

- Ubuntu, GNOME Wayland, multiple monitors
- Effective launcher platform: `--ozone-platform=x11` (XWayland)
- NVIDIA GPU
- Upstream app build: `26.803.41515`
- Representative workload: several active sessions, long transcripts, Browser Use tabs, and ongoing app-server work

## Root cause

The launcher determines the effective rendering profile separately from the compositing fallback. The automatic fallback checks raw `XDG_SESSION_TYPE=wayland`, not the effective Ozone platform, so this combination is produced:

```text
--ozone-platform=x11 --disable-gpu-compositing
```

Commit `5887e92fe65e992df9fbca0a40560132cd20a640` previously changed the default to keep GPU compositing enabled and retained an explicit opt-in fallback. Commit `d17d5d436eaf71380e046d36d71205e3031aaf91` later restored the raw Wayland-session branch, regressing that behavior.

## Same-host A/B evidence

The same installed app, profile, conversations, Browser Use state, and host were sampled before and after removing only the automatic `--disable-gpu-compositing` flag.

| Metric | Automatic fallback | GPU compositing enabled |
| --- | ---: | ---: |
| Electron GPU process CPU | 48-100% | 14-19% |
| `VizCompositorThread` CPU | 98.8% under load | 5.75-5.99% |
| GPU-process minor faults | ~154,420/s | ~351-1,105/s |

The host retained 60-69% aggregate CPU idle time and about 15 GiB available RAM during the baseline, so the stall was not caused by whole-system CPU, memory, or storage saturation. With compositing enabled, NVIDIA VRAM usage increased from about 28 MiB to 195 MiB, consistent with Electron using hardware compositing instead of continuously exercising the software/Viz path.

No renderer crash, `render-process-gone`, GPU-process failure, or unresponsive-window event appeared during the composited test run. Renderer and app-server CPU still varied with active work, as expected.

## Proposed behavior

- Keep GPU compositing enabled by default for both X11/XWayland and native Wayland profiles.
- Preserve `CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=1` as an explicit compatibility fallback.
- Preserve a caller-supplied `--disable-gpu-compositing` argument.
- Do not change the existing GNOME Wayland multi-monitor decision to use X11/XWayland in this fix; native Wayland remains a separate compatibility experiment.

## Acceptance criteria

- Default Wayland-session launches do not automatically add `--disable-gpu-compositing`.
- Explicit environment and command-line fallbacks continue to work.
- Launcher smoke tests cover the default and explicit fallback paths.
- Launcher performance documentation records the rationale and measured A/B result.

nxank4:
## Summary

- stop automatically adding `--disable-gpu-compositing` merely because the desktop session is Wayland
- preserve the explicit environment and command-line compatibility fallbacks
- add regression coverage for default Wayland and GNOME Wayland multi-monitor/XWayland launches
- document the same-host A/B result and the behavior intentionally left unchanged

## Root cause

The launcher chooses the effective rendering profile separately from its compositor fallback. On GNOME Wayland with multiple monitors, the auto profile intentionally uses `--ozone-platform=x11`, but the fallback then checks raw `XDG_SESSION_TYPE=wayland` and also appends `--disable-gpu-compositing`.

That raw-session check regressed the behavior introduced by `5887e92fe65e992df9fbca0a40560132cd20a640` when it was restored in `d17d5d436eaf71380e046d36d71205e3031aaf91`.

## Measured impact

Same host, installed app, profile, active sessions, and Browser Use state; only the automatic compositor-disable flag changed:

| Metric | Before | After |
| --- | ---: | ---: |
| Electron GPU process CPU | 48-100% | 14-19% |
| `VizCompositorThread` CPU | 98.8% under load | 5.75-5.99% |
| GPU-process minor faults | ~154,420/s | ~351-1,105/s |

The host still had 60-69% aggregate CPU idle time and about 15 GiB available RAM during the baseline, ruling out general CPU, memory, or storage saturation. No renderer crash, GPU-process failure, or unresponsive-window event appeared in the composited test run.

## Compatibility

`CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=1` and an explicit `--disable-gpu-compositing` argument still enable the side-panel flicker fallback. This PR does not change the GNOME Wayland multi-monitor X11/XWayland selection and does not opt users into native Wayland.

## Validation

- red/green launcher regression test: the updated expectation fails with the prior production branch and passes with this change
- `git diff --check`
- full `bash tests/scripts_smoke.sh` in an isolated PID namespace: all script smoke tests passed
- live same-app A/B sampling with `pidstat`; no crash/unresponsive/GPU failure observed

Closes #1291

### DIFF
--- a/docs/launcher-performance.md
+++ b/docs/launcher-performance.md
@@ -3,13 +3,13 @@
 ## Context
 
 This decision record captures a performance comparison between ChatGPT Desktop
-and another Electron 42 app (Claude Desktop) running side by side on the same
-X11 GNOME 4K host, what the launcher changed as a result, and — just as
-importantly — what was reviewed and deliberately left alone so future work
-does not re-litigate it without new evidence.
+and another Electron app (Claude Desktop) running side by side on the same
+GNOME Wayland multi-monitor host, what the launcher changed as a result, and —
+just as importantly — what was reviewed and deliberately left alone so future
+work does not re-litigate it without new evidence.
 
-Evidence came from live process command lines, `/proc/<pid>/maps`, the
-launcher log, and repository history rather than synthetic benchmarks.
+Evidence came from live process command lines, `pidstat`, `/proc/<pid>/maps`,
+the launcher log, and repository history rather than synthetic benchmarks.
 
 ## What Changed
 
@@ -31,10 +31,18 @@ launcher log, and repository history rather than synthetic benchmarks.
   ppid-guarded watchdog pattern capped at 0.5 s, so a broken session bus
   counts as "not detected" instead of delaying launch.
   Override: `CODEX_FORCE_RENDERER_ACCESSIBILITY=1|0`.
+- GPU compositing now stays enabled by default, including when Electron runs
+  through XWayland. In an eight-second same-app A/B sample with active sessions
+  and browser tabs, removing the automatic `--disable-gpu-compositing` flag
+  reduced GPU-process CPU from 48–100% to 14–19%, Viz compositor CPU from
+  about 99% to 6%, and GPU-process minor faults from roughly 154,000/s to
+  350–1,100/s. Claude Desktop also keeps GPU compositing enabled on this host.
+  The side-panel flicker workaround remains available with
+  `CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=1` or the matching command-line flag.
 
-Both decisions are visible at runtime in the `Electron launch mode:` line of
+These decisions are visible at runtime in the `Electron launch mode:` line of
 `~/.cache/codex-desktop/launcher.log` (`dev_shm_usage_disabled=`,
-`renderer_accessibility_forced=`).
+`renderer_accessibility_forced=`, `gpu_compositing_disabled=`).
 
 ## Reviewed And Deliberately Not Changed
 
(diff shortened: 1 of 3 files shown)


ilysenko:
Thanks for the detailed investigation, measurements, and implementation. We intentionally keep GPU compositing disabled by default on Wayland because this compatibility default was introduced after real rendering and stability problems. Users whose systems work better with GPU compositing can already opt in by setting `CODEX_ELECTRON_DISABLE_GPU_COMPOSITING=0`. We prefer to preserve the safer default and let users enable compositing explicitly for their environment, so we are closing this pull request.

PR #1295, opened 2026-08-10
fix launcher planned termination status

anervalens-netizen:
## Summary
- forward launcher signals to Electron
- treat planned SIGTERM as a successful service stop
- isolate launcher recovery tests from an installed Codex instance

## Validation
- bash syntax checks pass
- all launcher warm-start/recovery variants pass, including the planned TERM status regression
- packaging/script smoke coverage before the launcher recovery group passes


ilysenko (approved):
Reviewed the current head: the launcher lifecycle fix is focused, covered by the TERM regression test, and all required checks pass.

PR #1304, opened 2026-08-11
Keep GUI integrations off mise multicall shims

JPM1118:
## What changed

- Detect Codex CLI paths that are mise multicall shims.
- Resolve those shims to the concrete CLI installed under mise's active Node installation.
- Add a launcher smoke-test fixture covering the mise layout.

## Why

Chrome native hosts and other GUI children do not reliably have mise's shell dispatch context. Registering the shim caused `codex app-server` to exit before becoming ready, breaking the ChatGPT Chrome integration after desktop or session restarts.

## Impact

GUI integrations now receive a directly executable Codex CLI path while non-mise CLI resolution behavior remains unchanged.

## Validation

- `python3 -m py_compile launcher/cli-launch-path.py`
- Isolated mise-shim resolution fixture
- `git diff --check`
- Rebuilt and installed the pacman package locally
- Confirmed the Chrome runtime manifest registered the concrete CLI path
- Confirmed the exact Chrome app-server launch stayed running until the verification timeout

### DIFF
--- a/launcher/cli-launch-path.py
+++ b/launcher/cli-launch-path.py
@@ -31,6 +31,33 @@ def resolve_cli_launch_path(raw_path: str) -> Path:
 
     if not stat.S_ISREG(metadata.st_mode) or not os.access(canonical_cli, os.X_OK):
         raise LaunchPathError(f"Selected Codex CLI target {canonical_cli} is not an executable file")
+
+    # mise uses a multicall binary: its shims all point at the mise executable,
+    # which selects the requested tool from the shim's argv[0]. GUI children
+    # such as Chrome native hosts do not provide mise's normal shell context,
+    # so register the concrete npm-installed CLI instead of either the shim or
+    # the canonical mise multicall binary.
+    if selected_path.is_symlink() and canonical_cli.name == "mise":
+        mise_install = (
+            selected_path.parent.parent
+            / "installs"
+            / "node"
+            / "latest"
+            / "bin"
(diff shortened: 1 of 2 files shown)


ilysenko (changes requested):
Thanks for the focused fix. I found a blocker:

- `resolve_cli_launch_path()` hardcodes `installs/node/latest/bin/codex`. mise installs are versioned, the active toolset can differ from `latest`, `latest` can be stale or absent, and `MISE_INSTALLS_DIR` can relocate installs. This can register the wrong CLI or turn launcher preflight into a hard failure. Please resolve the active shim through a mise-supported contract and add regressions for a pinned active version, stale/missing `latest`, and a custom installs directory.

ilysenko:

ilysenko (changes requested):
Thanks for the focused fix. One additional end-to-end blocker remains:

- `verify_cli_launch_path()` keeps the mise shim in `CODEX_CLI_SOURCE_PATH`, and `run_cli_preflight()` passes that shim back to the native update manager. `stable_cli_launch_path()` then canonicalizes it to the `mise` binary; preflight can return that path and overwrite the concrete `CODEX_CLI_PATH`, or a required/synchronous preflight can fail startup. Please make the launcher/updater round trip preserve the resolved concrete Codex path and add an integration test that exercises the real updater preflight with a mise shim.

ilysenko:

Issue #963, opened 2026-07-14
Warm start keeps a live Electron process whose packaged webview server died with the launcher


M1racleShih:
## Summary

If the long-lived launcher is killed with `SIGKILL`, the packaged webview server exits through its Linux `PR_SET_PDEATHSIG` guard, but Electron stays alive. The next launch sees the live Electron PID, sends the normal warm-start IPC action, and exits without restoring the webview origin. The existing app remains unusable because its renderer URL still points at a closed localhost port.

This reproduced 3/3 times on an unmodified build from `0883c4e`.

## Environment

- Ubuntu 24.04.3 LTS
- KDE, X11
- side-by-side dev build (`codex-cua-lab`), no optional Linux features
- source commit: `0883c4ebc6028b3f2492f850ab16158eac7d850a`
- upstream app: `26.707.31428`
- Electron: `42.1.0`

## Minimal reproduction

Use a normal generated app and adjust `APP`, `APP_ID`, and `PORT` if needed:

```bash
APP=./codex-app
APP_ID=codex-desktop
PORT=5175
STATE_DIR="${XDG_STATE_HOME:-$HOME/.local/state}/$APP_ID"

"$APP/start.sh" &
launcher_pid=$!

# Wait for a healthy cold start before continuing.
until [ -s "$STATE_DIR/app.pid" ] && [ -s "$STATE_DIR/webview.pid" ] && \
      curl -fsS "http://127.0.0.1:$PORT/index.html" >/dev/null; do
    sleep 0.1
done

electron_pid="$(cat "$STATE_DIR/app.pid")"
webview_pid="$(cat "$STATE_DIR/webview.pid")"

kill -KILL "$launcher_pid"
wait "$launcher_pid" 2>/dev/null || true
sleep 1

kill -0 "$electron_pid"                         # succeeds
kill -0 "$webview_pid"                          # fails
curl -fsS "http://127.0.0.1:$PORT/index.html"   # connection refused

timeout 10 "$APP/start.sh"                      # exits after warm-start IPC
kill -0 "$electron_pid"                         # same Electron is still alive
curl -fsS "http://127.0.0.1:$PORT/index.html"   # still connection refused

kill "$electron_pid"                            # cleanup
```

## Actual behavior

- Electron keeps the same PID after the launcher is killed and after relaunch.
- The webview server exits and the configured port closes.
- The launch-action socket and `app.pid` still make the second launcher choose warm start.
[...]

PR #1307, opened 2026-08-11
chore(computer-use): satisfy stable Clippy

M1racleShih:
## Summary

- replace two negated `Option::is_some_and` checks with the equivalent `Option::is_none_or` form
- preserve the missing-or-empty `WAYLAND_DISPLAY` semantics used by native X11 session detection and xdotool pointer selection
- restore the workspace Clippy gate on Rust 1.97.1 stable

## Root cause

Rust 1.97.1's Clippy reports these expressions as `nonminimal_bool`. Because CI promotes Clippy warnings to errors with `-D warnings`, the latest-stable toolchain failed validation even though the runtime behavior was correct.

This is a maintenance follow-up in the native X11 logic exercised by #963; it does not reopen or close that issue.

## Validation

- `cargo fmt --check`
- `TMPDIR=/tmp cargo test --workspace --all-targets`
  - warm-start recovery and window-reopen scenarios passed
  - webview probe equivalence passed 21/21
- latest upstream DMG `26.803.81509` rebuilt with verdict `accepted`
- clean provenance verified at commit `30cdedfc` with `dirty: false`
- isolated X11 runtime acceptance reached Electron spawn, Codex CLI initialization, renderer route mount, and ready state

### DIFF
--- a/computer-use-linux/src/diagnostics.rs
+++ b/computer-use-linux/src/diagnostics.rs
@@ -490,9 +490,9 @@ fn desktop_env_hydration_updates(
         && current_env
             .get("DISPLAY")
             .is_some_and(|value| !value.trim().is_empty())
-        && !current_env
+        && current_env
             .get("WAYLAND_DISPLAY")
-            .is_some_and(|value| !value.trim().is_empty());
+            .is_none_or(|value| value.trim().is_empty());
 
     DESKTOP_ENV_KEYS
         .iter()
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the current head: the Clippy-only rewrite is semantically equivalent, existing X11 behavior is preserved, and all required checks pass.

PR #1003, opened 2026-07-15
Fix Nix store CLI trust regression

colonelpanic8:
## Summary

- allow root-owned sticky ancestors with standard Nix store 1775 permissions in the Computer Use runtime path validator
- preserve rejection of writable ancestors that are either non-sticky or not root-owned
- add focused coverage for 1775 and 1777 acceptance plus the unsafe cases

## Context

The original PR also adjusted launcher and updater trust validation. PR #1002 replaced that subsystem while this PR was open, so the branch is now rebased onto current main and intentionally retains only the still-needed Computer Use fix.

## Testing

- cargo fmt --all -- --check
- cargo test -p codex-computer-use-linux root_owned_sticky_directory_accepts_nix_store_permissions -- --nocapture
- cargo test -p codex-computer-use-linux runtime_paths_reject_an_unsafe_grandparent_directory -- --nocapture

PR #1308, opened 2026-08-11
Fix Chrome side-panel CLI trust selection

DmytroMitin:
## Summary

Plain Desktop launches can select an NVM/npm Codex CLI that is valid under the current Desktop launcher policy but fails the stricter Chrome native-host path policy. The launcher then exports that path into both Chrome v2 runtime manifests, so the side panel reports `Unable to start ChatGPT` even when a trusted same-version Codex CLI is already installed under the dedicated npm prefix. Users currently have to pin that second path with `CODEX_CLI_PATH` before launching Desktop.

This change keeps normal Desktop and updater selection intact and adds a separate Chrome-only reconciliation step:

- the shipped Rust Chrome native host validates the normal selected path with the existing ownership, executable, permission, canonicalization, and parent-chain checks;
- a safe normal selection remains selected;
- for an unsafe implicit npm selection, only the dedicated `~/.codex-cli-npm` candidate is eligible, and only when both package metadata files identify the exact same `@openai/codex` version;
- the candidates are not executed to determine trust or compatibility;
- Chrome registration receives the result through an internal `CODEX_CHROME_CLI_PATH`, while Desktop continues using `CODEX_CLI_PATH`;
- explicit `CODEX_CLI_PATH` remains authoritative and never silently falls through;
- without a safe compatible fallback, Desktop remains usable and Chrome still fails closed under its existing validator.

The source-of-truth changes are in:

- `computer-use-linux/src/bin/codex-chrome-extension-host.rs`
- `computer-use-linux/src/chrome_runtime.rs`
- `scripts/patches/impl/chrome-plugin.js`
- adjacent launcher, Chrome patcher, Electron 42, and Rust tests
- `README.md`, `docs/architecture.md`, and `docs/updater.md`

### Policy and security context

PR #962 originally made strict path trust a broad launcher/updater execution gate and recommended a fresh dedicated npm prefix instead of changing modes in an already-untrusted tree. PR #1002 later deliberately relaxed normal launch validation to a canonical regular-executable check so ordinary user installs created under `umask 0002` remain usable. The Chrome runtime retained its stricter trust policy, including through the Nix correction in #1003.

This PR does not blindly restore #962's obsolete global launch gate. It reconciles the policies only for Chrome registration and directly reuses the existing Rust Chrome validator. It does not chmod an NVM tree, set a global umask, reorder all CLI sources, change the user's terminal `codex`, or weaken Chrome trust.

PR #1286 fixed a different layer: once a trusted npm `codex.js` is accepted, its `#!/usr/bin/env node` child must receive a trusted managed Node path even when GUI `PATH` lacks Node. This PR does not alter that child-`PATH` behavior, and its regression test remains green.

The fallback cannot change install channels or versions: it is npm-to-npm only, the package name must be exactly `@openai/codex`, and the version strings must match exactly. Direct, standalone, system-package, AppImage, Nix, updater, and no-updater normal selections are otherwise unchanged.

Open draft PR #1304 addresses a separate mise multicall-shim dispatch problem. It does not apply Chrome path trust or provide a same-version trusted npm fallback, so this change does not duplicate it.

## Validation

TDD red before implementation:

- `node scripts/patches/impl/chrome-plugin.test.js` — 15 passed, 1 failed; the new registration regression received the NVM-style Desktop path instead of the trusted dedicated-prefix path.

Green after implementation:

- `node scripts/patches/impl/chrome-plugin.test.js` — 16 passed.
- `env TMPDIR=/tmp node --test scripts/patch-linux-window-ui.test.js` — 428 passed.
- `bash tests/scripts_smoke.sh` — passed, including launcher routing and cross-format packaging smoke checks.
- `env TMPDIR=/tmp cargo test -p codex-computer-use-linux --bin codex-chrome-extension-host` — 85 passed, including same-version fallback, explicit/safe-first/fail-closed semantics, no version/channel switching, unchanged trust checks, and the #1286 npm-shebang test.
- `env TMPDIR=/tmp cargo test --workspace --all-targets` — passed.
- `env TMPDIR=/tmp cargo check --workspace --all-targets` — passed.
- `env TMPDIR=/tmp cargo clippy --workspace --all-targets -- -D warnings` — passed.
- `cargo fmt --all -- --check` — passed.
- `bash -n launcher/start.sh.template tests/scripts_smoke.sh` — passed.
- `git diff --check upstream/main...HEAD` — passed.

Live installed-app validation of the automatic NVM-to-dedicated-prefix Chrome fallback has not been performed; the exact installed environment currently works with the equivalent explicit CODEX_CLI_PATH workaround.

### Non-goals

- Changing global Desktop/updater CLI discovery order or update policy.
- Repairing or changing permissions on existing npm/NVM installations.
- Weakening Chrome native-host trust validation.
- Switching versions or install channels to find any merely available CLI.
- Changing #1286's managed-Node child environment.
- Modifying Chrome profiles, native-host registrations, installed packages, or updater state as part of this source change.

## Checklist

- [x] If this fixes upstream drift, it targets only the latest `CODEX.DMG` and removes obsolete fallback code and tests from the affected area. (Not an upstream-drift fix; current-DMG fixtures pass.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,11 @@ platform binary is present. The launcher does not rank installed CLIs by
 version; it uses an explicit `CODEX_CLI_PATH` first, then the normal lookup
 order, and logs the resolved CLI path plus best-effort version so GUI PATH
 issues are visible. Set `CODEX_CLI_PATH=/path/to/codex` when you want to pin a
-specific binary.
+specific binary. Chrome's side-panel runtime keeps its stricter path trust
+checks. For an implicit unsafe npm selection, the launcher may register the
+same-version CLI from the trusted dedicated `~/.codex-cli-npm` prefix for
+Chrome only; Desktop continues using its normal selection. An explicit
+`CODEX_CLI_PATH` is never replaced.
 
 Local AppImage builds can optionally embed that CLI and its matching Linux
 platform package. Set `CODEX_CLI_BUNDLE_SOURCE` to an installed
--- a/computer-use-linux/src/bin/codex-chrome-extension-host.rs
+++ b/computer-use-linux/src/bin/codex-chrome-extension-host.rs
@@ -1,4 +1,4 @@
-use anyhow::{bail, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use serde_json::{json, Value};
 use std::{
     collections::HashMap,
@@ -510,6 +510,10 @@ impl RolloutTracker {
 }
 
 fn main() -> Result<()> {
+    if let Some(result) = run_cli_selection_mode() {
+        return result;
+    }
+
     let effective_uid = unsafe { libc::geteuid() };
     let socket_dir = socket_dir(effective_uid);
     let socket_path = socket_path(&socket_dir)?;
@@ -543,6 +547,24 @@ fn main() -> Result<()> {
     runtime_manager.shutdown();
     remove_socket_if_present(&socket_path)?;
     result
+}
+
+fn run_cli_selection_mode() -> Option<Result<()>> {
+    let args = env::args_os().skip(1).collect::<Vec<_>>();
+    if args.first().is_none_or(|arg| arg != "--select-chrome-cli") {
+        return None;
+    }
+    Some((|| {
+        if args.len() != 3 {
+            bail!("usage: codex-chrome-extension-host --select-chrome-cli SELECTED FALLBACK");
+        }
+        let selected = Path::new(&args[1]);
+        let fallback = Path::new(&args[2]);
+        let trusted = chrome_runtime::select_trusted_chrome_cli_path(selected, Some(fallback))
+            .map_err(|error| anyhow!(error.message().to_string()))?;
+        println!("{}", trusted.display());
+        Ok(())
+    })())
 }
 
 fn socket_dir(effective_uid: u32) -> PathBuf {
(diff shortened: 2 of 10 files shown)


ilysenko (changes requested):
Thanks for the update. I found several blockers on the current head:

- The launcher executes the bundled Chrome selector before the existing source-tree permission/symlink trust gate. Please validate the selector with the same trust policy before execution, or invoke an already trusted durable-runtime copy.
- The same-version decision is only a one-time check of a mutable `~/.codex-cli-npm` path. Background CLI preflight can upgrade that prefix before Chrome uses it, leaving Desktop on version X and Chrome on Y. Please make the selected identity durable through use or revalidate it after updater mutation.
- `select_chrome_cli_path` returns non-zero for an empty CLI and is called unguarded under `set -e`, aborting the supported warm/second-instance fallback. Please make the empty case a successful no-op and add coverage.
- Please make modern patch completeness require `CODEX_CHROME_CLI_PATH` and add a damaged-contract regression. Also make the new launcher smoke fixture architecture-neutral so it covers supported arm64 hosts.

ilysenko:
One process request for the follow-up: before creating or pushing the next commit, please run a full review of the complete final diff with your coding model at maximum reasoning effort. Keep iterating on the changes and do not commit until that review explicitly reports that no blockers remain. Then rerun the relevant tests and commit the reviewed result.

ilysenko:

PR #1310, opened 2026-08-11
Offer a local project when ChatGPT handoff would go projectless

### LINKED ISSUE #1309: Allow ChatGPT to Work handoffs on Linux to target a local folder
## Summary

This is solved in the linked PR by choosing a local folder at the ChatGPT -> Work handoff point.

When a ChatGPT conversation on Linux is handed off to Work mode and upstream would otherwise create a projectless Work thread, the app can ask for a local folder, create or reuse the local project for that folder, and start the Work thread with real filesystem hands.

## What the user sees

1. Start in a normal ChatGPT conversation.
2. Choose the existing ChatGPT -> Work handoff action.
3. If upstream has no synced ChatGPT Project to reuse, Linux prompts for a local folder.
4. Pick a folder and the new Work thread starts with that folder as its local project root.

This is slightly awkward, but it is the right place to make the choice: plain ChatGPT conversations do not have a project folder designation or cwd to edit in place. The handoff is the moment the Work thread exists.

## Why the handoff needs the folder prompt

ChatGPT conversations are projectless by design. For conversations outside synced ChatGPT Projects, upstream currently creates a `{type: "projectless"}` Work target. A projectless Work target has no local folder, cwd, or workspace-write boundary, so it cannot give the assistant hands over user files.

Asking for the folder during handoff turns that otherwise-projectless target into a local-project Work target instead.

## After handoff

Normal Work-mode shell behavior applies. The selected folder is the starting project root and workspace-write boundary. The user can still ask the assistant to `cd /some/path`, but that only changes the shell cwd inside the session; it does not retroactively change the project root or workspace-write boundary selected at handoff time.

## Fallback

Canceling the folder picker preserves upstream behavior and creates the same projectless target as before.

albertyang644:
Fixes #1309.

## Summary

This PR solves the Linux ChatGPT -> Work handoff case that would otherwise become projectless.

When upstream has no synced ChatGPT Project to reuse, the Linux wrapper now prompts for a local folder during the handoff, creates or reuses the corresponding local project, and starts the new Work thread with that folder as its project root. Canceling the picker preserves upstream behavior and continues projectless.

## User-visible behavior

1. Start in a normal ChatGPT conversation.
2. Use the existing ChatGPT -> Work handoff action.
3. If upstream would otherwise create a projectless Work thread, Linux asks for a local folder.
4. Picking a folder hands off into a local-project Work thread with filesystem hands for that folder.

The UX is slightly awkward because plain ChatGPT conversations do not have a project folder designation or cwd to edit in place. The handoff is the moment the Work thread exists, so it is also the right moment to choose the folder.

## Why this works

ChatGPT conversations are projectless by design. For conversations outside synced ChatGPT Projects, upstream currently chooses `{type: "projectless"}`, which gives the new Work thread no cwd or local workspace-write boundary.

This patch keeps the upstream handoff path intact, but replaces only that otherwise-projectless target with a local project selected by the user. After handoff, normal Work-mode shell behavior applies: users can still ask the assistant to `cd /whatever`, but that only changes the shell cwd; it does not change the project root or workspace-write boundary selected at handoff time.

## Source of truth files changed

- `scripts/patches/core/all-linux/main-process/window-shell/patch.js`
- `scripts/patches/core/all-linux/webview/chatgpt-handoff-local-project/patch.js`
- `scripts/patches/impl/main-process/misc.js`
- `scripts/patches/impl/webview/index.js`

## Validation

  - 438 pass, 0 fail
- `node --check` on changed patch/impl files
  - clean
- `bash -n install.sh scripts/lib/*.sh launcher/start.sh.template scripts/build-deb.sh scripts/build-rpm.sh scripts/build-pacman.sh`
  - clean from earlier pre-push validation; this PR does not change shell files
  - ok from earlier pre-push validation; this PR does not change Rust files

Environment caveats:

- `cargo test -p codex-update-manager` has one pre-existing failure on this machine: `trusted_standalone_installer_path_ignores_inherited_user_tools`. It fails the same way on stock `main`; this PR changes no Rust files.
- `tests/scripts_smoke.sh` is partial here because warm-start recovery is affected by local environment: `python3` resolves to miniconda 3.13.12 without `os.pidfd_open`, and the warm-start test's 5s Electron appearance window is tight for this machine. `launcher/`, `scripts/lib/`, and `install.sh` are byte-identical to `main` on this branch.

### DIFF
--- a/scripts/patch-linux-window-ui.test.js
+++ b/scripts/patch-linux-window-ui.test.js
@@ -100,6 +100,7 @@ const {
   applyLinuxHostProcessEnvironmentPatch,
   applyLinuxFileManagerPatch,
   applyLinuxGitOriginsSourceFallbackPatch,
+  applyLinuxPickLocalProjectHandlerPatch,
   applyLinuxLocalAppServerFeatureEnablementHandlerPatch,
   applyLinuxOwlFeatureBindingFallbackPatch,
   applyLinuxRemoteControlConfigPreservationPatch,
@@ -197,6 +198,8 @@ const {
   applyLinuxSidebarScrollPerformancePatch,
   applyLinuxSkillsListDedupePatch,
   applyLinuxThreadSidePanelNativeTooltipPatch,
+  applyLinuxChatGptHandoffLocalProjectPatch,
+  applyLinuxDesktopRequestGlobalPatch,
   applyLinuxTooltipWindowControlsCollisionPatch,
   applyLinuxWindowControlsSafeAreaPatch,
   applySubagentNicknameMetadataPatch,
@@ -1054,11 +1057,14 @@ test("default core patch descriptors are grouped and unique", () => {
     "linux-application-menu",
     "linux-app-reload-shortcuts",
     "linux-x11-project-picker",
+    "linux-pick-local-project-handler",
     "opaque-window-default-general-settings",
     "opaque-window-default-webview-index",
     "linux-window-controls-safe-area",
     "linux-tooltip-window-controls-collision",
     "linux-thread-side-panel-native-tooltip",
+    "linux-desktop-request-global",
+    "linux-chatgpt-handoff-local-project",
     "linux-fast-mode-model-guard",
     "subagent-nickname-metadata-shape",
     "local-environment-action-modal-draft",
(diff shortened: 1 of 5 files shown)


ilysenko (changes requested):
Thanks for the focused implementation and thorough tests. I found three blockers:

1. This intentionally differs from the standard macOS behavior. The current macOS app creates a generated projectless workspace under `Documents/Codex`; it does not show a folder picker during this handoff. Because this is a workflow-specific UX enhancement rather than Linux compatibility, please move the descriptors, implementation, and tests into a disabled-by-default `linux-features/chatgpt-handoff-local-project/` feature with `feature.json` and `README.md`.
2. Please validate or capture the complete minified contracts. The main transform injects uncaptured `n`/`V` aliases, and missing main/webview shapes can be reported as `applied` or `already-applied` without a warning. Add drift tests proving a warning and byte-identical skip instead of runtime-only failure.
3. Please correct the rationale and source comments. Standard projectless handoffs already receive a generated cwd, workspace root, and output directory. The feature’s actual value is selecting an existing local-project root, not introducing filesystem access.

Before requesting another maintainer review, please run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, rerun the review, and repeat until the model reports no remaining blockers.

ilysenko:

Issue #1257, opened 2026-08-08
Latest upstream ChatGPT DMG fails required patch detection (`linux-avatar-overlay-mouse-passthrough`) on v0.10.3


leviowern-cmyk:
### Environment
- Distribution: Nobara 44 (Fedora-based)
- Desktop: KDE Plasma
- codex-desktop-linux version: 0.10.3
- Source commit: bc7b92cf38c74b49dbff568257542eabbc62cf55

### Description
After the latest upstream ChatGPT Desktop update, the automatic local rebuild fails.

The DMG downloads successfully, extraction succeeds, Electron headers are downloaded, and all native modules rebuild successfully. The failure occurs during the Linux patch phase.

### Relevant output
```
Detected Electron version from package.json: 42.3.0
Native modules built successfully
Patching Linux window and shell behavior...
WARN: Could not find avatar overlay mouse passthrough policy — skipping Linux avatar overlay passthrough recovery patch
Critical patch failures (2):
linux-avatar-overlay-mouse-passthrough
main-process-ui
Aborting: these patches are required for a working Linux app.
```

Updater exits with `install.sh failed during local rebuild` and rejects the candidate build.

Everything before the patch stage succeeds:
- DMG downloaded successfully
- DMG extracted with 7z
- Electron version detected (42.3.0)
- better-sqlite3 rebuilt
- node-pty rebuilt
- Native module rebuild completed successfully

This appears to be upstream JavaScript layout drift causing the required patch to no longer match. The full install.log is available if needed.

ilysenko:
Thanks for the report. This exact upstream drift was fixed by #1242, which retargeted `linux-avatar-overlay-mouse-passthrough` to the new upstream `applyInputShape` contract. Your build uses commit `bc7b92c`, which predates that fix; current `main` successfully builds the same ChatGPT 26.803.41515 / Electron 42.3.0 DMG with both required patches applied. Please update the wrapper checkout to current `main` (or a release containing #1242) and retry. If the problem persists, please follow up or open a new report with a fresh patch/rebuild report from current `main`. Closing this as already fixed.

PR #1311, opened 2026-08-11
docs: explain required-patch rebuild failures

Lucacux:
## Summary

- document the `install.sh failed during local rebuild` signature when a newer upstream DMG outpaces the installed Linux wrapper
- point users to the updater service log and packaged wrapper provenance
- recommend updating the wrapper/package before retrying
- explicitly discourage bypassing required patches during routine updates

## Why

The acceptance and architecture docs explain required-patch rejection, but the troubleshooting table did not connect the user-facing updater notification to wrapper drift or give a safe recovery path. This is the failure mode reported in #1257 and addressed for that DMG family by #1242.

## User impact

Users can distinguish wrapper drift from a broken local install, preserve the working app, and collect the right reports if current `main` still rejects the DMG.

## Validation

- `git diff --check`
- documentation-only change; commands and paths were verified on a native Debian-family package install


ilysenko:

ilysenko (approved):
Reviewed the current documentation-only diff. The troubleshooting guidance matches the updater logs, packaged wrapper provenance, and required-upstream failure policy.

PR #1315, opened 2026-08-12
Fix updater drift for current DMG and replaced app liveness

wesleytsai:
[Local Agent]\n\n## Summary\n- update the read-aloud feature matcher for the current local-conversation-turn bundle and select the matching asset semantically\n- detect a still-running Electron process whose executable was replaced during package installation\n\n## Root cause\nThe cached upstream DMG moved the assistant render call from app-initial into local-conversation-turn, causing the enabled read-aloud feature to be silently omitted by acceptance. Replacing /opt/codex-desktop/electron exposes the existing process as a deleted executable in proc, which made diagnostics report the app as stopped.\n\n## Validation\n- 60 read-aloud tests passed\n- 428 patcher tests passed\n- targeted updater liveness tests passed\n- real DMG required-patch acceptance passed with warnings; read-aloud applied and integrity findings were 0\n- installed and verified locally as 2026.08.11.220000+d9e0901\n\nThe only build warning is the existing optional Browser Use external availability drift.


ilysenko:

Issue #1316, opened 2026-08-12
Temporary pause on pull requests and new issues: migrating to OpenAI’s official Linux app


ilysenko:
OpenAI has released an official native Linux build of ChatGPT Desktop. This project is already being migrated to that Linux build right now, and the migration is actively in progress.

This project is **not being deprecated**. It is moving away from the current macOS DMG conversion approach and being rebased on OpenAI’s official Linux build, while preserving the additional package formats, Linux compatibility work, and unique features provided by this repository.

## Please hold off on new pull requests and issues

Until this transition is complete:

- Please **do not open new pull requests**. Changes based on the current DMG conversion pipeline may conflict with the migration or become obsolete shortly afterward.
- Please **do not open new issues** unless they directly concern the migration and contain concrete findings or recommendations about adapting this project to OpenAI’s official Linux build.

General bug reports, feature requests, and support questions should wait until the migration is complete, since the underlying application and build process are changing significantly.

If you test the official Linux version and notice something relevant to the migration—missing functionality, packaging limitations, compatibility problems, or behavior that this project should preserve—please share it as a comment here. Open a separate issue only when the recommendation requires focused technical discussion.

## How you can help in the meantime

Please try OpenAI’s official Linux version and pay particular attention to:

- installation and startup problems;
- differences between distributions and package formats;
- Wayland, X11, desktop environment, and window manager compatibility;
- features available here that are missing or broken in the official version;
- theme, performance, rendering, and integration issues;
- functionality that may be lost or require special handling during the migration;
- packaging limitations, dependency problems, and anything else that could go wrong while adapting this project to the official Linux build.

When reporting a finding, please include your Linux distribution and version, desktop environment, session type, package used, and clear reproduction steps.

The migration is happening now. An update will be posted when the new foundation is ready and normal pull requests and issues can be accepted again.

This transition should reduce the amount of macOS-specific patching and allow the project to focus on Linux packaging, compatibility, and the features that make it useful across a wider range of Linux systems.

Thank you for your patience and for helping test the new upstream Linux version.

Related discussion: #1312

ilysenko:
<img width="954" height="150" alt="Image" src="https://github.com/user-attachments/assets/811efab3-6c9f-45cf-bf45-60d78e263b18" />

PR #1318, opened 2026-08-12
Document Chrome native host parent trust

anervalens-netizen:
## What

- document the exact Chrome side-panel `required path parent` failure
- explain the native host parent-chain ownership and permission invariant
- give targeted diagnostics and recovery without editing generated manifests

## Why

The v2 app-server and native-host manifests can be structurally valid while a group/world-writable ancestor still fails the native host trust check. The current error text points users toward the manifest instead of the filesystem chain.

## Impact

Maintainers and users can identify the real filesystem trust failure and preserve the security check instead of weakening it or patching generated manifests.

## Checks

- `git diff --check`
- reproduced on Dell with `/opt` effectively `0775` through its ACL
- native `hello` and `ensure` succeeded after restoring the intended ancestor mode to `0755`
- Chrome side panel loaded normally after retry


ilysenko:
https://github.com/ilysenko/codex-desktop-linux/issues/1316

PR #1335, opened 2026-08-13
Fix shared app-server startup without explicit CLI path

moxw:
## Summary

Enabling `shared-app-server-socket` patched Desktop to require
`CODEX_CLI_PATH`, but the feature's launcher hook exported only the shared
socket path. Fresh native-package launches therefore stopped before the main
window with `shared app-server socket requires CODEX_CLI_PATH`.

The hook now defaults `CODEX_CLI_PATH` to the official bundled
`resources/codex` executable while preserving an explicit override. Regression
coverage exercises both cases, and the feature documentation records the
launcher-owned default.

User-visible result: installations with this disabled-by-default feature
enabled start normally without requiring a manual environment override.

The declarative hook is shared by deb, RPM, pacman, AppImage, and Nix feature
staging. Manual package/install proof was performed on Arch x86_64; the other
formats and arm64 were not manually built in this change.

## Validation

- `CODEX_CLI_PATH=/opt/codex-desktop/resources/codex node --test linux-features/shared-app-server-socket/test.js` (41 passed, including the real authority/proxy lifecycle)
- `bash tests/scripts_smoke.sh` (46 passed)
- `git diff --check`
- Feature-only build from the latest signed stable OpenAI Linux package with only `shared-app-server-socket` enabled (one feature patch applied, hook staged, runtime diagnosis passed)
- `make pacman`, inspected the packaged hook, and installed `codex-desktop 2026.08.13.101330-1`
- Plain installed launch with caller `CODEX_CLI_PATH` unset: Codex CLI handshake succeeded, the window reached `ready-to-show`, and the owner-only shared socket was listening with mode `0600`

## Checklist

- [x] Upstream drift handling is not applicable; this fixes the launcher contract of one disabled-by-default Linux feature.

### DIFF
--- a/linux-features/shared-app-server-socket/README.md
+++ b/linux-features/shared-app-server-socket/README.md
@@ -17,6 +17,10 @@ PATH` byte tunnel and its existing WebSocket transport. Other local clients use
 the same stock proxy command to attach to the Unix socket and receive the normal
 WebSocket `/rpc` byte stream. Closing Desktop stops the authority.
 
+The launcher uses the official CLI bundled in `resources/codex` by default.
+An explicit `CODEX_CLI_PATH` remains supported and is preserved by the feature
+hook.
+
 The default socket is scoped by Linux app id under `XDG_RUNTIME_DIR`, preventing
 side-by-side Desktop instances from sharing an authority accidentally. Override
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed the current head. The feature-local launcher default restores startup without overriding explicit CLI paths, and the focused regression coverage and required checks pass.

PR #1345, opened 2026-08-13
Restore pre-migration hover scrollbars on official Linux

wy3z:
## Summary

Official Linux paints classic Chromium thumbs on every `overflow-*-auto` scroller. macOS overlay-hid these before the Linux package migration.

This restores that hover-hide on the project sidebar, chat thread, and other overflow surfaces.

Shipped as a disabled-by-default feature so a clean build still preserves `app.asar`. If maintainers prefer, keep it opt-in rather than a default fix.

## Validation

- `node --test linux-features/hover-scrollbars/test.js`
- Official `26.810.41047` `app-initial-iMhn6nFd.js` applies once and is idempotent
- Feature/smoke/updater tests from the earlier review pass
- Manual: hover hide on the project sidebar after rebuild

## Checklist

- [x] Ready for review
- [x] Focused source change, no generated output
- [x] Latest official Linux package; no obsolete fallbacks
- [x] Tests added and run
- [x] Reviewed the final diff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Added
 
+- An opt-in `hover-scrollbars` feature restores the pre-migration overlay-style
+  hover hide on official Linux overflow scrollers. Classic Chromium thumbs stay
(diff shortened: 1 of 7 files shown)


wy3z:
clanker issues

PR #1349, opened 2026-08-14
Fix bundled marketplace staging permission loop

0xdeafbeef:
## Summary

Nix store sources give bundled marketplace plugin files read-only modes. The signed app copies those modes into `openai-bundled.staging-*`, then fails when it writes plugin metadata or removes an interrupted staging tree. Repeated reconciliation can therefore leave additional full staging copies behind.

This PR adds one internal Nix-only ASAR repair. It wraps the current bundled-marketplace copy contract in `try/finally` and recursively adds owner-write permission to real directories and regular files in the copied destination. Symlinks and special nodes are skipped, a missing destination is harmless, and real `chmod` failures still propagate.

The repair is injected only by `mkCodexDesktop`. It is hidden from public Linux feature selection and replaces the narrower Computer Use-owned patch, so default, Computer Use, and maximal Nix profiles compose exactly one staging-permission descriptor. Native deb/RPM/pacman/AppImage builds, the updater, and public `.#installer` remain descriptor-free and preserve the official ASAR byte-for-byte.

## Drift policy

User Nix package builds treat this repair as best-effort. If the current upstream main-bundle contract cannot be found or the optional patch otherwise drifts, the report records `skipped-optional`, emits a warning, and preserves an unchanged official ASAR byte-for-byte. Required-upstream descriptors, integrity failures, and all other enabled features remain fail-closed.

Repository CI is deliberately stricter: runtime checks for default, Computer Use, and both maximal profiles require this exact descriptor to be `applied` on the current signed official ASAR on amd64 and arm64.

This change does not scan or remove staging directories that already accumulated. Existing leaked data still requires manual cleanup.

## Validation

- Full CI-equivalent Node suite: 849 passed, 1 pre-existing skip, 0 failed
- Shell smoke suite: 50/50 passed
- Shell syntax checks and `git diff --check`
- Three independent blocker reviews of the final base-to-head diff: no findings
- All GitHub checks passed, including real Nix builds and strict runtime audits on x86_64 and aarch64

## Checklist

- [x] Synchronized by merging current `main` without force-push
- [x] Current-DMG-only contract; no legacy fallback path
- [x] Optional consumer drift and strict repository validation are tested separately
- [x] Native package/updater/public-installer isolation is preserved
- [x] Final diff was independently reviewed until no blockers remained

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,8 @@ jobs:
           node --test scripts/lib/package-common.test.js
           node --check scripts/ci/verify-official-linux-features.js
           node --test scripts/automation/upstream-linux-package-watchdog/test.js
-          node --test scripts/patch-linux-window-ui.test.js scripts/lib/linux-features.test.js linux-features/*/test.js
+          node --test scripts/patch-linux-window-ui.test.js scripts/patches/descriptor.test.js \
+            scripts/lib/linux-features.test.js linux-features/*/test.js
       - name: Smoke tests
         run: bash tests/scripts_smoke.sh
 
@@ -71,6 +72,8 @@ jobs:
         run: nix build ".#checks.${{ matrix.system }}.modules" --no-link
       - name: Build audited default runtime
         run: nix build ".#checks.${{ matrix.system }}.nix-runtime" --no-link
+      - name: Build audited Computer Use runtime
+        run: nix build ".#checks.${{ matrix.system }}.nix-runtime-computer-use" --no-link
       - name: Build audited directory-watch feature profile
         run: nix build ".#checks.${{ matrix.system }}.nix-runtime-maximal-directory-watch" --no-link
       - name: Build audited shallow-watch feature profile
--- a/docs/linux-features-architecture.md
+++ b/docs/linux-features-architecture.md
@@ -59,6 +59,7 @@ Manifest fields:
 | `id` | Stable configuration ID matching the directory name |
 | `title`, `description` | User-facing wizard and documentation text |
 | `defaultEnabled` | Must be `false` for every repository and local feature |
+| `internal` | Optional boolean for build-owned plumbing hidden from public feature selection |
 | `entrypoints.patchDescriptors` | Optional ASAR descriptor module |
 | `entrypoints.stageHook` | Last-resort app staging script |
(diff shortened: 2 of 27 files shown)


ilysenko (changes requested):
Thanks for investigating and fixing the staging permission loop. The permission repair itself looks sound, but the current PR makes a Nix-store-specific compatibility issue part of every default build and updater rebuild.

A much smaller design should be possible:

- Move the descriptor into a self-contained disabled-by-default feature such as `linux-features/nix-store-bundled-marketplace-permissions/`.
- Enable that feature internally only for the immutable `mkCodexDesktop` Nix outputs. The ordinary native packages and `.#installer` produce writable application trees and do not need this repair.
- Reuse the existing pinned `${pkgs.asar}` path in the Nix derivation.
- Require the descriptor to be reported as `applied` before accepting the Nix output.
- Keep the core registry empty and revert the global baseline, native updater, package, and unrelated documentation changes.

This would keep the fix on the affected surface and remove most of the current blockers automatically.

With the current global design, these blockers still need to be addressed:

1. Every default native install and updater rebuild now executes `npx --yes @electron/asar`, but the deb/RPM/pacman updater does not bundle an ASAR CLI or guarantee `npm`/`npx`. It also introduces a live, unpinned npm-registry dependency into the default rebuild path.
2. If main-bundle discovery fails, `--enforce-critical` only warns, writes no descriptor entry, and exits successfully. A required patch must not allow an unpatched candidate to be accepted. The Nix-only design can close this narrowly with an explicit `--require-applied` validation.
3. The new `.#installer` source copy assumes `TMPDIR` is set under `set -u` and is never cleaned up. An unset `TMPDIR` currently aborts the installer. This entire change should be unnecessary with the Nix-only design.

Please verify the revised change against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant local tests before pushing the update, and wait for every required CI check to complete successfully.

Before requesting maintainer re-review, run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, rerun the review, and repeat until that model reports no remaining blockers.

ilysenko (approved):
Reviewed the current head after the Nix-only optional redesign, current signed-package update, independent blocker reviews, and full green CI. The earlier requested changes are resolved.

PR #1350, opened 2026-08-14
Autodetect native Wayland sessions

0xdeafbeef:
## Summary

The shared launcher now selects native Wayland automatically when
`XDG_SESSION_TYPE=wayland` and `WAYLAND_DISPLAY` resolves to a live listening
Unix socket. `DISPLAY` does not disable detection because normal Wayland
sessions commonly export it for XWayland compatibility.

Explicit backend selections from global or app-specific Electron flags,
features, launcher hooks, and command-line arguments always suppress automatic
defaults. Users can opt out for one launch with:

```bash
codex-desktop --ozone-platform=x11
```

The existing `NIXOS_OZONE_WL` compatibility opt-in remains supported. Backend
selection now lives in the shared launcher instead of the Nix wrapper, so the
same behavior applies to Nix, deb, RPM, pacman, and AppImage packages.

This fixes incorrectly tiny XWayland cursors on HiDPI GNOME sessions while
keeping an explicit X11 escape hatch.

## Validation

- `node --test launcher/start.test.js` — 10 passed
- Verified detection against the live GNOME Wayland socket with
  `NIXOS_OZONE_WL` unset
- Built and launched the Nix package; the app selected
  `--ozone-platform=wayland` and the cursor rendered at the expected size

## Checklist

### XWayland — before

Tiny cursor under XWayland

### Native Wayland — after

Correctly scaled cursor under native Wayland

The screenshot tool renders both cursors larger than they appeared on screen;
the relative size difference is what matters.

Tested end-to-end on GNOME Wayland with `XDG_SESSION_TYPE=wayland`,
`WAYLAND_DISPLAY=wayland-0`, and `NIXOS_OZONE_WL` unset. Native Wayland works
correctly. Explicit backend overrides were also verified from global and
app-specific Electron flags, feature arguments, launcher hooks, and direct CLI
arguments; `codex-desktop --ozone-platform=x11` remains a working opt-out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,7 +349,7 @@ output layout, parallelism, and payload inspection.
 | Browser/Chrome extension cannot connect after migration | Exit ChatGPT and Chrome completely, then follow the narrow cache repair in [Troubleshooting](docs/troubleshooting.md#browser-or-chrome-plugin-is-visible-but-cannot-connect) |
 | Signature or package verification fails | Do not bypass it; check time, network, `gpgv`, architecture, and disk space |
 | App does not launch | Run `/opt/codex-desktop/start.sh --diagnose` |
-| App uses XWayland or needs persistent Electron flags | Put one flag per line in `~/.config/codex-desktop/electron-flags.conf`; for example, `--ozone-platform=wayland` |
+| App uses XWayland or needs a backend override | Native Wayland is automatic for a Wayland session with a live display socket, or the explicit `NIXOS_OZONE_WL` compatibility opt-in. Use `codex-desktop --ozone-platform=x11`, or put one Electron flag per line in `~/.config/codex-desktop/electron-flags.conf` |
 | AppImage reports a sandbox error | Enable user namespaces or install a native package; `--no-sandbox` is not added automatically |
 | Enabled feature drifts after an upstream release | Disable that feature to confirm the clean baseline and attach its patch report to an issue |
 | Updater waits for application exit | Close official and Community processes, then inspect `codex-update-manager status --json` |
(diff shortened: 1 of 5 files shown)


ilysenko (changes requested):
Thanks for the contribution and the thorough test coverage. Please address these blockers before re-review:

1. Respect explicit Wayland IME overrides. Automatic detection currently adds `--enable-wayland-ime=true` even when `--disable-wayland-ime` was supplied; Chromium gives the enable switch precedence. Do not inject conflicting IME defaults, and add regression coverage.
2. Remove the retired backend selectors from detection, tests, and documentation. The current Electron 42.3.0 / Chromium 148 contract exposes `--ozone-platform`; `--ozone-platform-hint`, `--x11`, and `--wayland` must not suppress defaults as if they selected a current backend.
3. Update the branch onto current `main`, resolve the launcher test conflict while preserving the usage-reporting test isolation, and rerun the required CI on the resolved head.

Please run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, rerun the review, and repeat until that model reports no remaining blockers before requesting maintainer re-review.

0xdeafbeef:
@ilysenko do you want any else changes? I've fixed all the issues

ilysenko:
Thanks for the work here, and sorry this sat without maintainer follow-up. I'm closing this older inactive PR as part of a queue cleanup rather than asking you to keep rebasing a stale branch. The current head now conflicts with main and would need fresh current-runtime validation and CI. If you'd still like to pursue it, a new focused PR from the latest main is welcome.

PR #1354, opened 2026-08-14
Separate Chronicle Skysight as an opt-in feature

### LINKED ISSUE #1353: Separate Chronicle Skysight into an independent opt-in feature
## Problem

Chronicle/Skysight activity memory is currently coupled to the broader Record & Replay feature. Users who only want continuous activity-memory resources must enable the recording workflow and share its plugin lifecycle.

## Proposed fix

- Add a disabled-by-default `chronicle-skysight` feature with its own manifest, docs, patch surface, stage and cleanup hooks.
- Stage a restricted MCP plugin exposing only activity-memory tools.
- Make Record & Replay depend on and reuse the shared Skysight backend.
- Keep both features independently selectable and fail closed when the dependency is missing.

The implementation is already exercised in the PatchRaptor fork and is ready to contribute upstream.

joshyorko:
## Summary

- add a disabled-by-default `chronicle-skysight` feature with its own plugin, bridge, stage, cleanup, and documentation surface
- move the shared activity-memory backend out of Record & Replay ownership
- make Record & Replay declare and reuse the Skysight dependency
- retain and prebuild the shared backend for Chronicle-only native, Nix, and updater rebuilds
- fail closed when the dependency or current official-package contract is missing

Closes #1353.

## Validation

- `node --test scripts/lib/package-common.test.js` (9 passed)
- `node --test linux-features/chronicle-skysight/test.js linux-features/record-and-replay/test.js` (45 passed)
- `bash tests/scripts_smoke.sh` (46 passed)
- `CI_SKIP_PULL=1 ./scripts/ci-local.sh core` (738 passed, 1 skipped)
- `git diff --check`

## Checklist

- [x] I added or updated relevant tests and ran the validation listed above. Required remote CI checks are pending.


ilysenko (changes requested):
Thanks for separating Chronicle / Skysight from the broader Record & Replay workflow. Keeping this as a repository-owned, disabled-by-default feature is reasonable, and this review does not require a broader framework redesign. A few concrete blockers still need to be addressed:

1. **The standalone Chronicle plugin exposes the full Record & Replay MCP server.** `skysight mcp` currently uses the same router as `event-stream mcp`, including recording, bundle drafting, and skill import tools. Please provide a genuinely Skysight-only tool catalog and add a test that verifies the exact exported tools.

2. **The Chronicle tray cannot start capture.** The bridge passes `--source` and `--owner` to `skysight start`, but the Rust CLI does not support those arguments. Please either remove them or add end-to-end CLI support, with a regression test covering the actual tray start command.

3. **Native updater rebuilds cannot consume the retained backend.** The update-builder correctly carries `target/release/codex-record-replay-linux`, but the shared stage helper requires and invokes Cargo before checking that artifact. The minimal updater bundle has no Cargo workspace. Please use the retained executable first and add a focused test for the minimal update-builder rebuild path.

4. **The dependency contract and documentation disagree.** Existing configurations containing only `record-and-replay` now fail, while the Chronicle README says dependencies are enabled and disabled automatically. Automatic dependency management is not required for this PR: an explicit two-feature configuration is acceptable. Please document that contract, update the Record & Replay example and migration instructions, and add `chronicle-skysight` to the English and Chinese feature tables.

A permanent CI matrix for every package format is not required, but the updated PR should include focused regressions for the contracts above and evidence of a Chronicle-only build from the current signed package through the relevant native/update path.

Before requesting maintainer re-review, please verify the change against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md) and run the relevant tests locally. Then run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, rerun the review, and repeat until the model reports no remaining blockers. After pushing the update, wait for every required CI check to complete successfully before requesting review or merge.

ilysenko:
This review was generated by my bot. Please use your judgment and fix what you think is relevant.

ilysenko (approved):
Reviewed the current head after the Chronicle lifecycle work landed in #1355. The remaining exact MCP catalog regression is clean and all checks pass.

PR #1365, opened 2026-08-16
fix: reconcile updater state after self-upgrade

Lucacux:
## What this fixes

A self-upgrade replaces the updater itself. `codex-update-manager.prerm` stops every
active user's `codex-update-manager.service` before dpkg unpacks, so the daemon that
launched the privileged install can exit before it persists the `Installing` →
`Installed` transition. The next daemon reads a pending candidate whose upstream
release is already installed, and rebuilds and reinstalls it on every check.

## What changed

Startup reconciliation of an interrupted install, gated on evidence that identifies
**this** rebuilt artifact rather than the upstream input:

1. It runs only from the persisted `Installing` status, and only for the commands that
   act on state (`daemon`, `check-now`, `install-ready`, `rollback`). `status` and
   `diagnose` keep reporting what is persisted.
2. It reads the native package version out of `state.artifact_paths.package_path` with
   the existing deb/RPM/pacman parsers (`install::package_file_version`) and compares it
   against `install::installed_package_version()`.
3. It additionally requires the package manager to report that version for a payload it
   has committed (`install::installed_package_is_usable`). `dpkg-query -W -f=${Version}`
   answers from the moment unpacking starts, and `prerm` has already stopped the daemon
   by then, so a failed unpack or a failed `postinst` would otherwise be recorded as a
   successful install. The allowlist accepts `installed`, `half-configured`,
   `triggers-awaited` and `triggers-pending`; `half-configured` is accepted because
   `postinst` starts the service from inside the dpkg transaction, so the recovering
   daemon normally runs while the package it is recovering is still being configured.
   RPM and pacman commit the payload before their databases answer, so the distinction
   is dpkg-only.
4. Only when all of that holds does it move the candidate upstream version/SHA-256 into
   the installed fields and clear the candidate. A missing artifact, an uninspectable
   artifact, a differing native version, an uncommitted payload, or an `unknown`
   installed version all preserve the candidate and skip reconciliation.

The upstream version and SHA-256 are never used as the decision input: they identify the
signed OpenAI `.deb`, not the per-user package rebuilt from it, and different per-user
builds can share them while differing in package version or enabled features.

No new `build-info.json` parser, no new state schema, no new dependency.

### Supporting changes

- `record_new_candidate` now clears `artifact_paths.package_path`. `build_update` only
  republishes the artifact paths once it has produced a package, so a failed download or
  build previously left `package_path` pointing at the *previous* release's artifact
  while the candidate fields already described the new one. Both consumers of
  `package_path` (the `install-ready` retry and this reconciliation) read it as the
  artifact of the current candidate, so it has to be absent until it is. The rollback
  target is captured, and the cache pruned, before the reset.
- `install-ready` with no recorded artifact is now a routine outcome instead of a
  `context()` error; on a `Failed` status it reports the recorded `error_message` rather
  than printing "No rebuilt package is ready to install."
- `package_file_version` resolves symlinks through the existing `package_identity_path`
  helper before parsing, because the pacman version lives in the file name and
  `codex-desktop-latest.pkg.tar.zst` is a symlink — the same thing
  `stable_validated_package` does.

## Tests

`updater/src/app.rs` (reconciliation is exercised through an injected
`installed_package_is_usable` and package-version reader, so no package manager is
needed):

- recovers an install whose native package version is the installed one;
- **preserves a candidate that shares the upstream release but has a different native
  package version** (the case called out in review);
- preserves a candidate the package manager recorded but did not install (payload not
  committed);
- preserves the candidate when the artifact is missing from the state, cannot be
  inspected, or the installed version is `unknown`;
- **does not reconcile any status other than `Installing`** (loops over all eleven);
- a failed build cannot be recovered as an installed candidate;
- a new candidate drops the previous candidate's artifact.

`updater/src/install.rs`: reads the native version from a real `dpkg-deb`-built `.deb`
(skipped when `dpkg-deb` is absent), from a pacman artifact, through a pacman `latest`
symlink, rejects a missing artifact, and accepts only the dpkg states that have the
payload installed.

## Verification

- `cargo test -p codex-update-manager` — 63 passed
- `cargo clippy -p codex-update-manager --all-targets -- -D warnings` — clean
- `cargo test -p codex-record-replay-linux` and its clippy — clean
- `bash -n install.sh scripts/lib/*.sh launcher/start.sh.template` — clean
- `cargo fmt --check` reports the same 51 hunks on `main` and on this branch; a
  hunk-by-hunk diff shows this branch adds none. Per CONTRIBUTING the pre-existing
  formatting churn is left out of this behaviour change.
- `bash tests/scripts_smoke.sh` could not run locally (no `ripgrep` on this machine); CI
  installs it.
- Branch is rebased on current `main`.

## Self-review

The base-to-head diff was run through a code-review model repeatedly, fixing what it
found, until the remaining findings were all pre-existing conditions rather than
anything the diff introduces:

1. `record_new_candidate` was leaving the previous release's artifact reachable from a
   new candidate — fixed (see above).
2. `package_file_version` was not resolving the pacman `latest` symlink — fixed.
3. `install-ready` was reporting a `Failed` state as "nothing to install" — fixed.
4. `dpkg-query` reports a version for a payload it has not committed — fixed with the
   status allowlist.
5. Two findings left unfixed, both verbatim on `main` and unchanged by this diff:
   - `check`'s `same_failed_candidate` guard (identical on `main`) keeps the updater in
     `Failed` for as long as upstream serves the release whose download or build failed,
     so a transient network error suspends updates until the next upstream release. This
     PR does not introduce that; it removes the one escape hatch that "worked", which
     was `install-ready` installing the *previous* release's artifact and recording it
     under the new upstream version and SHA-256 — exactly the corruption this PR exists
     to prevent. Making that state retry is a retry-policy decision (a naive re-check
     would rebuild a deterministically failing candidate on every tick), so it is left
     to you rather than folded in here.
   - `state.installed_version` is still assigned from the raw package-manager answer at
     every entry point, as it is on `main` (three sites, unchanged), so during a dpkg
     transaction `status` can report a version whose payload is not committed. The
     reconciliation does not rely on that field alone, which is why the stricter test
     lives in the predicate. Applying it to the assignment as well would change what
     `installed_version` means for `rollback`'s known-good capture too, which is beyond
     this fix.

## Out of scope, reported rather than bundled

Two further pre-existing issues found while reviewing this change. CONTRIBUTING asks not
to mix unrelated cleanup into a behaviour change, and the one-open-PR rule rules out a
parallel PR, so they are only reported here:

- `builder.rs` `find_package` requires exactly one match in `DIST_DIR`, but
  `scripts/build-pacman.sh` leaves both the versioned package and the
  `codex-desktop-latest.pkg.tar.zst` symlink there, so every Arch rebuild on `main`
  fails the match. Happy to send a fix once this PR is resolved.
- An interrupted **rollback** has the same shape of problem, but rollback persists no
  candidate tuple, so recovering it needs the rollback target's upstream identity, which
  is not in the state today. That is a different mechanism from the candidate-scoped fix
  requested here.

### DIFF
--- a/docs/updater.md
+++ b/docs/updater.md
@@ -40,6 +40,25 @@ waits for process exit, then performs the same atomic candidate exchange used by
 manual rebuilds. A durable journal recovers interrupted promotion, and the
 immediately previous managed package remains the rollback target.
 
+Recording a new candidate drops the previous candidate's package artifact, so a
+recorded artifact always belongs to the pending candidate. A failed download or
+build therefore leaves nothing for `install-ready` to retry until a rebuild
+produces a package; an explicit `install-ready` in that state reports the
+recorded failure instead of succeeding silently.
+
+A self-upgrade replaces the updater itself, so the package lifecycle can stop
+the service before the daemon records the install. On startup an `Installing`
+state is recovered only when the native package version of the candidate
+artifact equals the package version the system reports as installed, and only
+when the package manager reports that version for a payload it has committed.
+dpkg answers with the new version from the moment it starts unpacking, so an
+`unpacked` or `half-installed` package is not treated as installed;
+`half-configured` is, because the service is started from `postinst` inside the
+same transaction. A missing or uninspectable artifact, any other native
+version, or any other status keeps the candidate pending: the upstream version
+and SHA-256 identify the signed OpenAI input, not the per-user package rebuilt
+from it.
+
 Automated user-local operations cannot override the running-app guard or
 silently accept unverified input. A failed privileged installation remains
 failed until an explicit retry or a newer candidate.
@@ -138,8 +157,8 @@ rollback path; it is not a normal troubleshooting step.
 ## Validation scenarios
 
 Tests cover new and unchanged releases, interrupted downloads, all trust
-failures, build-while-running, promotion-after-exit, rollback, cleanup, and old
-state migration. Run:
+failures, build-while-running, promotion-after-exit, interrupted self-install
+recovery, rollback, cleanup, and old state migration. Run:
 
 ```bash
 cargo test -p codex-update-manager
--- a/updater/src/app.rs
+++ b/updater/src/app.rs
@@ -12,7 +12,7 @@ use anyhow::{Context, Result};
 use chrono::Utc;
 use std::{fs::{self, OpenOptions}, path::Path, time::Duration};
 use tokio::time;
-use tracing::{error, info};
+use tracing::{error, info, warn};
 
 pub async fn run(cli: Cli) -> Result<()> {
     if let Some(result) = run_privileged_command(&cli.command) {
@@ -24,6 +24,24 @@ pub async fn run(cli: Cli) -> Result<()> {
     let config = RuntimeConfig::load_or_default(&paths)?;
     let mut state = PersistedState::load_or_default(&paths.state_file, config.auto_install_on_app_exit)?;
     state.installed_version = install::installed_package_version();
+    // Only the commands that act on the state recover it; `status` and
+    // `diagnose` report what is persisted and must not change the recorded
+    // status or drop a candidate. (They still rewrite the file with the
+    // refreshed `installed_version`, as they always have.)
+    //
+    // This is deliberately not serialised against a concurrent updater
+    // process. `check.lock` would not help: neither `install_ready` nor the
+    // daemon's deferred-install tick holds it, and failing to take it would
+    // silently skip the recovery for the life of the process — reinstating the
+    // rebuild loop. The predicate is safe unsynchronised because it fires only
+    // once the package manager already reports the candidate's own package
+    // version.
+    if matches!(
+        cli.command,
+        Commands::Daemon | Commands::CheckNow | Commands::InstallReady | Commands::Rollback
+    ) {
+        reconcile_interrupted_install(&mut state);
+    }
     state.save_updater(&paths.state_file)?;
 
     match cli.command {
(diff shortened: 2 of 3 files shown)


ilysenko (changes requested):
Thanks for fixing the repeated rebuild/install loop after a self-upgrade.

There is one merge blocker:

The startup reconciliation currently treats the upstream version and SHA-256 as proof that the exact rebuilt `codex-desktop` candidate was installed. That tuple identifies only the official upstream input, not the resulting native package. Different per-user builds can share it while having different package versions or enabled features. Because package lifecycle hooks can restart every active user's updater, installing one user's package can silently clear another user's pending candidate.

A simpler and safer fix does not require a new `build-info.json` parser or a new state schema:

- Reconcile only when the persisted status is `Installing`.
- Read the native package version from `state.artifact_paths.package_path` using the existing deb/RPM/pacman package-version parsers.
- Compare it with `installed_package_version()`.
- Only when they match, move the candidate upstream version/SHA into the installed fields and clear the candidate. If the artifact is missing or cannot be inspected, preserve the candidate and skip reconciliation.
- Add regression coverage for a same-upstream but different native package candidate, and verify that states other than `Installing` are not reconciled.

Please verify the updated change against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md) and run the relevant tests locally before pushing it.

Run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, rerun the review, and repeat until the model reports no remaining blockers before requesting maintainer re-review. After pushing the update, wait until every required CI check has completed successfully before requesting review or merge.

Lucacux:
Thanks for the review — the reconciliation no longer treats the upstream version and
SHA-256 as evidence of anything about the rebuilt package.

It now runs only from the persisted `Installing` status, reads the native package version
out of `state.artifact_paths.package_path` with the existing deb/RPM/pacman parsers, and
promotes the candidate only when that equals `installed_package_version()`. A missing or
uninspectable artifact, or any other native version, preserves the candidate and skips
reconciliation. No `build-info.json` parser and no state-schema change.

One addition beyond your outline, which my own review pass turned up: a reported version
is not proof of an installed payload. `dpkg-query -W -f=${Version}` answers from the
moment unpacking starts, and `prerm` has already stopped the daemon by then, so a
self-upgrade that fails to unpack or to configure would have been recorded as a
successful install. Promotion now also requires a committed payload via
`${db:Status-Status}`, accepting `half-configured` because `postinst` starts the service
from inside the same dpkg transaction. RPM and pacman commit before their databases
answer, so the check is dpkg-only.

Regression coverage for both cases you asked about is in `updater/src/app.rs`: a candidate
sharing the upstream release but not the native package version is preserved, and every
status other than `Installing` is left untouched (the test loops over all eleven).

I ran the diff through my code-review model repeatedly until the only remaining findings
were pre-existing conditions verbatim on `main`; the PR description lists them, along with
two unrelated pre-existing bugs I found and deliberately did not bundle in. The
`find_package` one breaks every Arch rebuild on `main` today — happy to send it as a
separate PR once this one is resolved.

`cargo test`, both clippy runs and `bash -n` are clean locally, `cargo fmt --check`
reports the same per-file hunk counts as `main` (so no formatting churn is mixed in), the
branch is rebased on current `main`, and all 13 CI checks are green.

ilysenko:
Thanks for the careful updater work here, and sorry this sat without maintainer follow-up. I'm closing this older inactive PR as part of a queue cleanup. Because this changes candidate promotion and package-manager recovery, any continuation should start from the latest main and include fresh coverage for exact artifact identity and failed or incomplete package states. A new focused PR is welcome.

PR #1368, opened 2026-08-17
Fix pacman updater package discovery

amandeepsp:
## Summary

- require rebuilt package candidates to be regular files
- ignore pacman's `codex-desktop-latest.pkg.tar.zst` symlink during artifact discovery
- add a regression test covering the versioned archive plus latest-symlink layout

The pacman builder intentionally writes both a versioned package and a stable
`latest` symlink. The updater previously matched entries by filename alone, so
it counted both paths and left Arch updates in `Failed` with
`expected one rebuilt package, found 2`.

This keeps the existing exactly-one-artifact invariant while excluding aliases,
directories, and other non-regular entries from the candidate set.

## Validation

- `cargo test -p codex-update-manager` (50 passed)
- `git diff --check`
- end-to-end Arch update from the signed `26.810.52044` package: rebuild,
  privileged pacman install, state transition to `Installed`, and subsequent
  check settling to `Idle`

`cargo fmt --all -- --check` was also inspected, but current unmodified `main`
already reports repository-wide formatting differences across updater files;
this PR does not broaden the diff to reformat them.

## Checklist

- [x] Upstream drift checklist is not applicable; this fixes local pacman updater artifact discovery.
- [x] I added relevant regression coverage and ran the validation listed above.

### DIFF
--- a/updater/src/builder.rs
+++ b/updater/src/builder.rs
@@ -170,7 +170,11 @@ async fn read_stream<R: tokio::io::AsyncRead + Unpin>(stream: R) -> Result<Strin
 fn find_package(dist: &Path) -> Result<PathBuf> {
     let mut matches = Vec::new();
     for entry in fs::read_dir(dist).with_context(|| format!("Failed to read {}", dist.display()))? {
-        let path = entry?.path();
+        let entry = entry?;
+        if !entry.file_type()?.is_file() {
+            continue;
+        }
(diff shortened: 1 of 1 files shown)


ilysenko:

eliliam:
@amandeepsp can we get this re-opened and merged? I'm hitting the same issue.

PR #1370, opened 2026-08-17
feat: show usage limits in Linux tray menu

ccueto36:
## Summary

- Add an opt-in `tray-usage` Linux feature that shows the existing usage-limit labels in the system-tray context menu.
- Reuse the usage data already delivered by the app's renderer; no new requests, credentials, background service, or updater behavior are added.
- Document the feature and add drift-safe patch tests.

## Problem

The renderer already sends `usageLimits` with the tray window state, but the main-process tray menu only renders those labels for macOS. Linux therefore receives the data but silently omits it.

This feature enables the same existing menu path on Linux while preserving the current macOS behavior and Windows exclusion. Left-click still opens the app; the usage labels are available in the Linux tray context menu.

## Scope

- Disabled by default; clean builds are unchanged unless `tray-usage` is explicitly enabled.
- No changes to the updater or tray icon.
- No new API surface or account/session handling.
- The patch is fail-soft when the upstream minified bundle drifts, with duplicate/mixed-match coverage.

## Validation

- `node --test linux-features/tray-usage/test.js` — 4 passed.
- `node --test scripts/lib/linux-features.test.js scripts/patch-linux-window-ui.test.js linux-features/*/test.js` — 757 tests, 756 passed, 1 skipped.
- `bash tests/scripts_smoke.sh` — passed.
- `cargo test -p codex-update-manager` — 49 passed.
- `cargo clippy -p codex-update-manager --all-targets -- -D warnings` — passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,7 @@ requirements, known limitations, configuration, and tests.
 | `shallow-repository-watches` | Avoid recursive main-thread walks for transient repository previews | [Docs](linux-features/shallow-repository-watches/README.md) |
 | `shared-app-server-socket` | Share one protocol-transparent Unix app-server socket | [Docs](linux-features/shared-app-server-socket/README.md) |
 | `thorium-chrome-plugin` | Add Thorium to the official bundled Chrome integration | [Docs](linux-features/thorium-chrome-plugin/README.md) |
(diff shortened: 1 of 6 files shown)


ilysenko (approved):
Reviewed the current head. The tray usage patch now uses a semantic, fail-closed ASAR anchor, the focused regression test covers the prior false-positive case, and all required checks are green.

PR #1372, opened 2026-08-18
fix(bootstrap): handle existing Rust installations on pacman systems

Axiaobo7788:

## Problem

`install-deps.sh` unconditionally installs `rustup` on pacman systems.
On Arch-derived distributions, `rustup` conflicts with the distro-provided
`rust` package, causing `make bootstrap-native` to abort when Rust is already
installed.

The existing `install_rust()` check also only verifies that the `cargo`
command exists. A rustup installation without an installed toolchain may
provide the command while Cargo itself is not usable.

## Solution

- Preserve an existing working distro-provided Rust/Cargo toolchain.
- Install `rustup` on pacman systems only when neither working Cargo nor
  rustup is available.
- Check `cargo --version` instead of only checking command presence.
- Initialize the stable toolchain when rustup exists but Cargo is not usable.

## Validation

Tested on Manjaro x86_64 with the distro `rust` package installed.

Before:
`make bootstrap-native` failed because pacman tried to install conflicting
`rustup`.

After:
`scripts/install-deps.sh` preserves the existing Rust toolchain and completes
successfully.

Also validated with:

    bash -n scripts/install-deps.sh
    git diff --check

## Checklist

### DIFF
--- a/.github/workflows/install-deps.yml
+++ b/.github/workflows/install-deps.yml
@@ -5,6 +5,7 @@ on:
     paths:
       - install.sh
       - scripts/install-deps.sh
+      - tests/install_deps_pacman_rust_matrix.sh
       - packaging/linux/control
       - .github/workflows/install-deps.yml
   push:
@@ -13,6 +14,7 @@ on:
     paths:
       - install.sh
       - scripts/install-deps.sh
+      - tests/install_deps_pacman_rust_matrix.sh
       - packaging/linux/control
       - .github/workflows/install-deps.yml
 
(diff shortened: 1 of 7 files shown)


ilysenko (changes requested):
Thanks for fixing the pacman/Rust bootstrap conflict. I found two blockers:

- `install-deps.sh` validates Cargo using the current `PATH`, but the bootstrap consumers prepend `$HOME/.cargo/bin` before the build. A working distro Cargo can therefore pass preflight while a rustup proxy without a toolchain shadows it during `install-native`. Please validate/use the same Cargo that the build will resolve, or stop the stale user-local proxy from shadowing the selected distro toolchain, and cover this mixed state.
- Please add automated regression coverage for the pacman/Rust state matrix: working distro Cargo, rustup without a toolchain, neither installed, and system Cargo shadowed by a user-local rustup proxy. The current install-deps CI only covers apt systems.

Please review the complete updated base-to-head diff with your own code-review model, fix every blocker it finds, rerun the review, and repeat until the model reports no remaining blockers before requesting maintainer re-review.

ilysenko (changes requested):
Thanks for addressing the original PATH-resolution and regression-coverage blockers; both now look fixed.

One blocker remains: `tests/install_deps_pacman_rust_matrix.sh` downloads the mutable `https://sh.rustup.rs` installer without a pinned version or verified hash and executes it as root after checkout. This allows unverified external code to modify the workspace and the regression result. Please build the user-local proxy fixture from a provenance-controlled source—for example, the signed Arch `rustup` package copied into `$HOME/.cargo/bin` before switching back to distro `rust`, or a versioned `rustup-init` artifact with a verified SHA-256.

Please review the complete updated base-to-head diff with your own code-review model, fix every blocker it finds, rerun the review, and repeat until the model reports no remaining blockers before requesting maintainer re-review.

Axiaobo7788:
Thanks for the follow-up. I’ve addressed the remaining blocker.

The shadowed proxy fixture now uses the `rustup` binary from Arch’s signed package instead of downloading and executing `sh.rustup.rs`. The pacman/Rust state matrix and the other relevant checks are passing.

I also reviewed the complete base-to-head diff again and found no remaining blockers.

The only red check left is the Official Linux workflow. Its Ubuntu runners got stuck during `apt-get update` and were cancelled after the 45-minute timeout, so the aggregate gate failed as expected. I don’t have permission to re-run workflows in the upstream repository. please re-run all jobs when convenient?

This should now be ready for re-review. Thanks!

ilysenko:
Thanks for addressing the review feedback, and sorry this sat without maintainer follow-up after you requested a CI rerun. I'm closing this older inactive PR as part of a queue cleanup. The required Official Linux workflow remained red on this head, so any continuation should be a fresh PR from the latest main with a clean current CI run.

PR #1384, opened 2026-08-20
Fix AppShots external target selection

### LINKED ISSUE #1383: AppShots composer targets ChatGPT instead of the previous external window
## Summary

When the disabled-by-default `appshots` Linux feature is used from the composer attachment menu, opening the menu gives focus to ChatGPT Community. The feature initially derives its target from current focus, so it can offer ChatGPT itself or an unrelated application instead of a deliberate external selection.

Adding an explicit Hyprland picker exposes a second part of the same failure: if the selected application is on another workspace, ordinary screenshot CLIs still capture the currently visible ChatGPT workspace. Cropping that image with the selected window's inactive-workspace bounds produces a resized or cropped image of ChatGPT rather than the selected application.

## Environment

- Repository: current `main` at `bf286d0b7a4080a2e4cb32cce17fb1fd5d86704c`
- Latest signed stable OpenAI Linux package: `chatgpt 26.818.22352 amd64`
- Distribution: Omarchy 4.0.0 (Arch family)
- Architecture: `x86_64`
- Desktop/session: Hyprland 0.56.2, native Wayland
- Installation method: pacman package built from source
- Relevant optional feature: `appshots`

The issue is isolated to `appshots`; its implementation and regression coverage do not depend on other optional features.

## Reproduction steps

1. Build and install ChatGPT Community with `appshots` enabled.
2. Open external applications on other Hyprland workspaces.
3. Return to ChatGPT Community and open the composer `+` menu.
4. Select the AppShots attachment entry.
5. Choose an external application in the Hyprland picker.

## Actual behavior

Without an explicit picker, the target is whichever window owns focus when the backend query runs. With selection added but without a workspace handoff, the selected metadata and bounds are combined with a screenshot of the still-visible ChatGPT workspace. Different external selections therefore attach ChatGPT itself, sometimes cropped or resized.

## Expected behavior

The composer workflow should let the user explicitly select an external window when the desktop provides a trustworthy selection mechanism. The selected window must be capture-visible with current verified geometry before the screenshot is taken. ChatGPT should be restored afterward, including on failure. The flow must skip ChatGPT and desktop portal windows and fail closed when no safe target can be selected or activated.

On Hyprland, `xdg-desktop-portal-hyprland` supplies the window-selection protocol used by `hyprland-preview-share-picker` and `hyprland-share-picker`. The picker itself is a Wayland layer-shell overlay, not a normal client controlled by ordinary floating rules. On EWMH-compatible X11, the existing topmost-external-window fallback can remain available until a desktop-neutral interactive picker is implemented.

## Scope

This concerns the repository-owned, disabled-by-default `linux-features/appshots` implementation, not upstream ChatGPT account or server behavior. Explicit selection is desktop/session dependent; making the selected target visible before a conventional screenshot is a separate correctness requirement.

I searched open and closed issues for AppShots focus, external-window, composer, selector, self-capture, and workspace capture reports and did not find a duplicate. The existing AppShots work in #329 established the feature but does not cover this composer focus and cross-workspace capture handoff.

## Diagnostics and privacy

No private conversation data, credentials, upstream payloads, or copyrighted bundles are attached. The proposed fix uses synthetic protocol and lifecycle regression coverage plus isolated builds against the latest signed stable package.

brianmsm:
mise ~/.config/mise/config.toml tools: gh@2.98.0
## Summary

Fixes #1383.

Opening the composer attachment menu focuses ChatGPT Community before AppShots queries the Linux window backend. An explicit picker is required, but selection alone is insufficient on Hyprland: ordinary screenshot CLIs capture the visible workspace, so cropping that image with bounds from another workspace can attach the wrong pixels.

This disabled-by-default `appshots` feature now:

- opens the installed Hyprland share picker only after the user activates the generic window entry;
- uses a temporary Windows-page config only for `hyprland-preview-share-picker`;
- leaves legacy and explicitly configured custom pickers argument-free;
- forces native GTK and Qt Wayland backends for the picker child process;
- resolves a live Hyprland instance from `$XDG_RUNTIME_DIR/hypr` when Electron did not inherit `HYPRLAND_INSTANCE_SIGNATURE`;
- requires a valid, visible ChatGPT return target before changing workspaces;
- activates the selected window and waits for stable identity, focus, visibility, and bounds;
- adds a render-settle interval and repeats the stable-state check before capture;
- revalidates identity, focus, visibility, and bounds after pixel acquisition;
- captures pixels and accessibility data concurrently;
- verifies ChatGPT focus restoration before delivering any renderer message;
- serializes capture transactions so requests cannot race compositor focus;
- keeps the EWMH stacking fallback for X11; and
- fails closed when any target, capture, or restoration invariant cannot be proven.

It prefers `hyprland-preview-share-picker`, falls back to `hyprland-share-picker`, and supports `CODEX_LINUX_APPSHOT_PICKER` as an explicit executable override.

### User-visible behavior

On Hyprland, the composer offers a generic `Attach window...` entry. Activating it opens the preview selector directly on Windows even when its persistent config defaults to Outputs or Region. Selecting a window on another workspace briefly displays that workspace while AppShots captures it, then restores ChatGPT Community before the attachment is delivered.

The preview selector is a centered native Wayland layer-shell overlay, not a normal tiled client. Output and Region selections remain intentionally fail-closed because the current AppShots contract accepts only a Window target.

### Scope and compatibility

- Hyprland native Wayland: explicit picker and verified focus/capture/restore transaction.
- EWMH-compatible X11 window managers: stacking-order fallback with synthetic coverage.
- Other Wayland compositors: unchanged safe failure; no compositor-specific target is guessed.
- Architectures and package formats: no architecture-specific or format-specific feature payload.

## Validation

Passed on the rebased branch:

```text
node --test linux-features/appshots/test.js
# 24 passed

node --test scripts/patch-linux-window-ui.test.js \
  scripts/lib/linux-features.test.js \
  linux-features/*/test.js
# 790 passed, 1 skipped

bash -n install.sh scripts/lib/*.sh launcher/start.sh.template
bash tests/scripts_smoke.sh
# 57 passed

./scripts/ci-local.sh pr
# deb, RPM, and pacman package builds passed

git diff --check
```

An isolated official-bundle build with only `appshots` enabled applied exactly its three descriptors against official `chatgpt 26.825.32147 amd64` (SHA-256 `986d38b690dd0310933ce61175b09c27434001f4e114332bb0f7b6ffdc3ca406`).

The updated regression coverage includes binary-aware picker arguments, temporary config cleanup, missing inherited Hyprland instance resolution, invalid return targets, target identity/focus/visibility/bounds drift, pre- and post-capture validation, verified restoration before delivery, restoration failure, concurrent pixel/accessibility acquisition, and serialized capture requests.

### Interactive Hyprland validation

Previously confirmed on Omarchy / Hyprland / native Wayland:

- the preview selector opens once as a centered native layer-shell overlay;
- windows from multiple apps and workspaces appear;
- explicit external selections attach that app instead of ChatGPT;
- ChatGPT regains focus before the thumbnail appears;
- successful Window selection does not show a failure toast;
- Output and Region fail closed.

Interactive X11 remains untested; synthetic coverage protects the EWMH fallback.

## Review status

The branch is rebased onto current `main`. I reviewed the complete updated base-to-head diff after the changes above and found no remaining blocker. Local validation and all required GitHub checks are green. The complete updated base-to-head diff was reviewed again before requesting maintainer re-review.

## Checklist

- [x] This pull request is ready for review.
- [x] I followed `CONTRIBUTING.md` and kept the change inside the optional feature.
- [x] Drift-prone bundle anchors fail closed and are covered by tests.
- [x] Relevant local tests and cross-format local CI pass.
- [x] I reviewed the final base-to-head diff.
- [ ] Newly triggered required GitHub Actions checks pass.


ilysenko (changes requested):
Thanks for completing the native Wayland picker work and interactive Hyprland validation. Before this can merge, please address the remaining capture-safety issues:

- Establish that the selected workspace is actually rendered before capturing, and revalidate the selected window’s identity, focus, visibility, and bounds before accepting the acquired pixels.
- Serialize capture transactions or add explicit ownership and cancellation so concurrent requests cannot race global compositor focus and attach pixels from another target.
- Reuse the repository’s Hyprland session/instance resolution for picker execution and focus/restore when the Electron environment lacks `HYPRLAND_INSTANCE_SIGNATURE`.
- Require a valid ChatGPT return target before activating another workspace, and fail closed unless restoration is successfully verified before delivering the attachment.
- Make picker invocation binary-aware. Preview-only config arguments must not be passed to legacy or custom pickers, and the documented Windows-first behavior must match each supported picker.

Please add regression coverage for these failure and concurrency paths.

Verify the updated PR against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant tests locally before pushing the update, and wait for every required CI check to pass. Run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, rerun the review, and repeat until the model reports no remaining blockers before requesting maintainer re-review.

brianmsm:
Thanks for the detailed review. I addressed the requested capture-safety changes in `eccfb7c`:

- selected-window identity, focus, visibility, and bounds are stabilized before capture and revalidated after pixel acquisition;
- capture transactions are serialized;
- picker and focus/restore paths resolve a live Hyprland instance when the Electron environment lacks the signature;
- capture now requires a valid visible ChatGPT return target and verified restoration before renderer delivery; and
- picker invocation is binary-aware, so preview-only config arguments are not passed to legacy or custom pickers.

Regression coverage now includes the failure, drift, restoration, and concurrency paths. The branch is rebased, the complete base-to-head diff was reviewed again with no remaining blockers, local validation passed, and all required CI checks are green. Re-requesting review.

ilysenko (approved):
Reviewed the exact current head. The requested transaction, instance-resolution, restoration, and picker fixes are present, and all required checks pass. The remaining compositor-specific capture edge cases are accepted as follow-up work for this disabled-by-default feature.

PR #1385, opened 2026-08-20
fix(browser): preserve proxy env for Browser Use helpers

redfive-cn:
**IMPORTANT: This is my only open pull request to this repository.**

## Summary

The latest signed Linux package (26.818.22352) can start Browser Use `node_repl` helpers with a filtered environment. On systems where the direct route cannot reach ChatGPT but the Codex app-server has explicit proxy variables, Chrome can still connect, list, and claim tabs while navigation and DOM commands hang during the Browser Use site-status security check.

This adds a disabled-by-default `browser-proxy` Linux feature that:

- wraps the bundled helper and recovers only missing standard proxy variables from its immediate app-server parent;
- preserves explicit child values and enables Node environment-proxy support only when a non-empty proxy exists;
- retains the official helper byte-for-byte as `node_repl.codex-linux-original`, with ownership checks, atomic replacement, rollback, and cleanup;
- documents configuration, scope, and credential exposure in English and Chinese.

The feature does not patch `app.asar`, log proxy values, alter Chrome page routing, set `BROWSER_USE_SECURITY_MODE`, or bypass site-status, URL-policy, or consent checks.

The cross-platform connect/list-but-DOM-timeout symptom also appears in [openai/codex#36278](https://github.com/openai/codex/issues/36278).

## Validation

- `node --test linux-features/browser-proxy/test.js`: 9 passed.
- `node --test scripts/lib/linux-features.test.js linux-features/browser-proxy/test.js`: 39 passed.
- `bash -n install.sh scripts/lib/*.sh launcher/start.sh.template`
- `node --test scripts/patch-linux-window-ui.test.js scripts/lib/linux-features.test.js linux-features/*/test.js`: 751 passed, 1 skipped, 0 failed.
- Feature-only build from signed `chatgpt 26.818.22352` on Ubuntu 24.04 / GNOME X11 / amd64; official `app.asar` preserved byte-for-byte.
- Installed and inspected a Debian package built with `PACKAGE_WITH_UPDATER=0`.
- Real Chrome extension validation:
  - Baidu navigation completed in 1.48 s.
  - visible DOM and full DOM snapshot completed in 51 ms and 68 ms.
  - five consecutive DOM + fill/read/clear rounds completed in 212–250 ms each with no timeout.
  - running helpers contained the expected proxy variable names; values and credentials were not printed.

Not tested: arm64, RPM, pacman, AppImage, or Nix runtime packaging. The stage hook is format-independent, but only the Debian payload was built and installed manually.

Security/rollback notes: proxy URLs may contain credentials, so the feature is opt-in and exposes them only to the same-user helper whose immediate parent already has them. Unreadable parent environment leaves the helper environment unchanged. Tests cover idempotency, explicit-child precedence, no-proxy behavior, failed-stage rollback, cleanup restoration, and foreign-wrapper collision preservation.

## Checklist

- [ ] I added or updated relevant tests, ran the validation listed above, and confirmed that required CI checks pass. (Relevant local validation passed; GitHub CI is pending.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,7 @@ requirements, known limitations, configuration, and tests.
 | `appshots` | Capture and crop the focused Linux window from the composer | [Docs](linux-features/appshots/README.md) |
 | `authenticated-proxy` | Username/password support for HTTP proxies | [Docs](linux-features/authenticated-proxy/README.md) |
 | `automation-extensions` | Multi-time schedules and eager `automation_update` exposure | [Docs](linux-features/automation-extensions/README.md) |
+| `browser-proxy` | Pass explicit proxy settings to Browser Use network helpers | [Docs](linux-features/browser-proxy/README.md) |
 | `chronicle-skysight` | Opt-in Linux desktop activity memory and restricted Skysight MCP tools | [Docs](linux-features/chronicle-skysight/README.md) |
 | `codex-micro` | Work Louder Codex Micro hotplug and hidraw policy using upstream `node-hid` | [Docs](linux-features/codex-micro/README.md) |
 | `computer-use-linux` | Linux desktop-control UI and native MCP backend | [Docs](linux-features/computer-use-linux/README.md) |
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -203,6 +203,7 @@ Nix 用户应从 profile、Home Manager 配置或 NixOS module 中删除该包�
 | `appshots` | 从 composer 捕获并裁剪当前 Linux 窗口 | [文档](linux-features/appshots/README.md) |
 | `authenticated-proxy` | 带用户名和密码的 HTTP proxy | [文档](linux-features/authenticated-proxy/README.md) |
 | `automation-extensions` | 多时间调度和 eager `automation_update` | [文档](linux-features/automation-extensions/README.md) |
+| `browser-proxy` | 让 Browser Use 的网络辅助进程继承显式代理设置 | [文档](linux-features/browser-proxy/README.zh-CN.md) |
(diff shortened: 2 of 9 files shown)


ilysenko (changes requested):
Thanks for the work on this feature. One blocker needs to be addressed before merge:

- Treat upper- and lower-case proxy variables as one proxy family. The wrapper currently checks each variable name independently. If the child has `HTTP_PROXY` and the parent has `http_proxy`, the parent value is imported and takes precedence in Node, overriding the child’s explicit route. The reverse combination also exposes an unnecessary credential-bearing variable. Apply family-level precedence to HTTP, HTTPS, ALL, and NO_PROXY, and add cross-case regression tests.

Before requesting re-review, verify the update against [CONTRIBUTING.md](https://github.com/ilysenko/codex-desktop-linux/blob/main/CONTRIBUTING.md), run the relevant tests locally, and wait for every required CI check to pass. Run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, and repeat until the model reports no remaining blockers.

redfive-cn:
Thanks for the review. I’ve addressed the proxy-family precedence blocker in 65325e9.

HTTP_PROXY/http_proxy, HTTPS_PROXY/https_proxy, ALL_PROXY/all_proxy, and NO_PROXY/no_proxy are now each treated as one family. If the child already has either spelling, the wrapper imports neither spelling from the parent. I also added cross-case regression coverage for all four families in both directions, including assertions that the child value is preserved and credential-bearing parent values are not exposed.

I reviewed the complete updated base-to-head diff with my code-review model and found no remaining blockers. The relevant local tests pass, and all required CI checks are green.

This should now be ready for re-review. Thanks!

ilysenko (approved):
The previously requested cross-case proxy-family fix is present on the current head, the focused regression coverage covers all proxy families in both directions, and all exact-head checks are green. Approved.

PR #1393, opened 2026-08-22
Fix remote-control provider alias capture

stijnhoste:
## Problem

Outbound remote-control authorization can fail after the enrollment-start response with:

```text
n.isAbsolute is not a function
```

The device-key patch searched the entire minified main bundle for existing Node module aliases. On the current official `26.818.41705` bundle, that search can capture a function-local `node:path` alias and then reference it from the separately injected provider, where it resolves to an unrelated value.

This is a follow-up to #919: the dedicated private key directory remains unchanged, but the provider no longer relies on drift-prone aliases from the upstream bundle.

## Solution

- import `node:path`, `node:fs`, `node:crypto`, and `node:child_process` through provider-scoped, feature-owned names
- remove the whole-bundle alias scanner
- add a regression fixture with an earlier function-local `node:path` alias
- keep the existing child-process and crypto collision coverage aligned with the owned module names

## User-visible behavior

Clicking **Authorize on chatgpt.com** no longer crashes locally while preparing the Linux device key. This affects the disabled-by-default `remote-mobile-control` feature across package formats and architectures.

## Validation

- `node --test linux-features/remote-mobile-control/test.js` — 101 passed
- `node --test scripts/patch-linux-window-ui.test.js scripts/lib/linux-features.test.js linux-features/remote-mobile-control/test.js` — 136 passed
- `bash tests/scripts_smoke.sh` — 51 passed
- feature-only inspection against official Linux package `26.818.41705` — 15 descriptors applied, 3 already applied, no failures
- `git diff --check`

Runtime reproduction and patched-package launch were verified on Arch Linux x86_64 with Hyprland. Other architectures and package formats were not manually exercised; the changed code is format-independent injected JavaScript.

### DIFF
--- a/linux-features/remote-mobile-control/patch.js
+++ b/linux-features/remote-mobile-control/patch.js
@@ -3,12 +3,6 @@
 const fs = require("node:fs");
 const path = require("node:path");
 
-function requireName(source, moduleName) {
-  const escaped = moduleName.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
-  const match = source.match(new RegExp(`([A-Za-z_$][\\w$]*)=require\\("${escaped}"\\)`));
-  return match?.[1] ?? null;
-}
-
 const DEVICE_KEY_CLIENT_MARKER = "codexLinuxRemoteControlDeviceKeyClient";
 const DEVICE_KEY_GUARD =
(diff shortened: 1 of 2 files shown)


ilysenko (approved):
Reviewed the complete base-to-head diff at 39647b16aae722518e31f51f83b8011243e9934d. The focused provider-owned module imports fix the alias-capture regression, and all exact-head checks pass.

PR #1407, opened 2026-08-28
Fix symlinked CODEX_HOME launcher identity

shishi:
## Summary

- Canonicalize an existing `CODEX_HOME` (or the default `$HOME/.codex`) after environment hooks and before the official ChatGPT/Codex consumers start.
- Ignore inherited `CDPATH` while resolving relative paths, and handle the `-` and `/` boundary cases explicitly.
- Keep missing paths unchanged and preserve Codex's reserved-marketplace validation.
- Make launcher tests portable to NixOS and cover default, explicit, environment-hook, missing-path, `CDPATH`, `-`, and `/` cases.

On NixOS, `~/.codex` may be a symlink into a dotfiles checkout. The desktop app staged `openai-bundled` under the logical path while the bundled CLI saw the physical path. Codex then correctly rejected the reserved marketplace because the lexical paths did not match. Giving both processes the same physical `CODEX_HOME` fixes the Brave/Chrome side-panel connection without weakening marketplace trust checks.

User-visible behavior: the official ChatGPT browser extension can connect to ChatGPT Community when `~/.codex` or an explicit `CODEX_HOME` is a symlink.

Scope: the shared launcher, so this applies to deb, RPM, pacman, AppImage, and Nix packages on amd64 and arm64. There are no ASAR or dependency changes. I found no exact existing issue or pull request; #1266/#1267 concern retired plugin-cache paths rather than logical-versus-physical `CODEX_HOME` identity.

## Validation

- Red: the original symlink regression test failed with an empty `CODEX_HOME` instead of the physical path.
- Red: the maintainer-requested relative-path test selected the `CDPATH` alternate and contaminated command-substitution output.
- Red: complete-diff review found that exact `CODEX_HOME=-` resolved `OLDPWD`; its regression test reproduced that behavior.
- Red: the next complete-diff review found that the first `-` fix changed `CODEX_HOME=/` to `//`; its regression test reproduced that behavior.
- `node --test launcher/start.test.js` — 13/13 passed.
- `bash -n launcher/start.sh.template` — passed.
- `node --check launcher/start.test.js` — passed.
- `git diff upstream/main...HEAD --check` — passed.
- Complete updated `upstream/main...HEAD` code-review loop: 1 blocker, then 1 blocker, then 0 blockers after both fixes.
- GitHub Actions — all 13 checks passed on `8b52592`.

Not manually tested: a live installed deb/RPM/pacman package. Security risk is limited: only existing directories are canonicalized; missing paths and reserved-marketplace checks remain unchanged.

## Checklist

- [x] This is not an upstream-drift fix; it targets launcher path identity and adds no fallback code.
- [x] I reviewed the complete updated base-to-head diff until the review reported no remaining blockers.

### DIFF
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -102,6 +102,15 @@ load_environment_hooks() {
     done < <(find "$HOOK_ROOT/env.d" -maxdepth 1 -type f -print0 | sort -z)
 }
 
+canonicalize_codex_home() {
+    local requested_codex_home="${CODEX_HOME:-$HOME/.codex}"
+    local physical_codex_home
+
+    [ -d "$requested_codex_home" ] || return 0
+    physical_codex_home="$(cd -P -- "$requested_codex_home" && pwd -P)"
+    export CODEX_HOME="$physical_codex_home"
+}
+
 refresh_legacy_bundled_plugin_cache() {
     local plugin_id="$1"
     local bundled_plugin="$CODEX_LINUX_APP_DIR/resources/plugins/openai-bundled/plugins/$plugin_id"
@@ -228,6 +237,7 @@ esac
 
 ELECTRON_ARGS=("--class=$CODEX_LINUX_APP_ID")
 load_environment_hooks
+canonicalize_codex_home
 refresh_legacy_bundled_plugin_caches
 load_user_electron_args
 load_electron_args
--- a/launcher/start.sh.template
+++ b/launcher/start.sh.template
@@ -107,7 +107,10 @@ canonicalize_codex_home() {
     local physical_codex_home
 
     [ -d "$requested_codex_home" ] || return 0
-    physical_codex_home="$(cd -P -- "$requested_codex_home" && pwd -P)"
+    if [ "$requested_codex_home" = "-" ]; then
+        requested_codex_home="./-"
+    fi
+    physical_codex_home="$(CDPATH='' cd -P -- "$requested_codex_home" && pwd -P)"
     export CODEX_HOME="$physical_codex_home"
 }
 
--- a/launcher/start.test.js
+++ b/launcher/start.test.js
@@ -253,6 +253,69 @@ test("launcher exports a physical explicit CODEX_HOME when it is a symlink", (t)
   assert.equal(fs.readFileSync(path.join(root, "codex-home"), "utf8").trim(), physicalCodexHome);
 });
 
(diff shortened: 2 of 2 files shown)


ilysenko (changes requested):
Thanks for the focused fix and regression tests.

Please fix `canonicalize_codex_home()` in `launcher/start.sh.template:110` so it ignores inherited `CDPATH`. With an existing relative `CODEX_HOME=profile` and exported `CDPATH=.`, `cd -P` writes its destination to stdout, so the command substitution exports `./profile\n/absolute/path/profile` instead of one valid path. A different CDPATH entry can also select another directory. The base launcher preserves the valid relative path; upstream accepts and canonicalizes it. Clear CDPATH inside the subshell and add a regression test covering a relative CODEX_HOME with a nonempty CDPATH.

Before requesting maintainer re-review, run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, rerun the review, and repeat until it reports no remaining blockers.

shishi:
@ilysenko, I addressed the requested changes in `8b52592`.

- Clear inherited `CDPATH` while resolving `CODEX_HOME`, with a regression test that proves a relative path neither selects a `CDPATH` alternate nor contaminates command-substitution output.
- Treat exact `CODEX_HOME=-` as the literal `./-` directory instead of `OLDPWD`.
- Preserve root `CODEX_HOME=/` as `/` rather than `//`.
- Keep missing paths unchanged and retain the existing reserved-marketplace checks.

Validation:

- `node --test launcher/start.test.js` — 13/13 passed.
- `bash -n launcher/start.sh.template` — passed.
- `node --check launcher/start.test.js` — passed.
- `git diff upstream/main...HEAD --check` — passed.
- GitHub Actions — all 13 checks passed on `8b52592`.

I ran the complete updated `upstream/main...HEAD` diff through independent code review after each fix:

1. Pass 1: 1 blocker (`CODEX_HOME=-`). Fixed and added a regression test.
2. Pass 2: 1 blocker (`CODEX_HOME=/` became `//`). Fixed and added a regression test.
3. Pass 3: 0 blockers and no non-blocking code findings.

Ready for maintainer re-review.

ilysenko (approved):
Reviewed the complete updated diff at 8b525929. The prior CDPATH blocker is fixed and covered by a regression test; the literal dash and root cases are also covered. No remaining blockers found, and all current-head required checks pass.

PR #1410, opened 2026-08-29
fix(updater): prune stale build workspaces

Fr4nzz:
## Problem

The update manager's cache cleanup only scans `packages/` for obsolete `.deb` files. It never scans `workspaces/`, so each completed or failed rebuild can leave a full build workspace behind indefinitely. On an actively updated installation this accumulated 7.6 GB of unreferenced build data.

## Solution

Extend cache pruning to remove unreferenced directories under `workspaces/` while preserving:

- the workspace explicitly referenced by persisted state;
- any workspace containing the current candidate package;
- any workspace containing the rollback package.

Directory entries are checked with `DirEntry::file_type`, so symlinked directories are not traversed or removed as workspaces.

## User-visible behavior

Routine updater cleanup now reclaims stale build workspaces instead of allowing them to accumulate across updates. Current candidate and rollback artifacts remain available.

## Scope

This changes the Rust update manager and applies across package formats and architectures that use it. It does not change the signed upstream package, package building, promotion, or rollback behavior.

## Validation

- `cargo test -p codex-update-manager` (51 passed)
- `git diff --check`

The regression test verifies that cleanup removes a stale package and stale workspace, retains the current package and workspace, retains the rollback workspace, and does not follow a workspace symlink.

### DIFF
--- a/updater/src/cache_cleanup.rs
+++ b/updater/src/cache_cleanup.rs
@@ -10,20 +10,109 @@ pub fn prune(cache_root: &Path, state: &PersistedState) -> Result<usize> {
         state.artifact_paths.upstream_package_path.as_ref(),
         state.artifact_paths.package_path.as_ref(),
         state.artifact_paths.rollback_package_path.as_ref(),
-    ].into_iter().flatten() {
-        if let Ok(path) = path.canonicalize() { retained.insert(path); }
+    ]
+    .into_iter()
+    .flatten()
+    {
+        if let Ok(path) = path.canonicalize() {
+            retained.insert(path);
+        }
     }
     let mut removed = 0;
     let package_dir = cache_root.join("packages");
-    if !package_dir.is_dir() { return Ok(0); }
+    if !package_dir.is_dir() {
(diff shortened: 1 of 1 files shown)


ilysenko (changes requested):
Thanks for restoring stale updater workspace pruning. I found three blockers:

- Please serialize workspace pruning with every updater entrypoint that can build, install, or roll back, and reload authoritative persisted state inside that boundary. Otherwise stale in-memory state can recursively delete another process's active workspace. Add a deterministic interleaving regression test.
- Please prune the configured `workspace_root/workspaces`, not only `cache_dir/workspaces`, and cover distinct package-cache and workspace roots in the test.
- Please reject or safely skip a symlinked `workspaces` root. The current `is_dir`/`read_dir` flow follows it and can recursively delete directories outside the updater-managed cache. Add a root-symlink regression test.

Finally, please run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, rerun the review, and repeat until that model reports no remaining blockers before requesting maintainer re-review.

PR #1413, opened 2026-08-31
fix(computer-use): use terminal paste shortcuts on KDE Plasma

thisisryanswift:
Transparently, agent/ai generated, but real issue I was having. Feel free to reject as needed!

## Summary

- port the KDE terminal paste fix from agent-sh/computer-use-linux#113
- preserve the embedded backend's cancellation-safe input and clipboard guards
- re-verify focus after portal setup before selecting and sending the paste chord
- add terminal identity, stale-selector, and shortcut-routing regression tests

## Validation

- `cargo fmt -p codex-computer-use-linux -- --check`
- `cargo test --locked -p codex-computer-use-linux kde_clipboard_`
- `cargo test --locked -p codex-computer-use-linux terminal_paste`
- `cargo test --locked -p codex-computer-use-linux --lib` - 274 passed
- `cargo clippy --locked -p codex-computer-use-linux --lib -- -D warnings`
- independent code review: no findings


ilysenko (changes requested):
Thanks for improving KDE terminal paste and preserving the input guards. Please fix these two regressions before merge:

1. In computer-use-linux/src/server.rs:4843, preserve ordinary GUI-field paste inside terminal windows. Konsole's Find field is a QLineEdit, but window identity now selects Ctrl+Shift+V instead of its working Ctrl+V. Use the actual input focus when selecting the shortcut and add coverage for the terminal view versus Find in the same window.

2. In computer-use-linux/src/terminal.rs:239, preserve Ghostty PTY detection with a custom X11 class. The default instance remains ghostty, which the new exact identity list omits. With a non-brand shell title, the same one-window/one-PTY fixture gets metadata on base and none on head, breaking terminal selectors. Retain this identity and add an enrichment regression test.

Please run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, rerun the review, and repeat until the model reports no remaining blockers before requesting maintainer re-review.

ilysenko (approved):
Verified the requested regressions are fixed, the complete updated diff has passed repeated review with no remaining blockers, and all local and CI checks are green.

thisisryanswift:
Sorry for missing your feedback and comment here! Tend to do my GitHub notifications in a batch each week. Appreciate you cleaning this up for me.

PR #1414, opened 2026-08-31
fix: require Node.js 24 for native updater

carbon-teq:
## Summary

- require Node.js 24 or newer in deb, RPM, and pacman updater packages
- execute the packaged system runtime at /usr/bin/node so user version-manager shims cannot select an older Node.js
- keep the Debian update-manager service alive during self-upgrades so successful promotion state is recorded
- document the Node.js runtime requirement in English, Chinese, packaging, updater, and changelog documentation

## Validation

- cargo test -p codex-update-manager: 50 passed
- cargo clippy -p codex-update-manager --all-targets -- -D warnings
- node --test scripts/lib/package-common.test.js: 12 passed
- bash syntax and git diff checks
- built and installed codex-desktop 2026.08.31.172424 on Pop!_OS
- verified nodejs 24.20.0, active updater service, signed upstream metadata check, idle status, and no updater error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- Native updater packages now require Node.js 24 or newer, and the update
+  manager rejects obsolete runtimes before invoking the signed-package
+  verifier at `/usr/bin/node`, so user version-manager shims cannot select an
+  older runtime. Debian bootstrap installs the latest available Node.js 24 LTS
+  from the signed NodeSource repository when the distro candidate is too old.
+- Debian upgrades no longer stop the update-manager service from inside its own
+  privileged package transaction, allowing it to record successful promotion
+  instead of rebuilding the same upstream package on the next start.
 - The opt-in `frameless-titlebar` feature again hides official Linux overlay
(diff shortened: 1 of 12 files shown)


ilysenko (changes requested):
Thanks for improving native updater reliability. Two issues block this change:

1. Skipping the Debian service stop leaves the old executable running after replacement. `current_exe()` then returns a path ending in ` (deleted)`; `builder.rs` passes it as `UPDATER_BINARY_SOURCE`, and `ensure_updater_binary` rejects the next rebuild. I reproduced this with the real packaging helper. Please move the independent Debian lifecycle change, its test and documentation to a subsequent PR. That fix needs safe process refresh after state persistence and a regression test covering the next rebuild.
2. `Requires: nodejs >= 24` does not enforce the runtime floor on Fedora: its epoch-1 Node 22 package satisfies this dependency, while the updater rejects `/usr/bin/node` 22. Make the dependency and selected runtime consistent across supported RPM distributions, with coverage rejecting Node 22 and accepting supported Node 24 providers. Do not assume a single epoch across distributions.

Please run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, rerun the review, and repeat until it reports no remaining blockers before requesting maintainer re-review.

PR #1417, opened 2026-09-01
fix(bootstrap): handle existing Rust installations on pacman systems

Axiaobo7788:

## Problem

`install-deps.sh` unconditionally installs `rustup` on pacman systems.
On Arch-derived distributions, `rustup` conflicts with the distro-provided
`rust` package, causing `make bootstrap-native` to abort when Rust is already
installed.

The existing `install_rust()` check also only verifies that the `cargo`
command exists. A rustup installation without an installed toolchain may
provide the command while Cargo itself is not usable.

## Solution

- Preserve an existing working distro-provided Rust/Cargo toolchain.
- Install `rustup` on pacman systems only when neither working Cargo nor
  rustup is available.
- Check `cargo --version` instead of only checking command presence.
- Initialize the stable toolchain when rustup exists but Cargo is not usable.

## Validation

Tested on Manjaro x86_64 with the distro `rust` package installed.

Before:
`make bootstrap-native` failed because pacman tried to install conflicting
`rustup`.

After:
`scripts/install-deps.sh` preserves the existing Rust toolchain and completes
successfully.

Also validated with:

    bash -n scripts/install-deps.sh
    git diff --check

## Checklist


ilysenko (approved):
Reviewed the complete base-to-head diff. The pacman Rust bootstrap fix is focused, preserves existing working toolchains, and is covered by the live Rust-state matrix. No blockers found.

Issue #1038, opened 2026-07-16
Expose Dock icon and Suggested prompts as opt-in UI Tweaks


PinguuSS:
## Summary

I would like to make two useful capabilities already present in the official macOS bundle available to Linux users as independently opt-in, default-off UI Tweaks:

- **Dock icon selection**: choose between the official ChatGPT and Codex icons.
- **Suggested prompts**: expose the upstream setting and show generated project-aware suggestions on Home.

I have both capabilities working in a local Linux build against the current DMG and have the implementation, tests, staging/cleanup behavior, and runtime evidence ready. I am willing to own the ongoing maintenance of both tweaks as upstream DMGs change.

## Problem

The current bundle contains the UI, resources, storage contracts, and runtime paths for both capabilities, but Linux users cannot access them through the port:

- Dock icon selection is gated to the macOS path and Linux has no `app.dock` implementation.
- Suggested Prompts is present in Settings search and the bundle contains its generation/rendering path, but the row and Home experience remain unavailable behind platform, rollout, and account eligibility gates.

PR #784 correctly hid the settings search results while these rows were unavailable. This request proposes making the underlying capabilities genuinely available when a user explicitly enables the corresponding Linux tweak, then allowing the search results to appear for that enabled configuration.

### Dock icon selection

A user can keep the current ChatGPT identity or select the official Codex icon. The chosen icon is reflected by Electron windows, the system tray, and supported desktop launchers/taskbars, and persists after restart.

### Suggested prompts

A user opening a project sees generated suggestions based on project files and connected apps. Selecting a card fills the composer with a concrete next action, using the upstream generation, storage, and card-rendering contracts.

## Proposed behavior

Add two independent, default-off settings under the existing `linux-features/ui-tweaks` feature, for example:

```json
{
  "enabled": ["ui-tweaks"],
  "settings": {
    "ui-tweaks": {
      "tweaks": {
        "appearance": {
          "dockIcon": { "enabled": true }
        },
        "home": {
          "suggestedPrompts": { "enabled": true }
        }
      }
    }
  }
}
```

Each tweak would:

- remain disabled by default;
- target only the current DMG contract;
- use descriptor-level `enabled` callbacks;
- patch atomically, idempotently, and fail-soft on drift;
- leave the asset byte-identical when the complete current contract is absent;
- clean up only its own staged payload when disabled;
- participate in normal updater acceptance whenever the user enables it.

[...]

PR #1421, opened 2026-09-02
Retarget Suggested Prompts to split Linux renderer bundles

PinguuSS:
## Summary

- Retarget Suggested Prompts to the current official Linux 26.901.20858 renderer layout.
- Split the renderer work into semantic descriptors for the feature-sync bridge, normal Home, and Work Home.
- Preserve the existing main-process, General Settings, and generated-source patches.
- Keep the Dock Icon implementation out of this focused change; its current-DMG descriptors are already covered by #1420.

## Root cause

After #1420, the official Linux package moved the relevant renderer contracts out of the previous monolithic `app-initial` shape:

- `app-initial-d237ce07a86a.js`: renderer feature-sync bridge
- `app-primary-609f2338596a.js`: normal Home
- `page-b7f6adbc0daf.js`: Work Home

The previous app-page descriptor expected both rollout gates in `app-initial`, so an enabled `tweaks.home.suggestedPrompts` configuration skipped the active Home asset on the current package.

The package identity used for validation was `chatgpt 26.901.20858 amd64`, Electron 42.3.0, DEB SHA-256 `42a6477f22f4136d62321eda7b4697a79da1eb66d61dcb85ab0420860a1a5223`, 411817410 bytes, and upstream ASAR SHA-256 `6be4f6074b300590459f9c5e1e8ea22bd57db08557a1186c521f055302d99c8a`.

## Safety

- Each descriptor selects one exact current asset family and then requires its semantic contract.
- Gate and eligibility aliases must agree before a patch is applied or reported as already applied.
- Multi-point contracts are validated before any write.
- Complete marker bodies are required for the already-applied state.
- Missing, ambiguous, mixed, or drifted input remains byte-identical and emits a warning.
- No legacy bundle fallback or generated output is included.

## Validation

- Focused Suggested Prompts tests: 16/16 passed.
- Full ui-tweaks and framework-related suite: 108/108 passed.
- `bash -n install.sh scripts/lib/*.sh launcher/start.sh.template`.
- Isolated `ui-tweaks` build against the package above: all six Suggested Prompts descriptors applied.
- Built patch report second pass: all five renderer descriptors reported already-applied.
- `node --check` passed for the patched main bundle and all five affected renderer assets.
- Side-by-side smoke opened the isolated candidate to the login screen, completed app-server startup, and left no candidate process running.
- The installed daily driver and its feature configuration were not modified.

## Maintenance

As agreed in issue #1038 and the prior retarget PRs, I will continue monitoring and maintaining both the Dock Icon and Suggested Prompts integrations against future official Linux package drift. This PR is intentionally limited to Suggested Prompts.


ilysenko (approved):
Reviewed the complete diff at cda5eb2. No blockers found. Required checks are green; all six Suggested Prompts descriptors applied to the current official Linux ASAR, were idempotent, and passed syntax checks.

PR #1430, opened 2026-09-05
Fix updater self-termination during deb upgrades

kassi0-peia:
## Problem

During updater-initiated Debian upgrades, `prerm` stopped `codex-update-manager.service` while it owned the apt/dpkg transaction. That left stale `(deleted)` binaries and caused retry loops.

## Solution

- `prerm upgrade` now leaves the updater service running; `remove` and `deconfigure` cleanup is unchanged.
- Install/rollback journaling is durable and ownership-aware, including boot-ID identity, bounded abandoned recovery, schema-2 manual migration, and `Installed` save/readback before replacement restart.
- Debian, RPM, and pacman recovery require verified package evidence; RPM verification uses `rpm -V --noscripts`.
- Public recovery documentation is updated. Advisory lockfile refreshes are isolated in a separate commit.

## Scope

Updater/package lifecycle for signed official Linux package `26.908.40834` on amd64 and arm64: Debian, RPM, pacman, candidate promotion, and rollback. No core ASAR or enabled-feature behavior changed.

## Validation

- CI green: source/Node, Rust, signed baseline amd64/arm64, Dock-icon feature, package matrix amd64/arm64, Nix x86_64/aarch64, Nix VM, watchdog, and release gates.
- `cargo test -p codex-update-manager --locked -- --test-threads=1`: 106 passed; workspace tests and updater Clippy passed.
- `cargo audit --no-fetch --deny warnings`: root and optional lockfiles clean.
- Shell/Node checks passed: `scripts_smoke.sh` 75 passed; broad Node suite 916 passed, 1 intentional skip.
- Fedora amd64 RPM and Arch amd64 pacman build/install/verify/remove passed; Fedora arm64 smoke passed.

## Limitations

- Arch arm64 native package validation was unavailable because the pinned Arch image has no arm64 manifest.
- One host-Wayland fixture is environment-sensitive; neutral-X11 and CI pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -166,7 +166,7 @@ checksum = "3b43422f69d8ff38f95f1b2bb76517c91589a924d1559a0e935d7c8ce0274c11"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -201,7 +201,7 @@ checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -421,7 +421,7 @@ dependencies = [
  "heck",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -623,7 +623,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "strsim",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -634,7 +634,7 @@ checksum = "ac3984ec7bd6cfa798e62b4a642426a5be0e68f9401cfc2a01e3fa9ea2fcdb8d"
 dependencies = [
  "darling_core",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -740,7 +740,7 @@ checksum = "67c78a4d8fdf9953a5c9d458f9efe940fd97a0cab0941c075a813ac594733827"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -903,7 +903,7 @@ checksum = "e835b70203e41293343137df5c0664546da5745f82ec9b84d40be8336958447b"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -1442,7 +1442,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "479ca8adacdd7ce8f1fb39ce9ecccbfe93a3f1344b3d0d97f20bc0196208f62b"
 dependencies = [
  "proc-macro2",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -1471,21 +1471,11 @@ checksum = "e0c5ccf5294c6ccd63a74f1565028353830a9c2f5eb0c682c355c471726a6e3f"
 
 [[package]]
 name = "quick-xml"
-version = "0.37.5"
+version = "0.41.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "331e97a1af0bf59823e6eadffe373d7b27f485be8748f71471c662c1f269b7fb"
+checksum = "e660451e55124f798a69a5af3f49ccfbefbd41910eefd25caf2393e1f3473ec1"
 dependencies = [
  "memchr",
-]
-
-[[package]]
-name = "quick-xml"
-version = "0.39.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "958f21e8e7ceb5a1aa7fa87fab28e7c75976e0bfe7e23ff069e0a260f894067d"
-dependencies = [
- "memchr",
- "serde",
 ]
 
 [[package]]
@@ -1572,7 +1562,7 @@ checksum = "b7186006dcb21920990093f30e3dea63b7d6e977bf1256be20c3563a5db070da"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -1607,7 +1597,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "serde_json",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -1652,7 +1642,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "serde_derive_internals",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -1688,7 +1678,7 @@ checksum = "d540f220d3187173da220f885ab66608367b6574e925011a9353e4badda91d79"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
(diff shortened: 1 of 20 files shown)


ilysenko (changes requested):
Thanks for isolating the Debian updater lifecycle issue and documenting the reproduced failure. Two blockers remain:

1. Skipping service shutdown during `prerm upgrade` leaves the old daemon running after `/usr/bin/codex-update-manager` is replaced. The current updater has no replacement detection, `postinst` does not restart an active unit, and the next rebuild receives a `current_exe()` path ending in ` (deleted)`. Please preserve the in-flight package transaction, then safely exit/restart the daemon only after the install result and `Installed` state have been persisted.
2. The added smoke assertions only check for strings and still pass if the actual `exit 0` guard is removed. Please add behavioral coverage for the staged `DEBIAN/prerm`: upgrade must avoid service side effects, removal/deconfiguration must retain cleanup, and a successful self-upgrade must restart onto the new binary and allow the next rebuild.

Please run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, rerun the review, and repeat until that model reports no remaining blockers before requesting maintainer re-review.

kassi0-peia:
Addressed the updater self-termination and stale-binary lifecycle issues from the review.

The Debian `prerm` now treats `upgrade` differently from removal/deconfiguration: an upgrade leaves the running updater service alone so an updater-initiated `apt/dpkg` transaction cannot terminate its own parent service mid-install. `remove` and `deconfigure` retain the existing stop, disable, and daemon-reload cleanup.

The updater now detects executable replacement by device/inode rather than content. This handles the normal `/proc/self/exe -> ... (deleted)` case and also catches byte-identical package replacement. After a managed install succeeds, the updater persists `Installed` state and reads it back before exiting with the service restart code; systemd then relaunches the new binary. If persisted state is still `Installing`, replacement restart is blocked rather than exiting early.

The builder path is also hardened so a subsequent rebuild does not use a stale `current_exe()` path ending in `" (deleted)"`; it resolves the live replacement updater instead while preserving the healthy current executable for normal dev/test processes.

Behavioral coverage added for the requested cases:

- staged `DEBIAN/prerm upgrade` produces no user-service side effects;
- staged `remove` and `deconfigure` retain stop/disable/reload cleanup;
- successful self-replacement runs the old inode, atomically replaces the updater, exercises the production restart exit path, starts the replacement binary, and verifies the next rebuild uses the live updater.

I also mutation-tested the Debian regression harness by removing the `upgrade` guard from a temporary staged `prerm`; the test then observed the forbidden `systemctl --user stop codex-update-manager.service` side effect, confirming the coverage is behavior-sensitive rather than string-only.

I ran repeated adversarial review passes over the updated diff and fixed the issues they exposed, including production-script staging in the test path, healthy-current-vs-installed binary resolution, exercising the real restart exit path, non-root CI skip behavior, and test-shell quoting.

Validation on the rebased branch:

- `cargo test -p codex-update-manager`: 60/60 passing
- repeated full updater suite: 10/10 passing before rebase finalization, plus 15/15 in the final paranoid pass
- repeated self-replacement + staged Debian `prerm`: 50/50 combined paranoid rounds, in addition to the earlier 25/25 runs for each
- updater Clippy with `-D warnings`: 3/3 earlier plus 5/5 final paranoid passes
- script smoke suite: 3/3 earlier plus 5/5 final paranoid passes
- Core CI: passing
- Debian package CI: passing
- RPM package CI: passing
- Pacman package CI: passing
- record/replay tests and Clippy: passing
- final `git diff --check`: clean

The branch is rebased onto current `origin/main` and the worktree is clean.

ilysenko (changes requested):
Thanks for the thorough lifecycle fix. The original stale-daemon path and the staged `prerm` behavior are substantially improved, but three blockers remain:

1. `passive_replacement_disposition()` blocks forever whenever the persisted status is `Installing`. If a separate `install-ready` or rollback process replaces the updater and then dies before saving `Installed`, the still-running service daemon will see the replacement every 15 seconds but will neither recover the interrupted transaction nor restart onto the installed binary. Please add bounded, ownership-aware recovery (for example, a lock/lease/liveness signal) so an active installer is protected while an abandoned `Installing` state can converge safely.
2. `tests/deb-prerm.test.js` chooses `/run/user/${40000 + pid % 20000}` without proving that the directory was created by the test. Both privileged branches accept a pre-existing path; the sudo branch chmods it to 0777, and cleanup recursively removes the whole directory. A UID collision can therefore modify or delete another user's active runtime directory. Please reserve a guaranteed-new numeric directory atomically, track/validate that it belongs to this fixture, and clean up only what the test created.
3. `successful_self_replacement_restarts_on_new_binary_and_next_build_uses_it` does not exercise the production install lifecycle: the child calls `exit_for_replacement()` directly on a marker and the parent manually starts the replacement. It never runs `install_ready()`, observes the package result, or proves the required `save Installed -> read back -> exit` ordering. Moving or deleting the production `restart_after_persisted_install()` call would leave this test green. Please add behavioral coverage through the production lifecycle so that exact ordering regression is caught.

Please run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, rerun the review, and repeat until that model reports no remaining blockers before requesting maintainer re-review.

kassi0-peia:
Since the last review, I reworked the updater replacement/recovery path around the three issues you called out and then did another adversarial pass over the resulting state machine.

The main changes are:

- Added a gated package-command ownership layer that persists the exact child PID/start-time identity before privileged mutation is released, instead of publishing the long-lived updater daemon as the transaction owner.
- Make abandoned `Installing` recovery bounded and ownership-aware, including future/rolled-back wall-clock timestamps.
- Reconcile abandoned Debian installs only from strong package evidence: fully configured package state, matching artifact identity, matching candidate version, and an observed install effect.
- Preserve the pre-transaction installed version so update and rollback recovery can reproduce the same bookkeeping as their successful paths.
- Exercise the real `install_ready()` replacement lifecycle in the self-replacement tests, including `Installed` persistence/readback before exit and restart.
- Harden the Debian `prerm` fixture so it only creates and cleans up runtime directories it can prove it owns.
- After the gated package command has been released, preserve `Installing` plus the durable transaction on a later nonzero exit because package mutation may already have occurred. The same rule now applies to rollback.
- Treat Debian `prerm failed-upgrade` like `upgrade` so upgrade failure/unwind cannot stop the updater that owns the package transaction.

I also added regressions for the new ownership/recovery cases, the production replacement lifecycle, post-release install/rollback failures, and the Debian `failed-upgrade` path.

ilysenko (changes requested):
Thanks for the update. I found blockers:

- The gated package command is released immediately after `save_updater()`, but that function only writes and renames `state.json`; it does not `sync_all()` the temporary file or fsync the parent directory. A power loss can therefore make package mutation durable without the transaction ownership journal that recovery relies on. Please make the pre-release journal durably ordered before sending the `go` token and cover that ordering.
- `ProcessIdentity` contains only PID and `/proc` start-time ticks, with no boot identity, while `identity_is_alive()` maps every `/proc` read error to a dead owner. After the grace period, a hidepid/permission or transient procfs failure can let the daemon recover over a still-running privileged transaction, and the identity is ambiguous across reboots. Please persist boot identity and distinguish confirmed process absence/mismatch from an unreadable owner.
- Successful abandoned RPM and pacman transactions cannot reconcile because `installed_package_version_for_recovery()` deliberately returns `None` for both. If the owner disappears after either package manager commits, the updater permanently records `Failed` with stale installed and rollback bookkeeping. Please add strong format-specific completion evidence and behavioral recovery coverage for both supported formats.

Please run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, and repeat until that model reports no remaining blockers before requesting maintainer re-review.

kassi0-peia:
@ilysenko All required CI checks are now green on rebased commit `46b943b`, including Dock, Rust, both Nix builds, package matrix, and the official Linux gate. The requested updater fixes, legacy schema-2 recovery handling, RPM `--noscripts`, dependency audit, and documentation updates are included. Please re-review PR #1430.

ilysenko (approved):
The three previously requested recovery and durability fixes are verified on the current tree. The history-only rewrite preserves the reviewed tree, focused local updater and Debian lifecycle tests pass, and the complete GitHub check matrix is green.

PR #1440, opened 2026-09-06
fix: restore Linux Computer Use shared descriptor gate

nakasyou:
## Summary

With `computer-use-linux` enabled, the latest signed stable Linux package
(`26.901.51231`) still excluded Computer Use from the runtime plugin registry.
The patch matched obsolete inline descriptors and could report success after
rewriting only the skill selector.

- Match the current shared-descriptor registry and enable Linux availability.
- Override the inherited installation opt-in requirement while preserving the
  opt-out key, migration callback, and existing macOS/Windows availability.
- Reject missing, ambiguous, or changed registry gates.
- Move the gate patch into the feature directory and remove obsolete fallbacks.
- Keep the feature disabled by default.

## Validation

- `node --test linux-features/computer-use-linux/test.js scripts/patch-linux-window-ui.test.js`
  — 16 tests passed.
- Verified the latest official amd64 package through signed repository metadata.
- Built the official bundle with only `computer-use-linux` enabled; all six
  feature patches applied.
- Built and launched the Nix package on NixOS with Niri/Wayland.
- Confirmed the patched gate in the running app.asar and Computer Use tool
  availability. Verified screenshots, clicking, and text input in native apps.
- Broader feature validation recorded 783 passing tests and two failures in
  Flatpak native messaging and MCP helper reaping, outside the changed feature.
- ARM64, X11, and other package formats were not manually tested.
  PR CI has not run yet.

## Checklist


ilysenko (changes requested):
Thank you for updating the gate to the current shared-descriptor registry. Two blockers remain:

1. Make the marketplace selector part of the enforced patch contract (`linux-features/computer-use-linux/plugin-gate.js:3-10,35`). The helper validates exactly one descriptor, but `rewriteComputerUseMarketplaceSelector()` does not verify that it matched anything. With the current descriptor and a missing or drifted selector, the function returns changed source successfully; the patch report is therefore `applied` and enabled-feature acceptance accepts a candidate where Linux cannot select `legacy-mcp`. I reproduced this directly with the current descriptor fixture and no selector. Require exactly one pristine or already-patched selector, reject zero/ambiguous/changed selectors, and add the corresponding negative tests.

2. Remove `parseDestructuredParamAliases()` from `scripts/patches/impl/computer-use.js:155-165`. It has no remaining consumer after this PR removes the obsolete inline/flexible gate fallbacks, so retaining it leaves dead compatibility code behind in a current-stable drift cleanup.

Before requesting maintainer re-review, please run the complete updated base-to-head diff through your own code-review model, fix every blocker it finds, rerun the review, and repeat until that model reports no remaining blockers.

ilysenko (approved):
Reviewed the current head after the requested fixes. The selector contract now fails closed for missing, duplicate, changed, pristine, and already-patched forms; the obsolete parser was removed. No blockers remain.

Issue #960, opened 2026-07-14
mcp-helper-reaper terminates live node_repl after thread resume, causing Transport closed


gtstore:
## Summary

With the optional `mcp-helper-reaper` feature enabled, a live `node_repl` MCP server is terminated shortly after resuming an existing Desktop thread. Subsequent `node_repl/js` calls fail with `Transport closed`, and Browser Use becomes unavailable for the rest of that thread.

Launching the same installed build with `CODEX_MCP_HELPER_REAPER_DISABLE=1` keeps `node_repl` and the in-app browser transport alive. The separate `node-repl-reaper` feature remains enabled in both cases.

## Environment

- OS: KDE neon 24.04, Wayland session with Electron running through X11/XWayland
- Installed Linux wrapper: `2026.07.13.134819`
- Wrapper commit: `bcec9e7313d3b5ff606d14c54eed5ba3d1ff6951`
- Upstream Desktop app: `26.707.62119`
- Electron: `42.1.0`
- Codex CLI: `0.144.3`
- Relevant enabled features:
  - `mcp-helper-reaper`
  - `node-repl-reaper`

## Reproduction

1. Start Codex Desktop normally with `/opt/codex-desktop/start.sh`.
2. Resume an existing thread.
3. Invoke a minimal Node REPL tool call:

   ```js
   nodeRepl.write("node-repl-live")
   ```

4. Bootstrap the bundled Browser client and obtain the IAB browser. The initial call may succeed.
5. Wait for the delayed cleanup passes and invoke `node_repl/js` again.

Observed result:

```text
tool call error: tool call failed for `node_repl/js`

Caused by:
    Transport closed
```

The `node_repl` process is no longer present after the failure.

### Feature enabled

- Initial browser bootstrap succeeded once.
- The next functional call failed with `Transport closed`.
- `node_repl` disappeared from the process table.

### Feature disabled for one launch

Desktop was started with:

```bash
CODEX_MCP_HELPER_REAPER_DISABLE=1 /opt/codex-desktop/start.sh
```

Verification sequence:

1. Minimal `node_repl/js` returned `node-repl-live`.
2. IAB initialization succeeded.
3. `browser.tabs.list()` returned `[]`.
4. After an 8-second delay, covering the default delayed passes at approximately 3, 5, and 7 seconds, another `browser.tabs.list()` call returned:

   ```json
   {
     "status": "still-live",
     "tabs": []
   }
   ```

[...]

PR #1447, opened 2026-09-08
fix(node-repl-reaper): preserve helpers with live Codex ancestors

numberyy:
## Summary

Fix `node-repl-reaper` terminating a live Browser/Computer Use helper when a Node launcher sits between the helper and its Codex owner.

The change is limited to the optional, disabled-by-default `node-repl-reaper` feature: its cleanup script, tests, and README. It does not change `mcp-helper-reaper`, the upstream app payload, package configuration, or sandbox permissions.

### Problem and cause

On Fedora 43 x86_64 with signed OpenAI Linux package `26.901.51231` and bundled Codex `0.153.4`, browser calls failed with `Transport closed`. The cleanup log recorded live `node_repl` processes being killed at the same times.

The bundled unified Computer Use plugin starts this process chain:

```text
codex app-server → node launch.mjs → node_repl
```

The old cleanup rule checked only the helper's immediate parent. It rejected the Node launcher and classified the live helper as leaked. Direct Codex children were already protected; indirect children were not.

### Fix

- Follow up to 64 process ancestors to find a live Codex owner.
- Keep the helper when process information cannot be read or the walk cannot establish that it reached the root.
- Check identity and ancestry again before both SIGTERM and SIGKILL.
- Pass through `codex-linux-sandbox` and `codex-mcp-helper-reaper` without treating them as Codex owners.
- Preserve cleanup after the Codex owner exits, including when a non-Codex launcher remains alive.

The orphan fixtures now launch from a shell that exits and wait for the helper's executable path before running cleanup. This avoids treating the test runner's own live Codex ancestor as an orphan and removes an exec timing race.

### Related work

- #482 introduced this optional feature.
- #960 and #965 addressed a similar live-helper termination in the separate `mcp-helper-reaper` feature. This PR does not change that implementation or reopen that resolved report.
- #1427 corrected lifecycle hook context. This PR retains those hooks and changes the process ownership check only.

I checked open PRs and searched issues and PRs for `node_repl`, `node-repl`, `reaper`, and `Transport closed`; I found no open duplicate for this ownership case.

## Validation

- The two Node-launcher regression cases failed against the original script before the fix.
- `node --test linux-features/node-repl-reaper/test.js`: 14 tests passed on two consecutive runs after review changes. Coverage includes script and executable Codex owners, both excluded intermediate process names, wrapped helpers, owner exit with a surviving launcher, unreadable/cyclic/bounded ancestry, and checks before signals.
- `bash tests/scripts_smoke.sh`: passed.
- `bash -n linux-features/node-repl-reaper/*.sh`: passed.
- `node --check linux-features/node-repl-reaper/test.js`: passed.
- `git diff --cached --check`: passed.
- A feature-only build against signed OpenAI Linux `26.901.51231` passed. The built cleanup script matched source, and upstream `resources/app.asar` remained byte-identical. Subsequent review changes touched tests and documentation only.

### Live A/B check

In a disposable Agent Workspace with networking disabled and a read-only project mount, the real bundled Codex app-server launched the real bundled Node CUA launcher and `node_repl`. MCP initialization and `tools/list` succeeded. The same helper replied to an MCP ping after each of three fixed cleanup passes. Running the original script then killed that helper (exit 143), and further transport writes failed. The test workspace was stopped afterward; the installed host app was not replaced.

The built ChatGPT window also opened in that workspace with a fresh user-data directory. This was separate from the transport check, not a full signed-in browser navigation test.

### Limits and risk

Not run: installed-host browser recovery, the full five-minute timer interval in the live A/B check, live subreaper scenarios, arm64, other distributions, or a package-format install matrix. The shared feature script applies across package formats when enabled, but those installations were not tested here. The script requires Bash 4.4 or newer. An uncertain ancestry can leave a leaked helper alive; `/proc` reads and signals are not atomic, so this does not claim to eliminate all process races.

## Checklist

- [x] The reproduced failure uses the current signed stable OpenAI Linux package; no alternate upstream format or version-specific fallback was added.
- [ ] I added or updated relevant tests, ran the validation listed above, and confirmed that required CI checks pass. Local checks pass; PR CI is pending.
- [ ] I reviewed the final diff with my coding agent using maximum reasoning effort, addressed all findings, and reran the relevant tests. Review findings were supplied and addressed, and tests were rerun; the maximum-effort setting has not been verified.

## Disclosure

This contribution is AI-assisted. An AI coding agent helped investigate the failure, implement the fix and regression tests, run the isolated live check, address review findings, and prepare this description. I reviewed the work and requested submission. The validation above distinguishes measured results from checks that were not run.

### DIFF
--- a/linux-features/node-repl-reaper/README.md
+++ b/linux-features/node-repl-reaper/README.md
@@ -6,12 +6,22 @@ reap them: helpers accumulate over long sessions and survive their owner
 over a day under a hidden-to-tray instance). Each holds memory and file
 descriptors indefinitely.
 
-This feature reaps **leaked** helpers — those whose parent is no longer a
-live Codex owner process. Helpers with a live Desktop `codex app-server` parent
-or a live CLI Codex parent such as `codex resume` are never touched, so active
+This feature reaps **leaked** helpers — those with no live Codex owner in their
+process ancestry. Helpers with a live Desktop `codex app-server` ancestor
+or a live CLI Codex ancestor such as `codex resume` are never touched, so active
 Browser Use sessions are unaffected. Matching is scoped to this install's
 `resources/cua_node/bin/node_repl` path, so side-by-side installs reap
 independently.
+
+The unified Computer Use launcher starts helpers through Node, so the direct
+parent need not be Codex. The reaper follows up to 64 ancestors and leaves the
+helper alone if process information is unreadable, an ancestor disappears,
+or traversal does not reach a root within that bound. It checks ownership again before
+each signal. A surviving launcher does not protect the helper after its Codex
+ancestor exits.
+
+The walk also passes through `codex-linux-sandbox` and
+`codex-mcp-helper-reaper`. These processes do not count as Codex owners.
 
 ## How it runs
 
@@ -25,9 +35,11 @@ independently.
 - **App exit**: the after-exit hook runs one immediate pass.
 - Reaping sends SIGTERM, then SIGKILL after a grace period
   (`CODEX_NODE_REPL_REAPER_KILL_GRACE` seconds, default 5), re-checking
-  process identity before escalating to guard against pid reuse.
+  process identity and ancestry before escalating.
 
 ## Compatibility
+
+This feature requires Bash 4.4 or newer.
 
 This feature can be enabled together with `mcp-helper-reaper`. If that feature
 wraps `resources/cua_node/bin/node_repl`, this reaper also matches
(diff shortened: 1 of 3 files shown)


ilysenko (approved):
Reviewed the complete diff and process ancestry/signal behavior. The feature remains disabled by default, preserves helpers on uncertain ancestry, and rechecks before both signals. Regression cases are confirmed passing in CI; no merge blockers found.

PR #1453, opened 2026-09-09
Fix filesystem-root task follow-ups with an opt-in feature

DPS0340:
## Summary

An existing local task rooted at `/` can accept its initial prompt but rejects follow-ups with **Unable to send message — Select a project to continue**. The current renderer treats `["/"]` as the missing-workspace placeholder even for an existing local follow-up.

Add the disabled-by-default `filesystem-root-follow-ups` feature. It recognizes the complete current submission guard and permits `/` only when both the composer mode and follow-up type are local. New-task placeholders, empty roots, projectless policy, cloud/worktree behavior, and other submission checks are preserved. Task records and filesystem permissions are not changed. This does not reattach a detached task to its former project.

Related upstream reports: https://github.com/openai/codex/issues/43845 and https://github.com/openai/codex/issues/36548. This is an opt-in downstream workaround; it does not close those upstream issues.

The matcher captures minified identifiers instead of pinning bundle hashes or function names, is idempotent, and leaves ambiguous, drifted, or partially patched guards untouched. Existing enabled-feature enforcement rejects a candidate on a missed match. English and Chinese feature listings and an adjacent README document activation and removal.

Scope: one optional ASAR feature for the current signed stable Linux package, `26.903.61454`, on `amd64` and `arm64`. No launcher, package-format, updater, dependency, or default-feature changes.

## Validation

- Adjacent regression tests: **18 passed**, including reproduction of the original root blocker, surrounding submission cases, renamed aliases, duplicate/partial guards, idempotence, disabled behavior, and patch-runner enforcement.
- Patch/descriptor/feature suite: **898 passed** using a headless environment:

  ```bash
  env -u CODEX_OZONE_PLATFORM -u WAYLAND_DISPLAY -u DISPLAY -u XDG_SESSION_TYPE \
    node --test scripts/patch-linux-window-ui.test.js scripts/patches/descriptor.test.js \
    scripts/lib/linux-features.test.js linux-features/*/test.js
  ```

  Clearing those inherited desktop variables avoids an existing MCP-reaper test assuming no Ozone flag; the test itself is unchanged.
- Verified signed `InRelease`, indexed package hashes, and both official `26.903.61454` packages with the repository's resolver. Ran `install.sh` into isolated output directories with only this feature enabled on each architecture; both candidates passed enforcement and built successfully.
- Compared all **9,023 logical ASAR file contents** per architecture: only the primary webview bundle contents changed. `ChatGPT`, `codex`, `rg`, and the code-mode host remained byte-identical. The final matcher produces exactly the verified built bundle on both architectures.
- `node --check` for the new JavaScript files, new documentation links, and `git diff --check` passed.
- Manual evidence: an equivalent local workaround restored actual follow-up submission on CachyOS/X11 with Nix and package `26.901.51231`, confirmed by the affected user. The latest-package checks above validate both builds and their bundled guard; GUI submission on `26.903.61454` and an ARM GUI session were not exercised.

GitHub CI is pending. The final diff was reviewed; a separately configured maximum-reasoning-effort review has not been run.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,7 @@ requirements, known limitations, configuration, and tests.
 | `computer-use-linux` | Linux desktop-control UI and native MCP backend | [Docs](linux-features/computer-use-linux/README.md) |
 | `copilot-reasoning-effort` | Persistent reasoning-effort defaults for Copilot-auth sessions | [Docs](linux-features/copilot-reasoning-effort/README.md) |
 | `directory-only-working-tree-watch` | Bounded Watchbound working-tree watching | [Docs](linux-features/directory-only-working-tree-watch/README.md) |
+| `filesystem-root-follow-ups` | Allow follow-ups in existing local tasks rooted at `/` | [Docs](linux-features/filesystem-root-follow-ups/README.md) |
 | `flatpak-chrome-native-messaging` | Bridge the official Chrome extension into Flatpak Google Chrome | [Docs](linux-features/flatpak-chrome-native-messaging/README.md) |
 | `frameless-titlebar` | Hide official Linux overlay buttons for compositor-managed decorations | [Docs](linux-features/frameless-titlebar/README.md) |
 | `global-dictation` | X11 and XDG portal global dictation hotkeys | [Docs](linux-features/global-dictation/README.md) |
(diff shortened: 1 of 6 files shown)


ilysenko (approved):
Reviewed the focused opt-in feature and regression coverage. Verified the guard, rootless predicate, and followUp call path in the current 26.903.61454 bundle; required CI is green. No merge blockers found.

PR #1475, opened 2026-09-14
fix(computer-use): handle Linux Sky setup handshake

danielcadev:
## Summary

Fix Linux Computer Use startup after the current unified runtime began initializing the configured Sky service with a `{ type: "setup" }` handshake.

The Linux trusted service only accepted its native `{ method, app, params }` requests, so the official `tinyskyAlt` initialization failed with `This native Linux Computer Use operation is not supported` before the Linux client adapter could be installed.

This change:

- answers the official Sky setup handshake as a Linux target;
- keeps native operations routed through the existing Linux adapter;
- updates both `getState` and its `initialize` alias;
- bumps the bundled plugin revision to `linux-native.5` so stale cached copies are replaced.

This is a focused follow-up to the unified runtime changes in #1466.

## Validation

- Added a regression test reproducing the rejected Sky setup handshake.
- 59/59 focused Computer Use tests passed.
- `MAX_BUILD_THREADS=8 ./install.sh` passed against signed Linux package `26.908.40834`.
- Exercised the packaged `cua-repl`/`node_repl` runtime:
  - `cua.getState()` discovered 7 native applications.
  - `cua.getApp()` and `getAXState()` successfully queried a real KDE window.
- After installing the generated RPM and restarting the desktop app, Computer Use connected to a real Chrome tab and captured a running Ghidra window.
- `git diff --check` passed.

## Checklist

- [ ] Required GitHub CI checks pass.
- [x] I reviewed the final diff with my coding agent, addressed all findings, and reran the relevant tests.


ilysenko (approved):
Reviewed the focused Linux Computer Use handshake fix and its regression coverage. The current head is scoped, policy-compliant, and all required checks are green.

PR #1486, opened 2026-09-17
fix(nix): let the bundled headless LibreOffice load on NixOS

Konakonai:
## Summary

On NixOS, document features that go through the primary runtime's bundled headless LibreOffice (`~/.cache/codex-runtimes/codex-primary-runtime/dependencies/native/libreoffice-headless`, LibreOffice `25.2-headless-codex.1`) fail to start. Running `soffice.bin` under the packaged Bubblewrap adapter's exact `NIX_LD_LIBRARY_PATH` reproduces two errors in sequence:

1. `liblcms2.so.2: cannot open shared object file: No such file or directory`
2. after adding lcms2: `libcurl.so.4: version 'CURL_OPENSSL_4' not found (required by .../libsvllo.so)`

The bundled LibreOffice ships its own cairo, pixman, ICU, libxml2 and libxslt, but expects `libfontconfig.so.1`, `libfreetype.so.6`, `liblcms2.so.2`, `libcurl.so.4` (with `CURL_OPENSSL_4` symbol versions) and NSS from the host. The workspace library path added in #1400 carries fontconfig and NSS, but no freetype or lcms2, and its only `libcurl.so.4` is the GnuTLS-compat build whose symbol versions were renamed to `CURL_GNUTLS_3` for the bundled Git.

This change:

- adds `freetype` and `lcms2` to `workspaceRuntimeLibraries`;
- places the stock `curl` ahead of the GnuTLS-compat build, so `libcurl.so.4` resolves with OpenSSL versions while the distinct `libcurl-gnutls.so.4` name still reaches the compat build for Git;
- adds a `document-runtime-probe` (a small C program linking libcurl with `CURL_OPENSSL_4`, FreeType and lcms2, with the generic interpreter and no RUNPATH, mirroring `generic-runtime-probe`) that runs in every `nix-runtime*` check and in the VM smoke test;
- documents the fix in `CHANGELOG.md` and `docs/nix.md`.

Affected: Nix package on NixOS only (`x86_64-linux` tested; `aarch64-linux` is the same code path). Other formats are unchanged.

## Validation

- `nix build .#checks.x86_64-linux.nix-runtime` passes with the new probe. Before the change, the cached `soffice.bin` fails under the unmodified adapter's `NIX_LD_LIBRARY_PATH` with the two errors above; the probe encodes the same requirements (`CURL_OPENSSL_4`, `libfreetype.so.6`, `liblcms2.so.2`).
- With the rebuilt adapter's `NIX_LD_LIBRARY_PATH` and `nix-ld`, the cached `soffice.bin --headless --convert-to pdf` produces a PDF, and the bundled `git ls-remote https://…` (via `git-remote-http`, which needs `libcurl-gnutls.so.4`) still succeeds.
- `nix build .#checks.x86_64-linux.nixos-vm` passes (the VM smoke script runs the probe through the packaged Bubblewrap adapter).
- `git diff --check` clean.
- Not tested: `aarch64-linux`.

Environment: NixOS unstable, Niri (Wayland), `x86_64-linux`, official package `26.911.61220`, features `computer-use-linux`, `directory-only-working-tree-watch`, `frameless-titlebar`.

Disclosure: this change was developed with Claude Code (Claude Opus 5); I reviewed the diff and ran the validation above.

## Checklist

- [ ] If this fixes upstream drift, it targets only the latest signed stable OpenAI Linux package and removes obsolete fallback code and tests. (not a drift fix)
- [x] I added or updated relevant tests, ran the validation listed above, and confirmed that required CI checks pass. (tests added and local validation run; will tick once CI is green)


ilysenko (approved):
Reviewed the current head. The focused Nix runtime-library ordering fix is covered by dedicated OpenSSL curl, FreeType, and lcms2 probes, both architecture builds, and the NixOS VM smoke test.

PR #1488, opened 2026-09-17
fix(nix): publish workspace runtime libraries through nix-ld

### LINKED ISSUE #1487: NixOS: Codex shell snapshot restores host NIX_LD_LIBRARY_PATH over the Bubblewrap adapter's value
## Summary

On NixOS with `programs.nix-ld.enable = true`, the library path injected by the packaged Bubblewrap adapter never reaches sandboxed commands. Codex snapshots the login shell when a thread starts and sources that snapshot before every sandboxed command; the snapshot re-exports the host `NIX_LD` / `NIX_LD_LIBRARY_PATH`, which overrides the adapter's `--setenv` values. Cached generic runtimes therefore resolve libraries through the system nix-ld path (`/run/current-system/sw/share/nix-ld/lib`), and the libraries added in #1486 are not visible to the primary runtime's headless LibreOffice.

## Environment

- NixOS unstable, `x86_64-linux`, Niri (Wayland)
- Nix package at `7852e92` (includes #1486), features `computer-use-linux`, `directory-only-working-tree-watch`, `frameless-titlebar`
- Official package `26.911.61220`, Codex CLI 0.154.0, `sandbox_mode = "workspace-write"`
- `programs.nix-ld.enable = true` (NixOS default library set)

## Evidence

Full argv of the `bwrap` process spawned for a sandboxed `sleep 20`, captured from `/proc/<pid>/cmdline`. The adapter did run and injected its arguments right before the first `--`:

```
--ro-bind /nix/store/…-nix-ld-2.0.6/libexec/nix-ld /nix/store/…-nix-ld-2.0.6/bin/nix-ld
--setenv NIX_LD /nix/store/…-glibc-2.42-84/lib/ld-linux-x86-64.so.2
--setenv NIX_LD_LIBRARY_PATH /run/opengl-driver/lib:/nix/store/…-alsa-lib-1.2.16.1/lib:…
--
/nix/store/…-codex-0.154.0/bin/codex --sandbox-policy-cwd … --apply-seccomp-then-exec --
/nix/store/…-bash-interactive-5.3p15/bin/bash -c '<script>'
```

The `<script>` Codex passes to bash contains:

```
if . '/home/lilya/.codex/shell_snapshots/<thread-id>.<nonce>.sh' >/dev/null 2>&1; then :; fi
…
exec '/nix/store/…/bin/bash' -c 'sleep 20'
```

and the snapshot file contains the host values captured from the login shell:

```
declare -x NIX_LD="/run/current-system/sw/share/nix-ld/lib/ld.so"
declare -x NIX_LD_LIBRARY_PATH="/run/current-system/sw/share/nix-ld/lib"
```

Environment of the sandboxed `sleep` (from `/proc/<pid>/environ`):

```
NIX_LD=/run/current-system/sw/share/nix-ld/lib/ld.so
NIX_LD_LIBRARY_PATH=/run/current-system/sw/share/nix-ld/lib
CODEX_SANDBOX_NETWORK_DISABLED=1
```

`codex sandbox -c 'sandbox_mode="workspace-write"' -- …` from a terminal (no shell snapshot) does see the adapter's path, which is why the existing `nix-runtime` and VM checks pass: they exercise the adapter directly, and the VM does not enable nix-ld, so its login shell exports nothing that the snapshot could restore.

## Impact

Any NixOS host with `programs.nix-ld` enabled. The default nix-ld library set covers zlib, curl, openssl and the GCC runtime, so most cached runtimes still work; the headless LibreOffice in the primary runtime fails with `libnss3.so: cannot open shared object file` (then `liblcms2.so.2`, `libfreetype.so.6`, `libfontconfig.so.1`). Hosts without nix-ld are unaffected.

## Workaround

```nix
programs.nix-ld.libraries = with pkgs; [ fontconfig freetype lcms2 nspr nss ];
```

With that, `soffice --headless --convert-to pdf` from a Desktop thread succeeds under `workspace-write` (PDF Producer `LibreOfficeDev 26.8.0.0.alpha0`).

## Proposed fix

Publish the package's `workspaceRuntimeLibraries` via `passthru`, and have the NixOS module add them to `programs.nix-ld.libraries` whenever `programs.nix-ld.enable` is set. That keeps the fix declarative and independent of whether the adapter's environment survives. An adapter-side alternative (bind-mounting an augmented directory over the system nix-ld path) would also cover `nix run` users without the module, but it has to merge the host's own library set at runtime and hard-codes the NixOS path, so I have not attempted it. PR to follow.

Disclosure: investigated with Claude Code (Claude Opus 5); I reproduced the capture and the workaround myself.

Konakonai:
## Summary

Fixes #1487.

On NixOS with `programs.nix-ld.enable = true`, the library path the packaged Bubblewrap adapter injects never reaches sandboxed commands: Codex sources a login-shell snapshot before every command, and that snapshot re-exports the host `NIX_LD` / `NIX_LD_LIBRARY_PATH`, overriding the adapter's `--setenv`. Cached generic runtimes then resolve libraries through `/run/current-system/sw/share/nix-ld/lib`, so the libraries added in #1486 are not visible to the primary runtime's headless LibreOffice, which fails with `libnss3.so: cannot open shared object file` (the captured `bwrap` argv, snapshot contents and sandboxed environment are in the issue).

This change:

- exposes `workspaceRuntimeLibraries` as `passthru.workspaceRuntimeLibraries` on the Desktop package;
- makes the NixOS module add that list to `programs.nix-ld.libraries` whenever `programs.nix-ld.enable` is set, so the system nix-ld path can load what the adapter would have provided;
- adds module-test assertions for both nix-ld states and for the package's `passthru`;
- enables `programs.nix-ld` in the NixOS VM test (mirroring a host whose login shell exports `NIX_LD_LIBRARY_PATH`) and runs `document-runtime-probe` through the system nix-ld path there. That assertion fails without the module change (`libfreetype.so.6` missing) and passes with it;
- documents the behaviour in `docs/nix.md` and `CHANGELOG.md`. Home Manager cannot set the system option; the docs tell those users to add the `passthru` list to `programs.nix-ld.libraries` themselves.

Affected: Nix package on NixOS with nix-ld enabled; `x86_64-linux` tested, `aarch64-linux` shares the code path. Hosts without nix-ld and the other package formats are unchanged.

## Validation

- `nix build .#checks.x86_64-linux.modules` passes.
- `nix build .#checks.x86_64-linux.nixos-vm` passes; with `nix/nixos-module.nix` reverted to `main` the new assertion fails on `libfreetype.so.6`.
- On the reporting host, the equivalent `programs.nix-ld.libraries` entry makes `soffice --headless --convert-to pdf` from a Desktop thread succeed under `workspace-write` (PDF Producer `LibreOfficeDev 26.8.0.0.alpha0`).
- `git diff --check` clean.
- Not tested: `aarch64-linux`; Home Manager path (documentation only).

Environment: NixOS unstable, Niri (Wayland), `x86_64-linux`, official package `26.911.61220`, Codex CLI 0.154.0, features `computer-use-linux`, `directory-only-working-tree-watch`, `frameless-titlebar`.

Disclosure: this change was developed with Claude Code (Claude Opus 5); I reviewed the diff and ran the validation above.

## Checklist

- [ ] If this fixes upstream drift, it targets only the latest signed stable OpenAI Linux package and removes obsolete fallback code and tests. (not a drift fix)


ilysenko (approved):
Reviewed the current head. The NixOS module now publishes the package runtime libraries only when nix-ld is enabled, with module assertions and a VM regression test covering the shell-snapshot override path.

PR #1497, opened 2026-09-20
Surface npm/npx updater build failures instead of stalling silently

### LINKED ISSUE #1496: Unattended updater stalls silently when npm/npx is not on the user-service PATH: check_deps omits npx, apt advice installs no npm, build failures never notify
## Impact

The unattended updater can stall silently for days. On this machine it failed every check from 15 Sep to 20 Sep across three upstream releases (26.908.70816 → 26.911.61220 → 26.915.31945) with no desktop notification, leaving the app pinned to the 2026.09.15 build while upstream moved three times.

## Environment

- Ubuntu 26.04, amd64, GNOME Wayland
- codex-desktop 2026.09.15.121500, updater from commit 5f7310d7
- Node via fnm (user shell only); distro `nodejs` 22 deb installed (provides `node`, not `npm`/`npx`)
- Updater runs as `codex-update-manager.service` (systemd user unit) with the stock service PATH

## Root cause chain

Three independent gaps compound:

1. **`check_deps()` does not verify `npx`/`npm`.** `scripts/lib/install-helpers.sh` checks `node curl dpkg-deb gpg gpgv sha256sum flock`, then prints "All system dependencies found". But `scripts/lib/asar-patch.sh:103` runs `npx --yes @electron/asar extract …`, so the build dies at exit 127 after the dependency check already passed.
2. **The remediation advice is insufficient on Ubuntu/Debian.** `dependency_help()` recommends `sudo apt install nodejs curl dpkg-dev gnupg`. The Debian `nodejs` package does not ship `npm`/`npx` (separate `npm` package), so following the documented fix reproduces the failure.
3. **Build failures never notify.** `notify::send()` is only invoked on success paths ("update ready", "codex-desktop updated"). `fail()`/`fail_check()` write `state.json` and `service.log`, but journalctl shows nothing since the unit logs nothing either — the only signal is `status: "failed"` buried in state.json. A user-mode service also inherits the stock systemd PATH, so Node installed via nvm/fnm/volta in the user's shell is invisible to the updater even though the interactive installer works fine.

Log evidence (same signature across all three releases):

```
[INFO] All system dependencies found
…/builder/scripts/lib/asar-patch.sh: line 103: npx: command not found
[ERROR] Failed at line 103 (exit code 127)
```

(A second, independent failure mode appeared once fnm's Node was made visible: enabled-feature patch drift against 26.915, already repaired on main by #1495 — not the subject of this issue.)

## Suggested fixes

Any one of these would have surfaced or prevented the stall; ideally all three:

1. Add `npx` (or `npm`) to the `check_deps()` command list, and fix `dependency_help()` to recommend `npm` alongside `nodejs` on Debian/Ubuntu (or use `node … javascript` invocation instead of `npx` for the asar tool).
2. Emit a desktop notification (rate-limited) when a check transitions to `failed`, or when a candidate has been failed for >N consecutive checks — the updater already tracks state for this.
3. Consider resolving Node from common version-manager locations (nvm/fnm/volta) when running as a user service, or document a PATH drop-in as the supported workaround.

Local workaround applied here, in case it helps others:

```
systemctl --user edit codex-update-manager
[Service]
Environment=PATH=/home/<user>/.local/share/fnm/node-versions/<ver>/installation/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

timrichardson:
## Problem

The unattended updater can stall silently for days. On Ubuntu 26.04 with the distro `nodejs` package (which ships `node` but not `npm`/`npx`), or with a version-manager Node invisible to the systemd user service, every update check failed with no notification while upstream moved through three releases (#1496).

Three compounding gaps:

1. `check_deps()` verified `node curl dpkg-deb gpg gpgv sha256sum flock`, printed "All system dependencies found", then `asar-patch.sh:103` died at `npx: command not found` (exit 127) — after the ~400MB upstream download.
2. `dependency_help()` recommended `sudo apt install nodejs curl dpkg-dev gnupg`, which reproduces the failure on Debian/Ubuntu since npm is a separate package there.
3. `fail()`/`fail_check()` persisted the error but never notified; `notify::send()` only ran on success paths, and the user unit writes nothing to the journal, so the only signal was `status: "failed"` inside `state.json`.

## Approach

- **asar-patch.sh**: fail closed at the point of use — when ASAR descriptors are active but `npx` is missing, `error` with actionable guidance instead of exit 127. The feature-free byte-preservation path is unaffected (never invokes npx).
- **check_deps()**: early `warn` when `npx` is missing (not a hard gate, matching the feature-free escape hatch); `dependency_help()` apt/dnf/zypper lines now include `npm` (pacman `nodejs` bundles npm already).
- **Updater**: desktop-notify on a *transition* into `Failed` — a healthy updater failing notifies once; repeat failures for the same candidate stay silent so a stalled candidate doesn't re-notify on every periodic check. A new candidate or a recovered-then-failed-again updater notifies again.
- **docs/troubleshooting.md**: the `npx: command not found` signature, the service-PATH cause, and the `systemctl --user edit` remedy.

## Tests

- `tests/scripts_smoke.sh`: anchored assertions pin the corrected apt line and both npx guards (message line / condition line, so removing the guard fails the smoke run — comment prose alone can't satisfy them). Proven by sabotage: reverting each fix individually makes the smoke run fail.
- `updater`: new `failure_transition_notifies_once_per_stall` unit test covering notify-on-transition, silence-on-repeat, and notifications-off. Proven by sabotage: inverting `should_notify_failure` fails the test.
- `cargo test -p codex-update-manager`: 115 passed (116 with the new test), 0 failed.
- `cargo clippy -p codex-update-manager --all-targets -- -D warnings`: clean.
- `bash tests/scripts_smoke.sh`: pass.

## Risk

Low. Shell changes add warnings and one new early-exit in a path that previously exited 127 with a worse message; the updater change adds a notification on an existing failure path, guarded by the existing `config.notifications` flag. No state-schema, trust-root, or packaging changes.

Closes #1496


timrichardson:
The `rust` check failure is unrelated to this diff: `successful_self_replacement_restarts_on_new_binary_and_next_build_uses_it` died at its fixture (`spawn installed updater` → `Text file busy (os error 26)`) — a race writing over a still-running binary under parallel test load. That test exercises no `fail_check`/notify path touched here, and the full suite (116 tests incl. this one) passes locally on the branch head. Could use a rerun of the failed job.

ilysenko:

ilysenko (approved):
Reviewed the complete updated diff at 58ccde71. The updater now notifies on metadata, download, and build failures, including new candidates; tests and required checks pass.

PR #1499, opened 2026-09-20
fix(installer): accept positional .deb equal to UPSTREAM_DEB env

### LINKED ISSUE #1498: Documented `UPSTREAM_DEB=… make build-app` always fails with "Only one upstream .deb path may be provided"
## Impact

The exact invocations documented in `README.md:96`, `README.zh-CN.md:91`, `docs/build-and-packaging.md:40`, and `docs/native-setup.md:98` fail 100% of the time when a local upstream `.deb` is supplied:

```
UPSTREAM_DEB=/path/chatgpt_26.915.31945_amd64.deb make build-app
```

```
============================================
  codex-desktop — official Linux upstream
============================================
[ERROR] Only one upstream .deb path may be provided
```

(exit 2 from `make build-app`, exit 1 from `install.sh`.)

All five targets that forward `$(UPSTREAM_ARG)` are affected: `build-app`, `rebuild`, `rebuild-install`, `inspect-upstream`, `rebuild-next` — via either the documented env-prefix form or the make-variable form `make build-app UPSTREAM_DEB=…` (command-line make variables are exported to recipe environments too). The default path (no `UPSTREAM_DEB`, signed-stable discovery) is unaffected, which is presumably why this hasn't surfaced: only builds against an explicit local package hit it.

## Root cause

`scripts/lib/install-helpers.sh:45` seeds the provided-deb path from the environment:

```bash
PROVIDED_UPSTREAM_DEB_PATH="${UPSTREAM_DEB:-}"
```

and `parse_args()` (line 108) rejects a positional argument whenever that variable is already non-empty:

```bash
[ -z "$PROVIDED_UPSTREAM_DEB_PATH" ] || error "Only one upstream .deb path may be provided"
```

Meanwhile `Makefile:20` expands `UPSTREAM_DEB` into a positional argument:

```make
UPSTREAM_ARG = $(if $(strip $(UPSTREAM_DEB)),"$(UPSTREAM_DEB)",)
```

so when `UPSTREAM_DEB` is set, `install.sh` receives the same path twice — once through the inherited environment and once as `"$1"` — and the duplicate guard fires on identical values.

## Reproduction

```
$ git rev-parse --short HEAD   # 1ef0ece6 (current main)
$ UPSTREAM_DEB=/tmp/x.deb make build-app
============================================
  codex-desktop — official Linux upstream
============================================
[ERROR] Only one upstream .deb path may be provided
make: *** [Makefile:96: build-app] Error 1

$ UPSTREAM_DEB=/tmp/x.deb ./install.sh /tmp/x.deb   # same signature without make
[ERROR] Only one upstream .deb path may be provided
```

## Suggested fixes (either works; the first is one site for all five targets)

1. In `parse_args()`, accept a positional that equals the environment-provided path (error only when the two differ — that's the genuinely ambiguous case):
   ```bash
   [ -z "$PROVIDED_UPSTREAM_DEB_PATH" ] || [ "$1" = "$PROVIDED_UPSTREAM_DEB_PATH" ] \
       || error "Conflicting upstream .deb paths: $PROVIDED_UPSTREAM_DEB_PATH and $1"
   ```
2. Or scrub the variable at the recipe boundary in the `Makefile` (`env -u UPSTREAM_DEB ./install.sh $(UPSTREAM_ARG)`) — requires touching all five recipes.

Happy to turn either into a PR with a `tests/scripts_smoke.sh` assertion if wanted.

## Workaround

Call `install.sh` directly, as the Makefile recipe does, without the variable in the environment:

```
CODEX_INSTALL_DIR="$PWD/codex-app" ./install.sh /path/chatgpt_26.915.31945_amd64.deb
```

timrichardson:
## Problem

The documented invocation from `README.md`, `README.zh-CN.md`, `docs/build-and-packaging.md`, and `docs/native-setup.md` fails 100% of the time when a local upstream `.deb` is supplied:

```
UPSTREAM_DEB=/path/chatgpt_26.915.31945_amd64.deb make build-app
[ERROR] Only one upstream .deb path may be provided
```

All five targets that forward `$(UPSTREAM_ARG)` are affected (`build-app`, `rebuild`, `rebuild-install`, `inspect-upstream`, `rebuild-next`), via either the env-prefix form or the make-variable form `make build-app UPSTREAM_DEB=…`. Only builds with an explicit local package hit it, which is why it goes unnoticed on the default signed-stable path.

Fixes #1498.

## Root cause

The Makefile forwards `UPSTREAM_DEB` to `install.sh` twice — through the recipe environment (GNU make also exports command-line variable assignments into recipe environments) and as the positional `$(UPSTREAM_ARG)` expansion — and `parse_args()` in `scripts/lib/install-helpers.sh` rejected any positional when the env-seeded `PROVIDED_UPSTREAM_DEB_PATH` was already set, even when the two values were identical.

## Fix

`parse_args()` rejects a positional `.deb` only when it **differs** from the environment-provided path; identical values are accepted as one input.

## Verification

- **Reproduced on a pristine shallow clone of `main` @ `1ef0ece6`** (no local changes): the documented command exits 2 at argument parsing.
- **RED/GREEN regression coverage in `tests/scripts_smoke.sh`**: with only the fix reverted, the new assertions fail (`parse_args rejected a positional .deb identical to UPSTREAM_DEB`); with the fix, the whole suite passes.
- **End to end**: `UPSTREAM_DEB=<verified 26.915.31945 deb> make build-app` on the patched clone completes (`Installation complete`, exit 0); the same command on `main` fails at argument parsing.
- `bash -n install.sh scripts/lib/*.sh launcher/start.sh.template` clean; `bash tests/scripts_smoke.sh` fully green (includes the `node --test` suites); `scripts/lib/upstream-linux-package.test.js` 4/4.
- A genuinely conflicting pair (env `a.deb` + positional `b.deb`) is still rejected, with a clearer message.

## Notes

- No payload changes: `resources/app.asar` untouched; the change is argument parsing only. No new dependencies.
- CHANGELOG entry added under Unreleased/Fixed.


ilysenko (approved):
The documented local-package invocation now accepts the same path from the environment and positional argument while still rejecting conflicting paths. The focused regression tests and required CI checks pass.

